PR #11, opened 2021-01-03
Tiny css tweaks, big css framework

octoshrimpy:
Closing in on a specific style. 

* Set up variable calculation for better color cohesion
* fixed status icon for games in library sticking to the top of the name bar
* made name bar for library games a bit transparent
* styled logout button

---

There is no easy way to leave the logout screen, as the navbar does not have the "back to library" button there.
Clicking on username from game info screen does nothing; expected: goes to same logout screen. 

image
image


octoshrimpy:
image
sneak peek into what might be getting added next, design-wise :)

flavioislima:
Nice changes! I think the colors of the buttons are better now.

I think for the icons on the GameCard, would be better to add a Play Button, and then we can launch the game directly from the library page. If the Game is not installed we can add an Install button. So clicking on the Game image could lead to the configuration page, but clicking on the buttons could play/install the game.

We will have a UX/UI designer to work with us pretty soon and we already have a draft of the final design here for the Library, but this is still in discussion. But it is something close to what you're doing already:

image
image

Also, we could have some kind of badge to indicate that the game is installed, like you're doing now with the icon, but on the top of the Game Card.

And that's true, I implemented the Login function thinking only of making it work but I could not make an easier way of getting out of it for now.

O logout button on the username is a great idea as well.

I'll make a simple change today to close the Login screen after it (since it is really annoying) and will add the logout button to the username as well with he refresh library because it seems that for some people it is not happening automatically.

flavioislima:
@octoshrimpy here when you have time:
https://www.figma.com/file/jJTpm7UewEkzAshneJczvr/Heroic-wireframes?node-id=5%3A41

I'm working on the Navbar and on the Login Page now.
If you want, you can make the changes on the GamePage and on the Library. 💪🏽

octoshrimpy:
Thanks! Is there some place we can all communicate that isn't in a PR?

Here are a few tweaks and consistencies I've added to Biliane's design.

https://www.figma.com/file/35Rn0hzZrRSAiOkblSl6fc/Heroic-wireframes?node-id=0%3A1

chernyaknikita:
Made my version too :D
https://www.figma.com/file/odGRvmPIKNmTxeTy3507IJ/Heroic-wireframes-Copy?node-id=0%3A1

flavioislima:
> Thanks! Is there some place we can all communicate that isn't in a PR?
Sure, search for @flaviofearn on Telegram.

> Here are a few tweaks and consistencies I've added to Biliane's design.
> 
> https://www.figma.com/file/35Rn0hzZrRSAiOkblSl6fc/Heroic-wireframes?node-id=0%3A1

Nice, I'll take a look after my work hours.
If possible, take a look at my current branch https://github.com/flavioislima/HeroicGamesLauncher/tree/add_settings
I have the new Login and the Navbar almost complete there. I also changed the routers. 

I don't think you will have too many conflicts because I haven't touch the library component and the CSS.

I also was finally able to get rid o Xterm for good. So we will have the installation and uninstallation feedback to use on the Game Page. 

I'm planning to release a new version with all the changes and fixes next Wednesday so if everything goes well we can release all the changes together :D

PR #63, opened 2021-01-18
PKGBUILD: Creates a symbolic link of the executable to /usr/bin/heroic

tomascarreira:
A small change to PKGBUILD that creates a symbolic link to be able to launch heroic from the terminal or with a launcher like dmenu.


flavioislima (approved):
Nice, thanks for that. I'll need to update the aur scripts as well because I noticed that there is no link as well.

PR #71, opened 2021-01-19
Changed some of the pixel measurements to clamp()s on the header

wiryfuture:
I thought that the header looked a big small when I resized the window because it wasn't changing the font-size, so I set it to a clamp() on `.App .header`. I also thought the toggle switch looked a bit too small so I clamped that too.

I have some doubts about the header font size being a tad too big now, but if you change the preferred size in the clamp on `.App .header` you can make it a bit smaller easily.

How it looks currently: 
image
image
image
image

How it looks with the clamps:
image
image
image
image

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -21,7 +21,7 @@
   font-family: Cabin;
   font-style: normal;
   font-weight: normal;
-  font-size: 14px;
+  font-size: clamp(18px, 1vw, 25px);
   line-height: 17px;  
   color: #C1C4C3;
 }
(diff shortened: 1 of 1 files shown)


flavioislima:
@wiryfuture nice catch. For me looks good but I know that @octoshrimpy is also making a lot of changes on the design and the CSS part. 

@octoshrimpy what do you think?

octoshrimpy:
This is fixing old css that will be (has already been) nuked. fontsize will be set by rems and ems, so it _should_ scale correctly on the next design iteration with the sidebar instead of header. for now this can be merged as a hotfix, but that header is going to be refactored into a sidebar, a la discord. :)

flavioislima:
Ok, I'll merge it since I'm planning to release a bug fix release later today.

PR #75, opened 2021-01-19
Show installed games first in the library

alexheretic:
Hello, thanks for doing this project! It made sense to me that installed games should show first in the library. _(If I recall correctly that's how the windows launcher does it too)_.

If you don't fancy this, or want it to be configurable only, feel free to close!

### DIFF
--- a/src/state/GlobalState.tsx
+++ b/src/state/GlobalState.tsx
@@ -34,6 +34,14 @@ export class GlobalState extends PureComponent<Props> {
   refresh = async (): Promise<void> => {
     this.setState({refreshing: true})
     const { user, library } = await getLegendaryConfig()
+
+    // Show in alphabetical order but with installed games first
+    library.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


flavioislima:
Hi!

Thanks for the PR.

Actually, it was like this before we added the filter. Now, I'm changing the filter again because I'm adding a 'downloading' one on the next release.

So I don't think we need to sort it by installed by default.

wiryfuture:
Maybe instead of a "downloaded only" toggle you could have a small drop-down with sort/filter options... unsure of how that might look though.

flavioislima:
@wiryfuture yeah, i just added it. Not a dropdown but something similiar with the options: All, Installed and Downloading. I'm testing it to be able to release it later today with other fixes.

alexheretic:
Ok np, I guess I'd settle for it remembering my filter choice.

flavioislima:
> Ok np, I guess I'd settle for it remembering my filter choice.

No worries, thanks man!

PR #90, opened 2021-01-21
Remove .eslintcache from repo and fix about version number

wiryfuture:
`.eslintcache` is already in `.gitignore` but it's still in the repo meaning it still gets fetched from the repo when downloaded it; we should just remove it from the repo.


wiryfuture:
Also just found that the version number is hard-written for the about page, fixed that by importing it from `package.json` and using that instead.

flavioislima:
I did a lot of changes related with lint, typescript, etc. Would be better to rebase your branch. But the getVersion was already in my branch as well. And I removed the eslintcache as well :D

PR #92, opened 2021-01-22
python3-requests

wiryfuture:
I gave building a shot today and it seems that `python-requests` isn't available anymore, however afaik `python3-requests` should be a drop-in replacement.
 `python-wheel-common` and `python-setuptools` appear to be working fine.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
         "curl",
         "gconf2",
         "gconf-service",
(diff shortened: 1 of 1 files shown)


flavioislima:
Yeah, noticed that when was checking the dependencies yesterday.
I have already removed it on my current branch

PR #108, opened 2021-01-24
Fix localStorage persist after app close

alexheretic:
Fixes non-persistent localStorage, which caused filter memory to not work.


alexheretic:
Looks like the underlying issue is multiple processes running at the same time not properly sharing the localStorage. So maybe only allowing a single heroic process is a better fix.

PR #111, opened 2021-01-25
Add Proton GE installation path

beudbeud:
Heroic search for wine version in many path but there aren't the one is recommanded for Proton GE.

~/.steam/root/compatibilitytools.d

Thanks


electron/utils.ts
```diff
@@ -41,6 +41,7 @@ const getAlternativeWine = () => {
   const steamPaths: string[] = [
     `${home}/.local/share/Steam`,
     `${home}/.var/app/com.valvesoftware.Steam/.local/share/Steam`,
+    `${home}/.steam/root`,
```
flavioislima, line 44: This is actually the same as .local/share/Steam since .steam/root is a symlink.
If you click on the properties of this folder you will see that is just a symlink.

PR #116, opened 2021-01-27
Run EXE on Directory and Fixes

Linux4Good:
Fix Long Downloads being interrupted when app on tray.
Added Option to run EXE on Prefix
Added Mangohud and Pulse LAtency as toggles since they are quite common.


electron/main.ts
```diff
@@ -213,13 +214,16 @@ ipcMain.handle('writeFile', (event, args) => {
   )
 })
 
-ipcMain.on('lock', () =>
+let powerId = null
+ipcMain.on('lock', () => {
   writeFile(`${heroicGamesConfigPath}/lock`, '', () => 'done')
-)
+  powerId = powerSaveBlocker.start('prevent-app-suspension')
```
flavioislima, line 220: Nice that was so simple. I read about this before but forgot to implement. Good that it fixed it.

electron/main.ts
```diff
@@ -343,19 +347,30 @@ ipcMain.on('getAlternativeWine', (event) =>
...
     winePrefix = `'${protonPrefix}/pfx'`
   }
 
-  const command = `WINE=${wineBin} WINEPREFIX=${winePrefix} ${
+  let command = `WINE=${wineBin} WINEPREFIX=${winePrefix} ${
     tool === 'winecfg' ? `${wineBin} ${tool}` : tool
   }`
 
+  if (tool === 'runExe') {
+    command = `WINEPREFIX=${winePrefix} ${wineBin} ${exe}`
```
flavioislima, line 371: Good. Thinking far away in time, this is something that we can do to launch games outside epic. so actually this will be the command and we can call this channel to launch them. Maybe changing only to be a async call.

electron/main.ts
```diff
@@ -507,11 +522,16 @@ ipcMain.handle('egsSync', async (event, args) => {
...
   const command = isLink ? linkArgs : unlinkArgs
 
-  const { stderr, stdout } = await execAsync(
-    `${legendaryBin} egl-sync ${command} -y`
-  )
-  console.log(`${stdout} - ${stderr}`)
-  return `${stdout} - ${stderr}`
+  try {
+    const { stderr, stdout } = await execAsync(
+      `${legendaryBin} egl-sync ${command} -y`
```
flavioislima, line 527: Good catch, literally! 😂

electron/utils.ts
```diff
@@ -126,17 +127,28 @@ const launchGame = async (appName: any) => {
...
+  } = settings[settingsName] as AppSettings
 
   let wine = `--wine ${wineVersion.bin}`
   let prefix = `--wine-prefix ${winePrefix}`
 
-  envVars = otherOptions
   const isProton = wineVersion.name.startsWith('Proton')
   prefix = isProton ? '' : `--wine-prefix ${winePrefix}`
 
+  const options = {
```
flavioislima, line 141: Liked this approach since it will make it easier to add more options in the future.

package.json
```diff
@@ -43,33 +43,12 @@
         "curl"
       ]
     },
-    "pacman": {
```
flavioislima, line 46: Hmm, maybe is for the best, but would be good to listen to the community, maybe someone is actually downloding the pacman files. Even though I never did, always used AUR for that. 🤔

flavioislima (approved):
Good PR! Liked the Drag/Drop approach since sometimes we have a lot of exe to run on the prefix, it will make it a lot easier :D

PR #134, opened 2021-02-02
Fix punctuation in handleRepair

Serial-ATA:


flavioislima:
Good catch. But I just added i18n support and fixed that when adding the translations. Thanks

PR #150, opened 2021-02-09
Added new feature 'back to top btn' gameList layout + minor tweaks

onadrog:
I've added a button to scroll at the top of the app + a list layout for the games and new context to choose between grid or list.
Also a label for the search icon.
For the list layout I've removed games logo while some seems not exists.

Capture d’écran de 2021-02-09 13-31-27

Capture d’écran de 2021-02-09 13-15-34


src/components/Library.tsx
```diff
@@ -1,49 +1,100 @@
...
-              art_square,
-              is_dlc,
-              art_logo,
-              app_name,
-              isInstalled,
-            }: Game) => {
-              if (is_dlc) {
-                return null
+    <>     
+      {layout === 'grid' ? (
```
flavioislima, line 35: Since the game list depends on this check, maybe we could use a ternary expression on the `className` and on the `return` 
Instead of repeating the whole code.

src/components/UI/GameCard.tsx
```diff
@@ -87,13 +87,15 @@ const GameCard = ({ cover, title, appName, isInstalled, logo }: Card) => {
             filter: isInstalled ? 'none' : `grayscale(${effectPercent})`,
           }}
           className="gameLogo"
+          loading="lazy"
```
flavioislima, line 90: Good. Did you notice any difference? I'm thinking of a way of caching the images instead of downloading them every time. But I haven't find the best way yet.

src/components/UI/GameList.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { Link } from 'react-router-dom'
+import { useTranslation } from 'react-i18next'
+
+import ContextProvider from '../../state/ContextProvider'
+import { GameStatus } from '../../types'
+import { getProgress } from '../../helper'
+const { ipcRenderer } = window.require('electron')
+interface List {
+    cover: string
+    logo: string
```
flavioislima, line 11: We can remove the logo here since is not being used.

src/components/UI/GameList.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const [progress, setProgress] = useState({
+        percent: '0.00%',
+        bytes: '0/0MB',
+        eta: '',
+    } as InstallProgress)
+    const { t } = useTranslation()
+
+    const { libraryStatus } = useContext(ContextProvider)
+
+    const gameStatus: GameStatus = libraryStatus.filter(
```
flavioislima, line 33: I'll about a different way of storing this info since now we are repeating it in two different places.

src/components/UI/GameList.tsx
```diff
@@ -0,0 +1,105 @@
...
+        if (isMoving) {
+            return t('gamelist.moving', 'Moving')
+        }
+        if (isReparing) {
+            return t('gamelist.repairing', 'Repairing')
+        }
+        return ''
+    }
+
+    return (
```
flavioislima, line 74: I think maybe having a smaller component for Gamecard and another one for the GameList would be better. Then we could reuse all functions and change only the small component depending on the user selection.

flavioislima:
Left a few comments but looks good in general. 
Some thought I have:
- We could avoid code repetition if we could have an intermediate component that could be called like: GameLink and inside this component we had all the logic that is already inside the GameCard and GameList and then those two become dumb components that could be rendered based on the user selection inside the GameLink.
- Another thing is that would be good to have more info on the list, as the size of the installed game and also the status when something is happening like on GameCard it shows if the game is being installed or moving or repairing.
- I have already a branch with some code to launch the game from the card, that would be good as well. But I can do this after merging your PR.

other than that, it is good. thanks!

onadrog:
I've made lot of changes i found many bugs and added more language support eg for games description, refined regex etc, changed the way the game list where readed, refractor the grid/list component, still need som test, but still ok for now.

flavioislima:
Looks really good now. Just need some small changes and I think would be ready to be merged :)

onadrog:
Yes I was thinking about removing ```legendary(lis-games)``` or maybe just use it to wait for the response of the new function. And yes code need some changes to be ready for prod, I'll do that  before final commit. (This is the first time I use electron it's a bit hard to understand all the functions)
But there is some things I can't change like call to the EG api where some game names or wrong, don't know if the call can be done by gameId but it will be better.

flavioislima:
> Yes I was thinking about removing `legendary(lis-games)` or maybe just use it to wait for the response of the new function. And yes code need some changes to be ready for prod, I'll do that before final commit. (This is the first time I use electron it's a bit hard to understand all the functions)
> But there is some things I can't change like call to the EG api where some game names or wrong, don't know if the call can be done by gameId but it will be better.

Dont worry, looks really good. This is also my first experience with Electron and I just read the docs that are really good.
The ipcRendrer.invoke always returns a promise so you can add the `await` argument to them to wait for the command to finish.

onadrog:
Yes, i've checked the website, there is some graphQL things but nothings we can use in in. 
I've added a new library 'react-lazyload' so instead of loading all images at launch it use lazy load.
I think this commit is pretty good now for merge.

flavioislima:
@onadrog 
So, I'm testing here your branch and one thing I saw is that if there is no library.json, the game list is empty, so it will be ok if someone never used the app before but if they had used, they will see a blank screen until they run the refresh library function. 
But that is easy to deal with. Basically, just need to check if the file exists and if not, return the 'writeLibrary' function.

The performance is far better btw good job. The lazy load made quite a difference.

Another thing I noticed is that the Return button on settings and on the game page is aligned on the middle.
Was this intentional?
image

flavioislima:
One last thing is that on the list option, the items aren't using a fixed size so on my screen, that is a ultrawide monitor, they are really big. So maybe a fixed size aligned on the center would be a better approach. What do you think?
image

src/App.css
```diff
@@ -12,12 +12,11 @@
 
 .App .header {
   display: grid;
-  width: 1194px;
```
flavioislima, line 15: So, removing this was what broke the return button. To fix this, add `width: 390px;` to the leftCluster class.

flavioislima:
Another necessary change to avoid breaking the app is adding a fallback image to the art_cover as well on this line on main.ts:
`const art_cover = gameBox ? gameBox.url : null`
Because some game can come without it and then react will break when try to render null on the frontend.

flavioislima:
> Another necessary change to avoid breaking the app is adding a fallback image to the art_cover as well on this line on main.ts:
> `const art_cover = gameBox ? gameBox.url : null`
> Because some game can come without it and then react will break when try to render null on the frontend.

just tested again and I think is not necessary because the art_cover fallback is the art_square and also with the lazyLoading, if the image is null it won't break, it will just not render anything.

flavioislima:
For some weird reason, the LazyLoading is not working for the logos as well. I tried changing a few settings without success. 
The covers as well don't load right away, a scroll is needed for them to work. I don't know the reason.

onadrog:
Ok, i'm not using a wide monitor so can't really test I think just get rid of ```small``` tags will be ok? for the return 'button' i think i was beacause when i test with french language wich is very verbose the fixed header size was not enough anf got word-break. Ok nice if it not break anything for images. I'll investigate for the return button. But I think for sizing text you'll be more efficient than me since i'ev 24" monitor here. And yeah Lazyloading need a fisrt scroll don't know why did we get rid of that or deal with for perfomance ?

flavioislima:
> Ok, i'm not using a wide monitor so can't really test I think just get rid of `small` tags will be ok? for the return 'button' i think i was beacause when i test with french language wich is very verbose the fixed header size was not enough anf got word-break. Ok nice if it not break anything for images. I'll investigate for the return button. But I think for sizing text you'll be more efficient than me since i'ev 24" monitor here. And yeah Lazyloading need a fisrt scroll don't know why did we get rid of that or deal with for perfomance ?

For me the lost size was fine when changing the item to 8vh, text to 1em and icons to 2.5em. Making the screen smaller looked good to me as well. About the lazy loading, I think there is no better solution than caching the images locally. Don't worry about that, I'll try to implement this on my current branch.

onadrog:
Modifications done. So I've removed react-lazyload, and Css changes you said.

flavioislima (approved):
Looks good now. I'll implement and improve other things on my branch before the next release.
 
Thanks

:D

Issue #172, opened 2021-02-18
The application loads forever after the latest update


Ahmed-E-86:
The application loads forever after the latest update. That has never happened before. The application freezes while listing games, but the command `legendary list-games` works fine from the terminal.

Kubuntu 20.10
HeroicGamesLauncher version 1.5.1.deb

rodrigo21:
I'm having the same problem. Waited the whole night and it was still loading.

I'm on Arch Linux with version 1.5.1.

Just downgraded to version 1.4.0 and it worked.

wiryfuture:
I have kubuntu 20.10 as well and just installed 1.5.1 and it loads almost instantly, however the game thumbnails take a moment longer to load than they used to. 
Since I don't have this issue, I'd need some more info before finding a fix.

dfireBird:
I have the same error on Arch Linux with version 1.5.1, here is what logged in the terminal: https://gist.github.com/dfireBird/a93cd9287349c7395ac239acf8c443d1

wiryfuture:
I think that "stdout" is too big and isn't being handled correctly (the json object with all your games) or the async logic surrounding it isn't being run properly and there's an infinite loop. (library.ts, line 23-64) 
```typescript
export async function getLegendaryGames() {
  const { stdout, stderr } = await execAsync(
    `${legendaryBin} list-games --json`
  )
  if (stdout) {
    const results = JSON.parse(stdout)
    // NEED DOUBLE CHECK IF SOMETHINGS INSIDE IS NEEDED OR ADD MORE STUFF IT REDUCE SIZE OF THE FINAL FILE
    // There is maybe a better way to do that tried with filter but still not work
    results.map((res: any) => {
      delete res.asset_info
      delete res.metadata.creationDate
      delete res.metadata.developerId
      delete res.metadata.dlcItemList
      delete res.metadata.endOfSupport
      delete res.metadata.entitlementName
      delete res.metadata.entitlementType
      delete res.metadata.lastModifiedDate
      delete res.metadata.releaseInfo
      res.metadata.keyImages.map((k: any) => {
        delete k.height
        delete k.md5
        delete k.size
        delete k.uploadedDate
        delete k.width
      })
    })
    const json = JSON.stringify(results)
    return writeFileSync(`${heroicFolder}library.json`, json)
  } else if (stderr) {
    // return a file to avoid infinite loop
    return writeFileSync(
      `${heroicFolder}library.json`,
[...]

PR #176, opened 2021-02-18
Recreated new function to have smaller library file

wiryfuture:
I saw the new function @flavioislima was making to decrease the size of the library file and implemented it myself. Only issue I see is that this requires a refresh of the library if it already exists.
If any more data is needed for each game, it can just be added as a key under `games` (key names preserved from stdout).
I don't know what additional customattributes are necessary, so all of them are present.


electron/legendary_utils/library.ts
```diff
@@ -21,45 +21,33 @@ const execAsync = promisify(exec)
...
-      `${heroicFolder}library.json`,
-      '{"result":"noGamesFound"}'
-    )
-  } else {
-    // return a file to avoid infinite loop
-    return writeFileSync(
-      `${heroicFolder}library.json`,
-      '{"result":"noGamesFound"}'
-    )
+  const {stdout, stderr} = await execAsync(`${legendaryBin} list-games --json`)
```
flavioislima, line 24: The issue here wasn't the size of the file but the stdout of the list-games command. IF you have too many games, the output will be really big and the process will complain so this solution will still have the same issue unfortunatelly.
I think the current one of reading the metadata directory is not the ideal but at least it has no issues with the stdout.
Maybe there is a way to surpass this limit but I'm not sure if it is possible to do it on the node service or it needs to be a configuration on Linux.
This was also the reason why I removed the output of the instalation after finishing it, because it was giving me stdout errors.

package.json
```diff
@@ -70,7 +70,7 @@
   },
   "scripts": {
     "start": "nf start -p 3000 && electron-start",
-    "electron": "yarn build-electron && electron .",
+    "electron-start": "yarn build-electron && electron .",
```
flavioislima, line 73: I'm not really using the start command anymore because with it is not possible to debug electron. So is better to open two terminal windows and run yarn react-start on one and yarn electron on the other. Then you can read console.log that you put on electron. So maybe I'll just get rid of the start command.

electron/legendary_utils/library.ts
wiryfuture: If stdout is a string type that gets too long, you could get the command output to be saved to a temporary .json file, then parse that file as an object directly to avoid it being a string at any point until you do JSON.stringify()

package.json
wiryfuture: Alright, I've just been doing start in vscode but I'll do that instead.

electron/legendary_utils/library.ts
flavioislima: I thought about doing that but I guess it will have too much overhead. Because this command is called everytime you refresh the library so can be slow. Like on first login, for instance. So on slower machines could take too long, I'm not sure about it.

wiryfuture:
The maxbuffer has been increased on `execAsync()` at `getLegendaryGames()`, which sould prevent the bug which stopped game libraries larger than the 1MB buffer default from loading.    (Issue #172) 
https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback
 
I've removed package-lock.json as we are no longer using npm (yarn.lock).

OPINION: I think we should change `yarn electron` to `yarn electron-start`, and `yarn start` to `yarn start-both` to clarify to new devs what each does (and for the sake of the naming scheme being consistent with `react-start`)

package.json
flavioislima: I changed my mind because I was able to fix the start command. So it is working as before. Check on the main branch.

electron/legendary_utils/library.ts
```diff
@@ -20,46 +19,33 @@ import { Game, InstalledInfo, KeyImage } from '../types'
...
-  } else {
-    // return a file to avoid infinite loop
-    return writeFileSync(
-      `${heroicFolder}library.json`,
-      '{"result":"noGamesFound"}'
-    )
+export async function getLegendaryGames(): Promise<void> {
+  const {stdout, stderr} = await execAsync(`${legendaryBin} list-games --json`, {maxBuffer: Infinity})
+  const stdoutjson = JSON.parse(stdout.toString())
+  const result = []
```
flavioislima, line 25: Even though this works, would be more cleaner here if you could use an Array.map function:
`const result = stdoutjson.map(game => {"app_name": game.app_name, ...})`
I think the above will work.
dont forget to set the types fro stdoutjson. Probably will be and array of Game, so `Array<Game>`

electron/legendary_utils/library.ts
```diff
@@ -20,46 +19,33 @@ import { Game, InstalledInfo, KeyImage } from '../types'
...
+          [k.url,k.type]
+        }),
+        "title": stdoutjson[index].metadata.title,
+        "developer": stdoutjson[index].metadata.developer,
+        "customAttributes": stdoutjson[index].metadata.customAttributes
+      }
+    }
+    result.push(game)
+  }
+  if (stderr || result.length == 0) {
```
flavioislima, line 43: is better to use tripple equals on Javascript like `===`

adityaruplaha:
I see this PR is old, conflicted, and not actively maintained. Close this?

PR #180, opened 2021-02-19
Use local material icons

### LINKED ISSUE #174: Icons missing
**Describe the bug**
All icons are displayed as their name instead of the actual icon

**To Reproduce**
I'm not entirely sure

**Expected behavior**
Icons should be displayed

**Screenshots**
a
b

**Desktop:**
 - OS: Arch Linux
 - Version: 1.5.2 (from AUR `heroic-games-launcher-bin`)

**Additional context**
I don't know why it happens but anyway I think it would be better to bundle the font assets instead of importing them like here:
https://github.com/flavioislima/HeroicGamesLauncher/blob/f7c89b678b6024d70c60769e7eabb77c1e1be18b/public/index.html#L6-L10

I tried to debug it but I got `/bin/sh: line 1: nf: command not found` and I don't remember how to fix this

D3SOX:
I had to rename `Update` to `UpdateComponent` because of a name conflict.
I wanted to import it as a different name but it didn't work with `import Update as UpdateIcon from ...`
(fixes #174)

Google Fonts removed in #181

I don't know how adding those two npm packages affect the build size. I hope it uses treeshaking.

### DIFF
--- a/src/components/GamePage/GamePage.tsx
+++ b/src/components/GamePage/GamePage.tsx
@@ -19,7 +19,8 @@ import '../../App.css'
 import { AppSettings, Game, GameStatus, InstallProgress } from '../../types'
 import ContextProvider from '../../state/ContextProvider'
 import { useParams } from 'react-router-dom'
-import Update from '../UI/Update'
+import Settings from '@material-ui/icons/Settings';
+import UpdateComponent from '../UI/UpdateComponent'
 import GamesSubmenu from './GamesSubmenu'
 import { IpcRenderer, Remote } from 'electron'
 const { ipcRenderer, remote } = window.require('electron') as {
@@ -154,12 +155,11 @@ export default function GamePage() {
         <div className="gameConfigContainer">
           {title ? (
             <>
-              <span
+              <Settings
                 onClick={() => setClicked(!clicked)}
                 className="material-icons is-secondary dots"
               >
-                settings
-              </span>
+              </Settings>
               <GamesSubmenu
                 appName={appName}
                 clicked={clicked}
@@ -267,7 +267,7 @@ export default function GamePage() {
               </div>{' '}
             </>
           ) : (
-            <Update />
+            <UpdateComponent />
           )}
         </div>
       </>
--- a/src/components/GamePage/GamePage.tsx
+++ b/src/components/GamePage/GamePage.tsx
@@ -158,8 +158,7 @@ export default function GamePage() {
               <Settings
                 onClick={() => setClicked(!clicked)}
                 className="material-icons is-secondary dots"
-              >
-              </Settings>
+              />
               <GamesSubmenu
                 appName={appName}
                 clicked={clicked}
--- a/src/components/Library.tsx
+++ b/src/components/Library.tsx
@@ -1,6 +1,7 @@
 import React, { lazy, useContext } from 'react'
 import { useTranslation } from 'react-i18next'
 import cx from 'classnames'
+import ArrowDropUp from '@material-ui/icons/ArrowDropUp';
 import ContextProvider from '../state/ContextProvider'
 
 import { Game } from '../types'
@@ -72,7 +73,7 @@ export const Library = ({ library }: Props) => {
         )}
       </div>
       <button id="backToTopBtn" onClick={backToTop}>
-        <span className="material-icons">arrow_drop_up</span>
+        <ArrowDropUp className="material-icons"></ArrowDropUp>
       </button>
     </>
   )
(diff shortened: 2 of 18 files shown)


flavioislima:
Dont worry about the component name.
The only think you could change here is that, on React, if a component has no chields like the Icons, you can use then like this:
`<IconName />` instead of `<IconName><IconName />`

So I appreciate if you can change this to keep the consistency.
But overall looks good :D

src/components/GamePage/GamePage.tsx
```diff
@@ -267,7 +267,7 @@ export default function GamePage() {
               </div>{' '}
             </>
           ) : (
-            <Update />
+            <UpdateComponent />
```
flavioislima, line 270: Like this one. You can see that it closes itself since it has no children.

D3SOX:
I'll change this. Do you know if adding the icons package will export all icons and significantly increase the build size? (since you said you already used this in another project)

src/components/Library.tsx
```diff
@@ -72,7 +73,7 @@ export const Library = ({ library }: Props) => {
         )}
       </div>
       <button id="backToTopBtn" onClick={backToTop}>
-        <span className="material-icons">arrow_drop_up</span>
+        <ArrowDropUp className="material-icons"></ArrowDropUp>
```
flavioislima, line 76: And this one you can simple use:
`<ArrowDropUp className="material-icons" />`
😄

flavioislima:
> I'll change this. Do you know if adding the icons package will export all icons and significantly increase the build size? (since you said you already used this in another project)

I don't know, I think they use tree shaking so it should not be a big issue. You can run, for instance, `yarn dist AppImage` and compare the new size with the one of 1.5.2. But, even if it is bigger, it should not be an issue in my opinion.

flavioislima:
My only question is if we need the core package as well. I'm using the full material-ui icon to build another website but I have never used only the icons.
But, anyway, I was thinking in using material-ui on Heroic already because I'm not happy with the design of some elements, like the toggle, for instance, I think is very ugly, and would be good to have another one.
Also, the input boxes could be better and the ones in material-ui are good so can be something for us to try in the future.

D3SOX:
What I noticed is that the cursor here is now a `default` cursor instead of the `hand` one.
a

D3SOX:
> My only question is if we need the core package as well

I tried it without, but it kept complaining that it needs something from `@material-ui/core/utils`

flavioislima:
> > My only question is if we need the core package as well
> 
> I tried it without, but it kept complaining that it needs something from `@material-ui/core/utils`

I see, no worries then.

flavioislima:
> What I noticed is that the cursor here is now a `default` cursor instead of the `hand` one.
> a

Check if, on the App.css file, the class has a `cursor: pointer`. If not, create a class and add it to those clickable icons.

D3SOX:
> You can run, for instance, `yarn dist AppImage` and compare the new size with the one of 1.5.2. But, even if it is bigger, it should not be an issue in my opinion.

Hm it's about 2.9 MiB bigger
a

flavioislima:
> > You can run, for instance, `yarn dist AppImage` and compare the new size with the one of 1.5.2. But, even if it is bigger, it should not be an issue in my opinion.
> 
> Hm it's about 2.9 MiB bigger

Then its ok since is better now that will be offline.

One bug I found related with the `nogames` PR was this one:
image

For some reason is looking on the settings page.

D3SOX:
I think the reason is that it's using the same component `Header` and `numberOfGames` is falsy in this case. Previously the header did not display anything in this case

flavioislima:
Yeah, it is an easy fix. 
Use this one:
`const { pathname } = useLocation()`

And on the component add a check like this:
```
{numberOfGames ? (
          <span className="totalGamesText">
            {t('Total Games')}: {numberOfGames}
          </span>
        ) : (
          !pathname.startsWith('/settings') && (
            <div className="totalGamesText">{t('nogames')}</div>
          )
        )}
```

Is not beautiful but it works

flavioislima:
> I think the reason is that it's using the same component `Header` and `numberOfGames` is falsy in this case. Previously the header did not display anything in this case.
> See #182, the header is shown once. Not sure what's causing this yet

The header was disappearing because the state of the location was changing. But dont worry, I fixed it on my branch.

flavioislima (approved):
This one looks good as well unless you're doing any other changes, can be merged.
Can you try merging it, I'm not sure if it is only me or any approved PR can be merged.

D3SOX:
It is ready. Maybe you targetted the icon span anywhere else in the CSS like the one with the cursor pointer. I would take another look.
I can't merge it.

D3SOX:
Okay I checked all occurrences of `span` in the `App.css` and it's fine. Ready to be merged.

PR #181, opened 2021-02-19
Use local fonts instead of Google fonts

D3SOX:
Now uses local font assets. I hope this will not conflict with #180
(pertains #174)


flavioislima:
looks good. I don't think will have any conflicts.

flavioislima (approved):

PR #184, opened 2021-02-19
Fix CSS issues

### LINKED ISSUE #182: CSS Issues
**To Reproduce**
See screen records

**Expected behavior**
In the first video it should stay the same
In the second: I'm not sure if the header is intentional

**Screen records**

https://user-images.githubusercontent.com/24937357/108539933-e549ec80-72e0-11eb-91b9-e8a71db3fc18.mp4

https://user-images.githubusercontent.com/24937357/108539938-e67b1980-72e0-11eb-8eb7-b2e713ad4846.mp4

 - OS: Arch Linux
 - Version: 1.5.2

D3SOX:
fixes #182

### DIFF
--- a/src/components/UI/Header.tsx
+++ b/src/components/UI/Header.tsx
@@ -70,13 +70,14 @@ export default function Header({
             </span>
           </span>
         )}
-        {numberOfGames ? (
+        {numberOfGames !== undefined && numberOfGames > 0 && (
           <span className="totalGamesText">
             {t('Total Games')}: {numberOfGames}
           </span>
-        ) :
+        )}
+        {numberOfGames !== undefined && numberOfGames === 0 && (
           <div className="totalGamesText">{t('nogames')}</div>
-        }
+        )}
         {title && <div className="headerTitle">{title}</div>}
         {handleLayout && (
           <div className="layoutSelection">
--- a/src/App.css
+++ b/src/App.css
@@ -386,6 +386,7 @@
 }
 
 .settingsWrapper {
+  width: 569px;
   display: grid;
   place-self: center;
   place-items: flex-start;


src/components/UI/Header.tsx
```diff
@@ -70,13 +70,14 @@ export default function Header({
...
           </span>
         )}
-        {numberOfGames ? (
+        {numberOfGames !== undefined && numberOfGames > 0 && (
           <span className="totalGamesText">
             {t('Total Games')}: {numberOfGames}
           </span>
-        ) :
+        )}
+        {numberOfGames !== undefined && numberOfGames === 0 && (
```
flavioislima, line 78: I think is best than using the location argument. 👍🏽

D3SOX: Yeah this is also possible and I also think it's better. Changing it.

flavioislima:
The Css issue with the width can be fixed changing the width of the settingsWrapper to 569px.

src/components/UI/Header.tsx
flavioislima: I mean, keep this solution. I think useLocation approach is not good.

D3SOX: I tried it but when adding a `useLocation` and using `location.pathname === "/"` instead of `numberOfGames !== undefined` it kept showing the `numberOfGames` (`0`)
a
and also TypeScript complained that `numberOfGames is possibly 'undefined'`

D3SOX:
The issue that the game title in the `Headers` is hidden after switching to another page in the game settings is still open. I think somehow the `title` prop gets lost

src/components/UI/Header.tsx
flavioislima: Yeah. I tested here and your solution is better. Let me know when it is ready to merge.

flavioislima:
> The issue that the game title in the `Headers` is hidden after switching to another page in the game settings is still open. I think somehow the `title` prop gets lost. Doesn't happen in the `Global Settings` though.

The header was fixed here: https://github.com/flavioislima/HeroicGamesLauncher/pull/183/files#diff-092b2bf0646c9c297e8677b8e0e561921caf0c95bd5de4b80bb1b301168d0296R126

So dont worry about it. I'll merge my changes after you finish yours so we can avoid conflict.

src/components/UI/Header.tsx
D3SOX: The 0 showing probably had something to do with the way I did the inline conditional rending.
This is the first React code base I'm working with and I honestly don't like jsx/tsx. It feels weird coming from Vue where it works [a little different](https://v3.vuejs.org/guide/conditional.html#v-if)

D3SOX:
If you fixed the header this is ready to be merged

flavioislima (approved):
Nice!

Issue #189, opened 2021-02-20
Able to change install path


ianwijma:
**Is your feature request related to a problem? Please describe.**
I have selected the wrong game, and I want a way to change this.

**Describe the solution you'd like**
An menu option to change the install location.

**Describe alternatives you've considered**
none

**Additional context**
This issue is for the PR I'm writing ATM. As i'd love to help out with this project. (if wanted)

flavioislima:
Hi. 

You mean move the game or change only the folder?
Because moving is already implemented but just to change the info about the folder, this is a feature that I want to implement as well. And there is already code for that on the moving game part. So the idea would be just to change the json file and do not move the game.

But I would like to have this on another settings page with other game information and then have the option to change the install path.

flavioislima:
And, of course, I need more people to work on this project and send me PRs because I'm starting to going crazy with so much work. Also would be good to have people to test before releases because is hard to test everything and some bugs happens o my on some distros.

So feel free to send pull requests. I'm planning to have a dev channel on telegram or discord for us to discuss the development.

ianwijma:
So I've selected the wrong game when importing a game. So it doesn't 100% works, but it looks really broken. 

![image](https://user-images.githubusercontent.com/17692119/108599884-de59d180-73e7-11eb-9a83-182e2079a1e3.png)


I'd think I'd love to help out here and there. One think I want to also work on is cleaning up the code and implementing tests where needed. 

I was looking for a discord or some sort of communication platform is check how you'd see the feature implemented. So i'll look out for this :)

I have started working on some form of UI for this. so a PR will be on its way!

flavioislima:
I see. Unfortunately Legendary does not give an error when importing a game from a wrong folder, that would be a good feature request there.

The code that you're looking for is this one:
https://github.com/flavioislima/HeroicGamesLauncher/blob/main/electron/main.ts#L439

If you remove the move command, it will only change the path on the legendary installed.json file.
so it will fix it.

It is a good PR for you to start in my opinion.

I'll create the channel and share it here and with other devs for us to talk better..

flavioislima:
Here: https://discord.gg/J5SYmqb9
@onadrog @D3SOX @wiryfuture guys, I created this discord channel for Heroic, at first for us devs to help each other with the development and also for testing the app before releases. I appreciate if you all could be there :D

onadrog:
@flavioislima Hi, the link doesn't work for me

[...]

PR #201, opened 2021-02-23
Feature change install location

ianwijma:
Fix for #189


ianwijma:
Looks like removing the old fork wasn't the smartest idea. (relatively new to OSS :P) I will make a new PR where I can still make changes after the fact.

PR #204, opened 2021-02-23
Add Spanish translation

iWeaker4you:
I have added the support to the Spanish language, here I have some screenshots so that it is appreciated in detail.

Captura de pantalla_2021-02-23_04-48-21
Captura de pantalla_2021-02-23_04-35-45
Captura de pantalla_2021-02-23_04-36-09

### DIFF
--- a/public/locales/es/translation.json
+++ b/public/locales/es/translation.json
@@ -1,0 +1,163 @@
+
+{
+  "All": "Todos",
+  "box": {
+    "appupdate": {
+      "message": "Hay una nueva versión de Heroic disponible, ¿quieres actualizar ahora?",
+      "title": "Actualización disponible"
+    },
+    "choose": "Elegir",
+    "choose-egs-prefix": "Elija el Wine-Prefix donde está instalado EGS",
+    "default-install-path": "Elija la ruta de instalación predeterminada",
+    "error": {
+      "launch": "Error al iniciar el juego, ¡revisa los registros!"
+    },
+    "importpath": "",
+    "installpath": "",
+    "no": "No",
+    "quit": {
+      "message": "Hay operaciones pendientes, ¿estás seguro?",
+      "title": "Salir"
+    },
+    "runexe": {
+      "title": "Seleccione EXE para ejecutar"
+    },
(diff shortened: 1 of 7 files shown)


flavioislima:
@iWeaker4you thanks a lot man, I was really missing the Spanish translations.

public/locales/es/translation.json
```diff
@@ -0,0 +1,163 @@
...
+    "quit": {
+      "message": "Hay operaciones pendientes, ¿estás seguro?",
+      "title": "Salir"
+    },
+    "runexe": {
+      "title": "Seleccione EXE para ejecutar"
+    },
+    "select": "Seleccionar",
+    "stopInstall": {
+      "keepInstalling": "",
```
flavioislima, line 27: It think there are some missing translations on this file like this one. Can you check? 
Other than that looks fine 👍🏽

src/components/Login.tsx
```diff
@@ -174,7 +174,16 @@ export default function Login({ refresh }: Props) {
...
+              Türkçe 🇹🇷 -{' '}
+            </span>
+            <span
+              className={cx({
+                ['selectedLanguage']: currentLanguage === 'es',
+                ['language']: currentLanguage !== 'es',
+              })}
+              onClick={() => handleChangeLanguage('es')}
+            >
+              Español 🇲🇽
```
flavioislima, line 186: I know this is a difficult question, but maybe we should use the Spanish flag instead? 
Even though then we should change the Portuguese one to the Portuguese flag 🤔

public/locales/es/translation.json
iWeaker4you: @flavioislima I don't know what goes there, in all languages ​​it is empty including English :0
I have already changed the flags, the one for Spanish and Portuguese.

flavioislima: Hmm, that's weird. I need to check then because maybe this is missing somewhere.

flavioislima: You're right, those keys were duplicated for some reason. I'll fix soon.

PR #217, opened 2021-02-27
Fix translation generation

ianwijma:
Fixed the Polish and Turkish locale JSON's.
The Polish JSON was fully formatted, this probably happened with a merge, and are easily overlooked.
The Turkish document had weird characters, this can be prevented to copy text from the web. Into a text editor like gedit or alike. and then copy that, to past it in to the document.


flavioislima:
Like I said on discord, the same fixes are on this PR so we can close that since the other one is covering other things as well:
https://github.com/flavioislima/HeroicGamesLauncher/pull/218

PR #219, opened 2021-02-27
GamePage: Show file size and amount downloaded.

adityaruplaha:
~~Also fixes an issue with a ghost installation process.
`sendKill` was being called after `removeFolder` was invoked.
This caused residual files to remain in some cases.
This made the app think installation was still ongoing when it wasn't.~~

This isn't true, but it still simplifies the code and presents more recent progress than before.

### DIFF
--- a/src/components/GamePage/GamePage.tsx
+++ b/src/components/GamePage/GamePage.tsx
@@ -1,29 +1,47 @@
-/* eslint-disable complexity */
-import React, { Fragment, useContext, useEffect, useState } from 'react'
-import { useTranslation } from 'react-i18next'
-import {
-  getGameInfo,
-  legendary,
-  install,
-  sendKill,
-  importGame,
-  launch,
-  syncSaves,
-  updateGame,
-  getProgress,
-  fixSaveFolder,
-  handleStopInstallation,
-} from '../../helper'
-import Header from '../UI/Header'
 import '../../App.css'
-import { AppSettings, Game, GameStatus, InstallProgress } from '../../types'
-import ContextProvider from '../../state/ContextProvider'
+
+/* eslint-disable complexity */
+import React, {
+  Fragment,
+  useContext,
+  useEffect,
+  useState
+} from 'react'
+
+import {
+  IpcRenderer,
+  Remote
+} from 'electron'
+import { useTranslation } from 'react-i18next'
 import { useParams } from 'react-router-dom'
+
 import Settings from '@material-ui/icons/Settings'
+
+import {
+  fixSaveFolder,
+  getGameInfo,
+  getProgress,
+  handleStopInstallation,
+  importGame,
+  install,
+  launch,
+  legendary,
+  sendKill,
+  syncSaves,
+  updateGame
+} from '../../helper'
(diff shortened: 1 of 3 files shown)


src/components/GamePage/GamePage.tsx
```diff
@@ -326,7 +326,7 @@ export default function GamePage() {
   }
 
   function getInstallLabel(isInstalled: boolean): React.ReactNode {
-    const { eta, percent } = progress
+    const { eta, bytes, percent } = progress
```
flavioislima, line 329: So, the thing here is that the information we get from Legendary doesn't really mean that they are the real size. For instance, if you have BArony on your library, this app is just 200MB but Legendary shows 4000MB, I don't know if this is a conversion issue or something else. But Heroic used to have this info until someone points out that those are not the real download size.

So, if you want, you could test more or reach out the guys from Legendary to understand what these number really means :)

src/helper.ts
```diff
@@ -240,11 +240,11 @@ export async function handleStopInstallation(
     buttons: [t('box.stopInstall.keepInstalling'), t('box.yes'), t('box.no')],
   })
   if (response === 1) {
-    return sendKill(appName)
+    sendKill(appName)
   }
-  if (response === 2) {
+  else if (response === 2) {
+    sendKill(appName)
```
flavioislima, line 246: The order here doesn't really makes difference because Javascript is async and these functions are both async.
On Electron, the send and on command are sync and the invoke/handle are async. so unless we change this on the main process to make the removeFolder to use a handle and here you change to invoke. So then the app would wait for the removal and then send the kill. But if the kill come first, even with these changes, it would make no diffrence since the remove folder is the last command 👍🏽

adityaruplaha: This definitely fixed the issue, but I'll have to check.

src/components/GamePage/GamePage.tsx
adityaruplaha: https://github.com/derrod/legendary/blob/master/legendary/downloader/manager.py#L710

Shouldn't be too difficult to reimplement.

adityaruplaha: Done and dusted.

flavioislima (dismissed):
Looks good, man!
Have you tested the basic stuff like install, update, uninstall, etc? Since you touched the main file would be good to know if nothing broke.
But I'll clone your repo and test as well.
If everything is fine we can merge 😄

flavioislima:
Found only one issue: the `supportURL` is missing on the imports on main.ts.
this variable comes from utils.ts.

adityaruplaha:
This should've fixed everything. Tested a bit as well, looks good to go.

adityaruplaha:
Made a mistake while rebasing, please wait...

src/helper.ts
adityaruplaha: Since this doesn't functionally change anything and editing old commits is a massive pain, I'm just marking this as resolved.

adityaruplaha:
There. Done. Sorry about the ugly rebase, commit signing was an issue.

flavioislima (approved):
LGTM 👍🏽

PR #220, opened 2021-02-27
Types + Lint fixes and config

flavioislima:
- Changed some rules on Eslint to avoid using Any types.
- Also changed the lint command to proper check TSX files, before it was only on TS.
- Fixed several functions and variables that were with any types. Some of them are quite complex and still need work but I changed them to _unknown_ that I think is less harm than any.


onadrog (approved):
Looks good to me

PR #234, opened 2021-03-06
Fix: Missing translations at Game Save manual syncing, hungarian translation correction

BiRo96:


public/locales/hu/translation.json
```diff
@@ -106,7 +106,7 @@
   "Return": "Vissza",
   "search": "Add meg a játék nevét itt...",
   "setting": {
-    "audiofix": "Önálló javítás (Pulse Audio Latency)",
+    "audiofix": "Hangzás javítás (Pulse Audio Latency)",
```
BiRo96: I first read auto instead of audio :sweat_smile:

BiRo96:
Manual Sync options and button is translatable now
image

public/locales/hu/translation.json
```diff
@@ -115,7 +115,7 @@
     "exit-to-tray": "Kilépés a Rendszer Tálcára",
     "gamemode": "GameMode használata (Feral Game Mode-nak telepítve kell lennie)",
     "language": "Válasszd ki az Alkalmazás Nyelvét",
-    "mangohud": "Engedélyezd a Mangohud-ot (Mangohud-nak telepítve kell lennie)",
+    "mangohud": "Mangohud engedélyezése (Mangohud-nak telepítve kell lennie)",
```
BiRo96: In hungarian, there is a huge difference between imperative sentence and declarative sentence.
I first thought it would be an imperative sentence, as you have to enable mangohud, not as an option to enable or disable it.
Tthen I figured out I can real-time debug Heroic with yarn, so I continued translating not just watching the file, but the place of the text. And this mistake was left here :( 
sry

BiRo96:
Btw, it is ready to get merged :)

flavioislima (approved):
Looks good!  👍🏽 
One tip: next time, instead of putting the translations on every single file, you can simply put a default value when defining the key.
For instance:
`t('setting.manualsync.download', 'Download')`, then it would replicate to all other files.

BiRo96:
Thanks! :)
I didn't know that it works like this :O
To be honest, it's the first time I edited a Typescript file :) 
Never worked with that before

PR #242, opened 2021-03-13
Refactor Electron completely. [Part 1] 

adityaruplaha:
This refactors the code in `electron` to be much more readable and maintainable, using a more object oriented approach.

## Objectives
- Enforce consistency in coding style.
- Reduce function spam (i.e. make the code more intuitive and less overwhelming).
- Make the interface more hierarchical using OOP so that we can extend it more easily when support for other stores is added.
- Clear boundaries between the middle-end (which does all the checking, logging, handling, etc), and the true backend which should just do what it says.
- Backend should be implicitly safe as much as possible.
- Overall QOL improvements.

## Changes
- Functions have been split into multiple files, utils is now much smaller.
- Library is now a class with instance caching.
- Game (`LegendaryGame`) is now a class with instance caching.
- `GameInfo` contains structured game data, which can be created from `RawGameJSON` (previously just `Game`) which is internally stored in the files.
- DXVK is now a seperate module.
- Config, GameConfig are now classes with instance caching.
- Direct calls to Legendary are now deprecated and will be removed in the future. 
- ~~[RFC] Change the invoke notation to make more sense. Something like "game:launch" instead of "launchGame". That way its clear what the scope of the function is.~~

### Library (Done)
- [x] Restructure.
- [x] Refactor.
- [x] Remove legacy code.
- [x] Add JSDoc comments.

### Game
- [x] Restructure
- [x] Refactor.
- [x] Remove legacy code.
- [ ] Add a few bells and whistles that might be used later.
- [ ] Standardise interface.
- [ ] Add JSDoc comments.

### DXVK (Done)
- [x] Restructure.
- [x] Refactor (not much to do really).

### Config
- [x] Restructure.
- [x] Refactor (again, not much I can do).
- [ ] Add comments.

All the extra bells and whistles, cleanup, commenting etc will be in a follow up PR.

## Tests
- [x] Login
- [x] Logout
- [x] Library Loading
- [x] Settings Load
- [x] Settings Save
- [x] Install (works, progress indicators are pretty bugged.)
- [x] Uninstall (works but has an existing bug where the library would still show it as installed)
- [x] Launch
- [x] Repair (works but progress indicators are bugged, which happens in current version anyways)
- [ ] Import (someone else needs to test this)
- [x] Move Install


adityaruplaha:
Should include all changes from #243 as well now.

adityaruplaha:
Ready for some initial review, most of the things are done.

electron/types.ts
```diff
@@ -21,7 +21,7 @@ export interface AppSettings {
   wineVersion: WineProps
 }
 export interface ContextType {
-  data: Game[]
+  data: GameInfo[]
```
adityaruplaha, line 24: This should be RawGameJSON.

package.json
```diff
@@ -133,6 +133,45 @@
       "pre-commit": "yarn lint-fix"
     }
   },
+  "importSorter": {
```
adityaruplaha, line 136: This section should removed since we're dropping that extension.

electron/config.ts
```diff
@@ -0,0 +1,219 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
adityaruplaha, line 1: This will be removed.

adityaruplaha (commented):
I want to modularise `config` further, like `DXVK`.

flavioislima:
Likes the structured. Didn't liked the OOP approach. Turns things more difficult to read and complex. More things to keep in mind when adding more stuff like the GOG implementation. Since following the idea, we will need to follow the same idea for it.

I don't have strong feelings about that. I will need to dig more into it. 👍

adityaruplaha:
Yep the classes aren't complete yet so they look really complex. It will be refactored further.

adityaruplaha:
The Library stuff is almost done now. I am planning to write a standardised interface for Game in general now. Then LegendaryGame should implement it.

adityaruplaha:
I'll probably have to follow this up with a PR deciding on `camelCase` vs `snake_case` for variable names. It's super annoying.

flavioislima:
> I'll probably have to follow this up with a PR deciding on `camelCase` vs `snake_case` for variable names. It's super annoying.

This JS/TS so that's not even a question. Camel Case always. 😬

flavioislima (commented):
No need to use JSdoc  since we are using TS already. JSdoc is a more simpler version to TS, a way to type code without enforcing it like TS does.

electron/games.ts
```diff
@@ -0,0 +1,232 @@
...
+      return {
+        about: about.data.about,
+        reqs: about.data.requirements.systems[0].details
+      }
+    } catch (error) {
+      return {}
+    }
+  }
+
+  public async getSettings() {
```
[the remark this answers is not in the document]
adityaruplaha: No, since we'll be calling from an object of type Game, not GlobalConfig.

adityaruplaha:
> No need to use JSdoc

TS only gives type annotations, not in-code documentation. I find JSDoc to be quite useful.

electron/main.ts
```diff
@@ -489,11 +489,14 @@ ipcMain.handle(
...
+  switch (config_class) {
+  case 'library':
     return Library.get().getGames('info')
-  }
-  else {
-    getLegendaryConfig(config_class)
+  case 'user':
+    return GlobalConfig().getUserInfo().displayName
+  default:
+    console.log(`Which idiot requested '${config_class}' using readConfig?`)
```
flavioislima, line 498: 😂

electron/new_config.ts
```diff
@@ -0,0 +1,294 @@
...
+  public config : AppSettings
+
+  /**
+   * Get the global configuartion handler.
+   *
+   * @param version ConfigVersion to use for parsing config file. Default: 'auto'
+   * @returns GlobalConfig instance or undefined.
+   */
+  public static get(version : ConfigVersion = 'auto') : GlobalConfig {
+    let configVersion : ConfigVersion = undefined
```
flavioislima, line 46: Any specific reason o use the space before the `:`?

flavioislima: Why start as undefined? can't it just start as 'auto'?

electron/new_config.ts
```diff
@@ -0,0 +1,294 @@
...
+      if (version === 'auto') {
+        version = currentConfigVersion
+      }
+    }
+    // Config file exists, detect its version.
+    else {
+      // Check version field in the config.
+      configVersion = JSON.parse(readFileSync(heroicConfigPath, 'utf-8'))['version']
+      // Legacy config file without a version field, it's a v0 config.
+      if (configVersion === undefined) {
```
flavioislima, line 62: Maybe, instead of checking for undefined, you could just check if configVersion returns something, because undefined is a falsy value.
so this would work:
`if (configVersion) {`

electron/main.ts
adityaruplaha: :P

electron/new_config.ts
adityaruplaha: No particular reason. I'll fix the colon.

Also I just start as `undefined` because that's convention for uninitialised variables. `auto` has a different meaning altogether.

electron/new_config.ts
adityaruplaha: I mean, I just need to check if the version field doesn't exist. I thought this was the way to do it?

electron/new_config.ts
flavioislima: What convention is that? I never heard of it. All variables that do not start with a value, are undefined by default. 

I think you can just declare the variable or start with the default value that I thought was 'auto' or 'v0'.

electron/new_config.ts
flavioislima: No need since a variable that does not exist or do not have a value will return false. :P

electron/new_config.ts
```diff
@@ -0,0 +1,294 @@
...
+    // Check for version mismatches.
+    // Should never happen for 'auto'.
+    if (version !== configVersion) {
+      console.log(`Config version mismatch! Requested: ${version}. Found: ${configVersion}.`)
+    }
+
+    // Select loader to use.
+    switch (version) {
+    case 'v0':
+      if (GlobalConfig.globalInstance === undefined) {
```
flavioislima, line 81: same here, you can remove the equals undefined. 
The thing here is that if GlobalConfig for some reason does not exist, you will have an error like `there is no globalnstance in undefined` or something like that.
If GlobalConfig can be undefined, try doing something like this:
`if (GlobalConfig?.globalInstance) {`
Then it will run if both are undefined as well.

electron/new_config.ts
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  public getUserInfo() {
+    if (this.isLoggedIn()) {
+      return JSON.parse(readFileSync(userInfo, 'utf-8'))
+    }
+    return { account_id: '', displayName: null }
+  }
+
+  protected abstract getSettings() : Promise<AppSettings>
```
flavioislima, line 204: That's the first time ever I saw anyone using `protected` or `abstract` in Javascript.
I'll need to do some research on how this features is implemented and how it works because, as far as I know, `private` and `public` are experimental, not approved by ECMA so far, but it is 'supported' on a experimental stage in chrome and should work on Electron. BUT protected and abstract I'm not so sure.

I'm just afraid that this can lead to unexpected behavior and more bugs that we can think that might be related to the code but it is actually a JS issue.

I know that you want to make this looks nice and all, using OOP, but unfortunately, JS is not an OOP language, it wasn't created for that and most of the things we have, like using TS to improve this compatibility, are workarounds.

Even classes are no more than objects in JS.

so let's use it with care. In my opinion, it is ok if you want to use classes, but I don't really know if the other features will work as you might think they will, like on C++ or Java, for instance.

flavioislima:
Man, the idea of using OOP might be good, but at the same time the complexity of the project is increasing a lot and then I start worrying about a few things:

1. Maintainability - If you become busy with your job, school stuff, etc or decide to leave the project, this code looks quite difficult to maintain in my opinion. The idea is that Heroic will grow with time, having new stores and all. 
2. Lack of simplicity - If we will need to write that much stuff when adding support to new things, need to worry about classes, private, public, abstract, inheritance methods, etc, those are not things that Javascript (web developers in general), are used to. Even though this is a Desktop App, it uses Web Tech, not Java, not Python and not C++. So this will make other people afraid of sending contributions to the project what can lead it to be abandoned in the future.

Personally, I think the file and functions refactoring that you are doing are really nice and look good. But I think would be good to have a balance between simplicity and being organized. I worked on several big projects that worked really well and were simple at the same time. 

Bottomline, since we are already using TS and having several rules for it and ESLINT, etc. With build and linting checks on PR, etc. I don't think we need to use OOP here to increase even more the complexity. 
The project can be well organized, safe, tested and also simple for new devs to contribute to it 😃

electron/new_config.ts
adityaruplaha: I tested and it looks to be working just fine. These are standard JS constructs, don't know why they're not used as often.

>Even classes are no more than objects in JS.
Not after ES6. And isn't Heroic ES6?

electron/new_config.ts
adityaruplaha: GlobalConfig is always defined. Shouldn't be an issue. I'll remove the unnecessary `undefined` parts.

electron/new_config.ts
adityaruplaha: Will fix.

electron/new_config.ts
adityaruplaha: Alright, will fix.

electron/new_config.ts
flavioislima: Classes are ES6 but not private, public, abstract, protected, etc.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes

flavioislima: Better: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields
first square :D

adityaruplaha:
I appreciate the comments, but honestly, all of this is standard (see ES6) well documented JavaScript, and it will simplify adding thing to the project instead of a mess of functions it currently is. Since ES6, JS has full OOP support, so I think using it where applicable is a smart idea.

As for maintainability, this code is written in way so that people can add things to it easily. It looks complex because of the boilerplate code that manages everything. Which honestly, will probably never need to be touched again. I'm still commenting various parts of the code so that any future dev with basic JS OOP experience can work on it. It's not difficult, all someone needs to know is a Google search away. I really don't know why JS OOP isn't used as often.

> 2\. need to worry about classes, private, public, abstract, inheritance methods,

Precisely, we don't. Once it's in place, the only thing someone needs to change is the content of the function. Copy paste, edit.

electron/new_config.ts
adityaruplaha: As I see, it should work well with Electron v11 or higher. Also access modifiers can be dropped, I just use them since its broadly supported.

adityaruplaha:
Ready for review. A bit of refactoring is left for Game, otherwise I'm done. Nothing major.

All the extra bells and whistles, cleanup, commenting etc. Will be in a follow up PR.

flavioislima (dismissed):
Ok, looks good. Just make sure to test everything before merging.
Would be good to have some tests as well since you have refactored most things, at least the basic file readings would be good to have some basic tests

adityaruplaha:
So, I tested almost everything, looks good. Cleanup is needed, but that comes in Part 2. Also, progress indicators are bugged and unreliable, which is nothing new. Plans are to refactor that eventually.

adityaruplaha:
Tested everything, works, good for merge.

flavioislima (approved):

adityaruplaha:
🚀

PR #244, opened 2021-03-14
refactor: Sort imports, exports, keys, etc alphabetically. Disallow trailing whitespace and dangling commas.

adityaruplaha:
This PR shouldn't change any code functionally.
---

The following are sorted alphabetically:
- [x] Imports
- [x] Module exports
- [x] Variables
- [x] Keys
- [x] Interface fields.

Also, trailing spaces and dangling commas are now disallowed (and autofixed).

### Why?
Consistency + much more readable. Now finding things in code is easier, and I'm working on further refactors. This is just the beginning.

### Do I need to keep this in mind while developing?
No. These are mostly autofixed (except type exports, they need to manually put in the right place I think). This PR also adds `yarn lint-fix` for this purpose, which also runs pre-commit.
**NOTE:** with this, commiting unlinted code is no longer possible. I think that's a good thing, but if you really need to push, use `--no-verify`. But I strongly recommend against it.

### Design Decisions
`export { ... }` at the end of the file is preffered over exporting at definition.

### Why was the inbuilt `sort-keys` not used?
No autofix. That's why. If the extra dependency is deemed not worth it, I'll use `sort-keys` instead or maybe remove this (or reduce it to a warning).

### Great. More conflicts.
I made a way to automatically resolve conflicts as well. Look [here](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/244#issuecomment-800206626).

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -3,43 +3,43 @@ interface About {
   shortDescription: string
 }
 export interface AppSettings {
-  wineVersion: WineProps
-  winePrefix: string
-  otherOptions: string
-  useGameMode: boolean
-  showFps: boolean
-  offlineMode: boolean
+  audioFix: boolean,
+  autoInstallDxvk: boolean,
+  autoSyncSaves: boolean,
+  customWinePaths: Array<string>,
+  darkTrayIcon: boolean,
+  defaultInstallPath: string,
   egsLinkedPath: string
-  savesPath: string
-  autoSyncSaves: boolean
-  exitToTray: boolean
-  launcherArgs: string
-  audioFix: boolean
-  showMangohud: boolean
-  defaultInstallPath: string
-  language: string
-  maxWorkers: number
-  darkTrayIcon: boolean
-  autoInstallDxvk: boolean
-  customWinePaths: Array<string>
+  exitToTray: boolean,
+  language: string,
+  launcherArgs: string,
+  maxWorkers: number,
+  offlineMode: boolean,
+  otherOptions: string,
+  savesPath: string,
+  showFps: boolean,
+  showMangohud: boolean,
+  useGameMode: boolean,
+  winePrefix: string,
+  wineVersion: WineProps
 }
 
 
 export interface ContextType {
-  user: string
-  data: Game[]
+  data: Game[],
+  error: boolean,
   filter: string
-  layout: string
-  refreshing: boolean
-  error: boolean
-  libraryStatus: GameStatus[]
-  gameUpdates: string[]
-  refresh: () => Promise<void>
-  refreshLibrary: () => void
-  handleGameStatus: (game: GameStatus) => Promise<void>
-  handleFilter: (value: string) => void
-  handleSearch: (input: string) => void
-  handleLayout: (value: string) => void
+  gameUpdates: string[],
+  handleFilter: (value: string) => void,
+  handleGameStatus: (game: GameStatus) => Promise<void>,
+  handleLayout: (value: string) => void,
+  handleSearch: (input: string) => void,
+  layout: string,
+  libraryStatus: GameStatus[],
+  refresh: () => Promise<void>,
+  refreshLibrary: () => void,
+  refreshing: boolean,
+  user: string
 }
 
 interface ExtraInfo {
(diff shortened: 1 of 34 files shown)


ianwijma:
Just me thought on the matter of sorting:

It looks like this also sorts any object keys, where I'm personally not 100% a fan of, Sometimes you want to have items in a certain order. an unrelated example could be an object with {firstname, lastname, gender} which will be sorted to `{firstname, gender, lastname}` Which is annoying to read.

Same story with class fields and variables. I sometimes group stuff together in a certain order. Which is not possible with this PR.

adityaruplaha:
> Sometimes you want to have items in a certain order. an unrelated example could be an object with {firstname, lastname, gender} which will be sorted to `{firstname, gender, lastname}` Which is annoying to read.
> 
> Same story with class fields and variables. I sometimes group stuff together in a certain order. Which is not possible with this PR.

In the vast majority of cases though, we don't. Most orders I saw in the code are essentially random. Also alphabetical order means its definitely going to be easy to locate. 

In case we have such an instance, we can always locally disable. I think consistency should be enforced, and when not, we should be explicit about it. That way, new contributors can find their way around code more easily. I'm gonna follow this up with further cleanup, of course. Just sorting doesn't make any sense.

src/types.ts
```diff
@@ -78,20 +79,19 @@ export interface GameStatus {
...
-  eta: string
 }
 export interface InstalledInfo {
   executable: string | null
-  version: string | null
+  install_path: string | null,
   install_size: string | null
-  install_path: string | null
-  is_dlc: boolean | null
+  is_dlc: boolean | null,
```
flavioislima, line 93: Have noticed these trailing commas here and there.
It is weird to have them on interfaces though so I did set up my VSCode to remove trailing comma when formating and I think there is also a linting/prettier rule to remove them as well.

flavioislima (changes requested):
Yeah, looking at the code after sorting the keys, it looks ok. Some places I miss following the context structure, for instance:
AltWine
SetAltwine

Ordering, they stay afar from each other. Personally I like to have the variable and the action to change it. But as far as I can see, this happened just in a few places and on the other ones they look nice well organized.

So far, the only thing you should take a look is the trailing commas 👍🏽

adityaruplaha:
> Yeah, looking at the code after sorting the keys, it looks ok. Some places I miss following the context structure, for instance:
> AltWine
> SetAltwine

Yeah that's king of a bummer, but honestly all the setters stay together and in the same order so I don't really mind.



Also I'm curious, you requested changes but didn't say what you wanted changed. Misclick?

> only thing you should take a look is the trailing commas

Looking into it now.

src/types.ts
adityaruplaha: I did as well, kind of weird.

adityaruplaha: Fixed.

flavioislima (approved):

adityaruplaha:
Ready for merge?

adityaruplaha:
Resolving conflicts with this PR:
```
git fetch
git rebase upstream/main
# a conflict happens
git checkout --ours
yarn lint-fix
git add -A .
git rebase --continue
# finished! :D
```

adityaruplaha:
Aw hell, the force push invalidated the review. :(

adityaruplaha:
Resolving conflicts was pretty difficult (since `yarn lint-fix` wasn't introduced until a later commit). Hope to merge this soon.

flavioislima:
> Resolving conflicts was pretty difficult (since `yarn lint-fix` wasn't introduced until a later commit). Hope to merge this soon.

It might be good to wait to merge this until we release 1.5.6. 
then we can start adding new features with this refactoring in mind. So, after 1.5.6 release, we merge this and then start working on something else.

flavioislima:
I'm also doing refactoring on the React and CSS part, and both would be good to wait until 1.5.6.

adityaruplaha:
I'll resolve conflicts after 1.5.6 is released.

adityaruplaha:
Conflicts resolved. Merge please?

flavioislima (approved):
👍🏽

PR #245, opened 2021-03-14
Add GitHub Actions.

adityaruplaha:
Currently runs on PRs and pushes to `main`.
Builds and lints code.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,28 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install modules.
+        run: yarn
+      - name: Build 'src'.
+        run: yarn build
+      - name: Build 'electron'.
+        run: yarn build-electron
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install modules.
+        run: yarn
+      - name: Lint code.
+        run: yarn lint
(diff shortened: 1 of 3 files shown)


adityaruplaha:
Looks good to merge. There's not much in this, just enforcing good code.

.github/workflows/main.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install modules.
+        run: yarn
+      - name: Build 'src'.
+        run: yarn build
+      - name: Build 'electron'.
+        run: yarn build-electron
+  lint:
+    runs-on: ubuntu-latest
```
flavioislima, line 22: This job could run on the previous docker. Also, why do we need to build the app first to run the linting?
I think you could just create the docker machine, run yarn and then run the linting and the prettier as well. :)

adityaruplaha: Okay I'm gonna split it into 2 different job files entirely then. That is what you want right?

flavioislima: No. I dont think you need to actions for that. You just need to remove the first one.

flavioislima: I think this should work:

```
name: Linting

on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]
  workflow_dispatch:

jobs:
  lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: Install modules.
        run: yarn
      - name: Lint code.
        run: yarn lint
```

adityaruplaha: So no need to run build? I thought it would be nice to check for that as well.

ianwijma (approved):
Actions look good.

adityaruplaha:
@flavioislima have a look then, I'm ready to merge.

adityaruplaha:
Talked to him on Discord, looks good, merging.

PR #250, opened 2021-03-17
Frontend Refactoring + CSS Split

flavioislima:
- [x] Split CSS per component
- [x] Add components to specific folders to have a better consistency
- [x] Remove unused CSS


D3SOX (dismissed):
Not entirely sure why you requested my review but I had a quick look and clicked around and it looks good.

flavioislima:
> Not entirely sure why you requested my review but I had a quick look and clicked around and it looks good.

Thanks!
I'll ask for review from most devs and it is good that everyone could take a look, sometimes when developing we can let some error pass, etc. Or forget something. :)

src/index.tsx
```diff
@@ -8,7 +8,7 @@ import LanguageDetector from 'i18next-browser-languagedetector'
 import './index.css'
 import App from './App'
 import GlobalState from './state/GlobalState'
-import UpdateComponent from './components/UI/UpdateComponent'
+import UpdateComponent from 'src/components/UI/UpdateComponent'
```
adityaruplaha, line 11: IMO this is inconsistent since we use './' everywhere else.

src/screens/Game/GamePage/index.tsx
```diff
@@ -21,13 +19,14 @@ import {
...
-import InfoBox from '../UI/InfoBox'
-import UpdateComponent from '../UI/UpdateComponent'
-import GamesSubmenu from './GamesSubmenu'
+} from 'src/helpers'
+import ContextProvider from 'src/state/ContextProvider'
+import { AppSettings, Game, GameStatus, InstallProgress } from 'src/types'
+import Header from 'src/components/UI/Header'
+import InfoBox from 'src/components/UI/InfoBox'
+import UpdateComponent from 'src/components/UI/UpdateComponent'
+import GamesSubmenu from '../GameSubMenu'
```
adityaruplaha, line 28: Another one you missed.

adityaruplaha (commented):
Few minor changes are required. Also there's a lot of absolute units in the CSS that won't scale well, but that's for another PR.

src/screens/Game/GamePage/index.tsx
flavioislima: This one is different since this component is in the folder above and not on src

src/components/Navbar/components/SearchBar/index.css
```diff
@@ -0,0 +1,35 @@
+.SearchBar {
```
ianwijma, line 1: I personally don't think its great to have repetitive folders inside each other. `components/Navbar/components` where the last `components` folder could just be omitted IMO.

Then you just straight see the index.tsx, and its child components. without needing to open an additional folder.

ianwijma: This also goes for the other files in this PR that have the same structure. :) Again, this is just an idea, not sure how the others think about it.

src/components/UI/LanguageSelector/index.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import './index.css'
```
ianwijma, line 2: The index.css does not contain any styling.

ianwijma (commented):
Looks good overall, only not 100% sold on having components outside the components dir.

src/components/Navbar/components/SearchBar/index.css
flavioislima: I personally like this pattern because we can then add the `tests` folders for each component. Also if we have some need of adding another folder with another thing, like constants, another helper file with some functions that can be shared across those components, we do not mix them. So `components` is one folder, but then we can have other things. So, when the project grows and we start supporting other stores, with different behaviors, this will be necessary 😄

src/components/UI/LanguageSelector/index.tsx
flavioislima: Good catch!

.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["amatiasq.sort-imports"]
```
adityaruplaha, line 2: Well I'm not sure if you noticed my message on Discord but this is what I use currently since it does exports as well, and respects ESLint rules. Plus removes unnecessary imports.

https://marketplace.visualstudio.com/items?itemName=dozerg.tsimportsorter#vscode-settings

adityaruplaha (commented):
Code looks okay, I don't know about the CSS. Overall looks good.
However, the extension stuff doesn't look right. Plus some import orders for changed, which don't look quite right to me? I'll check again on PC.

.vscode/extensions.json
flavioislima: Ignore this, I'll remove it. I'm using the one you have recommended. Which files have wrong import order?

flavioislima: Updated the recommended extensions, take a look when you have time. Also I could not find any file with problems of sort importing. I'll try to find a prettier rule for that and maybe we can fix all imports with one command

flavioislima: I guess we have already a rule on eslint for sort imports, but it conflicts with the extension on vscode. So if run yarn lint-fix, it will sort the imports. If I save the file, the imports will be sorted by the extension, with different criteria. Maybe it is better to not use the extension and do a lint-fix before pushing the commit.

adityaruplaha: Weird. I'll have a look into it later. Drop the extension for now I guess. Let's get this merged.

adityaruplaha (approved):

PR #258, opened 2021-03-21
Fix: Not starting on new installations

### LINKED ISSUE #256: [Bug] Starts but unable to open a window
**Describe the bug**
Heroic starts, spawns a process but doesnt open a window.

**To Reproduce**
1. Download the deb or appimage from the release section
2. chmod it, (install), run it
3. Nothing except the process spawning happens

**Expected behavior**
The heroic launcher window opens

**Screenshots**

 - OS: Ubuntu 20,04 & Kubuntu 20.04
 - Version 1.5.6 & 1.5.5 (didnt test the others)
-  Game: - 
-  Wine/Proton version: -

**Additional context**
Tried it on an clean install of kubuntu and my main ubuntu. When starting via cli it starts but immediately end itself after.

**Image of the process but no window:**
image

**Image of it starting and ending immediately on my main system**
(ignore the gtk errors, that just my system theme that shouldnt impact it)
image

flavioislima:
Fixed #256 
The bug was introduced with the latest fix I sent related with the custom wine 🤦🏽 . Shit happens.
It happens not only in ubuntu, but all distros since Heroic starts a never-ending loop when looking for custom wine on new installations.
Added a check to skip looking for custom wine/proton when there is no config file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "heroic",
-  "version": "1.5.6",
+  "version": "1.5.7",
   "private": true,
   "main": "public/main.js",
   "homepage": "./",
@@ -132,7 +132,7 @@
       "post-checkout": "yarn",
       "post-merge": "yarn",
       "post-rebase": "yarn",
-      "pre-push": "yarn lint && pretty-quick --staged lint-staged",
+      "pre-push": "yarn lint-fix && pretty-quick --staged lint-staged",
       "pre-commit": "yarn lint-fix"
     }
   },
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
       "post-checkout": "yarn",
       "post-merge": "yarn",
       "post-rebase": "yarn",
-      "pre-push": "yarn lint-fix && pretty-quick --staged lint-staged",
+      "pre-push": "yarn lint && pretty-quick --staged lint-staged",
       "pre-commit": "yarn lint-fix"
     }
   },
--- a/src/components/Settings/index.tsx
+++ b/src/components/Settings/index.tsx
@@ -1,25 +1,12 @@
-import React, {
-  useEffect,
-  useState
-} from 'react'
+import React, { useEffect, useState } from 'react'
 
 import { IpcRenderer } from 'electron'
 import { useTranslation } from 'react-i18next'
-import {
-  NavLink,
-  useLocation,
-  useParams
-} from 'react-router-dom'
+import { NavLink, useLocation, useParams } from 'react-router-dom'
 
-import {
-  getGameInfo,
-  writeConfig
-} from '../../helper'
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -132,7 +132,7 @@
       "post-checkout": "yarn",
       "post-merge": "yarn",
       "post-rebase": "yarn",
-      "pre-push": "yarn lint && pretty-quick --staged lint-staged",
+      "pre-push": "yarn lint-fix && pretty-quick --staged lint-staged",
```
adityaruplaha, line 135: This should be reverted as fixes will not be committed and will cause checks to fail. I didn't change this on purpose.

adityaruplaha (changes requested):
One major issue not actually related to this PR, not sure why it was changed.

package.json
flavioislima: good catch, I didn't want to push this, was just testing 😂

flavioislima: fixed

src/components/Settings/index.tsx
```diff
@@ -1,24 +1,11 @@
-import React, {
-  useEffect,
-  useState
-} from 'react'
+import React, { useEffect, useState } from 'react'
```
adityaruplaha, line 1: Yeah we need a standardised styling guideline for these...

src/components/Settings/index.tsx
```diff
@@ -62,7 +49,7 @@ function Settings() {
   const [maxWorkers, setMaxWorkers] = useState(0)
   const [egsPath, setEgsPath] = useState(egsLinkedPath)
   const [language, setLanguage] = useState(
-    () => storage.getItem('language') || ''
+    () => storage.getItem('language') || 'en'
```
adityaruplaha, line 52: Good catch, missed this one before.

src/components/Settings/index.tsx
```diff
@@ -72,7 +59,11 @@ function Settings() {
     setOn: setUseGameMode
   } = useToggle(false)
   const { on: showFps, toggle: toggleFps, setOn: setShowFps } = useToggle(false)
-  const { on: offlineMode, toggle: toggleOffline, setOn: setShowOffline } = useToggle(false)
+  const {
```
adityaruplaha, line 62: Gonna need standardised style guidelines on this one too....

adityaruplaha (approved):
PR looks good, left some miscellaneous comments that need to be addressed in upcoming PRs.

src/components/Settings/index.tsx
flavioislima: Yeah, the thing is that I'm not sure If the prettier config is working for everyone the same way.
Every time I save a file, something moves. I set up my vscode to format on save, but using the prettier rules we have on the prettierrc file.

Issue #246, opened 2021-03-14
Better MacOS support: Native games, etc.


LukeVader-IV:
**Is your feature request related to a problem? Please describe.**
Wine already has MacOS support and currently there's no good option to play games like Rocket league on the platform

**Describe the solution you'd like**
A port of this application for MacOS users

**Additional context**
I have no clue If this would be hard to achieve but I think it would be appreciated by the MacOS crowd.
PS: I love the project, makes gaming on Linux that much better, keep it up (:

adityaruplaha:
You'll be happy to hear that Heroic might just already work out of the box. We would greatly appreciate it if you could run it on macOS and let us know where things start going wrong. You might need to build it yourself to see the app crash dialogs.

flavioislima:
Hi. There is no way to support mac right now since legendary doesn't work on mac. And heroic depend on it to login, download the games and launch them.

entaromia:
@flavioislima I managed to run heroic on mac os. Nearly all features are working after some fixes.

![Heroic on mac](https://user-images.githubusercontent.com/30384045/111902666-9ada9d80-8a4f-11eb-98b8-1e1ba7f395d0.png)

Only cancelling game installation isn't working for now, it continues to download on background.

flavioislima:
Nice, and for some reason Legendary is working. 
So you can login and download games?
So, I know legendary can download the Mac versions of the games that supports it.
The question is: download the native games when there is any or download the windows version and run with wine, even though the compatibility is not that good.

@entaromia I'll reopen this issue then.

entaromia:
Yes, even there are no references to mac on legendary readme, it just works. I can login, download games, launch them using wine etc. Since now heroic works on mac, i will try to implement downloading native versions on mac today.

flavioislima:
> Yes, even there are no references to mac on legendary readme, it just works. I can login, download games, launch them using wine etc. Since now heroic works on mac, i will try to implement downloading native versions on mac today.

Nice to hear! 
I know there is a flag to pass the OS version that you want to download. Would be good to have a filter on heroic to show only native versions.

adityaruplaha:
>legendary doesn't work on mac

It does. No official support maybe, but the code does seem to support Mac.
Sorry I didn't notice this issue got closed.

adityaruplaha:
> Only cancelling game installation isn't working for now

I would've into it, but I don't have a Mac to test against :/

[...]

PR #260, opened 2021-03-21
Add support for Mac OS

entaromia:
This pull request adds support for mac os requested at #246.

Probably most of the questions has answers in commit messages, but let me explain some more:

**Where is the support for downloading native version of games?**
As i said before, it requires much more changes than I expected, so I will split them into two pull requests.
I will probably do it after all refactorings are done.

**What is the easiest way to distribute heroic for mac?**
Run `yarn dist-mac dir` then zip the output (`heroic.app`)

**What's broken?**
Canceling game installation stops every installation going, not only the cancelled game.
Legendary owner does not want to support mac and it is only solution I can find now. Still better than not being able to stop it at all in my opinion.


flavioislima:
Great job, man! I'll take a look carefully on this, but looks promising! 😁

entaromia:
Thanks!

README.md
```diff
@@ -78,4 +78,5 @@ To be able to run you will need to have NodeJS installed locally and follow the
 - On the project folder run `yarn install`.
 - Run `yarn build` to build the React Assets.
 - Run `yarn start`.
-- To build the binaries run `yarn dist {package to create}`.
+- To build the binaries for linux, run `yarn dist {package to create}`.
+- To build the binaries for mac os, run `yarn dist-mac {package to create}`.
```
adityaruplaha, line 82: I would prefer Linux, and MacOS (or OSX).

adityaruplaha (dismissed):
Looks good code-wise. Can't test.
One minor nit.

README.md
entaromia: Should I change `dist` to `dist-linux` and `dist-mac` to `dist-osx`?
Or leave `dist` as is and only change `dist-mac` to `dist-osx`?

adityaruplaha: Oh I meant in the documentation. Since Linux and macOS are proper nouns, the grammar Nazi in me gets triggered when it sees their capitalisation changed. :P

entaromia: It should be good now

electron/main.ts
```diff
@@ -355,14 +357,14 @@ ipcMain.handle('install', async (event, args) => {
...
-  const logPath = `${heroicGamesConfigPath}${game}.log`
+  const logPath = `"${heroicGamesConfigPath}${game}.log"`
   let command = `${legendaryBin} install ${game} --base-path '${path}' ${workers} -y &> ${logPath}`
   if (path === 'default') {
     command = `${legendaryBin} install ${game} --base-path ${defaultInstallPath} ${workers} -y |& tee ${logPath}`
   }
   console.log(`Installing ${game} with:`, command)
   try {
-    await execAsync(command, { shell: '/bin/bash' })
+    await execAsync(command, { shell: shell })
```
flavioislima, line 367: The only thing we need to check if this breaks HEroic on Ubuntu, because it requires that everything runs on bash. So, if this breaks, we will need create a global variable like: `const isMac = os.platform() === 'darwin'` and use it to choose the shell.

entaromia: Heroic on linux continues to use bash, it only changes to zsh if running on mac as you said.
Check 'shell' variable at electron/utils.ts

package.json
```diff
@@ -21,7 +21,8 @@
       "node_modules/**/*"
     ],
     "asarUnpack": [
-      "build/bin/legendary",
+      "build/bin/linux/legendary",
+      "build/bin/darwin/legendary",
```
flavioislima, line 25: Since the same binary works for both, why do we need to duplicate it?

entaromia: It's not the same binary actually, linux binary does not even launch on mac. I built mac one from legendary source using pyinstaller

flavioislima: Ah, I didn't know that, I guess I missed this part on your explanation. 
Nice then. :D

flavioislima (approved):

electron/utils.ts
```diff
@@ -25,13 +25,13 @@ const { showErrorBox, showMessageBox } = dialog
 
 const home = homedir()
 const legendaryConfigPath = `${home}/.config/legendary`
-const heroicFolder = `${home}/.config/heroic/`
+const heroicFolder = process.platform === 'darwin' ? `${home}/Library/Application Support/heroic/` : `${home}/.config/heroic/`
```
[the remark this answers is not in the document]
entaromia: Well if someone has interest in porting to windows, the correct path is {user}/AppData/Local.
But there are many things will also not work on windows such as curl, mv, tar commands

electron/utils.ts
```diff
@@ -25,13 +25,13 @@ const { showErrorBox, showMessageBox } = dialog
...
 const heroicConfigPath = `${heroicFolder}config.json`
 const heroicGamesConfigPath = `${heroicFolder}GamesConfig/`
 const heroicToolsPath = `${heroicFolder}tools`
 const userInfo = `${legendaryConfigPath}/user.json`
 const heroicInstallPath = `${home}/Games/Heroic`
-const legendaryBin = fixPathForAsarUnpack(join(__dirname, '/bin/legendary'))
+const legendaryBin = fixPathForAsarUnpack(join(__dirname, '/bin/', process.platform, '/legendary'))
```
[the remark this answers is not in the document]
entaromia: I'm also thinking that we should not include mac binary on linux build, or linux binary on mac builds. (Also windows maybe later on.) 
So we can reduce app size a little bit.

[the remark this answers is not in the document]
flavioislima: I think with automation we can delete the unnecessary files. So, locally we can have all 3, and on the build script, we delete the others, or maybe instead of delete, just copy the necessary one.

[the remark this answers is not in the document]
entaromia: That's a great idea, we can implement something like that. Moving all binaries to one folder, then copying only required one to public/bin folder on github actions should do it

adityaruplaha (approved):

flavioislima:
@entaromia all good from my side if you want to merge.

entaromia:
Absolutely, you can merge

PR #269, opened 2021-03-23
Remove Prettier.

adityaruplaha:
I found that prettier was (still) not set up properly, and was conflicting with some ESLint rules (I couldn't figure out the root cause). I found one of them in #263. 

Also, all the rules we defined in Prettier were available and autofixable in ESLint anyways, so I decided that it was more hassle than it was worth, and removed it. It will also simplify our pre-commit and pre-push workflows.

This is just a proposal so that we can discuss the merits and demerits of this.

----

Functionally, only a few tabs got shifted around and the dangling commas got removed.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,8 +7,7 @@
   "extends": [
     "eslint:recommended",
     "plugin:react/recommended",
-    "plugin:@typescript-eslint/recommended",
-    "prettier"
+    "plugin:@typescript-eslint/recommended"
   ],
   "root": true,
   "parser": "@typescript-eslint/parser",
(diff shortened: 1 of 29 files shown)


adityaruplaha:
Any comments on this?

flavioislima:
> 
> 
> Any comments on this?

I'll try to take a look later.

flavioislima:
Yeah, I think we might not need prettier then. Like you said, we are using eslint to fix most of the things already.

flavioislima (dismissed):

flavioislima (approved):
👍

PR #286, opened 2021-03-31
feat: Add initial support for Unreal Marketplace.

### LINKED ISSUE #257: Unreal Engine marketplace integration. Requirements provided.
I wrote a blog which can help you bring the same functionality in the Heroic launcher (legendary back-end does support the unreal engine marketplace) however a directory in the wine prefix needs to be created inside Documents/Unreal Projects and a minimal .uproject file (a simple text file must be added to the wine prefix)

You can see the directory structure required, in my blog https://blog.gamedev.tv/ue4-marketplace-linux/
In my blog EGS requires .uproject file inside a project folder which should be inside the Unreal Projects directory. .uproject file is just a text file and its minimal content is also provided in my blog. Heroic can simply create these when someone installs a marketplace asset.

ALSO/ OR

Better option for the best Linux native expirience will be to detect Documents/Unreal Projects  directory as thats the default on Unreal Engine Linux build. And detect projects and then install assets inside the native Linux projects directory. This should be possible on Heroic as it can see Linux directories unlike EGS inside wine prefixes. as in the previous method we had to install assets inside wine(fake project as told you above) and copy paste it into our native Linux Projects.

Kjell291:
Added initial support for the unreal engine marketplace as requested in #257.
There still is much room for improvement, but it's working.
To install a package one should select the project folder located in ~/Documents/Unreal Projects.

Fixes #257.

Heroic01
Heroic02
Heroic03
Heroic04
Heroic05


flavioislima (commented):
Good work, man! I think this feature will be really useful for devs in general and I received this request since Heroic's first release but I had no idea on how to implement it!

codewise looks good, I just think that we might add another menu or switch for the market place since it is something specific for game devs.

I was thinking that we can have a submenu under Library with a link to show only the items under market place. What do you think? We can discuss that and ask for other devs review on our discord development channel.

electron/legendary_utils/library.ts
```diff
@@ -56,85 +56,161 @@ export async function getLegendaryConfig(file: string): Promise<unknown> {
...
-            info,
-            install_path,
-            install_size: convertedSize,
-            isInstalled,
-            is_dlc,
-            namespace,
-            saveFolder,
-            title,
-            version
+            const art_cover = gameBox ? gameBox.url : null
```
flavioislima, line 168: I noticed that this part of the code is repeated almost entirely, we could maybe have a helper function to that, or check the namespace inside it. You could have a variable called: `isUEAsset`, for instance, and use it where you need to check different things.

src/screens/Game/GamePage/index.tsx
```diff
@@ -193,6 +195,9 @@ export default function GamePage(): JSX.Element | null {
                   <div className="title">{title}</div>
                   <div className="infoWrapper">
                     <div className="developer">{developer}</div>
+                    {namespace == 'ue' && (
```
flavioislima, line 198: Maybe use a variable for this check to make It more readable since you're using in other places as well :)

flavioislima: If you don't want for the asset to have settings, maybe its better to disable the settings button. Make more sense I guess.

src/screens/Game/GamePage/index.tsx
```diff
@@ -478,7 +483,7 @@ export default function GamePage(): JSX.Element | null {
         return refresh()
       }
 
-      if (installPath === 'default') {
+      if (installPath === 'default' && gameInfo.namespace != 'ue') {
```
flavioislima, line 486: Same here. Maybe is better to just remove the import and install on default path from the option list.

jatin-cbs:
Thanks for this ! The things that I think are currently missing -

1. We can download assets to the project however Marketplace also has whole project samples which simply gets installed in "Unreal Projects" directory. like this one 

https://www.unrealengine.com/marketplace/en-US/product/cesium-samples

2. Content Details - they are just important when comes to marketplace.    But overall great work !!

adityaruplaha:
This will cause severe conflicts with the Electron refactor as this entire module is being overhauled. So I strongly advise against merging this right now.
Also the OOP approach should clean up this code further.

So please, hold your merge for a week or two. I'll try to review it in a few hours.

If after the refactor the code looks too complex to add this (which it shouldn't  really), I'll do it myself, and add @Kjell291 as co-author. Sounds good? Sorry about stalling development.

Kjell291:
Thanks for all the feedback.
I will continue working on this.

> codewise looks good, I just think that we might add another menu or switch for the market place since it is something specific for game devs.
> 
> I was thinking that we can have a submenu under Library with a link to show only the items under market place. What do you think? We can discuss that and ask for other devs review on our discord development channel.

Another menu would be a good idea with the option to filter for specific Engine versions.

ianwijma (dismissed):
No complains from my end beside the points already made

adityaruplaha (approved):
LGTM. If the tests go fine, I'll merge this.

PR #294, opened 2021-04-09
electron: Refactor Electron. [Part 2]

adityaruplaha:
This continues where #242 left off. 

## Changes

- User has its own interface.
- Frontend calls to Legendary are removed.
- Everything has JSDoc comments explaining.
- A couple of **MAJOR** bugfixes.
- `syncSaves` integrated into `LegendaryGame`.
- Reordered stuff in main.ts.
- A fix for Malayalam.

Please check the checklist in the original PR as well.


electron/games.ts
```diff
@@ -217,8 +217,8 @@ class LegendaryGame {
     let prefix = `--wine-prefix '${fixedWinePrefix.replaceAll("'", '')}'`
 
     const isProton =
-      wineVersion.name.startsWith('Proton') ||
-      wineVersion.name.startsWith('Steam')
+      wineVersion.name.indexOf('Proton') >= 0 ||
```
flavioislima, line 220: But when this changes? Before, all Protons were Proton, Steam was removed. I missed this on your latest PR then. I dont think we should use Steam and Lutris, only Wine and Proton, like before.

adityaruplaha: Oh? I didn't know. Yeah you missed it probably. I split it up to "Steam Proton", "Lutris Wine" etc.

flavioislima: It was like this at first and then I changed it to be more simple to understand. :)
Because then, now, you need to do 2 checks for isProton instead of one.
For the user also is more information.

adityaruplaha: I think indicating where the Wine version comes from may be a good idea. Not sure though, I'm okay with changing it back.

flavioislima: Please change it back. It is also easier for us to maintain. I think Proton and Wine are the best indications we will have.

electron/games.ts
```diff
@@ -101,16 +136,31 @@ class LegendaryGame {
...
+  /**
+   * Helper for `checkGameUpdates().contains(this.appName)`
+   *
+   * @returns If game has an update.
+   */
   public async hasUpdate() {
-    return (await Library.get().listUpdateableGames()).find((app_name) => {
-      return app_name == this.appName
-    }) !== undefined
+    return (await Library.get().listUpdateableGames()).indexOf(this.appName) >= 0
```
flavioislima, line 154: indexOf could be replaced with a includes here. More readable and uses ES6.
so `.includes(this.appName)`

electron/legendary_utils/library.ts
```diff
@@ -99,17 +129,37 @@ class Library {
...
 
+  /**
+   * Update all updateable games.
+   * Uses `listUpdateableGames` along with `LegendaryGame.update`
+   *
+   * @returns Array of results of `Game.update`.
+   */
   public async updateAllGames() {
-    return (await this.listUpdateableGames()).map(LegendaryGame.get).map(
+    return Promise.allSettled((await this.listUpdateableGames()).map(LegendaryGame.get).map(
```
flavioislima, line 139: Nice! Just need a frontend integration for that then?

electron/legendary_utils/user.ts
```diff
@@ -0,0 +1,33 @@
...
+  userInfo
+} from '../constants'
+
+export class User {
+  public static login(sid: string) {
+    exec(`${legendaryBin} auth --sid ${sid}`)
+  }
+
+  public static logout() {
+    exec(`${legendaryBin} auth --delete`)
```
flavioislima, line 19: I think to avoid a racing condition here. The ideal would be to use `await execAsync` on the first one.

adityaruplaha: Alright.

electron/legendary_utils/library.ts
adityaruplaha: Yep, this isn't used anywhere as of now.

electron/games.ts
adityaruplaha: I forgot that function existed. Jeez I need coffee.

electron/main.ts
```diff
@@ -270,17 +251,147 @@ ipcMain.on('unlock', () => {
...
+    console.log(`Which idiot requested '${config_class}' using readConfig?`)
+    return {}
+  }
+})
+
+ipcMain.handle('requestSettings', async (event, appName) => {
+  if (appName === 'default') {
+    return await GlobalConfig.get().config
+  }
+  return await GameConfig.get(appName).getSettings()
```
flavioislima, line 377: Can we have something like `Game.get(appName).settings()` ?
Since we are requesting a settings with `get` already? 
So we could have `get.settings()` and `set.settings(settingsObj)`

adityaruplaha: I will change that to `.config`. getSettings shouldn't be called here.

electron/main.ts
```diff
@@ -270,17 +251,147 @@ ipcMain.on('unlock', () => {
...
+})
+
+ipcMain.handle('writeConfig', (event, [appName, config]) => {
+  if (appName === 'default') {
+    GlobalConfig.get().config = config
```
flavioislima, line 382: Like in this case, since we are writing the config. Should we not use `set` instead of `get`?

src/components/Navbar/components/UserSelector/index.tsx
```diff
@@ -18,9 +18,8 @@ export default function UserSelector() {
 
   const { user, refresh, refreshLibrary } = React.useContext(ContextProvider)
   const handleLogout = async () => {
-    if (confirm('are you sure?')) {
-      await legendary(`auth --delete`)
-      await legendary(`cleanup`)
+    if (confirm('Are you sure?')) {
```
flavioislima, line 21: Hmm, one label we missed translating.
can you add the label here with this text as default and then we can remember translating it?

flavioislima (changes requested):
Yeah, just added a few comments but otherwise looks nice! 👍🏽

electron/main.ts
adityaruplaha: We're get getting the settings, we're getting the Config object.

electron/main.ts
adityaruplaha: > requesting a settings with `get` already

No we're not, we're getting the Config handler.

electron/main.ts
flavioislima: I don't 'get' it. 🤔

src/components/Navbar/components/UserSelector/index.tsx
adityaruplaha: Sure.

electron/games.ts
adityaruplaha: Done.

electron/main.ts
adityaruplaha: Yeah `.get()` returns the global config handler. Hovering over it will tell you that on VSCode. `.config` is the actual settings object.

electron/main.ts
```diff
@@ -291,16 +292,16 @@ ipcMain.on('getLog', (event, appName) =>
 
 ipcMain.on('removeFolder', async (e, [path, folderName]) => {
   if (path === 'default') {
-    const defaultInstallPath = (await GlobalConfig.get().getSettings()).defaultInstallPath.replaceAll("'", '')
+    const defaultInstallPath = (await GlobalConfig.get()).config.defaultInstallPath.replaceAll("'", '')
     const folderToDelete = `${defaultInstallPath}/${folderName}`
     return setTimeout(() => {
-      exec(`rm -Rf ${folderToDelete}`)
+      rmdirSync(folderToDelete, {recursive: true})
```
flavioislima, line 298: Have you tested this approach? It was my first try and it didn't work, but maybe I missed the recursive option, I don't remember now.

flavioislima (dismissed):

electron/main.ts
adityaruplaha: According to the spec, it should behave exactly as `rm -R`. So I think we're good here.

flavioislima (dismissed):

electron/main.ts
adityaruplaha: Nope, I tested, it doesn't work. Nice catch.

adityaruplaha: Great. `exec` doesn't work as well. I think there's something wrong with my setup. Anyway, I trust the spec, and we can fix it before release if we find out the root cause.

flavioislima: have you tried to check if the folder to delete is coming correctly?
Because the exec should work since it worked before.

electron/game_config.ts
```diff
@@ -178,7 +178,7 @@ class GameConfigV0 extends GameConfig {
   }
 
   public async getSettings(): Promise<GameSettings> {
-    const settings = JSON.parse(readFileSync(this.path, 'utf-8'))
+    const settings = existsSync(this.path) ? JSON.parse(readFileSync(this.path, 'utf-8')) : {}
```
flavioislima, line 181: So if there is no game settings file we just return an empty object? 
Before, if there is no game settings file we returned the default settings. Otherwise, this might break the launch of games and the loading of the settings.

electron/legendary_utils/library.ts
```diff
@@ -319,7 +319,7 @@ class Library {
       is_ue_plugin,
       is_ue_project,
       namespace,
-      save_folder: saveFolder,
+      save_folder: saveFolder || '',
```
flavioislima, line 322: `saveFolder `already returns an empty string look at line 259.

src/helpers/index.ts
```diff
@@ -146,6 +146,11 @@ async function fixSaveFolder(
   prefix: string,
   isProton: boolean
 ) {
+
+  if (!folder) {
```
flavioislima, line 150: Why do we need this now?

electron/game_config.ts
adityaruplaha: Yeah this needs fixing, my last commit is really quite broken.

adityaruplaha:
Reverting last commit and applyinmg suggested fixes.

electron/main.ts
adityaruplaha: It isn't, I'm investigating.

adityaruplaha: Fixed as discussed on Discord.

flavioislima (approved):
Looks good. Merge it if you cant find any other bugs 👍🏽

PR #301, opened 2021-04-12
unreal marketplace improvements

Kjell291:
I added a seperate category and filters for the unreal engine marketplace.
Heroic


src/components/UI/Header/index.tsx
```diff
@@ -74,12 +99,56 @@ export default function Header({
...
+              className={filter === 'project' ? 'selected' : ''}
+              onClick={() => handleFilter('project')}
+            >
+              {t('Projects')}
+            </span>
+            <select
+              className={filter.includes('UE_') ? 'selected' : ''}
+              onChange={(event) => handleFilter(event.target.value)}
+            >
+              <option value='UE_4.26'>4.26</option>
```
flavioislima, line 135: I guess one improvement would be having this list of versions on an array and then just map it here. Because in the future can be easier to add/remove versions.

Kjell291:
Sorry that this takes so long, I didnt have much time.
I will try to do this during the next few days.

electron/legendary_utils/library.ts
```diff
@@ -209,10 +209,7 @@ class Library {
       dlcItemList,
       releaseInfo,
       categories,
```
Kjell291, line 211: important bugfix, heroic does not start when the library includes unreal assets, because they do not have a cloudsavefolder

Kjell291:
I added an array for unreal versions, I hope this is what you had in mind

Also Heroic 1.5.8 fails to start when the library includes unreal assets, but I fixed it

electron/legendary_utils/library.ts
```diff
@@ -209,10 +209,7 @@ class Library {
...
-      customAttributes : {
-        CloudSaveFolder,
-        FolderName
-      }
+      customAttributes
```
Kjell291, line 212: Important bugfix, heroic does not start, because unreal assets do not have a cloudsavefolder

flavioislima:
@Kjell291 whats the state of this PR?
I'm asking since we have issues on latest release, as you can see here on #367 and then we could think about merge it, or use the part of the code that improves the UI to the next release
Because what I did on 1.6.1 was only to hide the filter if no Asset is found, but it seems that it didn't work as it should 🤦🏽

electron/legendary_utils/library.ts
flavioislima: Yeah, we noticed this when we released 1.5.8. It was fixed on 1.5.9

flavioislima:
Fixed the conflicts :)

Kjell291:
@flavioislima 
Everything should work, I think it is safe to merge it.

flavioislima:
> @flavioislima
> Everything should work, I think it is safe to merge it.

Can you test again, because I asked for @warriormaster to test and he said that the Unreal Marketplace is not showing for him and I wanted to release this on 1.7.0 but since I could not make it work since I have no way to test it, I couldn't.

Kjell291:
@flavioislima 
I fixed it, although I do not know why why changing the name from "Unreal Marketplace" to "Unreal" helped.

flavioislima:
> @flavioislima 
> I fixed it, although I do not know why why changing the name from "Unreal Marketplace" to "Unreal" helped.

I guess translations keys can't contain spaces. Possibly that.

Kjell291:
> > @flavioislima
> > I fixed it, although I do not know why why changing the name from "Unreal Marketplace" to "Unreal" helped.
> 
> I guess translations keys can't contain spaces. Possibly that.

But why worked it with 1.5.8?

flavioislima:
> > > @flavioislima
> > > I fixed it, although I do not know why why changing the name from "Unreal Marketplace" to "Unreal" helped.
> > 
> > I guess translations keys can't contain spaces. Possibly that.
> 
> But why worked it with 1.5.8?
> 

Good question. We updated i18next so maybe that was a breaking change that we missed. 🤔

flavioislima (approved):
Looks good! 👍🏽

flavioislima (approved):

PR #308, opened 2021-04-15
Implement a basic URL scheme.

### LINKED ISSUE #307: Implement a registered Heroic protocol.
For integration purposes, we need to implement a standardised protocol for communicating with Heroic from other applications. The simplest way to achieve this is using Electron's [protocol](https://www.electronjs.org/docs/api/protocol). 

Candidates:
- `heroic://` (going with this one.)
- `heroicgl://`

I'm not sure which kind of protocol we should register, but I'm thinking string.

Usecases:
- Desktop Shortcuts
- Integration into other services like Discord Rich Presence. (I'm thinking: Open Game button to the Rich Presence so that anybody can play along.)

adityaruplaha:
Fixes #307.

Implementation based on what I read here:
https://github.com/electron-userland/electron-builder/issues/4035
https://github.com/electron/electron/issues/4857

Scheme is a lot like HTTP, except that there's no domain name.
Like `heroic://launch?appName=Hazelnut`.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -171,10 +171,14 @@ const contextMenu = () =>
 if (!gotTheLock) {
   app.quit()
 } else {
-  app.on('second-instance', () => {
+  app.on('second-instance', (event, argv) => {
     // Someone tried to run a second instance, we should focus our window.
     if (mainWindow) {
       mainWindow.show()
+    }
+    if (argv[1]) {
+      const url = argv[1]
+      handleProtocol(url)
     }
   })
   app.whenReady().then(async () => {
@@ -209,9 +213,7 @@ if (!gotTheLock) {
     createWindow()
 
     protocol.registerStringProtocol('heroic', (request, callback) => {
-      const [command, args_string] = request.url.split('://')[1].split('/')
-      console.log('Received:', request.url)
-      handleProtocol(command, args_string.split('::'))
+      handleProtocol(request.url)
       callback('Operation initiated.')
     })
     if (!app.isDefaultProtocolClient('heroic')) {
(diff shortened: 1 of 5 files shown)


flavioislima:
Maybe to test you just need to put
`open heroic://`
on the terminal
or xdg-open, like we are doing with the urls to open the epic login page.

adityaruplaha:
Done. Should work perfectly.

electron/main.ts
```diff
@@ -204,6 +206,21 @@ if (!gotTheLock) {
       ]
     })
 
+    protocol.registerStringProtocol('heroic', (request, callback) => {
+      const [command, args_string] = request.url.split('://')[1].split('/')
+      handleProtocol(command, args_string.split('::'))
+      callback('Operation initaited.')
```
flavioislima, line 212: just a typo

adityaruplaha:
@flavioislima this is ready to be merged, but problem is that the frontend has no idea what just happened. I'll make the backend state-aware and the client can just query it. Would also reduce the global state.

electron/protocol.ts
```diff
@@ -0,0 +1,69 @@
...
+          'defaultPath': path,
+          'message': 'Select Installation Path',
+          'properties': ['createDirectory', 'openDirectory']
+        })
+        path = filePaths[0] || path
+      }
+      return await game.install(path)
+    }
+  }
+}
```
adityaruplaha, line 69: Whoopsie.

flavioislima (approved):
Looks good. I'm just not 100% on the user confirmation because if the desktop file will be clicked, then I think it was on porpuse. Steam and lutris don't ask for confirmation for instance. So I think this might not be necessary. But merge it and we can think about that before release.

adityaruplaha:
> I'm just not 100% on the user confirmation

I'm not either, but I did it just in case as its easier to remove something than to implement it pre-release.

electron/protocol.ts
adityaruplaha: Fixed by #314.

PR #312, opened 2021-04-17
ci: Upload artifacts, use Yarn for deployment.

adityaruplaha:
Build and upload artifacts for all pushes and PRs.
Binaries will be kept for a week.

Also switches to using Yarn instead of npm for releases, as recommended by electron-builder's README.

---
See also: https://github.com/electron-userland/electron-builder/issues/5463


adityaruplaha:
This is not financially feasible due to the limited time GitHub Actions offers.
I'm exploring other options for automated PR builds.

PR #314, opened 2021-04-17
eslint: Require trailing newline.

adityaruplaha:
For compliance with the UNIX philosophy.


Flonky (approved):
Dont see anything brealing

PR #334, opened 2021-04-27
Winston logger

wiryfuture:
I've only added some basic logging with winston for when the ipcMains are invoked and on all the catches I saw that were logging some sort of info in the console. This still needs work but is somewhere to start.

Logs will be stored in their own logs folder, with _heroic-issues_ for warnings and errors, _heroic-info_ for info, and _heroic-debugging_ for anyone who's developing and adds a .debug logger temporarily when working on something.

I wish I could try out some more tests but main is broken right now and the frontend isn't working for me 😂


electron/logger.ts
```diff
@@ -0,0 +1,27 @@
+import { createLogger, format, transports }from 'winston'
```
adityaruplaha, line 1: Formatting looks a bit weird.

electron/logger.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const Logger = createLogger({
+  //defaultMeta: { service: 'heroic-main' },
+  exceptionHandlers: [
+    new transports.File({ filename: 'logs/heroic-issues.log' })
```
adityaruplaha, line 6: Would prefer if this was changed to heroic-error.log. Also do we really need the `heroic-` prefix?

electron/logger.ts
```diff
@@ -0,0 +1,27 @@
...
+  ],
+  exitOnError: false,
+  format: format.combine(
+    format.timestamp({
+      format: 'YYYY-MM-DD HH:mm:ss'
```
adityaruplaha, line 11: Do we need a timezone here? Not sure.

electron/main.ts
```diff
@@ -280,14 +285,15 @@ ipcMain.on('unlock', () => {
 })
 
 ipcMain.on('kill', (event, game) => {
+  Logger.info({message: `Attempting to kill game ${game} `, service: 'kill game'})
```
adityaruplaha, line 288: I don't think service name should contain spaces. Seems a bit wrong to do.

adityaruplaha (commented):
Some thoughts on this. Also, I do not like that we have so many service names for small things, with no indication as to where this actually happens. Can't think of a better solution right now, maybe prefix `ipcMain:` to the service names?

Also does the logger print messages to terminal as well as file?

electron/main.ts
wiryfuture: I saw this when writing it myself and was pretty sure I change it to just kill, must have slipped past me

wiryfuture:
They're just the function names atm and if you search for the service it should take you to the function that is at issue. 

I believe some places still have console logs but I was considering combing through and commenting those all out or removing them.

adityaruplaha:
> They're just the function names atm and if you search for the service it should take you to the function that is at issue.

I still don't feel very comfortable not having any structure to them, but I don't have a strong opinion on this.
Also I'll add in a lot of debug logging to the parts which I refactored.

electron/logger.ts
wiryfuture: Was thinking about piping the output from wine or legendary to a log in the same dir so would've needed clarification for that.

electron/logger.ts
adityaruplaha: I'd definitely suggest adding a millisecond field tho.

electron/logger.ts
adityaruplaha: Oh alright. Please rename the exception handler to `heroic-crashes` and `heroic-issues` to `heroic-errors` though.

wiryfuture: Maybe each element should have its own folder inside of logs, ex, "/logs/heroic;/logs/wine;/logs/legendary;"

Issue #285, opened 2021-03-31
Winecfg and winetricks not using game specifig wineprefix


VortexAcherontic:
**Describe the bug**
Hey there if I launch winetricks of winecfg from the game settings tab it will not run those in the specified wine prefix but instead defaults to "~/.wine"

**To Reproduce**
1. Specifiy a dedicated wineprefix for a game
2. run winecfg or winetricks form the settings page
3. winetricks will list the current wineprefix in the window title

**Expected behavior**
- Using the right prefix

**Screenshots**

![Game settings](https://user-images.githubusercontent.com/11244092/113523978-dd51bd80-95ab-11eb-8df8-5683db007c84.png)

![Winetricks](https://user-images.githubusercontent.com/11244092/113099817-dbc67500-91fa-11eb-9c4a-a381f2a5aa04.png)

**Desktop (please complete the following information):**
- OS: openSUSE Tumbleweed
- Version: 1.5.7 (RPM)
- Game: Any
- Wine/Proton version: Lutris Wine 6.0

**Additional context**
I launched heroic via cli and noticed this output:

````
{
  command: "WINE='/home/vortex/.local/share/lutris/runners/wine/lutris-6.0-x86_64/bin/wine64' WINEPREFIX=/z004/sdd1/user/vortex/Spiele/Heroic/Rage2/.wine \n" +
    '    winetricks'
}
{
  command: "WINE='/home/vortex/.local/share/lutris/runners/wine/lutris-6.0-x86_64/bin/wine64' WINEPREFIX=/z004/sdd1/user/vortex/Spiele/Heroic/Rage2/.wine \n" +
    "    '/home/vortex/.local/share/lutris/runners/wine/lutris-6.0-x86_64/bin/wine64' winecfg"
}
````

I suspect it will not get the right prefix because the brackets ' are missing at WINEPREFIX=

flavioislima:
Fixed in #335

PR #335, opened 2021-04-28
Fix issue 285 in electron/main.ts

maxim-kruch:
This is a possible fix for #285. It also fixes a similar issue with the "Run EXE on Prefix" option.

This is the first time I ever made a pull request, so sorry for any mistakes.


electron/main.ts
```diff
@@ -354,11 +354,10 @@ ipcMain.on('callTool', async (event, { tool, wine, prefix, exe }: Tools) => {
     winePrefix = `'${protonPrefix}/pfx'`
   }
 
-  let command = `WINE=${wineBin} WINEPREFIX=${winePrefix} 
-    ${tool === 'winecfg' ? `${wineBin} ${tool}` : tool}`
+  let command = `WINE=${wineBin} WINEPREFIX='${winePrefix}' ${tool === 'winecfg' ? `${wineBin} ${tool}` : tool}`
```
adityaruplaha, line 357: What exactly is changing here?

electron/main.ts
```diff
@@ -354,11 +354,10 @@ ipcMain.on('callTool', async (event, { tool, wine, prefix, exe }: Tools) => {
...
 
   if (tool === 'runExe') {
-    command = `WINEPREFIX=${winePrefix} ${wineBin} ${exe}`
+    command = `WINEPREFIX=${winePrefix} ${wineBin} '${exe}'`
```
adityaruplaha, line 360: Good catch, thank you.

electron/main.ts
maxim-kruch: I think one of the causes for the problem was that the line was spilt into two, but I am not sure about that. The main change is from "WINEPREFIX=${winePrefix}" to "WINEPREFIX='${winePrefix}'".

flavioislima: To fix this kind of error, is good for you to:
1. Simulate the issue exactly as described
2. Add the fix
3. Repeat and check if the error is gone
4. Do the same but using different prefixes, wine and proton configurations.
5. Also, test with prefixes and wine paths that contains spaces on it.

If everything works fine after that, then the error might be gone and nothing was broken.

Because I played for several times with this before, and its hard to not break anything. Some people might have config from months before and we changed something that might have broken it. So updating their config files, should fix the issue most of the time.

maxim-kruch: Thanks for the reply, I followed your steps as far as I could right now (I am on mac right now, so I couldn't check if it works with proton) and I actually found a problem with spaces in different kinds of paths. I added the fix to the pull request.

electron/main.ts
```diff
@@ -354,11 +354,10 @@ ipcMain.on('callTool', async (event, { tool, wine, prefix, exe }: Tools) => {
     winePrefix = `'${protonPrefix}/pfx'`
```
flavioislima, line 354: You will need to remove those quotes from here otherwise you will fix the issue for wine prefixes but will break proton prefixes since we will have double single quotes.

flavioislima (approved):
NIce!
Thanks for the fix

Issue #344, opened 2021-05-07
[UI/UX] Idea: Transparent/Blur NavBar


DavidoTek:
**Is your feature request related to a problem? Please describe.**
More transparency/blur effects could be added for a more "modern" look.

**Describe the solution you'd like**
- Make the NavBar transparent and add a blur effect
- Show blurred game image in the background on the Game page
- Maybe add the option to disable the blur effect (not sure if it consumes too much system resources)

**Describe alternatives you've considered**
Not adding the effect for a simpler look and lower resource usage.

**Additional context**
Changes (NavBar): https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/compare/main...DavidoTek:blur-ui
![heroic2](https://user-images.githubusercontent.com/54072917/117438750-407b9a80-af32-11eb-8014-d856959c763a.png)
Mockup (Game page):
![heroic1](https://user-images.githubusercontent.com/54072917/117438754-41acc780-af32-11eb-8a6a-418fd290cd15.png)

adityaruplaha:
Fixed by #345.

PR #345, opened 2021-05-07
Make NavBar and SearchBar transparent with blur

DavidoTek:
See https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/344  
heroic2

### DIFF
--- a/src/components/Navbar/components/SearchBar/index.css
+++ b/src/components/Navbar/components/SearchBar/index.css
@@ -8,7 +8,7 @@
 .searchInput {
   width: 388px;
   height: 39px;
-  background: #1a1b1c;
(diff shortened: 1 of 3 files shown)


src/components/Navbar/components/SearchBar/index.css
```diff
@@ -8,7 +8,7 @@
 .searchInput {
   width: 388px;
   height: 39px;
-  background: #1a1b1c;
+  background: rgba(26, 27, 28, 0.5);
```
flavioislima, line 11: Can you create a variable for that? Since we are planning on making Heroic themes in the future, we want to use variables for all colors.

adityaruplaha (approved):

PR #349, opened 2021-05-08
electron: Move state management to the backend.

adityaruplaha:
This ended up being more of a refactor that I thought. Highly WIP.

Additionally, we add a lot of abstraction so that it becomes easier to add support for non-Legendary stuff.


adityaruplaha:
Test failure has nothing to do with the actual PR.

adityaruplaha:
This will be followed up by better progress management using child processes and promises.

adityaruplaha:
@flavioislima can we please get this merged quickly? These conflicts were inhumane.

flavioislima:
> @flavioislima can we please get this merged quickly? These conflicts were inhumane.

No, I'll need to look and test those changes. 
This is a big refactor.

If another work depends on this, start a new branch out of this one, and then you can just update it with main.

adityaruplaha:
Alright, but this should have exactly 0 functional code change.

electron/main.ts
```diff
@@ -29,11 +25,11 @@ import i18next from 'i18next'
 import isDev from 'electron-is-dev'
 
 import { DXVK } from './dxvk'
+import { Game } from './games'
 import { GameConfig } from './game_config'
 import { GlobalConfig } from './config'
-import { LegendaryGame } from './games'
-import { Library } from './legendary_utils/library'
-import { User } from './legendary_utils/user'
+import { LegendaryLibrary } from './legendary/library'
```
flavioislima, line 31: I'm wondering: should we really use Legendary or should we use Epic? EpicUser, EpicGame. I think makes more sense to me since Legendary is the backend now, but we might change and implement our own in the future.

flavioislima (approved):
Ok Looks good 👍🏽 
waiting for part 2

electron/main.ts
adityaruplaha: Yeah, that would have its own seperate folder.

PR #350, opened 2021-05-08
Improved Filter on Header

flavioislima:
Added Updates filter to list only updatable games and removed Unreal Market if no item was found


adityaruplaha (approved):
Looks good overall, I assume it has been tested.

flavioislima:
> Looks good overall, I assume it has been tested.

Yeah, tested and also this is just a new filter

PR #358, opened 2021-05-13
Update Electron + Continuous Progress + Wiki Link

flavioislima:
- [x] Update Electron to v12
- [x] Add continuous progress feature to install and updates
- [x] Add Wiki Link

Updated version to 1.7 since next version will be a feature release.

Header Change:
image


src/components/utils.ts
```diff
@@ -1,3 +1,4 @@
+import { InstallProgress } from './../types';
```
adityaruplaha, line 1:
```suggestion
import { InstallProgress } from '../types';
```

src/components/utils.ts
```diff
@@ -43,7 +46,7 @@ export async function handleInstall({
   if (installPath === 'import') {
     const { filePaths } = await showOpenDialog({
       buttonLabel: t('gamepage:box.choose'),
-      properties: ['gamepage:openDirectory'],
+      properties: ['openDirectory'],
```
adityaruplaha, line 49: What happened here?

src/components/Navbar/index.css
```diff
@@ -7,7 +7,7 @@
   box-shadow: 0px 4.00313px 10.0078px #161616;
   width: 100%;
   display: grid;
-  grid-template-columns: 3fr 5fr 1fr;
+  grid-template-columns: 5fr 5fr 1fr;
```
adityaruplaha, line 10: A screenshot would be helpful in visualising the difference. Minor thing though.

adityaruplaha (commented):
Did a general look through on mobile. Looks good mostly. Couple of things I would like explanation on.

src/components/utils.ts
flavioislima: Weird that the automatic import got the wrong file? 🤔

src/components/utils.ts
flavioislima: This is a bug. If you look at it, we should have a property but we are using a translation key. I fixed this long time ago but I think got back on some rebase.
This bug make it impossible to select a folder when installing from the library.

src/components/Navbar/index.css
flavioislima: Ok. But this basically increased the size of the navbar to accommodate the wiki button. Without it, in some languages the wiki button was colliding with the search bar.

src/components/utils.ts
flavioislima: Fixed

src/components/Navbar/index.css
flavioislima: Just added

flavioislima:
@adityaruplaha ignore the i18n changed files, thats because weblate uses 4 indentation spaces on JSON and there is no option to change that, and we were using 2 spaces, so I changed the config and generated the files again to avoid conflicts with it.

src/components/Navbar/index.css
adityaruplaha: > Ok. But this basically increased the size of the navbar to accommodate the wiki button. Without it, in some languages the wiki button was colliding with the search bar. 

Alright. Looks good then.

adityaruplaha (approved):
LGTM.

PR #370, opened 2021-05-21
Added Nvidia Prime Render Option to Settings

### LINKED ISSUE #369: Enable Nvidia Prime Render Offload/Prime-run
For us hybrid graphics laptops owners, we need to add `__NV_PRIME_RENDER_OFFLOAD=1  __GLX_VENDOR_LIBRARY_NAME=nvidia` to the global environment variables to force the use of the discrete GPU.

We could have options alongside MangoHud, GameMode ... to enable Nv Prime Render Offload or to prime-run a game. Example of Lutris:
Screenshot from 2021-05-21 13-00-35

The current alternative is to set environment variables, but it makes quite a lot of mess there and it would be cleaner and easier to manage to have just a switch to flip.

**Additional context**

flavioislima:
Added new option on Global and GameSettings to enable the Nvidia Prime render as requested on #369.

Fixes #369.


adityaruplaha (commented):
Code-wise looks okay, but I would strongly recommend not adding the stub text to all the locales when it's already passed as the default in t(). I have a feeling that it might cause issues with Weblate.

flavioislima:
> Code-wise looks okay, but I would strongly recommend not adding the stub text to all the locales when it's already passed as the default in t(). I have a feeling that it might cause issues with Weblate.

No, weblate checks that the translations are the same as english one and it warns that the text wasn't translated. What we need to do is to merge weblate changes first. Then the changes with new labels, them weblate will find them.

electron/games.ts
```diff
@@ -298,6 +299,7 @@ class LegendaryGame implements Game {
       audio: audioFix ? `PULSE_LATENCY_MSEC=60` : '',
       fps: showFps ? `DXVK_HUD=fps` : '',
       other: otherOptions ? otherOptions : '',
+      prime: nvidiaPrime ? '__NV_PRIME_RENDER_OFFLOAD=1 __GLX_VENDOR_LIBRARY_NAME=nvidia' : '',
```
adityaruplaha, line 302: I think there was another related issue where different environment variables had to be used. Not sure though.

adityaruplaha (approved):
Looks good then.

electron/games.ts
flavioislima: True. But in this case we will use the same behavior as lutris since I think they know best. 👍

PR #373, opened 2021-05-22
UI Improvements + Fixes

### LINKED ISSUE #195: Some games automatically cancel their installations while downloading.
**Describe the bug**
Some games' downloads cancel themselves after a few minutes of downloading (doesn't happen on every game, seemingly at random)

**To Reproduce**
Attempt to download a game and watch as the install is (sometimes) cancelled automatically after roughly 10 minutes.

**Expected behavior**
Games should reach 100% download and successfully install.

 - OS: Manjaro
-  Game: GTA V, Metro Last Light Redux, Rocket League, several others

### LINKED ISSUE #367: Unreal Marketplace tab is hidden even though I have marketplace items in my account.
**Describe the bug**
When I launch the application I should see Unreal Marketplace tab since I do have items in there but instead I don't see it because application assumes that I don't have them.

**To Reproduce**

1. Open application 
2. log into your account
3. See error

**Expected behavior**
Unreal marketplace tab should be visible

**Screenshots**
image

- OS: KDE Neon and Arch Linux

flavioislima:
Improve the logic for the continuous download and:

Fix #367 
Probably fix #195

### DIFF
--- a/public/locales/nl/translation.json
+++ b/public/locales/nl/translation.json
@@ -182,6 +182,5 @@
         "refresh": "Herladen",
         "support": "Steun ons"
     },
-    "wiki": "Wiki",
-    "Store": "Winkel"
+    "wiki": "Wiki"
 }
--- a/src/components/utils.ts
+++ b/src/components/utils.ts
@@ -5,10 +5,10 @@ import {
   install
 } from 'src/helpers'
 
-import { GameStatus, InstallProgress } from 'src/types'
+import { AppSettings, GameStatus, InstallProgress } from 'src/types'
 import { TFunction } from 'react-i18next'
 
-const { remote } = window.require('electron')
+const { ipcRenderer, remote } = window.require('electron')
 const {
   dialog: { showOpenDialog }
 } = remote
@@ -30,13 +30,22 @@ export async function handleInstall({
   t,
   progress
 }: Install) {
+  const config: AppSettings = await ipcRenderer.invoke(
+    'requestSettings',
+    'default'
+  )
+
   if (isInstalling) {
     const { folder_name } = await getGameInfo(appName)
-    return handleStopInstallation(appName, [installPath, folder_name], t, progress)
+    let path: string = installPath
+    if (installPath === 'default'){
+      path = config.defaultInstallPath
+    }
+    return handleStopInstallation(appName, [path, folder_name], t, progress)
   }
 
   if (installPath === 'default') {
-    const path = 'default'
+    const path = config.defaultInstallPath
     await handleGameStatus({ appName, status: 'installing' })
     await install({ appName, path })
     return await handleGameStatus({ appName, status: 'done' })
@@ -65,7 +74,7 @@ export async function handleInstall({
     })
 
     if (filePaths[0]) {
-      const path = filePaths[0]
+      const path = `${filePaths[0]}`
       await handleGameStatus({ appName, status: 'installing' })
       await install({ appName, path })
       return await handleGameStatus({ appName, status: 'done' })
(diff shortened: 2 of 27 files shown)


src/types.ts
```diff
@@ -85,6 +85,7 @@ export interface GameStatus {
 export interface InstallProgress {
   bytes: string
   eta: string
+  folder?: string
```
adityaruplaha, line 88: Please keep this consistent across frontend and backend.

src/components/utils.ts
```diff
@@ -65,7 +74,7 @@ export async function handleInstall({
     })
 
     if (filePaths[0]) {
-      const path = filePaths[0]
+      const path = `${filePaths[0]}`
```
adityaruplaha, line 77: Why this?

src/types.ts
adityaruplaha, line 88: Please keep this consistent across frontend and backend.

adityaruplaha (changes requested):
Some comments. A few really need to be addressed.

public/locales/nl/translation.json
```diff
@@ -182,6 +182,5 @@
         "refresh": "Herladen",
         "support": "Steun ons"
     },
-    "wiki": "Wiki",
-    "Store": "Winkel"
+    "wiki": "Wiki"
```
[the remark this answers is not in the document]
flavioislima: duplicated, see line 161

src/types.ts
flavioislima: You mean to add the folder to the BE part? but we don't need it. its dead code from the beginning. I'm using this only to check if the user is installing on the same folder, so the frontend can continue the percentage.
In the future, we could do the other way, and recalculate the percent on the BE, but for that, Electron should support storage on the backend or would need to use a file for that. Which I don't think is that good.

src/components/utils.ts
flavioislima: Because I changed the function on the BE to install the games.
if I don't do that, if the folder have spaces, legendary will install on the wrong folder.
It comes from the folder selection like '/mnt/game folder', so that's why I need to wrap it in another string.

src/types.ts
adityaruplaha: I know we don't need it anywhere in the backend. Probably wouldn't as well. This is just for maintaining consistency since one check I think we should have is that frontend and backend types have the same definitions.

src/components/utils.ts
adityaruplaha: So shouldn't you have done `"${...}"`?

src/types.ts
flavioislima: > I know we don't need it anywhere in the backend. Probably wouldn't as well. This is just for maintaining consistency since one check I think we should have is that frontend and backend types have the same definitions.

But that's another thing and then we should find a way to properly share the same types across both.
We can think about this later but for this PR I don't think its necessary

src/components/utils.ts
flavioislima: No, because we already get the folder name with `''`
Check this install command:
`/.../legendary install a579a08e71b341e6af106dbd170e2ab2 --base-path /mnt/ssd480/Games/Heroic/Folder with Spaces  -y |& tee "/home/flavio/.config/heroic/GamesConfig/a579a08e71b341e6af106dbd170e2ab2.log" "/tmp/heroic/install-a579a08e71b341e6af106dbd170e2ab2.sock"`

flavioislima: Yeah, you were right. Fixed 😛

src/types.ts
flavioislima: Fixed as well, Mr. consistency 👍🏽

PR #375, opened 2021-05-23
[breaking] Switch to Yarn v2, Husky v6, Node v15. Update ALL packages.

adityaruplaha:
While we're not PnP ready, we can still switch to the latest version. No behaviour should change.

### Checklist:
- [x] Frontend compile.
- [x] Backend compile.
- [x] Frontend runs.
- [x] Backend runs.


adityaruplaha:
This might introduce breaking changes. I'll test this.

flavioislima:
The upgrades look ok since I did most of the major version on my update electron PR. But I'm still not convinced about Yarn2.
So everyone that has yarn1 on their computers will need to upgrade to start working with Heroic.
I still think NPM 7 is a better alternative if we really need to leave yarn1 since it is not maintained.

adityaruplaha:
> So everyone that has yarn1 on their computers will need to upgrade to start working with Heroic.

Nope, should start working out of the box.

flavioislima:
We still have this other option that I never tried before, but saw some people talking. It is good since it doesn't download the same package twice if you're using it in other repo.
So to use it with different heroic folders with different branches and forks (like I do), would be interesting.

adityaruplaha:
Well whatever we do, we better decide quickly as this is basically a very efficient conflict generator.

flavioislima:
> Well whatever we do, we better decide quickly as this is basically a very efficient conflict generator.

I don't really want to decide things quickly, and this is probably a big change like you said. 
So maybe we could use yarn2 locally and keep npm for the scripts like I said on discord, since they are compatible now.

adityaruplaha:
So, are we going Yarn 2 or nah?

flavioislima:
> So, are we going Yarn 2 or nah?

I think it is an unnecessary dependency on local development. I think we should do it as I said above.
Locally we can use whatever we want, but for the scripts and actions, lets keep using npm, since both are compatible.

Also, I noticed that you haven't upgraded node, only types/node. The only way of upgrading node is upgrading electron, that is on version 13.0.1, so you might then revert the yarn changes and upgrade the other packages, including electron. Then we will have node 14.7.0, latest LTS

adityaruplaha:
Alright, so closing this. I'll just update all the packages.

PR #385, opened 2021-05-25
electron: Implement monitoring game progress using a TCP server.

adityaruplaha:
Create a TCP server ~~on port 40627~~ (see the 3rd commit) where the log gets tee'd (relies on bash magic).
Theoretically, can also be used to monitor progress of a remote machine.

I tried to use UNIX sockets but couldn't manage to find a way to easily shove data into the socket using readily available tools. We don't have netcat available on all machines unfortunately.

Breaks Windows compatibility. Also doesn't work on MacOS since Zsh doesn't have the ability to easily pass data into a Unix socket. However, it can still be acheived as shown here: https://who23.github.io/2020/12/03/sockets-in-your-shell.html

An easier solution would be to periodically shove the entire file into the progress management algorithm. However this has its downsides of higher disk usage and higher memory footprint.

----

## Checklist

- [x] Install/Update
- [x] Repair
- [x] Multiple Progress Servers

### DIFF
--- a/electron/constants.ts
+++ b/electron/constants.ts
@@ -6,6 +6,7 @@ import { join } from 'path'
 const currentGameConfigVersion : GameConfigVersion = 'v0'
 const currentGlobalConfigVersion : GlobalConfigVersion = 'v0'
 const home = homedir()
+const heroicPort = 40627
 const legendaryConfigPath = `${home}/.config/legendary`
 const heroicFolder = `${home}/.config/heroic/`
 const heroicConfigPath = `${heroicFolder}config.json`
@@ -46,6 +47,7 @@ export {
   heroicGamesConfigPath,
   heroicGithubURL,
   heroicInstallPath,
+  heroicPort,
   heroicToolsPath,
   home,
   icon,
--- a/electron/constants.ts
+++ b/electron/constants.ts
@@ -6,7 +6,6 @@ import { join } from 'path'
 const currentGameConfigVersion : GameConfigVersion = 'v0'
 const currentGlobalConfigVersion : GlobalConfigVersion = 'v0'
 const home = homedir()
-const heroicPort = 40627
 const legendaryConfigPath = `${home}/.config/legendary`
 const heroicFolder = `${home}/.config/heroic/`
 const heroicConfigPath = `${heroicFolder}config.json`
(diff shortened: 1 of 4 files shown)


flavioislima:
@adityaruplaha have you tried decrease the buffer size to check if the solution work? But even if it work I don't think it's a good change since this will break compatibility. So please don't merge until we decide about this.

adityaruplaha:
Yeah I'm not gonna merge until I'm sure the compatibility issues can be clearly resolved.

And sorry, I haven't tested it yet, was busy with other projects.

adityaruplaha:
Do NOT merge this.

On some careful reconsideration, I've decided that using a TCP server is a good idea, but the current implementation is dependent on bash, and outright doesn't work on Mac and Windows. So it's simply not worth it.

I'll look for a better solution. Still inclined on using something similar but using a different mechanism to get the log.

flavioislima:
@adityaruplaha if you really really want to fix that bug for good, and have a stream way of getting the logs and etc. Also, since node 14.7, having a way of killing the process independent of the OS. 
Start to experiment changing the install functions using this:
https://nodejs.org/dist/latest-v14.x/docs/api/child_process.html#child_process_child_process_spawn_command_args_options

Maybe, the main implementation wont be enough, so you might need to install other third-party libraries to make the async work.

Or, you can write your own async spawn function, using promises.
But, if you like challenges, that's a good one 😄

adityaruplaha:
I was actually thinking of doing something like that while in the shower today, I'll look into it. I don't like polling the filesystem.

flavioislima:
> I was actually thinking of doing something like that while in the shower today, I'll look into it. I don't like polling the filesystem.

Also, spawn is more efficient and doesn't use the shell, so should improve in other areas as well.

adityaruplaha:
Waiting on #399 so that I can use these.

adityaruplaha:
Parts of this will be integrated back into #349. Both are pretty linked to each other, and keeping one away from the other will be difficult. So I'll just do this as part of #349.

adityaruplaha:
On second thought, merging #349 first and then rebasing this might be better.

adityaruplaha:
This is gonna have crazy conflicts, I'll reimplement this after #385 gets merged.

PR #393, opened 2021-05-29
React frontend tests

Nocccer:
- [x] Add tests for src/components
- [ ] Add tests for src/screen
- [ ] add tests for src/helpers
- [ ] add tests for src/hooks
- [ ] add tests for src/state
- [ ] add tests for rest
- [ ] adapt to code guidelines
- [ ] comment tests


flavioislima:
HI! 
One thing: Maybe it's not a good idea to wait to write all the tests for that PR to be merged. Maybe you could open a PR for a bunch of tests. Otherwise, you can have conflicts and also need to fix the tests after we merge new stuff to those components.

So I think you could pick an item from the checklist above and merge it after finishing that item and then open another PR for the others :)

Nocccer:
Ok sounds good. Never opened a pull-request for specific files. How can i do that? 
I would merge the tests for the components first, because im done with every file there.

adityaruplaha:
I would say make a PR for each checklist item.
@flavioislima add a new project and give@Nocccer permissions?

PR #396, opened 2021-05-29
frontend tests for src/components

Nocccer:
- [x] tests for src/components
- [x] fix unreal version selection
- [x] fix components/Header handle functions

Open Questions:
1. How to mock react-i18next global ?
2. How to wait that useContext rerendered on change ?
3. How to mock electron BrowserWindow ?


flavioislima (approved):
Everything looks really good to me.
Only thing I would change but thats just personal preference, would be to have the test files inside a `__tests__` folder. Then, we could have snapshots tests as well and we could have specific helper files.

But we can think about folder structure in the future.

But other than that its amazing, thanks a lot :)

PR #399, opened 2021-06-01
Update all packages.

adityaruplaha:
Important changes:
* Moving to Husky v6.
* Electron 13 (Node 14.7). :tada:
* ~~Update to Jest 27.~~ (Incompatible with `react-scripts`)
* Added `@testing-library/dom` as a dev dependency as yarn was complainingg about it.
* Moved the lint plugins to `devDependencies`.


package.json
```diff
@@ -85,7 +82,7 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "4.0.3",
     "tslib": "^2.2.0",
-    "web-vitals": "^0.2.4"
+    "web-vitals": "^1.1.2"
```
flavioislima, line 85: I think this one can go away because Heroic is not a website. But on the code side, might have some import that was added by create-react-app. So remove the file that uses it and remove this dependency.

package.json
```diff
@@ -111,40 +109,35 @@
...
+    "@testing-library/dom": "^7.31.0",
+    "@testing-library/jest-dom": "^5.12.0",
+    "@testing-library/react": "^11.2.7",
+    "@testing-library/user-event": "^13.1.9",
     "@types/classnames": "^2.2.11",
     "@types/i18next-fs-backend": "^1.0.0",
-    "@types/jest": "^26.0.20",
-    "@types/node": "^14.14.22",
+    "@types/jest": "^26.0.23",
+    "@types/node": "^15.6.1",
```
flavioislima, line 119: I'm wondering if, since we are using node 14.7, would be better to use the types for that version, since we can have wrong information and types from methods that exists on node 15 and dont exists on node 14.

flavioislima (approved):

PR #275, opened 2021-03-25
OS/Windows: Initial support for Windows

flavioislima:
Here is a list of things that I'm working right now:
- [x] Sign In
- [x] Load Library
- [x] Install Games
- [x] Launch Games
- [x] Import Games
- [x] Game Statuses (Progress is working perfectly)
- [x] Remove unneeded settings.
- [ ] Stop Installation (Have no idea on how to do it - Will do on another PR, maybe using AbortController introduced on Node14.7 and available on Electron 13).

Problems that need fixes:
- [x] A Better way to handle paths. On Windows, if the path contains `'`, legendary cannot do anything, so for now will need to use a replace function when iswindows.
- [x] Test a lot on all platforms to check what could go wrong.
- [ ] Selective download is not working so its not possible to download Fortnite or Cyberpunk yet. (Something related to conversion between cp1252 and utf8, will fix on another PR).

PR #400, opened 2021-06-01
ci: Switch to Node 14 for all tasks, don't report coverage (for now).

adityaruplaha:
This should fix the issues we've been having with deploying tests on CI. Now we use Node 14 everywhere.
See https://github.com/actions/setup-node for details.

Also, since we're having issues with coverage plugins, I think we should just run tests for now, coverage can come later.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,9 @@ jobs:
     steps:
       - name: Checkout repository.
         uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
       - name: Install modules.
         run: yarn
       - name: Build 'src'.
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -10,7 +10,10 @@ jobs:
   draft-releases:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
       - run: sudo apt-get install --no-install-recommends -y libarchive-tools libopenjp2-tools rpm
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
       - run: yarn install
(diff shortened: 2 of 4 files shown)


adityaruplaha:
The issue is the plugin being used. I'm switching it out.

flavioislima:
But still, I wanted to remove yarn and use npm 7 there. Also, we can remove this coverage plug-in, if you check the logs you will see that it's downloading another docker image so it is a docker inside a docker. Ahaha

adityaruplaha:
> But still, I wanted to remove yarn and use npm 7 there. Also, we can remove this coverage plug-in, if you check the logs you will see that it's downloading another docker image so it is a docker inside a docker. Ahaha

Yes, not a good idea to use this plugin. `npm` though? I like Yarn better personally.

adityaruplaha:
@flavioislima the issue seems to be that the bot can't post coverage data. Maybe check repo perms?

adityaruplaha:
Since we're having issues with coverage I decided to just run tests.

adityaruplaha:
I'll backport the CI changes from #275 to here.

adityaruplaha:
Oh great, we can't. So we have to mix package managers? I'm really not seeing the point in using npm when yarn works just fine.

@flavioislima

adityaruplaha:
As expected, the issue is with repo permissions.
https://github.com/adityaruplaha/HeroicGamesLauncher/pull/1

Nocccer:
Check runs needs the write permission:
https://docs.github.com/en/rest/reference/checks#create-a-check-run

flavioislima:
About this. If you want to upgrade all packages, I noticed that the yarn command doesn't show all upgradable packages.
Check if you have updated typescript and all plugins related as well.
Also Electron is on 13.1 now.

adityaruplaha:
> If you want to upgrade all packages

That was a different PR.

flavioislima:
> > If you want to upgrade all packages
> 
> That was a different PR.

Sorry, I'm crazy 😂

flavioislima (approved):
Looks Good 👍🏽

PR #401, opened 2021-06-01
Try stuff out with CodeQL Analysis.

adityaruplaha:
I'm curious to see what static code analysis gives, don't merge this.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '28 19 * * 4'
+
+jobs:
+  analyze:
+    name: Analyze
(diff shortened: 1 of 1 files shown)


adityaruplaha:
@flavioislima you think this is useful? Vulnerability and performance analysis isn't bad imo.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
```
flavioislima, line 36: Are you sure it will work with Typescript as well?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '28 19 * * 4'
```
flavioislima, line 21: This means it will run on the main branch whenever it reaches this time?

adityaruplaha: It should.

.github/workflows/codeql-analysis.yml
adityaruplaha: I mean, the docs say so.

flavioislima:
I still don't get the point about this QL stuff, I think we could use our time to develop new features and improve the code overall with things more valuable to the user like:
**- Integration with other stores**
That should be the main focus since will bring a lot of value to the product and to the community since we are lacking a really good GOG client on Linux right now.

I think have tests is essential and have an organized codebase is also really good, but whats most important is to have a good features since its what the user sees in the end :P

Issue #406, opened 2021-06-05
macOS? Heroic does not load games when in a folder with spaces


StefanLobbenmeier:
**Describe the bug**
I just built this for macOS and installed it to the Applications folder. I opened it and it was stuck loading the games list.
I then started the binary via the terminal to see the logs:

```
➜  MacOS pwd                       
/Applications/Heroic Games Launcher.app/Contents/MacOS
➜  MacOS ./Heroic\ Games\ Launcher 
Default Wine not installed
Default Wine not installed
Protocol already registered.
Error occurred in handler for 'checkGameUpdates': Error: Command failed: /Applications/Heroic Games Launcher.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary list-installed --check-updates --tsv
/bin/sh: /Applications/Heroic: No such file or directory

    at ChildProcess.exithandler (child_process.js:317:12)
    at ChildProcess.emit (events.js:315:20)
    at maybeClose (internal/child_process.js:1048:16)
    at Socket.<anonymous> (internal/child_process.js:439:11)
    at Socket.emit (events.js:315:20)
    at Pipe.<anonymous> (net.js:673:12) {
  killed: false,
  code: 127,
  signal: null,
  cmd: '/Applications/Heroic Games Launcher.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary list-installed --check-updates --tsv',
  stdout: '',
  stderr: '/bin/sh: /Applications/Heroic: No such file or directory\n'
}
^C% 
```

**To Reproduce**

1. Launch Heroic from a folder with spaces (e.g. default installation on macOS)

**Expected behavior**
The game list is shown

**Screenshots**
Screenshot of Heroic when stuck loading:
![image](https://user-images.githubusercontent.com/36509607/120900449-1ef6f700-c635-11eb-941e-2ec765f34a03.png)

**Desktop (please complete the following information):**

- OS: macOS Big Sur
- Version 521620abae366b7d3725b07b94e100b275da25f9  (Actually on my fork: https://github.com/StefanLobbenmeier/HeroicGamesLauncher/commit/521620abae366b7d3725b07b94e100b275da25f9)
- Game: -
- Wine/Proton version: -
- Hardware specs (RAM, CPU, GPU): x86

**Additional context**

StefanLobbenmeier:
[...]

PR #409, opened 2021-06-08
Bugfixes and Optimizations

Linux4Good:
I wanted to develop the update all games feature but could not test it enough yet.
So this is a maintenance task only and I'll continue the update all games in the future:

- [x] Upgrade Electron to 13.1.1
- [x] Remove Remote modules completely since they are deprecated now
- [x] Fix Python check on other platforms as stated here at #406 
- [x] Fix a bug on importing a game the folder was not selectable (introduced on my previous PR)
- [x] Created a UpdateAllGames function to be used in the future


Linux4Good:
@flavioislima 
@Nocccer 
@adityaruplaha 

Please take a look when you guys have time.
The commit list is big because my branch was outdated I guess.
So it starts from commit about fixing Python

Nocccer:
Looks good. Only think i don't know if we should change this atm is the version of heroic.
@flavioislima should decide that.

Nocccer (approved):

flavioislima:
> Looks good. Only think i don't know if we should change this atm is the version of heroic.
> @flavioislima should decide that.

That's right, but 1.7 is gone, I'm not sure if we will release a bugfix version like 1.7.1 or go directly to 1.8, what makes sense since we changed a lot of stuff and added support for Windows.

@Linux4Good keep it 1.8 then, I think makes sense.
I think we might release a new version this week after proper testing.

src/helpers/index.ts
```diff
@@ -126,7 +121,7 @@ async function fixSaveFolder(
 ) {
   const { user, account_id: epicId } = await ipcRenderer.invoke('getUserInfo')
   const username = isProton ? 'steamuser' : user
-  const isWin = process.platform === 'win32';
+  const isWin = await getPlatform() === 'win32';
```
flavioislima, line 124: I think this will return a promise and not the platform.
can you test it, otherwise it will break the save-sync functionality,

adityaruplaha:
Would advise releasing a new version now, I have some refactors planned.

src/helpers/library.ts
```diff
@@ -195,11 +192,22 @@ const launch = (appName: string, t: TFunction<'gamepage'>, handleGameStatus: (ga
 const updateGame = (appName: string): Promise<void> =>
   ipcRenderer.invoke('updateGame', appName)
 
+// Todo: Get Back to update all games
```
adityaruplaha, line 195: Is this queue based? Otherwise exposing an endpoint for the updateAll function already present on the backend would be smarter imo.

adityaruplaha (dismissed):
Looks good code-wise, a few comments.

src/helpers/index.ts
Linux4Good: I'll check it out

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "heroic",
-  "version": "1.7.0",
+  "version": "1.8.0",
```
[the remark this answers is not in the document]
Linux4Good: I'm not releasing, just updated the version :P

src/helpers/library.ts
Linux4Good: the problem is the backend part is still not connected with the frontend, so if we update a game there, there will be no feedback.
But, you have a point. I tried to implement this using a queue system on GlobalState, so I was checking first if there was any game on the libraryStatus with a status different from playing. So if we have a game installing, updating or repairing, the update would wait for it to finish.

But I could not test it enough since during this all my games got updated :joy: 

I think we might then use the backend implementation once you have the status and the communication going on back and forth.

But can I keep this in case we decide to call this on FE?

electron/main.ts
```diff
@@ -252,9 +249,13 @@ if (!gotTheLock) {
...
 
-    if (pythonVersion < 3.8) {
-      console.log(`Python Version incompatible. Python needed: >= 3.8, Python found: ${pythonVersion}`);
-      dialog.showErrorBox('Python Error', `${i18next.t('box.error.python', 'Python needs to be higher than 3.8 for Heroic to work')}`)
+    if (process.platform === 'linux'){
+      const {stdout: pythonInfo} = await execAsync('python --version')
+      const pythonVersion: number | null = pythonInfo ? parseFloat(pythonInfo.split(' ')[1].replace('\n', '')) : null
+      if (pythonVersion < 3.8) {
+        console.log(`Python Version incompatible. Python needed: >= 3.8, Python found: ${pythonVersion}`);
+        dialog.showErrorBox('Python Error', `${i18next.t('box.error.python', 'Python needs to be higher than 3.8 for Heroic to work')}`)
```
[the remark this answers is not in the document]
Linux4Good: Dam, I thought I have fixed this on the other PR! :joy:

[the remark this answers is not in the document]
Linux4Good: fixed

src/helpers/index.ts
Linux4Good: fixed

flavioislima (approved):

PR #416, opened 2021-06-09
ci: Upload artifacts for PRs.

adityaruplaha:
Supersedes #312.

@flavioislima confirmed that we have unlimited Actions time, so building and uploading artifacts is now feasible for PRs.

Also, I'm only keeping AppImages, pacman binaries to cut down on time.


flavioislima:
Looks good but I don't think the upload is working:`
Warning: No files were found with the provided path: dist/heroic*.AppImage. No artifacts will be uploaded.`

I think would be good to build an AppImage and Pacman file when we merge something in main, then we can have an AUR package like heroic-git for people that want to try new stuff before release.

flavioislima:
I think the pacman package starts with a lowercase, different from the AppImage, don't ask me why 😂

adityaruplaha:
I realised this just as you commented lol.

adityaruplaha:
Works fine now.

flavioislima (approved):
nice :D

PR #418, opened 2021-06-10
Experimental controller support.

flavioislima:
I'll try to get back to where @onadrog stopped from here.


dragonDScript:
I think that gamepad support should be handled by an entire separate UI. The current UI is thought for mouse and keyboard only.

PR #420, opened 2021-06-10
Fixed python check

Nocccer:
- [x] catches not found calls
- [x] added check for python3
- [x] added more log messages

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -246,10 +246,26 @@ if (!gotTheLock) {
     })
 
     if (process.platform === 'linux'){
-      const {stdout: pythonInfo} = await execAsync('python --version')
-      const pythonVersion: number | null = pythonInfo ? parseFloat(pythonInfo.split(' ')[1].replace('\n', '')) : null
-      if (pythonVersion < 3.8) {
-        console.log(`Python Version incompatible. Python needed: >= 3.8, Python found: ${pythonVersion}`);
+      let found = false
+      let version: number | null = null
+      for(const value of ['python', 'python3']){
+        try {
+          const {stdout: pythonInfo} = await execAsync(value + ' --version')
+          version = pythonInfo ? parseFloat(pythonInfo.split(' ')[1].replace('\n', '')) : null
+          if (version >= 3.8) {
+            console.log('Found python version ' + version)
+            found = true
+            break
+          }
+        }
+        catch (error){
+          console.log('Catched ' + value + ' with: \n' + error)
+        }
+      }
+
+      if(!found)
+      {
+        console.log(`Python Version incompatible. Python needed: >= 3.8, Python found: ${version}`);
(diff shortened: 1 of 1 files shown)


flavioislima (commented):
I think the logic is quite complex for just a python check. To be honest, what happens if we add python 3.8+ as a dependency for the packages? We can do that on the package.json. So all packages created will check the dependency first when installing. Maybe its better than check on the code. It will also download the dependencies In case the user does not have it.

adityaruplaha (changes requested):
Would advise using a regex based solution. Floats are quite unreliable when parsing equalities, and we can use all 3 fields.

Nocccer:
Can you implement the regex solution? The floating point solution comes from @Linux4Good. I just added the python3 --version check.

flavioislima:
> Would advise using a regex based solution. Floats are quite unreliable when parsing equalities, and we can use all 3 fields.

What do you mean by that? Is this in Javascript you mean?
Even if not, this would be an issue if the float number need to be in some exact way. 
HEre it will be something like 3.8.3, 3.9.2, anyway.

Javascript is the best language in terms of coersion. I dont this should be a problem.

I think we are overthinking too much a simple check.

adityaruplaha:
Precisely, how do we check if its greater than 3.8.9 (say)? 3.8.9 is not a float. The existing implementation fails. Versions aren't floats and shouldn't be treated as such.
Now that the checking is getting a bit complicated, I suggest we should really make it properly, instead of just another hack.

I'll try to whip something up in a couple minutes, was in the middle of GSR Part 2, gimme some time.

flavioislima:
> Precisely, how do we check if its greater than 3.8.9 (say)? 3.8.9 is not a float. The existing implementation fails. Versions aren't floats and shouldn't be treated as such.
> Now that the checking is getting a bit complicated, I suggest we should really make it properly, instead of just another hack.
> 
> I'll try to whip something up in a couple minutes, was in the middle of GSR Part 2, gimme some time.

But why do you need to check if its bigger than 3.8.9? Also, have you tried?

Have you checked our current implementation to check heroic updates and dxvk updates?

They use the same solution and are working perfectly.

Javascript, for that, is better than most languages. Believe me.

Look for a workshop by Kyle Simpson about coersion.

adityaruplaha:
> They use the same solution and are working perfectly.

It'll break the moment we hit version 1.10 or higher, since we rely on ASCII comparisons.
For DXVK it doesn't matter since we're not using any ordering. Just equality checks.

flavioislima:
Anyway, we should have the dependency for the packages and for AppImage we could use any other check, I just think you guys are overthinking this.

Nocccer:
If python as a dependency doesn't update the system python, i would go for it. But this python problem should be fixed before release of 1.8.0, because it is a real problem for most system with python3 instead of python. My System included.

adityaruplaha:
Superseded by #424. Closing.

PR #337, opened 2021-04-30
backend: Implement logging using Winston.

wiryfuture:
Had to close and reopen this as I reforked heroic.

I need someone else to test my branch for this because the frontend is not working with it for me, even though _main_, the branch I forked from, works fine.

Fixes #422.

PR #423, opened 2021-06-11
Add logger

### LINKED ISSUE #422: Add a logger
When a bug occurs, usually we have to re-start Heroic in the terminal. If we added a logger and then saved the logs in a file, e.g. %appdata%/heroic/logs/2020-02-02-23:00.txt or .config/heroic/... it'd be much easier to report bugs.

In the main process, listen for changes in stdout and save them in an array. Then, when the program is exiting, array.join("\n") and save it to a file with the local time.

It is important to log everything important, e.g. config changed, started game X, ... and the time that it occurs.

(I'm free to work on this tomorrow or so, so I will submit a PR soon!)

dragonDScript:
Fixes #422


electron/main.ts
```diff
@@ -54,13 +54,20 @@ import {
...
 const { showErrorBox, showMessageBox,showOpenDialog } = dialog
 const isWindows = platform() === 'win32'
 
 let mainWindow: BrowserWindow = null
 
 function createWindow(): BrowserWindow {
+  listenStdout().then((arr) => {
+    const str = arr.join('\n')
+    const date = new Date().toDateString()
+    const path = `${app.getPath('crashDumps')}/${date}`
```
flavioislima, line 67: Where this path would be? its where Heroic is installed? Because then we could have permission issues. It might be good to save the logs under `$HOME/.config/heroic/` if that's the case.

dragonDScript:
It is platform dependant. I can modify it using app.setPath("crash dumps",
value)

electron/main.ts
dragonDScript: It is platform dependant. I can modify it using app.setPath("crashDumps",
value). But usually, it is under home.

adityaruplaha:
What exactly is the objective of this PR? #337 looks much better to me than just dumping stdout to a file.

dragonDScript:
> What exactly is the objective of this PR? #337 looks much better to me than just dumping stdout to a file.

This PR doesn't use any dependency.

Nocccer (dismissed):

adityaruplaha:
> > What exactly is the objective of this PR? #337 looks much better to me than just dumping stdout to a file.
> 
> This PR doesn't use any dependency.
> 
> *A little correction: it also saves things from stderr*
> 
> We could also add electron-log to this easily, because electron log uses stdout and stderr.

I don't really see the point. Just saving stdout and stderr isn't exactly what I have when I hear logging. Considering the much more refined version is ready for merge, I think I'll go with that.

dragonDScript:
Ok, i accept your opinions about this. Define 'refined'.  my solution
doesn't use any API or module.

wiryfuture:
> Ok, i accept your opinions about this. Define 'refined'. my solution doesn't use any API or module. Actually, it is possible to merge both.

I get where you're coming from, but the tad bit more formatting with servicenames and log levels like in #337  (and whatever else you choose to use from the winston module) are nice to have. 
Since the winston solution realistically has less maintenence for us, I'd personally lean towards using that one (but, PRs to that one would be appreciated if you see a way to improve it)

adityaruplaha:
I agree with @wiryfuture. @dragonDScript I appreciate your efforts but this is redundant. I really don't see why this solution is better, apart from the fact that this doesn't use a dependency, but how does that matter?

dragonDScript:
Thank you, guys

flavioislima:
> Thank you, guys

Yeah, thanks for that!
If you still want to contribute to the project, check our issues list. There are several there that you could pick up and we really appreciate any help :D

flavioislima:
@adityaruplaha @wiryfuture @dragonDScript 
After looking into #337 (that was reverted because of some bugs but now we have #438  to fix the bugs), I think this other PR looks better for 3 reasons:
- It doesn't use an external dependency like was said, but that would be ok perhaps.
- This one stores the log after finishing the process, so after heroic crashes or is closed as far as I can tell looking at the code. The other one is writing information every time, especially when we are calling functions for the games, like `getGameInfo`, that will make the app a lot slower especially on HDDs and for people with really long libraries. Also, testing here, most of the log is useless since we need logs that help us to debug but in the log, there is just a message that something went wrong, but what?
- With all this writing, on an SSD we could reduce the lifespan of it I'm afraid.

So, for now, I think we can keep this one, and if we elaborate more and have some other way of logging that it's more useful and have no performance impact on Heroic.

electron/main.ts
flavioislima: I noticed that we have a file under `.config/Heroic` called Crash Reports so it's probably that folder.

flavioislima (dismissed):
Looks good as a simple and first solution to the logs problem.
We can maybe in the future use a more sophisticated solution, but for now this one should suffice.

Nocccer:
Can we get something like a log function, which takes a log level,  message and maybe location ? And we could ad something like, a flag which turns on/off Debug logging.

Log Levels:
Error
Warning
Info
Debug

flavioislima:
> Can we get something like a log function, which takes a log level, message and maybe location ? And we could ad something like, a flag which turns on/off Debug logging.
> 
> Log Levels:
> Error
> Warning
> Info
> Debug

That would be nice

dragonDScript:
> Can we get something like a log function, which takes a log level, message and maybe location ? And we could ad something like, a flag which turns on/off Debug logging.
> 
> Log Levels:
> Error
> Warning
> Info
> Debug

That is possible to that.

I can implement two solutions:

- Using the electron-log library.
- Custom library specially written for Heroic.

Which do you prefer?

dragonDScript:
I've just fixed the conflicts and added log levels.

electron/main.ts
dragonDScript: Yeah!

Nocccer:
I will also look at it. Can you maybe write tests for the functions you implemented ?

electron/console.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Use this module to log things into the console.
+ * Everything will be saved to a file before the app exits.
+ * Note that with console.log and console.warn everything will be saved too.
+ * error equals console.error
+ */
+
+
```
Nocccer, line 8: Can you wrap this into a single function? Thinking of
```
type LogLevel = "error" | "warning" | "info" | "debug";
export function logMessage(level: LogLeveL, message: string) {
  switch(level){
    case 'error':
      console.log('ERROR: ' + message);
      break;
// ...
}
```

Also some documentation with jsdoc or tsdoc would be great :).

electron/logger.ts
```diff
@@ -0,0 +1,22 @@
+export function listenStdout(): Promise<string[]> {
```
Nocccer, line 1: Please add a tsdoc or jsdoc comment and describe a little bit what the function does.
This will help other people like me to understand in the future, what the purpose of this function is.

Nocccer (changes requested):
I really like how you writing code :). Pretty clean and well readable. Still i have some suggestions, which needs to be discussed.

electron/console.ts
dragonDScript: I think that the current solution is more idiomatic. Think of window.console or electron-log. They do the same.

dragonDScript: There are docs already :)

If you think that docs are missing, where? Thank you.

electron/logger.ts
dragonDScript: :+1: Sure

electron/console.ts
dragonDScript: I've just added documentation. If you think that docs are missing yet, tell me where! Thank you.

flavioislima (dismissed):
Lgtm 👍

dragonDScript:
> I will also look at it. Can you maybe write tests for the functions you implemented ?
> 

I'm not very good at writing tests. Console.ts already has tests (thanks to @Nocccer !).

flavioislima (approved):

dragonDScript:
I've made everything you guys told me to do. Can you (@adityaruplaha and @Nocccer ) review it? Thanks.

Nocccer:
I try to write a test for listenStdOut but i don't know how. Can you see my discord message. I really want to test everything before merging, else it could break some stuff makes all mad again, because everyone is doing only fixes. Please investigate how to write a test for that.

adityaruplaha (approved):

dragonDScript:
Do not merge yet! I am testing the performance & trying to test the code.

dragonDScript:
Performance is OK

Nocccer:
Ok let's merge then. I will come back to the test soon.

Nocccer (approved):

flavioislima (approved):

PR #424, opened 2021-06-11
More robust version checks.

adityaruplaha:
Now we can check for even the patch version.
Also, versions aren't floats, and they shouldn't be treated as such.

Also fixes the Python check, supersedes #420.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -37,6 +37,7 @@ import {
   handleExit,
   isOnline,
   openUrlOrFile,
+  semverGt,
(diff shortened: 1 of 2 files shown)


Nocccer (approved):
@adityaruplaha Looks good to me. So we can remove my python_fix branch ?
@flavioislima we can remove the python check later on anyways, but for a fast solution this should be good.

adityaruplaha:
@Nocccer yeah sounds good to me.

PR #427, opened 2021-06-11
Fix Stop/Kill Game Command

flavioislima:


electron/legendary/games.ts
```diff
@@ -389,15 +390,9 @@ class LegendaryGame extends Game {
...
-    const { install : { install_path, executable} } = await this.getGameInfo()
-    const exe = install_path + '/' + executable
-    console.log('killing', this.appName)
-    return await execAsync(`pkill -ef ${exe}`).then((v) => {
-      this.state.status = 'done'
-      return v
-    })
+    const pattern = process.platform === 'darwin' ? 'legendary' : this.appName
+    console.log('killing', pattern)
+    return spawn('pkill', ['-f', pattern])
```
adityaruplaha, line 395: This returns the ChildProcess, which doesn't match the type annotation.

Considering doing what I suggested in [#385 (comment)](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/385#issuecomment-857705880).

adityaruplaha (changes requested):

electron/legendary/games.ts
flavioislima: Yeah but the issue is only the type?
Because I changed the type.
I think for stop, since we are killing the process, we don't need to use a Promise.
In the worst case, pkill might not exists on some distro and the command won't work, but we don't need to worry about every single use case and distros and configurations. We will go crazy like this.

adityaruplaha: The type annotation doesn't seem to be fixed though.

flavioislima: Ok, I fixed it to use spawn the right way, so the return of the function will be a console.log, a void.

adityaruplaha (approved):

PR #404, opened 2021-06-04
Some Fixes, Feedback and UI improvements

Linux4Good:
HI!

**Ignore commits before the last merge since my branch was out of date**

Through the issues list I found a few things to work with:

- [x] Update some packages
- [x] Files not removed when canceling installation #288 
- [x] Add Warning Proton Message
- [x] Add Error when Python version is lower than 3.8
- [x] Add Error message when Credentials have expired #306 
- [x] Improve Refresh Library Performance
- [x] Added Weblate Link to attract more translators 
- [x] Added Launch function to Protocol (using `heroic://launch/{appName}`)
- [x] Refactor Install/uninstall functions on the frontend.
- [x] Refactor Play Function on FE
- [x] Refactor Dialogs

PR #438, opened 2021-06-12
Fixed issues introduced by winston-logging

wiryfuture:
- Removed logging from unnecessary places.
- TS server still can't tell that the Logger is async >:(
- Added much needed breaklines where desperately needed
- Directory of logs made dependent on isDev; dev environments get ./logs, production gets heroicFolder/logs


electron/logger.ts
```diff
@@ -0,0 +1,78 @@
+import {
+  createLogger,
+  format
+} from 'winston'
+import { heroicFolder } from './constants'
+import DailyRotateFile from 'winston-daily-rotate-file'
+import isDev from 'electron-is-dev'
+// If we're developing, logs will be in ./logs for easy access, in production they'll be in the user's heroic folder
+const logsFolder: string = isDev ? './logs/' : heroicFolder + 'logs/'
```
flavioislima, line 9: stop doing strings like that, use ``` instead, like:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/fbf5d6e63f5f70141fc34f9e157cf855a61679c4/electron/constants.ts#L18

electron/main.ts
```diff
@@ -224,12 +225,12 @@ if (!gotTheLock) {
...
       if (app.setAsDefaultProtocolClient('heroic')) {
-        console.log('Registered protocol with OS.')
+        Logger.info({message: 'Registered protocol with OS', service: 'ipcMain::registerHeroicProtocol'})
       } else {
-        console.log('Failed to register protocol with OS.')
+        Logger.warn({message: 'Failed to register protocol with OS', service: 'ipcMain::registerHeroicProtocol'})
       }
     } else {
-      console.log('Protocol already registered.')
+      Logger.info({message: 'Protocol already registered with OS', service: 'ipcMain::registerHeroicProtocol'})
```
flavioislima, line 233: I think is unecessary and doesn't help anything since it will just flood the logs with the same message everytime we open heroic.
Logging success and error is enough.
Since the Logs will be used to help Debug, the most import part are the errors, not the confirmation that something was already done 1000 times :P

electron/main.ts
```diff
@@ -336,7 +333,12 @@ ipcMain.on('getLog', (event, appName) =>
   openUrlOrFile(`${heroicGamesConfigPath}${appName}-lastPlay.log`)
 )
 
+/*
+  Used to remove the game folder if someone stops game installation and chooses not to keep the files
```
flavioislima, line 337: This was fixed here: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/404
Unless the issue was introduced again.

electron/main.ts
```diff
@@ -359,7 +361,9 @@ interface Tools {
   wine: string
 }
 
+// Neither of these want to work anyway on my distro so I can't test this
```
flavioislima, line 364: you need to have winetricks and wine installed for this to work

electron/main.ts
```diff
@@ -374,65 +378,97 @@ ipcMain.on('callTool', async (event, { tool, wine, prefix, exe }: Tools) => {
...
 // Not ready to be used safely yet.
-ipcMain.handle('updateAll', () => LegendaryLibrary.get().updateAllGames())
+ipcMain.handle('updateAll', async () => {
+  await Logger.info({message: 'Updating all apps', service:'Game::updateAllGames'})
+  return LegendaryLibrary.get().updateAllGames()
+})
 
-ipcMain.handle('checkVersion', () => checkForUpdates())
+ipcMain.handle('checkVersion', async () => {
+  await Logger.info({message: 'Checking for a newer heroic version', service: 'ipcMain::checkVersion'})
```
flavioislima, line 399: I'm worried that all those await calls will make the app slower. Because we always need to write something to the logs first to then do something.
Would be good to benchmark those.
Especially on an HDD that is slower.

electron/main.ts
```diff
@@ -374,65 +378,97 @@ ipcMain.on('callTool', async (event, { tool, wine, prefix, exe }: Tools) => {
...
-  new BrowserWindow({ height: 700, width: 1200 }).loadURL(url)
-)
+ipcMain.on('createNewWindow', (e, url) =>{
+  return new BrowserWindow({ height: 700, width: 1200 }).loadURL(url)
+})
 
-ipcMain.handle('getGameInfo', async (event, game) => {
-  const obj = Game.get(game)
+ipcMain.handle('getGameInfo', async (event, appName) => {
+  await Logger.info({message: `Getting app info for ${appName} `, service: 'ipcMain::getGameInfo'})
```
flavioislima, line 415: This await being called for +100 or + 200 games, will make the app a lot slower, I'm sure of that.

flavioislima (changes requested):
I havent finished the review because I'm worried about the performance impact we will have here.
So first find a way of benchmarking, specially those functions that are called several times.
You can use console.time or performance (from nodejs), to do that.
After we make sure that the performance impact on HDDs especially, but also on SSD since all those writting will compromise the lifespan of the device, we can continue with the review.

electron/main.ts
wiryfuture: Should I remove this else as it isn't doing anything then?

electron/main.ts
wiryfuture: Weird, I just cloned main from before the merge of the logging pr so it should have been there 😬

electron/main.ts
wiryfuture: This is only ran when a user goes into the game's page, not when they're in the library.

electron/main.ts
wiryfuture: Left a comment in it instead

electron/main.ts
wiryfuture: As the writes are very small, I doubt it actually makes a noticeable difference. However, I'm now looking into doing some sort of worker process for the logger instead.

flavioislima:
I'll close this one for now since merged #423 and will be redundant. We can get back to it if the other logger doesn't work as we expect.

Nocccer:
I think we don't need winston. But we need ti change all console.log and cosnole.error with the new functions from console.ts

PR #442, opened 2021-06-13
Use $SHELL instead of bash in Linux and Mac

dragonDScript:
Uses the custom shell (provided by every user) instead of bash by default. In case it's not set, it uses bash.

- [ ] Created / Updated Tests
- [x] Created / Updated documentation

- [x] Login
- [x] Download a Game
- [x] Import a Game
- [x] Launch Default config
- [x] Launch wine
- [ ] Launch Proton (Doesn't work with main branch on my pc)
- [ ] Launch Custom Proton
- [x] Uninstall Game
- [x] Move Game

### DIFF
--- a/electron/constants.ts
+++ b/electron/constants.ts
@@ -45,7 +45,7 @@ function getShell() {
   default:
     // $SHELL is set by login
     // If it's 0-value, use bash
-    return process.env.SHELL || "/usr/bin/bash"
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
Looks good. This is one change that was in my list but I was always forgot. Thanks for doing.

PR #445, opened 2021-06-13
Fix launch protocol

dragonDScript:

- [ ] Created / Updated Tests

- [ ] Login
- [ ] Download a Game
- [ ] Import a Game
- [ ] Launch Default config
- [ ] Launch wine
- [ ] Launch Proton
- [ ] Launch Custom Proton
- [ ] Uninstall Game
- [ ] Move Game

### DIFF
--- a/electron/protocol.ts
+++ b/electron/protocol.ts
@@ -1,5 +1,6 @@
 import { BrowserWindow, dialog, nativeImage } from 'electron'
 import { Game } from './games'
+import i18next from 'i18next'
 
 export async function handleProtocol(window : BrowserWindow, url : string) {
   const [scheme, path] = url.split('://')
@@ -21,10 +22,10 @@ export async function handleProtocol(window : BrowserWindow, url : string) {
     if (!is_installed) {
       console.log(`ProtocolHandler: "${arg}" not installed.`)
       const diag = await dialog.showMessageBox(window, {
-        buttons: ['OK', 'Not now'],
+        buttons: [i18next.t('box.yes'), i18next.t('box.no')],
         cancelId: 1,
         icon: nativeImage.createFromDataURL(art_logo),
-        message: `${title} is not installed. Install it now?`,
+        message: `${title} ${i18next.t('box.protocol.install.not_installed')}`,
         title: title
       })
       if (diag.response === 0) {
--- a/electron/protocol.ts
+++ b/electron/protocol.ts
@@ -28,7 +28,7 @@ export async function handleProtocol(window : BrowserWindow, url : string) {
         title: title
       })
       if (diag.response === 0) {
-        game.install(`${process.env.HOME}/Games`)
+        game.install('default')
       }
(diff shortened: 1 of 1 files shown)


electron/protocol.ts
```diff
@@ -6,24 +6,18 @@ export async function handleProtocol(window : BrowserWindow, url : string) {
...
+    return console.log('Received ping! Arg:', arg)
   }
-  const appName = args.get('appName')
-  const game = Game.get(appName)
   if (command === 'launch') {
+    const game = Game.get(arg)
     const { is_installed } = await game.getGameInfo()
     if (!is_installed) {
-      console.log(`ProtocolHandler: "${appName}" not installed, ignoring launch request.`)
+      console.log(`ProtocolHandler: "${arg}" not installed, ignoring launch request.`)
```
flavioislima, line 18: We could open a new dialog here saying that the game is not installed and ask to install. 
then we could send the install command to the frontend with the path argument 'default'.

electron/protocol.ts
```diff
@@ -1,29 +1,51 @@
...
+      console.log(`ProtocolHandler: "${arg}" not installed.`)
+      const diag = await dialog.showMessageBox(window, {
+        buttons: [i18next.t('box.yes'), i18next.t('box.no')],
+        cancelId: 1,
+        icon: nativeImage.createFromDataURL(art_logo),
+        message: `${title} ${i18next.t('box.protocol.install.not_installed')}`,
+        title: title
+      })
+      if (diag.response === 0) {
+        window.webContents.send('install', {
```
flavioislima, line 32: No I think might be confusing what I said, but here you can send the install command only with the appname, and on the frontend, on the GlobalState.tsx file, that line I said above, you need to copy paste how we are handling the launch, but instead call the install command with these arguments, but no on the backend, on the frontend.
Its two different things.
or maybe just call install, appName here and I can handle the frontend later.

flavioislima: Yeah, change this line to `...send('install', app_name), and I'll handle the rest.

flavioislima (approved):
Good job!

adityaruplaha (changes requested):
Please retain the map implementation, passing multiple arguments will be difficult in the future otherwise.

dragonDScript:
> Please retain the map implementation, passing multiple arguments will be difficult in the future otherwise. 

Ok

flavioislima:
> Please retain the map implementation, passing multiple arguments will be difficult in the future otherwise.

No, I have to disagree with that. I don't see any use case for that map function.
99.9% of the time, the protocol will be used for launching game, once you  do your PR, you can adapt to the other needs you think we need, but for now, the current solution @dragonDScript did is perfect.

flavioislima:
It is also made to work with the #443 that he did.
So this should be enough.

dragonDScript:
In the future, though, if there is any use for arguments i Will add It myself.

PR #446, opened 2021-06-14
Fix error coverage report

Nocccer:
- [x] adding jest.config.js for frontend
- [x] adding jest.config.js for backend

- [ ] Created / Updated Tests

- [ ] Login
- [ ] Download a Game
- [ ] Import a Game
- [ ] Launch Default config
- [ ] Launch wine
- [ ] Launch Proton
- [ ] Launch Custom Proton
- [ ] Uninstall Game
- [ ] Move Game

### DIFF
--- a/electron/game_config.ts
+++ b/electron/game_config.ts
@@ -24,7 +24,24 @@ abstract class GameConfig {
 
   public abstract version : GameConfigVersion
 
-  public config : GameSettings
+  public config : GameSettings = {
+    audioFix: false,
+    autoInstallDxvk: false,
+    autoSyncSaves: false,
+    launcherArgs: '',
+    nvidiaPrime: false,
+    offlineMode: false,
+    otherOptions: '',
+    savesPath: '',
+    showFps: false,
+    showMangohud: false,
+    useGameMode: false,
+    winePrefix: '',
+    wineVersion: {
+      bin: '',
+      name: ''
+    }
+  }
 
   readonly appName : string
   readonly path : string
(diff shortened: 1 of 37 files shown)


flavioislima:
Looks good for this class but the others also must be inialialized to fix all errors.

electron/game_config.ts
```diff
@@ -24,7 +24,24 @@ abstract class GameConfig {
 
   public abstract version : GameConfigVersion
```
dragonDScript, line 25: Declare version

adityaruplaha (approved):
Looks good, I'm not 100% sure we can get away with just `if (config)` after this though. I would initially set it to null and test if it was not null and only then proceed. 

But everything seems to be fine, so LGTM.

flavioislima:
The build will never work if you use `tsconfig` like this, they need to be different.
You need to keep in mind that the electron and the src folder are two complete different worlds, its the same logic we have on a web app, being the electron the backend and the src with react the frontend. We can have similar things, types, and share a few patterns, but they are not the same, that's why we have two different `tsconfig` files.

electron/utils.ts
```diff
@@ -13,6 +13,7 @@ import {
 import { promisify } from 'util'
 import i18next from 'i18next'
 
+import { error } from 'console'
```
dragonDScript, line 16: Same here. You should use './console'

dragonDScript (changes requested):
Please fix these

PR #452, opened 2021-06-15
Implement Button to create Desktop Shortcut 

### LINKED ISSUE #447: Implement Button to create Desktop Shortcut
The implementation is already done on the backend, so just need to add some way to call it from the frontend.
Also we need an option under settings to create or not Desktop Shortcuts automatically, since now this is always true.

- [ ] A button on the game page to create a Desktop Shortcut to launch the game
- [ ] A Setting to enable/disabled create Desktop shortcuts automatically when installing.

dragonDScript:

- [x] Created / Updated Tests
- [ ] Created / Updated documentation (self-explaining code)

- [ ] Login
- [ ] Download a Game
- [ ] Import a Game
- [ ] Launch Default config
- [ ] Launch wine
- [ ] Launch Proton
- [ ] Launch Custom Proton
- [ ] Uninstall Game
- [ ] Move Game

Fixes #447


electron/legendary/games.ts
```diff
@@ -200,6 +201,51 @@ class LegendaryGame extends Game {
...
   }
 
+  /**
+   * Adds a desktop shortcut to $HOME/Desktop and to /usr/share/applications
+   * so that the game can be opened from the start menu and the desktop folder.
+   * Both can be disabled with enableDesktopShortcutsOnDesktop and enableDesktopShortcutsOnStartMenu
+   * @async
+   * @public
+   */
+  public async addDesktopShortcut() {
```
Nocccer, line 211: Would be nice if you can write test for this function. Wouldn't be that hard. Atleast not harder then mocking process.exit.

Nocccer (changes requested):
On frontend WineSettings and Settings page we lost some coverage. Please adapt the frontend test to get 100 % again.

dragonDScript:
Alright. Sure. Already in my todos.

electron/legendary/games.ts
dragonDScript: I don't know...We'd need to mock this.getGameInfo first. Don't we?

dragonDScript: I will try to do so.

adityaruplaha (commented):
Soft-approving this, looks good at a glance but I haven't tested this or gone deep into the weeds.

dragonDScript:
> Soft-approving this, looks good at a glance but I haven't tested this or gone deep into the weeds.

Thank you very much!

src/screens/Game/GameSubMenu/index.tsx
```diff
@@ -140,6 +140,12 @@ export default function GamesSubmenu({
           >
             {t('submenu.log')}
           </span>
+          <span
+            onClick={() => ipcRenderer.send('addShortcut', appName)}
+            className="hidden link"
+          >
+            {t`submenu.addShortcut`}
```
flavioislima, line 147: this doesnt look right, should be:
{`${t('submenu.addShortcur', 'Add Shortcut to Desktop')}`}

Besides the function call, its good always to have a default.
After adding new i18n keys, is also good to run:
`yarn i18n` on the terminal, otherwise this new key wont be generated and weblate wont recognize it.

src/screens/Settings/components/GeneralSettings/index.tsx
```diff
@@ -246,6 +252,23 @@ export default function GeneralSettings({
       </span>
       <span className="setting">
         <span className="toggleWrapper">
+          {t('setting.adddesktopshortcuts', 'Add desktop shortcuts automatically')} (Linux)
```
flavioislima, line 255: Maybe it's better to add a platform check here. We have a platform from contextProvider, so just check if is Linux and render those options conditionally.

flavioislima: Check line 72 for an example

src/screens/Settings/index.tsx
```diff
@@ -274,6 +276,10 @@ function Settings() {
               setMaxWorkers={setMaxWorkers}
               toggleDarkTrayIcon={toggleDarkTrayIcon}
               darkTrayIcon={darkTrayIcon}
+              addGamesToStartMenu={addGamesToStartMenu}
+              addDesktopShortcuts={addDesktopShortcuts}
+              toggleAddDesktopShortcuts={() => setAddDesktopShortcuts(!addDesktopShortcuts)}
```
flavioislima, line 281: WE have that toggle hook for a reason, because using like this can lead to errors when writing the file if the user click several times. So use the same hook we have for the other toggles instead of using only `userState`.

flavioislima (changes requested):
Overall looks good but need to adjust a few more things to be perfect :D

src/screens/Game/GameSubMenu/index.tsx
dragonDScript: Ok

src/screens/Settings/components/GeneralSettings/index.tsx
dragonDScript: Ok

src/screens/Settings/index.tsx
dragonDScript: ok

src/screens/Settings/components/GeneralSettings/index.tsx
dragonDScript: What if I disable the field if platform != linux? and keep the (Linux) parenthesis.?

flavioislima: But why? Since it will appear only for Linux?

Nocccer: What about cheking (!isWindows) as we do for other options instead of implementing now (Linux). Makes the code more readable.

dragonDScript: > But why? Since it will appear only for Linux?

This way they'll know that the feature exists and they'll not ask "hey, add x"

dragonDScript: > What about cheking (!isWindows) as we do for other options instead of implementing now (Linux). Makes the code more readable.

This will make mac support even harder.

flavioislima: > > But why? Since it will appear only for Linux?
> 
> This way they'll know that the feature exists and they'll not ask "hey, add x"

So let it be the way it is, for all platforms and before next release I'll make it work for Windows and mac.
Don't add the Linux in parenthesis and also let is show for all platforms.
After you merge this I'll check how we can do on mac and Windows. 🙂

dragonDScript:
Continue with this on #459.

PR #454, opened 2021-06-16
os/windows: Windows improvements

flavioislima:
Did some changes already on main by mistake.
The rest of this PR is basically just some adjustments on the EGS sync feature and some other type checks

- [ ] Created / Updated Tests

- [x] Login
- [x] Download a Game
- [x] Import a Game
- [x] Launch Default config
- [ ] Launch wine - Not Aplicable
- [ ] Launch Proton - Not Aplicable
- [ ] Launch Custom Proton - Not Aplicable
- [ ] Launch Tools (Winetricks and Winecfg) - Not Aplicable
- [x] Uninstall Game
- [x] Move Game


src/screens/Settings/components/GeneralSettings/index.test.tsx
```diff
@@ -45,7 +45,7 @@ async function renderGeneralSettings(props: Partial<Props> = {})
     toggleDarkTrayIcon: () => {return;},
     toggleTray: () => {return;}
   };
-  const returnvalue = await render(<GeneralSettings {...{...defaultprops, ...props}} />);
+  const returnvalue = render(<GeneralSettings {...{...defaultprops, ...props}} />);
```
dragonDScript, line 48: Why?

dragonDScript (commented):
Aside from the comment I left over there, this PR looks good.

src/screens/Settings/components/GeneralSettings/index.test.tsx
adityaruplaha: Same question here.

src/types.ts
```diff
@@ -38,7 +38,7 @@ export interface ContextType {
   handleSearch: (input: string) => void
   layout: string
   libraryStatus: GameStatus[]
-  platform: string
+  platform: NodeJS.Platform | string
```
adityaruplaha, line 41: Can you safely drop the `| string` or change it to `| 'unknown'`?

dragonDScript: No, process.platform always returns a | string so we should conserve it.

src/screens/Settings/components/GeneralSettings/index.test.tsx
flavioislima: Simple, because render is not an async function.
On your VScode cant you see like 3 dots on the await word and putting the mouse over it says: the await here does any difference, or something like this?

src/types.ts
flavioislima: @adityaruplaha yes
@dragonDScript no, because this here is the function we receive from the Main process, so we are not calling the process itself from the frontend since it is impossible.

flavioislima: Remember, this is just a type, this should help us to understand what the function returns

adityaruplaha (approved):
LGTM

src/types.ts
flavioislima: @adityaruplaha this doesnt work, it says can't accept a string, because 'unknown' is a string, I think its something with TS that I dont want to go deeper into.

adityaruplaha (approved):
sudo LGTM

PR #480, opened 2021-06-25
Update game from clickable text on GamePage

flavioislima:
Got the changes from @dragonDScript branch, tested and fixed some issues I found

- [x] Created / Updated Tests
- [x] Created / Updated documentation

- [x] Login
- [x] Download a Game
- [x] Import a Game
- [x] Launch Default config
- [x] Launch wine
- [x] Launch Proton
- [x] Launch Custom Proton
- [x] Uninstall Game
- [x] Move Game


dragonDScript:
What is the color from var(--text-secondary) Could you attach a screenshot of how it looks?

dragonDScript:
Following with my PR:


No. A button is a thing you can click and text is text. We shouldn't mix it. We can have both if you want. But a button is a clear signal of a clickable action.

flavioislima:
> Following with my PR:
> 
> No. A button is a thing you can click and text is text. We shouldn't mix it. We can have both if you want. But a button is a clear signal of a clickable action.

No I don't think we need another button, clicking on the text looks good so far.
Also, you got some things wrong maybe because of some lack of explanation but we don't wanna remove the play button and replace it with an update button like you did. Because its possible to play the game without an update.
Second, I'm busy now but the secondary color is close to white, so the text comes from green to a greenish white on hover, not a big deal and also this is only for this old design, on the new design we will need to thing on something else.

dragonDScript:
Most users will think  its not clickable.

flavioislima:
> Most users will think its not clickable.

That's why I added another text: 'Click to Open'

flavioislima:
> Most users will think its not clickable.

Also, 3 buttons there will be a mess. We will have time to think on a better soltuion when working on the new design

dragonDScript:
It's unnecesary. Every launcher uses buttons for a reason.

flavioislima:
> It's unnecesary. Every launcher uses buttons for a reason.

Example? Steam just shows a button for that. Epic official launcher doesnt have anything since it update automatically. Origin its like steam, doesnt allow you to play without update.

flavioislima:
Anyway, lets think about this later, I'll merge this since @Nocccer needs to merge his as well.

PR #482, opened 2021-06-25
Windows fixes 2

flavioislima:
- [x] Fix Game folder not being deleted when stopping installation.
- [x] Fix some undefined messages on progress
- [x] Fix checking wine prefix and breaking game page when cloud saves is available

- [ ] Created / Updated Tests

- [x] Login
- [x] Download a Game
- [x] Import a Game
- [x] Launch Default config
- [x] Uninstall Game
- [x] Move Game


electron/legendary/games.ts
```diff
@@ -185,22 +185,37 @@ class LegendaryGame extends Game {
...
+    const workers = maxWorkers === 0 ? '' : ` --max-workers ${maxWorkers}`
     const logPath = `"${heroicGamesConfigPath}${this.appName}.log"`
     const writeLog = isWindows ? `2>&1 > ${logPath}` : `|& tee ${logPath}`
-    const command = `${legendaryBin} update ${this.appName} ${workers} -y ${writeLog}`
-
-    try {
-      return await execAsync(command, execOptions).then((v) => {
-        this.state.status = 'done'
-        return v
+    const command = `update ${this.appName}${workers} -y ${writeLog}`.split(' ')
```
Nocccer, line 191:
```suggestion
    const command = `update ${this.appName} ${workers} -y ${writeLog}`.split(' ')
```

electron/legendary/games.ts
```diff
@@ -185,22 +185,37 @@ class LegendaryGame extends Game {
   public async update() {
     this.state.status = 'updating'
     const { maxWorkers } = (await GlobalConfig.get().getSettings())
-    const workers = maxWorkers === 0 ? '' : `--max-workers ${maxWorkers}`
+    const workers = maxWorkers === 0 ? '' : ` --max-workers ${maxWorkers}`
```
Nocccer, line 188:
```suggestion
    const workers = maxWorkers === 0 ? '' : `--max-workers ${maxWorkers}`
```

electron/legendary/games.ts
```diff
@@ -185,22 +185,37 @@ class LegendaryGame extends Game {
...
+          const isVerifing = `${data}`.includes('Verification')
+          if (appName === this.appName){
+            if (isVerifing){
+              progress.bytes = `${String(data).split(' ')[2]}MiB`
+              progress.percent = `${data}`.split(' ')[3].split(')')[0].replace('(', '')
+              progress.eta = 'verifying'
+            }
+          }
+        })
+        console.log({progress})
```
Nocccer, line 211: Maybe also `logInfo` instead `console.log` ?

electron/main.ts
```diff
@@ -342,12 +342,15 @@ ipcMain.on('getLog', (event, appName) =>
 )
 
 ipcMain.on('removeFolder', async (e, [path, folderName]) => {
+  console.log({path});
```
Nocccer, line 345: Maybe also logInfo instead console.log ?

electron/main.ts
```diff
@@ -342,12 +342,15 @@ ipcMain.on('getLog', (event, appName) =>
...
   if (path === 'default') {
-    const defaultInstallPath = (await GlobalConfig.get()).config.defaultInstallPath.replaceAll("'", '')
-    const folderToDelete = `${defaultInstallPath}/${folderName}`
+    const { defaultInstallPath } = await GlobalConfig.get().getSettings()
+    const path = defaultInstallPath.replaceAll("'", '')
+    const folderToDelete = `${path}/${folderName}`
+    console.log({folderToDelete});
```
Nocccer, line 350: Maybe also logInfo instead console.log ?

Nocccer (changes requested):
I left some suggestions. Else looks really good.
Can't test the update behavior on linux, because i don't have games to update.

electron/legendary/games.ts
flavioislima: It was like this before but the thing is that it creates an empty space when the config is not set.

electron/legendary/games.ts
flavioislima: No because logInfo doesnt accept objects or anything other than strings.

flavioislima:
I'll drop the update function change for now and will try on another branch, this is out of scope

electron/legendary/games.ts
Nocccer: Maybe use then something like `join` instead of `$` syntax

electron/legendary/games.ts
Nocccer: progress is only strings.
logInfo can be
```typescript
logInfo([progress.bytes, progress.eta, progress.percent]);
```
or if there a short thing like progress.getKeyValues ?
```typescript
logInfo([progress.getKeyValues]);
```

flavioislima: Progress is not 'only strings', its an object. The values are strings but showing only the values doesnt make sense, we wouldn't know what is about.
We need to change those log functions to accept other types. Test if its an object and then call console log in between brackets, etc.
But I wont do that now.

Also, I'm not deploying this change anymore on this branch since its bugged and need more investigation.

electron/legendary/games.ts
flavioislima: How can I use join if I need to pass variables?
I think its completely readable. Its only one space that I added there because the output command is putting two spaces there.
Not that it was breaking anything, it was working before and its working now.

electron/main.ts
Nocccer: Is path a string ? I find this log really unreadable. What should tell me the log?
i would expect something like:
```typescript
const folderToDelete = path;
if (path === 'default') {
{
    const { defaultInstallPath } = await GlobalConfig.get().getSettings()
    const path = defaultInstallPath.replaceAll("'", '')
    folderToDelete = `${path}/${folderName}`
}
logInfo(`Removed folder ${folderToDelete}`)
```

flavioislima: i removed the log, check the updated files because you're commenting on old commits

PR #487, opened 2021-06-26
Refactor Update Function

flavioislima:
This PR solves one issue and introduces another bug that I'm trying for two days now and without success.

It fixes the fact that if an update has a repair routine first, the user will get feedback about it, it's also using spawn and not writing and reading log files.

The bad thing is that if you cancel the update and click on start update again, legendary will be called but there will be no feedback to the user, it will show as update canceled.
For some reason the process is called, the child is created but it dies at the same time, and then new processes are created and the frontend won't realize that because the promise was fulfilled.

Anyhow, I think the bug is not major since it's not common to start and stop updates like this.
Also, we have plans to update the games automatically soon.
So I think the benefits are good besides this bug.

OBS.: The changes were done only on the `games.ts`, on the `update` function. The other commits are from the other branch #482.

- [ ] Created / Updated Tests

- [x] Login
- [x] Download a Game
- [x] Import a Game
- [x] Launch Default config
- [x] Launch wine
- [x] Launch Proton
- [x] Launch Custom Proton
- [x] Uninstall Game
- [x] Move Game


electron/legendary/games.ts
```diff
@@ -185,22 +185,45 @@ class LegendaryGame extends Game {
...
-    const logPath = `"${heroicGamesConfigPath}${this.appName}.log"`
-    const writeLog = isWindows ? `2>&1 > ${logPath}` : `|& tee ${logPath}`
-    const command = `${legendaryBin} update ${this.appName} ${workers} -y ${writeLog}`
-
-    try {
-      return await execAsync(command, execOptions).then((v) => {
-        this.state.status = 'done'
-        return v
+    const workers = maxWorkers === 0 ? '' : ` --max-workers ${maxWorkers}`
+    const command = `update ${this.appName}${workers} -y`.split(' ')
```
Nocccer, line 189: Why you split here? Does spawn expect a array ?
Else you could make the command from the beginning an array of strings and append every thing which is defined.
Also solves the problem with the missing space for --max-workers, because you only append the argument if it is not 0.

Nocccer (changes requested):
I have some question, else looks good.

electron/legendary/games.ts
flavioislima: Yes, spawn expect and array of arguments.
Yeah but I'll still have and empty space on the command, it will be the same.

Nocccer (approved):

PR #515, opened 2021-07-03
Flatpak config

flavioislima:

- [ ] Created / Updated Tests

- [ ] Login
- [ ] Download a Game
- [ ] Import a Game
- [ ] Launch Default config
- [ ] Launch wine
- [ ] Launch Proton
- [ ] Launch Custom Proton
- [ ] Uninstall Game
- [ ] Move Game


iMonZ:
Any Update here?

flatpak/org.heroic.yml
```diff
@@ -0,0 +1,66 @@
...
+sdk: org.freedesktop.Sdk
+base: org.electronjs.Electron2.BaseApp
+base-version: '20.08'
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node14
+command: run.sh
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --socket=x11
```
iMonZ, line 13: There should be:
`- --socker=wayland`
Don't miss that

iMonZ (changes requested):
There should be:
`- --socker=wayland`
In  flatpak/org.heroic.yml 
Don't miss that

flavioislima:
> Any Update here?

So far: Heroic opens and I can Login, but it get stuck on the listing games part. I'm not sure if in a flatpak container Heroic will be able to execute legendary, it seems that is the issue but I need to investigate more.

Doomsdayrs:
Why not just compile legendary and bundle it with the application?

JakobDev:
> Why not just compile legendary and bundle it with the application?

It's against the rules of Flathub

flavioislima:
> > Why not just compile legendary and bundle it with the application?
> 
> It's against the rules of Flathub

Really? Didn't know that and Heroic already bundles the legendary binary. 
So we would need to add it as a dependency to be downloaded during the installation or legendary would need to be on flathub for us to do some kind of integration?

JakobDev:
Flathub [does not allow the use of pip install during instalations](https://discourse.flathub.org/t/is-the-use-of-pip-allowed/1857). The use of a network connection is not allowed during build I already ask for some of my Apps. You have to include all the whole dependency tree in the manifest. It's a lot of fun for me (not) to figure out how to build Qt from source or to figure out, why a build system that I never heard of before is not working during a Fltpak build.

Luckily for you, legendary depends only on requests, which is no problem to include in a Flatpak. More Luckily for you, [there is already a person who has a legendary Flatpak](https://gitlab.com/parcimonic/legendary-flatpak/-/tree/development), but without working wine yet. You should try to contact him.

electron/constants.ts
```diff
@@ -30,11 +31,18 @@ function getLegendaryBin() {
...
-const heroicFolder = `${home}/.config/heroic/`
+const flatPakHome = env.XDG_DATA_HOME.replace('/data', '')
+const home = isFlatpak ? flatPakHome : homedir()
+const legendaryConfigPath = isFlatpak
+  ? `${home}/config/legendary`
+  : `${home}/.config/legendary`
+console.log({ home, legendaryConfigPath })
+const heroicFolder = isFlatpak
+  ? `${home}/config/Heroic/`
+  : `${home}/.config/Heroic/`
```
catsout, line 45: @flavioislima 
You can use XDG_CONFIG_HOME, to get right config path in sandbox  
XDG_CONFIG_HOME -> ~/.var/app/<id>/config

flatpak/org.heroic.yml
```diff
@@ -0,0 +1,69 @@
+app-id: org.hgl.heroic
+runtime: org.freedesktop.Platform
+runtime-version: '20.08'
```
catsout, line 3: 21.08 is available

flatpak/org.heroic.yml
```diff
@@ -0,0 +1,69 @@
...
+      #  sha256: 5aa8112359ce892ae3e045a56ec90e8fe7a588139ae58ee52284586afac4c021
+      - type: dir
+        path: ..
+        dest: main
+      - generated-sources.json
+      # Wrapper to launch the app
+      - type: script
+        dest-filename: run.sh
+        commands:
+          - zypak-wrapper.sh /app/main/heroic "$@"
```
catsout, line 57:
```suggestion
          - export TMPDIR=$XDG_RUNTIME_DIR/app/$FLATPAK_ID
          - zypak-wrapper.sh /app/main/heroic "$@"
```
This is required for electron app to avoid some issues in flatpak

flatpak/org.heroic.yml
```diff
@@ -0,0 +1,69 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.node14
+command: run.sh
+separate-locales: false
+finish-args:
+  - --share=ipc
+  - --socket=x11
+  - --socket=wayland
+  - --socket=pulseaudio
+  - --share=network
```
catsout, line 16:
```suggestion
  - --allow=multiarch
  - --allow=devel
  - --share=network
```
For wine

catsout:
@flavioislima 
Why not open a pr on flathub?  
Flathub's people can help you there.

electron/constants.ts
flavioislima: Ah, I swear I looked for this one and could not find it. I'll check it.

flatpak/org.heroic.yml
flavioislima: Ok, will update!

flavioislima:
> @flavioislima Why not open a pr on flathub? Flathub's people can help you there.

Yes, good idea. We have a post on flathub forum but no one answered there yet.
I'll make more tests and will open a PR there.

mirkobrombin:
@flavioislima are you getting a log like this with a silent crash?

```bash
INFO: 
 Launch Command:              /app/main/resources/app.asar.unpacked/build/bin/linux/legendary launch a434dcb20f0d439b93aaa31dac9e3210   --wine '/var/home/mirko/.var/app/org.hgl.heroic/data/caffe-6.23/bin/wine64' --wine-prefix '/var/home/mirko/.var/app/org.hgl.heroic/Games/Heroic/Prefixes/Defense-Grid-The-Awakening'  
INFO: Checking for game updates
INFO: Found 0 game(s) to update
INFO: checking for heroic updates
hostname: invalid option -- 'f'
Try 'hostname --help' for more information.
INFO: Log was written
```

Firstly I was using the caffe wine runner from the bottles path but legendary was not able to find it (also giving specific permissions to the Heroic flatpak), so I've just moved it to the `org.hgl.heroic/data` path and it starts working. I'm able to launch the `winecfg` tool from Heroic but when starting a game it just silently crash with the above log. From the system monitor I can see the wine processes so legendary is able to reach and execute the binary.

I'm looking at the legendary [source code](https://github.com/derrod/legendary/blob/742d3a3b05f1fb0e5eef31a8797d0f1f00d65412/legendary/cli.py) and I see they set a lot of logs but any of these came in the output. Is this a normal Heroic behavior?

flavioislima:
> @flavioislima did you get a log like this with a silent crash?
> 
> ```shell
> INFO: 
>  Launch Command:              /app/main/resources/app.asar.unpacked/build/bin/linux/legendary launch a434dcb20f0d439b93aaa31dac9e3210   --wine '/var/home/mirko/.var/app/org.hgl.heroic/data/caffe-6.23/bin/wine64' --wine-prefix '/var/home/mirko/.var/app/org.hgl.heroic/Games/Heroic/Prefixes/Defense-Grid-The-Awakening'  
> INFO: Checking for game updates
> INFO: Found 0 game(s) to update
> INFO: checking for heroic updates
> hostname: invalid option -- 'f'
> Try 'hostname --help' for more information.
> INFO: Log was written
> ```
> 
> Firstly I was using the caffe wine runner from the bottles path but legendary was not able to find it (also giving specific permissions to the Heroic flatpak), so I've just moved it to the `org.hgl.heroic/data` path and it starts working. I'm able to launch the `winecfg` tool from Heroic but when starting a game it just silently crash with the above log. From the system monitor I can see the wine processes so legendary is able to reach and execute the binary.
> 
> I'm looking at the legendary [source code](https://github.com/derrod/legendary/blob/742d3a3b05f1fb0e5eef31a8797d0f1f00d65412/legendary/cli.py) and I see they set a lot of logs but any of these came in the output. Is this a normal Heroic behavior?
> 
You can check the whole log by going to the game page > tools > Latest log. So then you can see the whole Legendary Output.

> Regarding the packaging, you can also just extract the appimage with `unappimage` how [some users do](https://github.com/search?q=org%3Aflathub+unappimage&type=code).
That's a great tip so I can just ignore the build you say if I do something like this? would be great!

Thanks for helping with that

mirkobrombin:
Ok now I got a better log :D

```
wineserver: using server-side synchronization.
002c:fixme:winediag:LdrInitializeThunk Wine TkG (staging) 6.23 is a testing version containing experimental patches.
002c:fixme:winediag:LdrInitializeThunk Please don't report bugs about it on winehq.org and use https://github.com/Frogging-Family/wine-tkg-git/issues instead.
002c:err:wineboot:process_run_key Error running cmd L"C:\\windows\\system32\\winemenubuilder.exe -r" (2).
006c:err:module:open_builtin_so_file failed to load .so lib "/var/home/mirko/.var/app/org.hgl.heroic/data/caffe-6.23/lib/wine/x86_64-unix/wineusb.sys.so"
006c:err:ntoskrnl:ZwLoadDriver failed to create driver L"\\Registry\\Machine\\System\\CurrentControlSet\\Services\\wineusb": c0000142
0034:fixme:service:scmdatabase_autostart_services Auto-start service L"wineusb" failed to start: 1114
/lib/ld-linux.so.2: could not open
```

I was facing this problem on bottles too and was caused by missing i386 dependencies https://github.com/bottlesdevs/Bottles/blob/0f7f7a49fb37ea3d19a93eecaad5d0f798307126/com.usebottles.bottles.dev.yml#L21 

Idk if this will to the trick or your problem is different, I'll try

mirkobrombin:
I'm trying with the AppImage as source: heroic starts but I can't get into the library, as even though the access seems to go through (seeing the logs), it still takes me to the login view. I've tried both standard and SID login but I think the problem is related to something else.

This is my current manifest:

```yaml
id: org.hgl.heroic
sdk: org.freedesktop.Sdk
runtime: org.freedesktop.Platform
runtime-version: '21.08'
base: org.electronjs.Electron2.BaseApp
base-version: '21.08'
command: heroic-run
separate-locales: false

finish-args:
  - --share=ipc
  - --share=network
  - --socket=x11
  - --socket=wayland
  - --socket=pulseaudio
  - --allow=multiarch
  - --allow=devel
  - --talk-name=org.freedesktop.Notifications
  - --persist=.wine
  - --filesystem=host
  - --filesystem=xdg-data/Steam:ro
  - --filesystem=~/.var/app/com.valvesoftware.Steam:ro
  - --require-version=1.1.2
  
modules:
  - name: unappimage
    buildsystem: simple
    build-commands:
      - make -C squashfs-tools -j ${FLATPAK_BUILDER_N_JOBS} install INSTALL_DIR=/app/bin
    sources:
      - type: git
        url: https://github.com/refi64/unappimage
        commit: d7f86f2a0d7ec3a69211125207d5f127386b849a

  - name: heroic
    buildsystem: simple
    build-commands:
      - unappimage Heroic-2.0.2.AppImage
      - rm Heroic-2.0.2.AppImage
      - mv squashfs-root /app/bin/heroic
      - install -D heroic-run -t /app/bin
    sources:
      - type: script
        dest-filename: heroic-run
        commands:
          - zypak-wrapper /app/bin/heroic/heroic "$@"
          
      - type: file
        filename: Heroic-2.0.2.AppImage
        url: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/download/v2.0.2/Heroic-2.0.2.AppImage
        sha256: e20ab464270f231fc29053594cce40b6477f18feafd6ddcc3b6b626eb2a47194
```

image

flavioislima:
Thanks again to check on that. I could not test it as well since I'm sick for a few days now but I'll do as soon as I get better.

mirkobrombin:
Don't worry. As soon as I can I try to apply patches in this PR and see if it fixes.

mirkobrombin:
I did some tests with the latest release but the problem persists. Is it possible to build a release (maybe an artifact) which handle the flatpak paths for the config path?

TheEvilSkeleton:
You might want to directly push to Flathub. That's what OBS Studio did (https://github.com/obsproject/obs-studio/pull/5569), and currently Element is working on it https://github.com/vector-im/element-desktop/pull/293.

mirkobrombin:
> You might want to directly push to Flathub. That's what OBS Studio did ([obsproject/obs-studio#5569](https://github.com/obsproject/obs-studio/pull/5569)), and currently Element is working on it [vector-im/element-desktop#293](https://github.com/vector-im/element-desktop/pull/293).

But we are really in early state, the application opens and lets you log in and access the library, but the whole part of legendary is broken. Isn't it convenient to open the PR on Flathub when there is a minimally functional version?

---

PS. I'm not able to reach legendary logs. Is there a verbose mode in heroic?

tim77:
> Isn't it convenient to open the PR on Flathub when there is a minimally functional version?

I believe yes and this is not a problem. It's better to do this from the start and work on build with other members which could help and review.

mirkobrombin:
> > Isn't it convenient to open the PR on Flathub when there is a minimally functional version?
> 
> I believe yes and this is not a problem. It's better to do this from the start and work on build with other members which could help and review.

It definitely makes sense +1

flavioislima:
> > Isn't it convenient to open the PR on Flathub when there is a minimally functional version?
> 
> I believe yes and this is not a problem. It's better to do this from the start and work on build with other members which could help and review.

Yes, great idea. I will try to fix at least the paths today and generate build with basic functionalities and then we can go through the other issues. At least 'on my machine it works' 😅 . So I will check what might be going on with the paths and the other issues with dependencies and wine not being able to create window, etc. We can get help on flathub. :)

flatpak/flatpak-node-generator.py
```diff
@@ -0,0 +1,1839 @@
+#!/usr/bin/env python3
```
gasinvein, line 1: `flatpak-node-generator` is updated often to catch up with nodejs ~quirks~ updates. If you want it to be part of you app repo, you probably better of adding flatpak-builder-tools as a submodule instead of committing it directly in the app tree.

flatpak/org.heroic.yml
```diff
@@ -0,0 +1,65 @@
...
+  - --talk-name=org.freedesktop.Notifications
+  - --persist=.wine
+  - --filesystem=xdg-data/Steam:ro
+  - --filesystem=~/.var/app/com.valvesoftware.Steam:ro
+build-options:
+  append-path: /usr/lib/sdk/node16/bin
+  cflags: -O2 -g
+  cxxflags: -O2 -g
+  env:
+    NPM_CONFIG_LOGLEVEL: info
```
gasinvein, line 29: This isn't be needed.

flatpak/org.heroic.yml
```diff
@@ -0,0 +1,65 @@
...
+modules:
+  - name: heroic
+    buildsystem: simple
+    build-options:
+      env:
+        XDG_CACHE_HOME: /run/build/heroic/flatpak-node/cache
+        npm_config_cache: /run/build/heroic/flatpak-node/node-cache
+        npm_config_nodedir: /usr/lib/sdk/node16
+        npm_config_offline: 'true'
```
gasinvein, line 38: This shouldn't be needed for Yarn-based builds.

flatpak/org.heroic.yml
```diff
@@ -0,0 +1,65 @@
...
+      - generated-sources.json
+      # Wrapper to launch the app
+      - type: script
+        dest-filename: run.sh
+        commands:
+          - export TMPDIR=$XDG_RUNTIME_DIR/app/$FLATPAK_ID
+          - zypak-wrapper.sh /app/main/heroic "$@"
+    build-commands:
+      # Install npm dependencies
+      - HOME=$PWD yarn config --offline set yarn-offline-mirror $FLATPAK_BUILDER_BUILDDIR/flatpak-node/yarn-mirror
```
gasinvein, line 56: `HOME=$PWD` is a rather fragile hack. Instead, you can add an inline source with yarn config, like this:
```yaml
type: inline
contents: |
  yarn-offline-mirror /run/build/heroic/flatpak-node/yarn-mirror
dest: main
dest-filename: .yarnrc
```

flatpak/org.heroic.yml
```diff
@@ -0,0 +1,65 @@
...
+      #  sha256: 5aa8112359ce892ae3e045a56ec90e8fe7a588139ae58ee52284586afac4c021
+      - type: dir
+        path: ..
+        dest: main
+      - generated-sources.json
+      # Wrapper to launch the app
+      - type: script
+        dest-filename: run.sh
+        commands:
+          - export TMPDIR=$XDG_RUNTIME_DIR/app/$FLATPAK_ID
```
gasinvein, line 52: If large files are to be stored in `$TMPDIR`, better set it to persistent storage, e.g. `$XDG_CACHE_HOME/tmp`

flatpak/org.heroic.yml
```diff
@@ -0,0 +1,65 @@
...
+          - zypak-wrapper.sh /app/main/heroic "$@"
+    build-commands:
+      # Install npm dependencies
+      - HOME=$PWD yarn config --offline set yarn-offline-mirror $FLATPAK_BUILDER_BUILDDIR/flatpak-node/yarn-mirror
+      - yarn --offline
+      # Build the app
+      - |
+        . ../flatpak-node/electron-builder-arch-args.sh
+        yarn dist-flatpak
```
gasinvein, line 61: Since you don't pass `$ELECTRON_BUILDER_ARCH_ARGS` to `yarn`, you can remove the bach source thing
```suggestion
      - yarn dist-flatpak
```
Or, if you want the args, actually pass them to yarn
```suggestion
      - |
        . ../flatpak-node/electron-builder-arch-args.sh
        yarn dist-flatpak -- $ELECTRON_BUILDER_ARCH_ARGS
```

gasinvein (commented):
Here are some tips regarding building an electron app in flatpak

flatpak/org.heroic.yml
flavioislima: What about the idea to use the AppImage instead of budding it with Yarn?

gasinvein: You can repackage a binary package, sure, but it's safe only as long as the app doesn't have native node modules (otherwise no ABI compatibility guarantees).

flavioislima:
@mirkobrombin @gasinvein 
Hey guys!
I finally could fix the issues with the build (wrong config path, etc). 
most of the things are working now except launching games (wine issue)
But before opening a PR on Flathub can you guys take a look one more time if something can be improved?

Thanks :D

mirkobrombin:
> @mirkobrombin @gasinvein Hey guys! I finally could fix the issues with the build (wrong config path, etc). most of the things are working now except launching games (wine issue) But before opening a PR on Flathub can you guys take a look one more time if something can be improved?
> 
> Thanks :D

I'll take a look at it this weekend. Do legendary logs work now?

flavioislima:
> I'll take a look at it this weekend. Do legendary logs work now?

Yes, I can see the output on the log screen now.

mirkobrombin:
Nice 😎

flavioislima:
> Just an update: I've made some tests but legendary keep crashing. I'm trying to use legendary as flatpak to understand if it's a heroic problem or legendary one.

That's weird. can you see any log of those crashes?
can you try removing the `.config/legendary` and `.config/heroic` folders under `.var/app/org.hgl.heroic/config/` ?

mirkobrombin:
> > Just an update: I've made some tests but legendary keep crashing. I'm trying to use legendary as flatpak to understand if it's a heroic problem or legendary one.
> 
> That's weird. can you see any log of those crashes? can you try removing the `.config/legendary` and `.config/heroic` folders under `.var/app/org.hgl.heroic/config/` ?

I had some permission overrides from old tests. After cleaning up everything including the configuration files, I am no longer able to log in via Epic

TheEvilSkeleton:
> Flathub [does not allow the use of pip install during instalations](https://discourse.flathub.org/t/is-the-use-of-pip-allowed/1857). The use of a network connection is not allowed during build I already ask for some of my Apps. You have to include all the whole dependency tree in the manifest. It's a lot of fun for me (not) to figure out how to build Qt from source or to figure out, why a build system that I never heard of before is not working during a Fltpak build.
> 
> Luckily for you, legendary depends only on requests, which is no problem to include in a Flatpak. More Luckily for you, [there is already a person who has a legendary Flatpak](https://gitlab.com/parcimonic/legendary-flatpak/-/tree/development), but without working wine yet. You should try to contact him.

FYI Flathub only prohibits network access during build time. This doesn't mean you cannot download the dependencies beforehand. You can use [flatpak-pip-generator](https://github.com/flatpak/flatpak-builder-tools/tree/master/pip) to create a manifest that downloads these dependencies. Perhaps this should make things easier?

PR #536, opened 2021-07-13
Cache user library

flavioislima:
This changed several requests from the frontend to the store, making Heroic a lot more responsive and avoiding unnecessary requests.
What was done:
- [x] Create a new store for the library called `libraryStore`.
- [x] Fill the library with the games information and its statuses.
- [x] Change the frontend requests to check the store first instead of calling legendary right away.
- [x] Set the legendary calls to run on the background and check for the files changes.
- [x] Update the library when there are changes on the legendary files and folders.

With this, Heroic doesn't need to reload anymore after an install or uninstall, making the UI feel a lot more responsive.
There are still a few loads for when checking the game's updates, other than that most of the time the changes are doing without it.

About tests, we need to implement a mocked store and I could not figure it out how to do it for now so I'm having to skip a few tests until we fix this. @Nocccer 


dragonDScript (approved):
I don't have a lot of time to review, but looks good to me! Good job!

PR #572, opened 2021-07-28
[Backend] Add updater (2nd attempt)

### LINKED ISSUE #563: Automatically update the launcher in the background
Currently you need to download the .AppImage manually, It'd be nice to have auto updater functionality

Every X minutes check whetever there are available updates (time can be customized). If there are updates, download them in the background. Then ask the user if he wants to restart Heroic, if he wants, restart & install. If not, ask him again in 10 minutes. 

Plus: Have a "Check updates" button that does this.

I've not considered alternatives.

**Note: It is possible to auto-update on Linux appimages, snaps, and flatpak (not sure about this last one) via electron builder's autoUpdater's "AppImageUpdater"**

See the snippet below:
```ts
import { NsisUpdater } from "electron-updater"
// Or MacUpdater, AppImageUpdater
```

https://www.electron.build/auto-update.html

dragonDScript:
Closes #563 

### DIFF
--- a/electron/config.ts
+++ b/electron/config.ts
@@ -321,6 +321,8 @@ class GlobalConfigV0 extends GlobalConfig {
     const defaultWine = isWindows ? {} : await this.getDefaultWine()
 
     return {
+      checkUpdatesInterval: 10,
+      enableUpdates: true,
       addDesktopShortcuts: false,
       addStartMenuShortcuts: false,
       autoInstallDxvk: false,
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -82,7 +82,7 @@ const tsStore = new Store({
 })
 
 // Trigger the autoUpdater every X minutes
-const interval = setInterval(() => checkUpdates(), GlobalConfig.get().config.checkUpdatesInterval * 60000) // Converts minutes to milliseconds
+const interval = setInterval(() => checkUpdates(), (GlobalConfig.get().config?.checkUpdatesInterval || 10) * 60000) // Converts minutes to milliseconds
 
 if (GlobalConfig.get().config.enableUpdates === false) {
   clearInterval(interval)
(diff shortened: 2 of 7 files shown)


electron/config.ts
```diff
@@ -321,6 +321,8 @@ class GlobalConfigV0 extends GlobalConfig {
     const defaultWine = isWindows ? {} : await this.getDefaultWine()
 
     return {
+      checkUpdatesInterval: 10,
+      enableUpdates: true,
```
flavioislima, line 325: I saw that you did add the option to the config file but haven't on the frontend for the user to change it.

dragonDScript: I will add that in a second PR. I don't want to make this one too big, or to get many conflicts.

flavioislima (approved):
Ok looks good then

electron/main.ts
```diff
@@ -80,6 +81,13 @@ const tsStore = new Store({
   name: 'timestamp'
 })
 
+// Trigger the autoUpdater every X minutes
+const interval = setInterval(() => checkUpdates(), (GlobalConfig.get().config?.checkUpdatesInterval || 10) * 60000) // Converts minutes to milliseconds
+
+if (GlobalConfig.get().config.enableUpdates === false) {
```
flavioislima, line 87: @dragonDScript I think this line is the problem, just remove the `=== false`its unnecessary since if nothing it will return false anyway.

dragonDScript: It should be true by default, so if it's undefined JS will interpret that it's false

flavioislima: So then do: 
`GlobalConfig.get().config.enableUpdates === false || true`

flavioislima: Also, I dont know if updates should be forced, we have a culture in Linux that the user chooses to upgrade or not, some things I upgrade only if it looks safe, sometimes updates breaks stuff

flavioislima: so keep it as false by default the autoupdate, leave only the "checkUpdates" as true by default

dragonDScript: We should ask our users in the Discord. I'll leave the default as false by default by now

PR #579, opened 2021-07-31
Fix logs

dragonDScript:
Logs are finally working!

They are saved to:
`$HOME/.config/heroic/Crash Reports` on Linux

For example,
`/home/dragon/.config/Heroic/Crash Reports/Sat Jul 31 2021.txt`

# Note that

If you press Ctrl+C from the terminal twice, the operation to save the logs might be aborted. Please close the window or press Ctrl+C once.


flavioislima (approved):

flavioislima:
Looks good!

PR #583, opened 2021-08-02
feat(electron): Removed old check updates system and use the new one

dragonDScript:

### DIFF
--- a/src/state/GlobalState.tsx
+++ b/src/state/GlobalState.tsx
@@ -266,21 +266,9 @@ export class GlobalState extends PureComponent<Props> {
   }
 
   checkVersion = async () => {
-    const { t } = this.props
     const newVersion = await ipcRenderer.invoke('checkVersion')
     if (newVersion) {
-      const { response } = await ipcRenderer.invoke('openMessageBox', {
-        buttons: [t('box.yes'), t('box.no')],
-        message: t(
-          'box.appupdate.message',
-          'There is a new version of Heroic Available, do you want to update now?'
-        ),
-        title: t('box.appupdate.title', 'Update Available')
-      })
-
-      if (response === 0) {
-        ipcRenderer.send('openReleases')
-      }
+      ipcRenderer.send('updateHeroic')
     }
   }
 


src/state/GlobalState.tsx
```diff
@@ -266,21 +266,9 @@ export class GlobalState extends PureComponent<Props> {
   }
 
   checkVersion = async () => {
-    const { t } = this.props
     const newVersion = await ipcRenderer.invoke('checkVersion')
     if (newVersion) {
-      const { response } = await ipcRenderer.invoke('openMessageBox', {
```
flavioislima, line 272: In this case, what happens on the other versions besides appImage?
I also forgot to ask if the information about the updates on the installable version. How are we handling the check if the user is using appImage or has installed heroic in the system?

src/state/GlobalState.tsx
```diff
@@ -266,21 +266,9 @@ export class GlobalState extends PureComponent<Props> {
...
-          'box.appupdate.message',
-          'There is a new version of Heroic Available, do you want to update now?'
-        ),
-        title: t('box.appupdate.title', 'Update Available')
-      })
-
-      if (response === 0) {
-        ipcRenderer.send('openReleases')
-      }
+      ipcRenderer.send('updateHeroic')
```
flavioislima, line 271: so you will need some kind of check here, to see if its appImage or not. probably seeing the path, if it starts with /tmp/ will return true if appImage. But I don't know about installed appImages, maybe they will go to /opt, then it won't work.

src/state/GlobalState.tsx
dragonDScript: See https://stackoverflow.com/questions/66974771/see-build-target-of-electron-builder-in-running-application

Closing

flavioislima (approved):

src/state/GlobalState.tsx
dragonDScript: Hmmm...maybe check if it's not an app image? like, appimages aren't on /usr, they are on home or the hdd

PR #585, opened 2021-08-02
MAD Repo

rauldipeas:
Added command line to add **MAD Linux** repository to Ubuntu and derivatives systems, to get easily updates.

### DIFF
--- a/madrepo.sh
+++ b/madrepo.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 echo 'deb Https://sourceforge.net/projects/madlinux/files/repo core main'|sudo tee /etc/apt/sources.list.d/madlinux.list
 wget -qO- Https://sourceforge.net/projects/madlinux/files/repo/madlinux.key|gpg --dearmor|sudo tee /etc/apt/trusted.gpg.d/madlinux.gpg>/dev/null
 sudo add-apt-repository -y ppa:apt-fast/stable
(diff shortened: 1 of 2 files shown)


madrepo.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+echo 'deb Https://sourceforge.net/projects/madlinux/files/repo core main'|sudo tee /etc/apt/sources.list.d/madlinux.list
+wget -qO- Https://sourceforge.net/projects/madlinux/files/repo/madlinux.key|gpg --dearmor|sudo tee /etc/apt/trusted.gpg.d/madlinux.gpg>/dev/null
+sudo add-apt-repository -y ppa:apt-fast/stable
+sudo apt install -y apt-fast
```
flavioislima, line 6: Do you really think adding apt fast it's necessary? Because this would be one more dependency to install heroic.

madrepo.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+echo 'deb Https://sourceforge.net/projects/madlinux/files/repo core main'|sudo tee /etc/apt/sources.list.d/madlinux.list
```
flavioislima, line 3: I tried here on pop_os and got an error because of the capital H, so should be `https:` instead of `Https:`, no?

flavioislima:
Left a few comments, other than that looks good

madrepo.sh
rauldipeas: SourceForge is very slow in Brazil, that's why i include it.

madrepo.sh
rauldipeas: Sorry for this, i'll fix it.

madrepo.sh
flavioislima: yes but since this is not a rule of thumb everywhere, maybe we should add information that using apt-fast might be better if the downloads are too slow.
And we can check on the script if there is apt-fast or not, if yes, we use it, if not, use normal apt. What do you think?

rauldipeas: Well, it can be, i'll made the changes.

flavioislima (approved):
Nice! Thanks for that!

PR #621, opened 2021-08-18
Improve the developer experience on VS Code

dragonDScript:
- [x] Tested the feature and it's working on my computer (VS Codium 1.59.0).
- [x] Tested the main App features and they are still working.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,8 +2,8 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "command": "yarn test -i --watchAll",
-      "name": "Debug yarn test",
+      "command": "yarn react-start",
+      "name": "Start renderer process",
       "request": "launch",
       "type": "node-terminal"
     },
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,41 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"type": "npm",
+			"script": "dist",
+			"problemMatcher": [],
+			"label": "Test",
+			"detail": "yarn test",
+			"group": {
+				"kind": "test",
+				"isDefault": true
+			}
+		},
+		{
+			"type": "npm",
+			"script": "dist",
+			"problemMatcher": [],
+			"label": "Build for Linux",
+			"detail": "yarn build-electron && yarn build && electron-builder -c.extraMetadata.main=build/main.js --linux",
+			"group": {
+				"kind": "build",
+				"isDefault": true
+			}
+		},
+		{
+			"type": "npm",
+			"script": "dist",
+			"problemMatcher": [],
+			"label": "Build for Windows",
+			"detail": "yarn build-electron && yarn build && electron-builder -c.extraMetadata.main=build/main.js --win32"
+		},
+		{
+			"type": "npm",
+			"script": "dist",
+			"problemMatcher": [],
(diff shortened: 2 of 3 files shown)


.vscode/tasks.json
```diff
@@ -0,0 +1,41 @@
...
+				"kind": "test",
+				"isDefault": true
+			}
+		},
+		{
+			"type": "npm",
+			"script": "dist",
+			"problemMatcher": [],
+			"label": "Build for Linux",
+			"detail": "yarn build-electron && yarn build && electron-builder -c.extraMetadata.main=build/main.js --linux",
```
flavioislima, line 20: We are missing the targeting here. I'm not sure if the default is appImage, but if not, would be good to pass it after the `--linux`

README.md
```diff
@@ -81,6 +81,18 @@ Heroic is built with Web Technologies like: TypeScript, React, NodeJS and Electr
 
 ## Installation
 
+### Any OS (development environment)
+
+1. Download Yarn and Node.js
+2. Download the dependencies with `yarn`
+3. Start Heroic by pressing F5 on vs code or `yarn react-start && yarn electron`
```
flavioislima, line 88: In this case, pressing F5 will run only the react process? or will run both commands?

.vscode/launch.json
```diff
@@ -2,8 +2,8 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "command": "yarn test -i --watchAll",
-      "name": "Debug yarn test",
+      "command": "yarn react-start",
```
flavioislima, line 5: Because this will run only the react process and won't open electron

.vscode/tasks.json
dragonDScript: I think that it is a default since i generated the task from package.json. It'd be good to test it tho.

README.md
dragonDScript: It will run both commands, see .vscode/launch.json

.vscode/launch.json
dragonDScript: there is another task below, see 'Debug Main Process'. I already tested it.

README.md
dragonDScript: yes it is only running the first configuration working on a fix rn

.vscode/tasks.json
dragonDScript: confirmed that it generates an appimage

dragonDScript:
Everything done

Snivyer1910:
nice pr number btw

.vscode/tasks.json
```diff
@@ -0,0 +1,41 @@
...
+				"kind": "build",
+				"isDefault": true
+			}
+		},
+		{
+			"type": "npm",
+			"script": "dist",
+			"problemMatcher": [],
+			"label": "Build for Windows",
+			"detail": "yarn build-electron && yarn build && electron-builder -c.extraMetadata.main=build/main.js --win32"
```
flavioislima, line 31:
```suggestion
			"detail": "yarn build-electron && yarn build && electron-builder -c.extraMetadata.main=build/main.js --win"
```
To use the same command we have on package.json.

flavioislima:
Check if its possible to run the `yarn` command before every build.

dragonDScript:
> Check if its possible to run the `yarn` command before every build.

it's possible but only for linux since we would have to run bash or sh. It's better to run it in the shell directly.

PR #639, opened 2021-09-03
Update Ko-Fi link

Flonky:
Update Ko-Fi link

Testing not necessary


flavioislima:
Good catch!
I think your branch was not updated, that's why the conflicts. I added a downloads badge so that's might be the reason.

PR #641, opened 2021-09-04
Fix Korean language label on Language Selector

OctopusET:
"韓國語" same as "한국어" but in traditional Chinese characters.
When writing in Korean language, almost only Hangul (Korean character) is preferred.
So, like other languages' label, there is no need to add "韓國語".


flavioislima (approved):
Sounds good 👍

Issue #305, opened 2021-04-14
Show the free game of the week.


adityaruplaha:
A simple `openUrlOrFile` should suffice I feel.

dragonDScript:
We should show it in a separate window, like steam does

dragonDScript:
> A simple openUrlOrFile should suffice I feel.

What about a link to the launch protocol? If it's installed, it will be played. If not, it will prompt to install it.

flegald:
I would like to take this please.

flegald:
PR has been opened for this feature.

I did try and see if there was any way I could get the free games' info without scraping the page's HTML so we could have better integration into Heroic.

But I did not see anything exposed for this in legendary.

This means that we have to use Epic's own storefront for this popup unfortunately.

dragonDScript:
It's okay. At least they can get the free games.

wiryfuture:
@flavioislima I think this can be closed w/ the store page.

flavioislima:
Yes! 👍🏽

PR #653, opened 2021-09-15
Display EGS free products popup on startup

flegald:
# Display EGS free products on startup
issue: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/305

Added feature to open new window to (regionally correct) Epic Game's free games of the week page.

New window is a popup handled exactly the same as the store link. 

```
    new BrowserWindow({ height: 700, width: 1200 }).loadURL(freeProductsUrl)
```

The IPC render action occurs once at start up, and is triggered in `App.tsx`

This new feature is toggle-able with a new General Setting ToggleSwitch which defaults to ON.

freetoggle

freegamespopup

### Bug Fix 
There was also a bug I noticed regarding the behavior of `setCheckForUpdatesOnStartup` toggle.
Even when checked into the `OFF` position using the toggle, the setting always reverted back to `ON` by itself.
This was because the following line would always resolve to true.
```
setCheckForUpdatesOnStartup(config.checkForUpdatesOnStartup || true)
```
I have replaced the `true` with `false`

My assumption is this was done to have the default value be `true`.

But this seems to be accomplished [here](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/electron/config.ts#L329)

### DIFF
--- a/electron/config.ts
+++ b/electron/config.ts
@@ -329,6 +329,7 @@ class GlobalConfigV0 extends GlobalConfig {
       checkForUpdatesOnStartup: true,
       customWinePaths: isWindows ? null : [],
       defaultInstallPath: heroicInstallPath,
+      displayFreeProductsOnStartup: true,
       language: 'en',
       maxWorkers: 0,
       nvidiaPrime: false,
--- a/electron/constants.ts
+++ b/electron/constants.ts
@@ -38,6 +38,7 @@ const supportURL =
   'https://github.com/flavioislima/HeroicGamesLauncher/blob/main/Support.md'
 const discordLink = 'https://discord.gg/rHJ2uqdquK'
 const weblateUrl = 'https://hosted.weblate.org/projects/heroic-games-launcher'
+const epicStoreURL = 'https://www.epicgames.com/store/'
 
 /**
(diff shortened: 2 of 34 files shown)


Nocccer (approved):
LGTM.

But i would love more a new tab "Free Games" like the "Store" tab instead of a popup on start.

flegald:
> LGTM.
> 
> But i would love more a new tab "Free Games" like the "Store" tab instead of a popup on start.

Oh yeah for sure.

@flavioislima Do you have an opinion of this?

Nocccer:
What about two other options:
1. Make the "Store" tab a dropdown list which shows, "Home" and "Free Games of the Week". 
2. Fetch the Free Games from the site and show them directly in heroic launcher. (Don't know if this is possible?)

flegald:
I want to explore option 2. I originally thought it was not possible but I received an endpoint from CommandMC in Discord that may be able to do this. I agree it would be the nicest/classiest thing to do. I am going to edit this PR into WIP and see if we can get some integrated solution going.

flegald:
Removing this PR as it is no longer current

Issue #660, opened 2021-09-21
Improve support for CrossOver on Mac/Linux


mrpippy:
**Is your feature request related to a problem? Please describe.**
I work at CodeWeavers, we would like to make it easier for Heroic to launch games with CrossOver's Wine on Mac and Linux.

**Describe the solution you'd like**
I think the following changes would help with this:
- detect CrossOver installations and list them in the Wine Version dropdown (next to Proton, Lutris, etc.)
- add a setting for the user to specify the "bottle"/prefix for CrossOver to use
- fix any other bugs that arise

I have the first one working, and in-progress on the second. Will create a PR once ready.

**Describe alternatives you've considered**
An alternate method would be to run the Windows version of Heroic inside of Wine, this avoids any need to set up prefixes/Wine installs. Heroic itself runs (with a Wine-side hack needed for CEF), but Legendary is built with an older PyInstaller that doesn't work under Wine. Legendary upstream isn't interested in supporting this usage (https://github.com/derrod/legendary/issues/341)

**Additional context**

flavioislima:
Thanks for the initiative @mrpippy That would be great if it works. I do not play on mac so I don't have a Crossover installation (or wine) to test it. If you could do that would be of great value for Mac users. :)

Nocccer:
Implemented.

PR #664, opened 2021-09-22
Improve support for CrossOver

mrpippy:
To go with #660, this PR adds support for detecting and using CrossOver installations, and adds a setting to specify the CrossOver "bottle"/prefix that will be used.
(The setting defaults to "Heroic", in CrossOver the user will need to create a new bottle with this name)

Comments and suggestions are welcomed, this is my first time with JS/Electron.


dragonDScript:
I don't understand this PR, I'm sorry.

src/screens/Settings/components/WineSettings/index.tsx
```diff
@@ -189,6 +193,18 @@ export default function WineSettings({
           ))}
         </select>
       </span>
+      {wineVersion.name.includes('CrossOver') && <span className="setting">
+        <span className="settingText">{t('setting.winecrossoverbottle', 'CrossOver Bottle')}</span>
```
Nocccer, line 197: I think t('setting.winecrossoverbottle', 'CrossOver Bottle') needs to be set in the translation.json for each language. Just insert the english version for each language. People on webplate will translate it later.
Or if i'm wrong here @flavioislima ?

Nocccer (changes requested):
LGTM. Still found a small open point.

src/screens/Settings/components/WineSettings/index.tsx
flavioislima: You're wrong here :P
Actually, we should not touch the translations file. Just add the new key and then run: ´yarn i18n` and it will be created for all languages. Otherwise, we will have conflicts with the weblate branch.

electron/config.ts
```diff
@@ -186,12 +188,41 @@ abstract class GlobalConfig {
...
+              const version = info['CFBundleShortVersionString'] || ''
+              crossover.add({
+                bin: `'${crossoverMacPath}/Contents/SharedSupport/CrossOver/bin/wine'`,
+                name: `CrossOver ${version} - ${crossoverMacPath}`
+              })
+            }
+          })
+        })
+    }
+    else if (!isWindows)
```
flavioislima, line 209: This would be for Crossover for Linux, right?

electron/config.ts
```diff
@@ -186,12 +188,41 @@ abstract class GlobalConfig {
       }
     })
 
+    const crossover: Set<WineInstallation> = new Set()
```
flavioislima, line 191: I think for the lack of organization, we could put this logic into a separate function like we are doing with the`getDefaultWine` so maybe add one called `getCrossover` or something like this. 👍🏽

flavioislima (approved):
Looks really good. Just left a few comments in case you want to improve it.

src/screens/Settings/components/WineSettings/index.tsx
Nocccer: Ok good to know :).

electron/config.ts
mrpippy: Yes, I'll add a comment

electron/config.ts
mrpippy: Good point, I'll make that change too.

Nocccer (approved):
LGTM :)

flavioislima (approved):
That great work here, man!
Thanks for that. :D

If possible, could you just add a screenshot of the setting? Would be good just for the records, hehe.

mrpippy:
Sure, here's a screenshot. I should update the wiki too with a mention about creating the CrossOver bottle.

Screen Shot 2021-09-27 at 10 43 29 AM
Screen Shot 2021-09-27 at 10 43 16 AM

PR #665, opened 2021-09-22
Free products screen

flegald:
Screen to display free games currently on offer from Epic Games.

API endpoint: `https://store-site-backend-static.ak.epicgames.com/freeGamesPromotions` is used to obtain current promotion information

[Git Issue](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/305)

free-heroic


src/App.tsx
```diff
@@ -44,6 +45,11 @@ function App() {
               <Library library={library} />
             </div>
           </Route>
+          <Route exact path="/free-games">
```
Nocccer, line 48: Why not Route with component and also why not move the class "content" into the FreeProduct function ?

```suggestion
          <Route exact path="/free-games" component={FreeProducts} />
```

src/screens/FreeProducts/utils/apiResponseType.tsx
```diff
@@ -0,0 +1,143 @@
+/**
```
Nocccer, line 1: Do we really need all this interfaces ? 
Why not fetching the data you need in the backend and return a new interface which contains all information you need ?

Nocccer (changes requested):
Overall looks good. Still have some open suggestions/questions.

src/screens/FreeProducts/utils/apiResponseType.tsx
dragonDScript: Why not have all this?

dragonDScript (changes requested):
LGTM, but the frontend needs to be improved. I'd put it in the library.

src/screens/FreeProducts/utils/apiResponseType.tsx
flavioislima: I do agree that the ideal would be to do all requests on BE and just send the needed information to the frontend like we are doing with the other requests. This way we can avoid making the frontend too bloated. 👍🏽

flavioislima:
> LGTM, but the frontend needs to be improved. I'd put it in the library. Or in a separate window (popup) that is spawned when the app starts and when the user requests it. We should create a figma document and share our ideas there.

I think we can keep it here for now because since we are going to develop a sidebar, it will be only one more button there. 
But on the library, like a filter, I don't know, could work as well.

src/screens/FreeProducts/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const FreeProducts = () => {
+  const { t, i18n } = useTranslation()
+  let lang = i18n.language
+  if (i18n.language === 'pt') {
+    lang = 'pt-BR'
+  }
+  const [products, setProducts] = useState<Element[]>([])
+
+  const fetchProducts = async () => {
```
flavioislima, line 22: In my opinion, we could improve the `requestFreeProducts` request to return the `validProducts` already, most of this logic would work better on the backend and the frontend should be used only for rendering that response.

flavioislima (changes requested):
I think looks good overall, but the approach of treating the data on the frontend is not the best one I would say. So moving this logic to the BE could be better and then the frontend would just send a request and get the list and all the info needed to open the game page.

What do you think?

src/screens/FreeProducts/utils/apiResponseType.tsx
Nocccer: Doing exactly that with the wine ge implementation. Fetching information in the BE and just return the needed information.

dragonDScript:
> I think looks good overall, but the approach of treating the data on the frontend is not the best one I would say. So moving this logic to the BE could be better and then the frontend would just send a request and get the list and all the info needed to open the game page.
> 
> What do you think?

I think that we should split the backend and the frontend. Frontend needs to be refactored. So merge backend quickly and then work on frontend. what do you think?

flegald:
Sounds good. I will refactor the API response formatting to the backend and make a few smaller changes here and there. 
I do not know where exactly to put the page right now @Nocccer, I dont know if the Library is the best place?

electron/utils.ts
```diff
@@ -208,11 +208,18 @@ async function checkCommandVersion(
   return found;
 }
 
+async function fetchFreeProducts() {
```
Nocccer, line 211: Bring over all functions from FreeProducts/utils.tsx and prepare the data here.
The Return should be Element[].

dragonDScript:
I'm getting this error when cloning this branch: 
Unhandled Rejection (Error): Error invoking remote method 'requestFreeProducts': No handler registered for 'requestFreeProducts'

PR #695, opened 2021-10-12
Feature/free products screen

Phosphorus-M:
It's basically the same merge request that in #665 resolve the issue #305
I took some Commits of that branch to help in the update.
Just change some things that you request in the PR

image

The Image and the button open the link but I'm not sure, maybe it's better than the title could be clickable. 


src/screens/FreeProducts/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { IpcRenderer } from 'electron'
+import { FreeGameElement } from 'src/types'
+const { ipcRenderer } = window.require('electron') as {
+    ipcRenderer: IpcRenderer
+}
+
+const FreeProducts = () => {
+  const { t, i18n } = useTranslation()
+  let lang = i18n.language
+  if (i18n.language === 'pt') {
```
arielj, line 17: maybe this fix for the pt language can be moved inside the `generateLink` function? also, is this safe? can't this be `pt-PT` too for portugal?

src/screens/FreeProducts/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+              className='freeProductItem'
+              data-testid='test-free-product-image'
+            >
+              <a
+                onClick={() => createNewWindow(generateLink(p, lang))}
+                data-testid='test-free-product-link'
+              >
+                <img
+                  src={p.keyImages.filter(i => i.type === 'DieselStoreFrontWide')[0].url}
+                  style={{
```
arielj, line 48: there's no need to set background styles for an image

arielj (commented):
Hi! I'm new here, hope it's fine if I review stuff

This PR is adding a package-lock.json and updating the yarn.lock with a lot of `npm` urls. You should use yarn to run the project, not npm.

I added some more comments in the code.

flavioislima:
> Some new about this? Any comment?

I'll take a look. Tbh I was just thinking about if this is still relevant after we added the store inside the main Heroic window. Since the free games are there on the homepage. What do you think?

Phosphorus-M:
> > Some new about this? Any comment?
> 
> I'll take a look. Tbh I was just thinking about if this is still relevant after we added the store inside the main Heroic window. Since the free games are there on the homepage. What do you think?

I'm not sure if the decision is to keep it maybe I could redirect to the store but not open a new window but redirect to the store section with some parameters to show the game.

dragonDScript:
Did some research:
1. Request https://www.epicgames.com/store/en-US/free-games
2. `div.css-1myhtyb` contains the games
3. `div.css-1myhtyb` contains divs with the first element being an `a`, which redirects to the game page.
4. If `div.css-1myhtyb` contains a `div.css-1kggtxl` and it has a data-component which is equal to FreeOfferCard, the game is free and available. Either it's not yet ready.

dragonDScript:
Keep up the good work, as an idea we could show this in the library or in a separate window (popup)

flavioislima:
closing this one since the store is easily accessible from the main window now.

PR #722, opened 2021-10-31
Get the desktop folder from the system instead of hardcoding '/Desktop'

### LINKED ISSUE #721: Adding a game shortcut only creates the Start Menu shortcut but not the desktop shortcut on non-english system
**Describe the bug**

The `Add shortcut` action says `Shortcuts were created on Desktop and Start Menu`, but the shortcut is only added for the Start Menu.

The problem seems to be this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/b3418e38f76a7182ad8b87a994f1018f34d7ee39/electron/legendary/games.ts#L258

Which is using always `Desktop` as the directory name which, in my system (spanish), it's `Escritorio`.

This command can be used to get the name of the directory in linux `xdg-user-dir DESKTOP`, though I don't know if this affects other operating systems.

Note this also affects removing the desktop shortcut, since it also uses the hardcoded `Desktop` directory https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/b3418e38f76a7182ad8b87a994f1018f34d7ee39/electron/legendary/games.ts#L305

**To Reproduce**

1. Install a game
2. Click `Add shortcut` in the game details screen
3. See the success message
4. Check the Start Menu > Games (Icon is there)
5. Check the Desktop (Icon is not there)

**Expected behavior**

The Icon should be in both places.

- OS: Linux Mint
- Version: 20.1 Cinnamon
- Cinnamon version: 4.8.6
- Game: Not specific for one game, it happens with any game
- Wine/Proton version: Proton - Experimental
- Hardware specs (RAM, CPU, GPU): 16GB, AMD Ryzen 7, AMD 5700 XT

**Additional context**

System in Spanish (desktop dir is `Escritorio`, not `Desktop`).

arielj:
Use `xdg-user-dir DESKTOP` on linux systems to get the correct Desktop folder path that depends on the system language.

My system is in spanish, so my `Desktop` folder is actually `Escritorio`, using the `xdg-user-dir` command returns the full path to that directory.

This is only implemented for linux since adding shortcuts to desktop is only implemented on linux.

Closes #721

I couldn't add a test since I would have to stub the system command somehow which feels too hacky, but I accept any suggestion on how to add a test for this fix!

### DIFF
--- a/electron/constants.ts
+++ b/electron/constants.ts
@@ -1,5 +1,4 @@
 import { GlobalConfig } from './config'
-import { app } from 'electron'
 import {
   homedir,
   platform
@@ -20,7 +19,6 @@ const isMac = platform() === 'darwin'
 const isWindows = platform() === 'win32'
 const currentGameConfigVersion : GameConfigVersion = 'v0'
 const currentGlobalConfigVersion : GlobalConfigVersion = 'v0'
-const desktopFolder = app.getPath('desktop')
 const home = homedir()
 const legendaryConfigPath = `${home}/.config/legendary`
 const heroicFolder = `${home}/.config/heroic/`
@@ -89,7 +87,6 @@ const execOptions = {
 export {
   currentGameConfigVersion,
   currentGlobalConfigVersion,
-  desktopFolder,
   discordLink,
   execOptions,
   fixAsarPath,
--- a/electron/constants.ts
+++ b/electron/constants.ts
@@ -1,5 +1,5 @@
-import { execSync } from 'child_process'
 import { GlobalConfig } from './config'
+import { app } from 'electron'
 import {
   homedir,
   platform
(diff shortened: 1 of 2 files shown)


electron/constants.ts
```diff
@@ -15,10 +16,24 @@ function getLegendaryBin(){
   return bin
 }
 
+// Get the "Desktop" directory with multilingual support, fallback to "~/Desktop"
+function getDesktopPath() {
```
flavioislima, line 20: Instead of using this ou can simple get the desktop folder using:
`app.getPath('desktop')` importing `app` from electron.

flavioislima (changes requested):
Looks good but use the builtin electron function to the get desktop folder since it will work for other platforms as well.

electron/constants.ts
arielj: oh, first time ever using electron, I didn't know this, thanks! I'll update the PR

arielj: I couldn't use `app.getPath('desktop')` here in the constants.ts file because electron's `app` is undefined during tests and it makes the tests fail, so I used that directly on the games.ts file

flavioislima (approved):
Looks good 👍🏽

PR #766, opened 2021-12-04
Sanitize shortcuts

### LINKED ISSUE #760: [Windows] Shortcuts are not created properly if the game name has invalid characters
### Try to get help on our Wiki first: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Troubleshooting

**Describe the bug**
If you try to create a shortcut for a game that uses characters that are invalid in Windows File names, then the creation of the shortcuts do not complete correctly.

**Add Logs**
No Logs are produced when clicking on Add Shortcut

**To Reproduce**

1. Go to a game that uses special characters that Windows can't use in a file name, for example "while True: learn()"
2. Click on Add shortcut
3. Message will appear to say a shortcut has been created on the Desktop and Start Menu
4. Shortcut on the Desktop is created, however it is called while True, has no icon and is a zero byte file, not a shortcut.
5. No shortcut is made in the start menu at all

**Expected behavior**
Needs to have some logic on Windows that stops it from trying to use Invalid characters in the shortcut filenames, added example screenshot:
image

- OS: Windows 11 21H2
- Heroic Version 1.11.0 Alpha 
- Game: while True:learn()
- Wine/Proton version: N/A
- Hardware specs (RAM, CPU, GPU): N/A

dragonDScript:
Fixes #760 ~~and adds a steam prompt like this:~~
image

- [x] Adapt functions and methods
- [x] Remove settings
- [x] Sanitize desktop shortcut name
- [ ] ~~Prompt UI~~


electron/legendary/games.ts
```diff
@@ -275,25 +275,33 @@ class LegendaryGame extends Game {
...
       return
     }
 
     const gameInfo = await this.getGameInfo()
     const launchWithProtocol = `heroic://launch/${gameInfo.app_name}`
-    const [ desktopFile, menuFile ] = this.shortcutFiles(gameInfo.title)
-    const { addDesktopShortcuts, addStartMenuShortcuts } = await GlobalConfig.get().getSettings()
+    const sanitizedTitle = gameInfo.title
+    // TODO: use regexp
+    sanitizedTitle.replaceAll('/', ' ')
```
flavioislima, line 290: Doing like this will be too slow. Try using this website to find a regexp pattern to remove special character. I think we are using regex in some places across the code.
https://regexr.com/

flavioislima:
I thibk you have removed the options from backend but is keeping on frontend. Need to remove from settings screen as well.

electron/legendary/games.ts
dragonDScript: I don't know Regexp. Can you suggest a pattern?

flavioislima:
I think that if this is to fix the issue, the pr could focus only on that part. Using regex to remove the special characters.
After that you could use another pr to make the other changes and the ui prompt. Sounds good?

dragonDScript:
> I thibk you have removed the options from backend but is keeping on frontend. Need to remove from settings screen as well.

I've removed them. And tested locally and the options weren't visible.

dragonDScript:
> I think that if this is to fix the issue, the pr could focus only on that part. Using regex to remove the special characters. After that you could use another pr to make the other changes and the ui prompt. Sounds good?

OK I will work on the prompt on another PR but lets leave this PR as-is

electron/legendary/games.ts
dragonDScript: I think /:|\/|\*|\?|<|>|\|/g should work

dragonDScript: I think /:|/|*|?|<|>||/g should work

dragonDScript: `new RegExp('[:|/|*|?|<|>||]')` is working.

flavioislima:
> > I thibk you have removed the options from backend but is keeping on frontend. Need to remove from settings screen as well.
> 
> I've removed them. And tested locally and the options weren't visible.

Ah OK. I'm blind and missed the changes on the settings file.
OK then.

dragonDScript:
> I think it needs to be replaced by a empty string sonit will be removed instead of adding a empty space on it.

OK fixed

dragonDScript:
I think that this is ready to merge, what do you think @flavioislima

electron/legendary/games.ts
```diff
@@ -275,25 +275,27 @@ class LegendaryGame extends Game {
...
     if (process.platform === 'darwin') {
       return
     }
 
     const gameInfo = await this.getGameInfo()
     const launchWithProtocol = `heroic://launch/${gameInfo.app_name}`
-    const [ desktopFile, menuFile ] = this.shortcutFiles(gameInfo.title)
-    const { addDesktopShortcuts, addStartMenuShortcuts } = await GlobalConfig.get().getSettings()
+    const sanitizedTitle = gameInfo.title
+    const pattern = new RegExp('[:|/|*|?|<|>||]')
```
flavioislima, line 289:
```suggestion
    const pattern = new RegExp('[:|/|*|?|<|>|\\|&|{|}|%|$|@|`|!|+]')
```
To be safe, let's replace all forbidden characters for filenames.

flavioislima (changes requested):
Looks good. Just check my suggestion for regexp

electron/legendary/games.ts
dragonDScript: Let's merge this first. If anyone sees an error we will update the pattern.

dragonDScript:
@flavioislima OK I think we can merge now

electron/legendary/games.ts
```diff
@@ -275,25 +275,27 @@ class LegendaryGame extends Game {
...
+  public async addShortcuts({addToStartMenu, addToDesktop}: {addToStartMenu: boolean, addToDesktop: boolean}) {
     if (process.platform === 'darwin') {
       return
     }
 
     const gameInfo = await this.getGameInfo()
     const launchWithProtocol = `heroic://launch/${gameInfo.app_name}`
-    const [ desktopFile, menuFile ] = this.shortcutFiles(gameInfo.title)
-    const { addDesktopShortcuts, addStartMenuShortcuts } = await GlobalConfig.get().getSettings()
+    const sanitizedTitle = gameInfo.title
```
flavioislima, line 288: Since the issue at #760 is on Windows, this actually doesn't fix it, since this title is used only on linux.
We need another solution for windows.

dragonDScript: What title is used in windows to create the shortcut filename?

PR #798, opened 2021-12-19
Login improvements

flavioislima:
- [x] Use Webview with Epic Website for Login
- [x] Add Login Feedback on Update Component
- [x] Add Login Button and Icon on Sidebar when Logged Out

I'm not sure if we should keep the URL bar though. Not sure if it's relevant but some people would find it safer to have it since they will know it's not a fake website.

image

---


src/App.tsx
```diff
@@ -6,7 +6,7 @@ import { Library } from 'src/screens/Library'
 import ContextProvider from 'src/state/ContextProvider'
 import ElectronStore from 'electron-store'
 import Sidebar from 'src/components/UI/Sidebar'
-import Login from './screens/Login'
```
flavioislima, line 9: I'm keeping this here because I might add the old login page as a fallback or in case someone else wants to use it instead of using the webview.

src/screens/WebView/index.tsx
```diff
@@ -1,38 +1,116 @@
...
+  const startUrl = isLogin ? '/login' : pathname
   const urls = {
     '/epicstore': epicStore,
-    '/wiki': wikiURL
+    '/wiki': wikiURL,
+    '/login': loginUrl
   }
 
+  useLayoutEffect(() => {
+    const webview = document.querySelector('webview') as Webview
```
dragonDScript, line 62: You should not be using the <webview> tag as it's deprecated, I recommend using `BrowserView` instead. Plus it's better for this case because there is only 3 webviews, one for login, one for store, and one for wiki,

dragonDScript (changes requested):
LGTM but use browserview instead.

src/screens/WebView/index.tsx
flavioislima: Do you know how to use it to integrate with react? because I tried and could not.
Also, webview is not deprecated, it's just not recommended since chromium is always changing it and might break. But electron is warning this for years now and it wasn't broken so far.
Browserview is a `main` process so it's more complicated to deal with. I tried to style it and it was a mess.

PR #802, opened 2021-12-20
Native MacOSX games Support

flavioislima:
This Adds several handlers to deal with MacOSX native games:

- [x] Add type `is_mac_native`to differentiate between Mac and Windows games.
- [x] Add filter on the Library for platforms (only on Mac)
- [x] Deal with launching native games
- [x] Show Platform information on InstallModal
- [x] Remove unnecessary settings for native games (wine, show fps, etc)
- [x] Show correct platform for already windows installed games that have Mac support as well.

Screenshot 2021-12-20 at 21 25 21

---


src/screens/Settings/components/OtherSettings/index.test.tsx
```diff
@@ -32,6 +32,7 @@ interface Props {
   toggleDiscordRPC: () => void
   useGameMode: boolean
   targetExe: string
+  isNativeMac: boolean
```
Nocccer, line 35: sometimes you call it "isMacNative" here you call it "isNativeMac". Is this on porpuse ?

src/screens/Settings/components/OtherSettings/index.tsx
```diff
@@ -15,6 +15,7 @@ const { ipcRenderer } = window.require('electron') as {
 interface Props {
   audioFix: boolean
   isDefault: boolean
+  isNativeMac: boolean
```
Nocccer, line 18: Same here with "isNativeMac" and "isMacNative"

src/screens/Settings/index.tsx
```diff
@@ -207,9 +211,11 @@ function Settings() {
         const {
           cloud_save_enabled: cloudSaveEnabled,
           save_folder: saveFolder,
-          title: gameTitle
+          title: gameTitle,
+          is_mac_native
         } = await getGameInfo(appName)
         setTitle(gameTitle)
+        setIsNativeMac(is_mac_native)
```
Nocccer, line 218: Here is a good example for the variable spelling. Is this on porpuse ?

Nocccer (changes requested):
Open question left. Else LGTM

src/screens/Settings/index.tsx
flavioislima: `is_mac_native` is a field from the library so belongs to the GameInfo. So to follow the standard we have today I decided to keep like this.

Nocccer: I would love to have setIsMacNative. So it follows the spelling of is_mac_native.

Nocccer (approved):
LGTM

PR #809, opened 2021-12-22
Use buttons for buttons, a instead of span and add aria-label

### LINKED ISSUE #745: Some features can't be used with just a keyboard
**Describe the bug**

For accessibility reasons, it should be possible to use Heroic with just the keyboard, currently some things are just unreachable with a keyboard, and some are unclear when you are focusing them.

In the Library screen:
- as a Grid, each game card is an `a` tag, but when tabbing through them there's no indication on which is the current game card you are on. The currently focused card should trigger the game name overlay and the outline like it does in the list layout
- as a List, the play/download/settings icons are just SVGs, they should be inside `a` or `button` tags so they can be focused with the tab key
- in both layouts, the `Filter` selector has no outline to show you are focusing it with the keyboard
- in both layouts, the `refresh` icon is an SVG and can't be focused, it should probably be a `button` tag

In the GamePage screen:
- the Play Now / Install / Uninstall buttons have no outline so you can't tell if you are focusing them
- the `Install on ...` select has no outline
- in the submenu, only `Settings` can get focus, the rest of the elements (Verify/Move/Change/Log/Shortcuts/etc) are `span` tag and you can't focus them, they should be `a` or `button` tags
- the `Install Path:` and the `Prefix` path are clickable but one is a div and the other is a span, they should be `a` tags

In the Settings page:
- toggles and select tags have no online, you don't know what you are focusing
- the `Help !` element is clickable but can't be focused, it should probably be a button
- winecfg/winetricks/run exe on prefix elements are clickable but can't be focused
- the + / - / +folder can't be focused, they are only SVGs, they should probably be buttons

**To Reproduce**

Open Heroic and try to use only the keyboard to navigate the app using Tab, Shift+Tab, the arrow keys, and spacebar/enter to activate elements.

**Expected behavior**

Every clickable element should be also reachable with the keyboard.

- OS: Linux Mint
- Heroic Version: latest main branch
- Game: -
- Wine/Proton version: -
- Hardware specs (RAM, CPU, GPU): -

arielj:
This PR improves the accessibility of the app by providing better support for keyboard navigation with better focus targets, focus indicators and labels:
- Added an SvgButton to wrap the different svg tags and font awesome icons
- Replaced many `span.link` to be `a` tags with `href="#"` so they can be focused when tabbing
- Added aria-labels to the ToggleSwitch element so a screen reader can match what's each switch for
- I also fixed the focus indication for game cards in the grid layout

I tried the different screens and I was able to change settings, install/uninstall a game, start it, etc.

This closes #745 

---


flavioislima:
Nice!
I'll try to review it today or tomorrow.
Btw, about the gamecard, I noticed you haven't changed it from a div to something else. Don't worry about that because I fixed it on my #802 PR.
I removed the `div` and adapted the `Link` so it can be browsed with the keyboard.

arielj:
> Nice! I'll try to review it today or tomorrow. Btw, about the gamecard, I noticed you haven't changed it from a div to something else. Don't worry about that because I fixed it on my #802 PR. I removed the `div` and adapted the `Link` so it can be browsed with the keyboard.

oh great! yeah I just fixed that one with CSS to not break anything (using the focus-within selector), but it's better if it's a link tag instead

src/components/UI/Header/index.css
```diff
@@ -129,15 +128,22 @@ option {
   cursor: pointer;
 }
 
-.headerIcons > *:hover {
+.headerIcons > .svg-button {
+  color: white;
```
flavioislima, line 132: We should always use variables for the colors.
```suggestion
  color: var(--text-primary);
```

src/screens/Library/components/GameCard/index.tsx
```diff
@@ -141,11 +142,9 @@ const GameCard = ({
     }
     if (hasUpdate) {
       return (
-        <FontAwesomeIcon
-          size={'2x'}
-          icon={faRepeat}
-          onClick={() => handleUpdate()}
-        />
+        <SvgButton onClick={() => handleUpdate()}>
+          <FontAwesomeIcon size={'2x'} icon={faRepeat} />
```
flavioislima, line 146: When I tested this branch, for some reason this icon is Black instead of green. I think some class is overriding it.
so try to make it use the correct color.

flavioislima (changes requested):
Overall looks pretty good, left a few minor comments though.

flavioislima:
> > Nice! I'll try to review it today or tomorrow. Btw, about the gamecard, I noticed you haven't changed it from a div to something else. Don't worry about that because I fixed it on my #802 PR. I removed the `div` and adapted the `Link` so it can be browsed with the keyboard.
> 
> oh great! yeah I just fixed that one with CSS to not break anything (using the focus-within selector), but it's better if it's a link tag instead

After releasing the new beta I noticed that my solution doesn't work as expected because then we cannot click on the icons. Like, since the link is the wrapper, it has preference over the rest of the elements. So I'll revert it and we will use yours :)

src/screens/Library/components/GameCard/index.tsx
arielj: fixed this, thanks

arielj:
> > > Nice! I'll try to review it today or tomorrow. Btw, about the gamecard, I noticed you haven't changed it from a div to something else. Don't worry about that because I fixed it on my #802 PR. I removed the `div` and adapted the `Link` so it can be browsed with the keyboard.
> > 
> > 
> > oh great! yeah I just fixed that one with CSS to not break anything (using the focus-within selector), but it's better if it's a link tag instead
> 
> After releasing the new beta I noticed that my solution doesn't work as expected because then we cannot click on the icons. Like, since the link is the wrapper, it has preference over the rest of the elements. So I'll revert it and we will use yours :)

oh sure, I remove the css, then read this message and now I re-added the css

arielj:
@flavioislima done, just merged main into this branch and fixed the css

dragonDScript (commented):
Thanks for the effort, and for caring about accesibility. Would set user-select to none worsen accesibility? This PR LGTM but I'll re-review later

flavioislima (approved):
Nice, it's working perfectly now! thanks! :D

Issue #826, opened 2021-12-28
add an exit button.


LithRakoon:
**Is your feature request related to a problem? Please describe.**
When installing a game theres a pop up giving the option to import or to install the game but not to exit the installation, it is very unclear that you need to click outside the pop up for it to stop the installer

**Describe the solution you'd like**
the solution is adding an X button to the pop up so its easy to know where you need to click to exit it.

**Describe alternatives you've considered**
add a third button to the button on the left side preferably that says cancel or exit

**Additional context**
very rough example of the first solution:
![image](https://user-images.githubusercontent.com/41258148/147605820-5223a221-ace3-4fd9-b399-11265846ea0a.png)

rough example second solution:
![image](https://user-images.githubusercontent.com/41258148/147606346-e01b30f6-368e-4177-a220-72d300b705bd.png)

Artumira96:
Hello, I would like to work on this issue :)

flavioislima:
> Hello, I would like to work on this issue :)

Feel free to send a PR.
Any questions or if you get stuck just got to our Development channel on our discord server. :)

Artumira96:
Hi! there's my PR for this issue along a quick video showing it's functionality, if anything comes up please let me know, thanks! :)

PR #835, opened 2021-12-29
Add cancel button for exiting install modal

Artumira96:
Solves issue: #826 of adding the "Cancel" button for the install modal 

Here's a quick demo of the functionality:

https://user-images.githubusercontent.com/80984726/147704334-2a81705d-9aef-44aa-ba53-15e34fda0ae5.mov

Regards, Arturo


dragonDScript (approved):
LGTM, nice PR

flavioislima (changes requested):
Hi, thanks for the PR
I know it works fine but I think having three buttons there is not ideal. can you go with the Close button solution? 
An X icon that you can grab from FontAwesome (I think we already have a use case in this component). The icon name is: `faXmark`.

Artumira96:
image

Got it, I changed the button to a close icon on the upper left corner, hope that makes the trick :) Here's a quick video of it: 

https://user-images.githubusercontent.com/80984726/147781193-b2dd270f-0044-4b0c-9775-fc0689ff4d11.mov

flavioislima:
Nice, can you make it a big bigger and post a screenshot with it on the right corner as well, so we can decide which one is better?

Artumira96:
Sure, here's on the right a little bigger, what do you think? 
image

On the left:
image

flavioislima:
Thanks. Let's go with smaller icon on the right. I think it's the best way 🙂

Artumira96:
You got it, it's on my last commit 
image

src/screens/Library/components/InstallModal/index.tsx
```diff
@@ -174,6 +174,11 @@ export default function InstallModal({ appName, backdropClick }: Props) {
     <span className="modalContainer">
       {gameInfo?.game?.title ? (
         <div className="modal">
+          <FontAwesomeIcon
```
arielj, line 177: Please, use an SvgButton component with the FontAwesomeIcon nested. FontAwesomeIcon is just an `<svg>` tag that can't be focused with a keyboard

Here's an example https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/screens/Library/components/GameCard/index.tsx#L148-L150

Also an `aria-label` attribute with the text `Close dialog` is a good idea

src/index.css
```diff
@@ -21,7 +21,7 @@ html {
   --secondary-hover: #feba36;
   --tertiary-hover: #322b2b;
   --success-hover: #30eca8;
-  --danger-hover: #bd0a0a;
+  --danger-hover: #e03e3e;
```
arielj, line 24: is this color change needed? wound't if affect other places using this variable?

src/App.css
```diff
@@ -87,6 +87,20 @@ html,
   color: white;
 }
 
+.close-icon {
+  margin-left: 34rem;
```
arielj, line 91: I think it would be better to position this icon with `position: absolute; right: somenumber, top: somenumber;`,  that `margin-left` can be harder to maintain if for any reason that dialog changes size

src/App.css
```diff
@@ -87,6 +87,20 @@ html,
...
+  cursor: pointer;
+}
+
+.button.is-cancel {
```
arielj, line 95: I understand these .button.is-cancel and :hover styles are not needed anymore, right?

src/screens/Library/components/InstallModal/index.tsx
Artumira96: image
Got it, this works for me, what do you think?

src/index.css
Artumira96: You're right, I removed it

src/App.css
Artumira96: Removed them!

src/App.css
Artumira96: image
This is my approach for this one, it works fine, sounds good?

src/screens/Library/components/InstallModal/index.tsx
```diff
@@ -173,7 +173,10 @@ export default function InstallModal({ appName, backdropClick }: Props) {
   return (
     <span className="modalContainer">
       {gameInfo?.game?.title ? (
-        <div className="modal">
+        <div className="modal close-icon-container">
+          <SvgButton onClick={() => backdropClick()}>
+            <FontAwesomeIcon className={'close-icon'} icon={faXmark} />
```
arielj, line 178: You can do `className="close-icon"` in the SvgButton instead, that way you can position the `button` tag (here you are moving the `svg` tag outside of the button).

When you do that, you'll probably won't need the `cursor: pointer` property since it's alredy a `button` tag and the cursor is pointer by default.

src/screens/Library/components/InstallModal/index.tsx
```diff
@@ -173,7 +173,10 @@ export default function InstallModal({ appName, backdropClick }: Props) {
   return (
     <span className="modalContainer">
       {gameInfo?.game?.title ? (
-        <div className="modal">
+        <div className="modal close-icon-container">
```
arielj, line 176: I don't think we need to add the `close-icon-container` class here, you can add `position: relative` to `.modal` in src/screens/Library/components/InstallModal/index.css

I think you can also move the `.close-icon` css inside that InstallModal/index.css file instead of the App.css, since this close button is only used in the InstallModal

src/App.css
arielj: I added some comments to refactor the css and the classes

src/screens/Library/components/InstallModal/index.tsx
Artumira96: Got it, thanks for your feedback! it's on my last commit

.vscode/settings.json
```diff
@@ -1,5 +1,25 @@
...
+    "activityBar.inactiveForeground": "#e7e7e799",
+    "activityBarBadge.background": "#155e02",
+    "activityBarBadge.foreground": "#e7e7e7",
+    "sash.hoverBorder": "#fa1b49",
+    "statusBar.background": "#dd0531",
+    "statusBar.foreground": "#e7e7e7",
+    "statusBarItem.hoverBackground": "#fa1b49",
+    "statusBarItem.remoteBackground": "#dd0531",
+    "statusBarItem.remoteForeground": "#e7e7e7",
+    "titleBar.activeBackground": "#dd0531",
```
dragonDScript, line 19: why are all this settings?

electron/main.ts
```diff
@@ -96,9 +96,9 @@ async function createWindow(): Promise<BrowserWindow> {
     width: 1200,
     show: !(exitToTray && startInTray),
     webPreferences: {
+      nodeIntegration: true,
       webviewTag: true,
-      contextIsolation: false,
```
dragonDScript, line 100: Don't change these

dragonDScript (changes requested):

src/screens/Library/components/InstallModal/index.css
```diff
@@ -19,6 +19,7 @@
 }
 
 .modal {
+  position: relative !important;
```
arielj, line 22: Is the `!important` needed here? can you try without it?

arielj (commented):
Apart from dragonDScript's comments about those extra changes I only have a question about the use of `!important`.

The rest looks good to me

electron/main.ts
Artumira96: Yeah those 2 files weren't supposed to be in the commit, sorry about that

src/screens/Library/components/InstallModal/index.css
Artumira96: I removed it and it works fine 👍

dragonDScript (commented):
LGTM; nice PR. fix what the others told you and it'll be good to go

arielj (approved):
great!

flavioislima (approved):
LGTM 👍🏽 
thanks for that!

Artumira96:
Glad I could help! :) thanks

PR #837, opened 2021-12-30
Fixed typo with Discord RPC indicating macOS is "MacOS" when it should be "macOS"

andylin2004:
Just a typo fix with the Discord RPC saying what platform Heroic is running on (from "MacOS" to "macOS")

### DIFF
--- a/electron/legendary/games.ts
+++ b/electron/legendary/games.ts
@@ -536,7 +536,7 @@ Categories=Game;
           os = 'Windows'
           break
         case 'darwin':
-          os = 'MacOS'
+          os = 'macOS'
           break
         default:
           os = 'Unknown OS'


electron/legendary/games.ts
```diff
@@ -536,7 +536,7 @@ Categories=Game;
           os = 'Windows'
           break
         case 'darwin':
-          os = 'MacOS'
```
flavioislima, line 539: Perhaps should be MacOSX then?

andylin2004: Apple's official term for their desktop operating system is now macOS, regardless of if it was previously named OS X or Mac OS X.

For example: [this documentation](https://developer.apple.com/documentation/eventkit/ekreminder)

flavioislima: Ok, sounds good.

flavioislima (approved):

Issue #840, opened 2021-12-30
Wine prefixes are not created. ~/Games/ folder is empty.


allora:
### Is this a game issue? If YES please use our Discord to get help.
No

### Try to get help on our Wiki first: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Troubleshooting

**Describe the bug**
Heroic does not appear to create child directories of ~/Games/Heroic or ~/Games/Heroic/Prefixes or any installed games prefix under this directory. For example when i download and install The Pathless, the installation process seems to go fine. However when i hit play it does the following

```
INFO: Updating game list
INFO: Game List Updated
INFO: Checking for game updates
INFO: Found 0 game(s) to update
INFO: Getting More details with Legendary info command... 
INFO: launching The Pathless geum
INFO: creating new prefix /home/<user>/Games/Heroic/Prefixes/The-Pathless
ERROR: Error: Command failed: WINEPREFIX='/home/<user>/Games/Heroic/Prefixes/The-Pathless' '/usr/bin/wineboot' -i &&  '/usr/bin/wineserver' --wait
wine: chdir to /home/<user>/Games/Heroic/Prefixes/The-Pathless : No such file or directory

WARNING: DXVK cannot be installed on a Proton or a invalid prefix!
WARNING: GameMode not installed
INFO: 
 Launch Command:     WINEESYNC=1         /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary launch geum   --wine /usr/bin/wine --wine-prefix '/home/<user>/Games/Heroic/Prefixes/The-Pathless'  
INFO: Getting More details with Legendary info command... 
INFO: checking for heroic updates
/bin/sh: line 1: hostname: command not found
INFO: Log was written
INFO: Getting More details with Legendary info command... 
INFO: Getting More details with Legendary info command... 
```

Obviously, i have full write access to this folder, and if i create the folders manually, everything seems to work and the game launches.

Maybe the wrong mkdir (or similar) command is being used when trying to create the full path for the game (or Heroic on first boot isnt creating its folder? That is what I would suspect anyway.

**Add Logs**

Check on how to [run Heroic from Terminal or Powershell](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Troubleshooting#running-heroic-from-terminalpowershell-to-debug-errors) and put some logs here

Full log of attempting to run the game:
```
[...]

PR #845, opened 2021-12-31
Fix wine prefixes not being created

imLinguin:
Changed a way of creating prefix for Proton. Instead of preparing `mkdir` command it uses mkdirSync from graceful-fs library.
Same for Wine prefixes. Before running `wineboot`, creates directory recursively.  

Fixes one problem out of two in
#840 


dragonDScript (approved):

dragonDScript:
Thanks for the nice PR

PR #869, opened 2022-01-04
[General] Refactor and enable spatial navigation

arielj:
This PR adds the `enable-spatial-navigation` flag so the app can be navigated using the arrow keys so the keyboard accessibility is better (and this may enable the spatial navigation to use with the gamepad). The PR includes some refactoring that fixed the bugs with the browser's native spatial navigation, this way we don't have to use a javascript solution.

I had to make some refactoring to fix the spatial navigation:
- the library screen header is not position: sticky, now the listings are wrapped in a div and that div has scroll, so the header won't break the spatial navigation
- because of the previous change, I had to add a small fix to the install modal that was offset
- I made a big refactor of the gameCard/gameListItem component so the same html works for both formats
- I replaced some divs with spans, because the spatial navigation was focusing the divs too for some reason

I'll add some comments in the diff.

I tested moving around the app with the arrow keys, tried installing a game, cancelling the installation, opening a game details and executing a game.

EDIT: since I was already refactoring things on that screen, I did a small refactor to fix the `back to top` button that got missing and the html was duplicated

---

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -41,10 +41,12 @@ function App() {
                     renderBackButton={false}
                     numberOfGames={numberOfGames}
                   />
-                  {showRecentGames && (
-                    <Library showRecentsOnly library={recentGames} />
-                  )}
-                  <Library library={library} />
+                  <div className="listing">
+                    {showRecentGames && (
+                      <Library showRecentsOnly library={recentGames} />
+                    )}
+                    <Library library={library} />
+                  </div>
                 </>
               ) : (
                 <WebView isLogin />
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -42,6 +42,7 @@ function App() {
                     numberOfGames={numberOfGames}
                   />
                   <div className="listing">
+                    <span id="top" />
                     {showRecentGames && (
                       <Library showRecentsOnly library={recentGames} />
                     )}
--- a/src/screens/Library/components/GameCard/index.css
+++ b/src/screens/Library/components/GameCard/index.css
@@ -142,6 +142,7 @@
   max-height: 8vh;
   width: 100%;
   height: 100%;
+  overflow: hidden;
 }
 
 .gameListItem .gameImg {
--- a/src/screens/Library/components/GameCard/index.css
+++ b/src/screens/Library/components/GameCard/index.css
@@ -1,9 +1,6 @@
 .gameCard {
   background-color: var(--background-darker);
   text-align: left;
-  display: flex;
-  justify-content: center;
-  align-items: center;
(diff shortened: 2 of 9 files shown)


src/App.tsx
```diff
@@ -41,10 +41,12 @@ function App() {
                     renderBackButton={false}
                     numberOfGames={numberOfGames}
                   />
-                  {showRecentGames && (
-                    <Library showRecentsOnly library={recentGames} />
-                  )}
-                  <Library library={library} />
+                  <div className="listing">
```
arielj, line 44: This allows me to remove the position: sticky of the header

src/screens/Library/components/GameCard/index.tsx
```diff
@@ -198,80 +206,39 @@ const GameCard = ({
...
-              {logo && (
-                <img
-                  alt="logo"
-                  src={`${logo}?h=400&resize=1&w=300`}
-                  style={{
-                    filter: isInstalled ? 'none' : `grayscale(${effectPercent})`
-                  }}
-                  className="gameLogo"
-                />
+            <img src={imageSrc} className={imgClasses} alt="cover" />
```
arielj, line 213: I moved from a span to an img so it handles dimensions with the element, that helps with the dimensions of the wrapper `a` tag

src/screens/Library/components/InstallModal/index.css
```diff
@@ -4,6 +4,7 @@
   width: var(--content-width);
   height: 100%;
   position: fixed;
+  top: 0px;
```
arielj, line 7: I had to add this fix after adding the `.listings` wrapper

src/screens/Library/index.css
```diff
@@ -48,6 +48,11 @@
   animation: apparition 0.3s ease-in forwards;
 }
 
+.listing {
```
arielj, line 51: this puts the scroll here instead of in the `.content` element

src/screens/Library/components/GameCard/index.css
```diff
@@ -55,21 +52,31 @@
...
 }
 
-.gameCard > a > .gameImg {
-  width: 8vw;
-  height: 10vw;
-  max-width: 200px;
-  max-height: 266px;
-  min-width: 130px;
-  min-height: 173px;
+.gameCard > a {
```
arielj, line 55: There are many changes in this file, some are new styles, some are just me moving things around so similar elements are together, and are refactors like moving width, min-width and max-width to a widh: clamp(...) property and things like that

flavioislima (approved):
Tested here and works really good. Great job and thanks for that! :D﻿

PR #888, opened 2022-01-09
Fix typo and add extra styling to copy to clipboard button

TabulateJarl8:
First things first, I noticed that the "Copy All Settings to Clipboard" button in the settings menu actually says "Setting" instead of "Settings." I'm not entirely sure if I need to do anything to update the translations, but I've fixed it where it appeared in the code. 

The other thing I added was just a small styling improvement (in my opinion) to that same button. When clicked, it now changes the text to "Copied to Clipboard!" and changes the text color to green on hover (success variable) for 3 seconds, after which it changes back to normal.

Hope you like the changes, and let me know if I should fix anything

---

### DIFF
--- a/src/screens/Settings/index.css
+++ b/src/screens/Settings/index.css
@@ -188,3 +188,7 @@ a {
   margin-left: 8px;
   padding-top: 0;
 }
+
+.settingsWrapper > .button.success-override-color:hover {
+  color: var(--success) !important;
+}
--- a/src/screens/Settings/index.css
+++ b/src/screens/Settings/index.css
@@ -189,6 +189,6 @@ a {
   padding-top: 0;
 }
 
-.settingsWrapper > .button.success-override-color:hover {
+.settingsWrapper > .button.isSuccess:hover {
   color: var(--success) !important;
 }
--- a/src/screens/Settings/index.tsx
+++ b/src/screens/Settings/index.tsx
@@ -1,6 +1,7 @@
 import './index.css'
 
 import React, { useContext, useEffect, useState } from 'react'
+import classNames from 'classnames'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faWindows, faApple } from '@fortawesome/free-brands-svg-icons'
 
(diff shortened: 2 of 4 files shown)


src/screens/Settings/index.tsx
```diff
@@ -451,14 +453,28 @@ function Settings() {
...
               clipboard.writeText(
                 JSON.stringify({ appName, title, ...settingsToSave })
               )
-            }
+              // set copied to clipboard status to true if it's not already set to true
+              // used for changing text and color
+              if (!isCopiedToClipboard) {
+                setCopiedToClipboard(true)
+
+                setTimeout(() => {
```
arielj, line 468: Maybe you can move this to a useEffect, it will clean this a bit (the onClick callback is getting kinda big) and you can fix a bug. The bug is that, if you click the copy to clipboard button and change to another screen before the timeout ends, you'll get an error in the console because react will try to update an unmounted component.

You can do something like:

```js
useEffect(() => {
  const timer = setTimeout(() => {
    setCopiedToClipboard(false)
  }, 3000)

  return () => clearTimeout(timer)
}, [copiedToClipboard])
```

the returned function will take care of stoping the timeout if needed

flavioislima:
Nice! I'll take a look later today.

flavioislima:
@TabulateJarl8 to fix the typo you will need to manually edit the English json file or fix it on weblate project.

src/screens/Settings/index.tsx
```diff
@@ -451,14 +465,19 @@ function Settings() {
           )}
           <span className="save">{t('info.settings')}</span>
           <button
-            className="button is-text"
-            onClick={() =>
+            className={`button is-text ${
```
flavioislima, line 468: We have a pretty useful npm package to deal with this.
You can just `import classNamesfrom 'classnames'`
 and the use:
```
className={classNames('button', 'is-text', {
              isSuccess: isCopiedToClipboard
            })}
```
also, avoid using class names like that with `-`, react might go crazy with that sometimes. 
I added a reminder to myself to rename all classes that have a dash to be camel case.

flavioislima (changes requested):
Looks good. Just one small thing that I think could be better and we are good to go :)

src/screens/Settings/index.css
```diff
@@ -188,3 +188,7 @@ a {
   margin-left: 8px;
   padding-top: 0;
 }
+
+.settingsWrapper > .button.success-override-color:hover {
```
flavioislima, line 192: Ideally:
```suggestion
.settingsWrapper > .button.isSuccess:hover {
```

TabulateJarl8:
Should be good to go now, I just manually added the new translation to the English JSON file.

flavioislima (approved):
NIce. Good job and thanks for that!

PR #890, opened 2022-01-11
[General] Add clear button (x) to search bar (Closes #877)

### LINKED ISSUE #877: Add Clear Button to Search Bar
when you use the search bar and theres text in it there is no easy way to dismiss the text, an easy fix for that is to make the "Library" button act as that dismissal since this feels very natural 

make the "library" button dismiss text and return you to the home page

add an X to the end of the search bar

TabulateJarl8:
I saw issue #877, and I figured it was easy enough to implement, so I went ahead and added that feature. I've tested it with a bunch of different window sizes and it seems to react accordingly, staying within reasonable boundaries. I've attached a screenshot of the feature to this pull request.

I'm not sure why it's showing all 10 commits in my fork since the oldest 6 were merged, but the most recent 4 are the relevant commits to this PR.

<details>
<summary>Full window screenshot</summary>

</details>
<details>

</details>

---


src/components/UI/SearchBar/index.css
```diff
@@ -1,6 +1,8 @@
 .SearchBar {
   grid-area: search;
   width: 100%;
+  display: inline-block;
```
arielj, line 4: is this display: inline-block needed?

src/components/UI/SearchBar/index.tsx
```diff
@@ -23,6 +23,17 @@ export default function SearchBar() {
         placeholder={t('search')}
         id="search"
       />
+      <span className="clearSearchInputWrapper">
```
arielj, line 26: is this extra wrapper needed? I think you should be able to position the clickable element directly

src/components/UI/SearchBar/index.tsx
```diff
@@ -23,6 +23,17 @@ export default function SearchBar() {
...
+        <span
```
arielj, line 27: Can you replace this with a `button` tag? that makes it more accessible, `span` tags are not good to grab keyboard focus if navigating the app without a mouse

src/components/UI/SearchBar/index.css
TabulateJarl8: No, must've been a leftover from my testing, I'll fix that in an upcoming commit

src/components/UI/SearchBar/index.tsx
TabulateJarl8: I believe so, having the padding directly on the element would cause all the space to the left of the element to be clickable, unless you can think of a better way of doing it

src/components/UI/SearchBar/index.tsx
TabulateJarl8: Should be out in the next commit

TabulateJarl8:
> Nice, looks good. Can we show the X only when we have some text maybe?

Should be working now, just haven't been able to test it with controller support

src/components/UI/SearchBar/index.tsx
arielj: instead of doing right: 0px with padding, you can do right > 0 and no padding, or you can try with margin instead of padding so it's not part of the click area

src/components/UI/SearchBar/index.css
```diff
@@ -13,12 +14,32 @@
...
   background: var(--background);
   width: 80%;
+  padding-right: 2.5em;
 }
 
-.SearchBar > label {
-  color: var(--secondary);
+.clearSearchInputWrapper {
+  position: absolute;
+  top: 22%;
```
arielj, line 22: a common trick to absolute position this in the center ( so you don't have to use values like this 22%) is to do

```
top: 50%
transform: translateY(-50%)
```

src/components/UI/SearchBar/index.css
```diff
@@ -13,12 +14,32 @@
...
+  font-size: 1.4em;
+  text-decoration: none;
+  background: none;
+}
+
+.clearSearchInput:hover {
+  color: var(--text-primary);
+}
+
+.isEmpty {
```
arielj, line 43: maybe you can do `.clearSearchInputWrapper.isEmpty` for the selector here to prevent issues, `.isEmpty` is too generic and maybe somebody tries to use that class somewhere else to mean a different style

src/components/UI/SearchBar/index.css
TabulateJarl8: Thanks! That's super useful to know. Should be pushed soon

src/components/UI/SearchBar/index.tsx
TabulateJarl8: This should also be resolved now, thanks for the help

TabulateJarl8:
Current state of design
image

arielj:
> Current state of design image
> 
> Edit: if I set `top` in the CSS to 49% instead of 50%, it looks like this, which looks to me like its more centered, although I can't tell if my eyes are just playing tricks on me or not image

hmmm it can be that the `x` is not the full height of the line, so the line is centered but the x is not centered in that space. I think it's fine to use 49% then

TabulateJarl8:
Should be working now

adityaruplaha (approved):
Looks good to go.

flavioislima (approved):
Looks Good now :D

PR #898, opened 2022-01-12
[General] Add more Warnings and Error Dialogs and other fixes

### LINKED ISSUE #905: Apostrophe ( ' ) in prefix name causes a mismatch between actual prefix and Heroic, thus DXVK install and winetricks etc fails
**Describe the bug**

I noticed this with the game Sid Meier's Civilization VI from EGS. 

Heroic automatically creates prefixes with the name of the game. In this case with CIV 6, prefix it tries to create is `Sid-Meier's-Civilization-VI`.

At prefix creation Heroic creates two directories. 

`Sid-Meier's-Civilization-VI`

`Sid-Meiers-Civilization-VI`

First one is just blank directory that is not an actual prefix, yet still specified as prefix in Heroic GUI. Therefore tall DXVK installation process,winetricks,winecfg usage etc fails silently.

Latter directory without apostrophe is the actual prefix but since in Heroic GUI it is not specified as correct path, it is just a prefix without a DXVK installation which will lead to broken behaviour by default due to low success rate of wined3d. In case Civ6 game doesn't even work with wined3d which ultimately led noticing this issue.

**Add Logs**

There is no log needed for this.

**To Reproduce**

1. Install Civ 6 from EGS
2. Click on Play while DXVK autoinstall etc were enabled
3. Notice game fails at launch and check logs to see wined3d spam
4. Check `/home/yoursername/Games/Heroic/Prefixes` path and notice two Civ 6 directories.
5. Check one with apostrophe to confirm it is just a blank directory. 
6. Go to Heroic GUI and delete apostrophe from default prefix name for that game ( Sid-Meier's-Civilization-VI ) and launch the game again. Prefix will be updated with DXVK.  

**Expected behavior**

Not needing that manual deletion ( Preferably Heroic ignoring special chars etc at prefix creation )

- OS: EndavourOS
- Heroic Version : 2.0.2
- Game: Sid Meier's Civilization VI
- Wine/Proton version: Any Wine version
- Hardware specs (RAM, CPU, GPU): Doesn't matter

flavioislima:
### Summary
Messages:
- [x] show Error dialog if Python > 3.8 not found
- [x] show error message if something goes wrong with Login
- [x] show error message if no wine version was found on play

Updates:
- [x] Update Legendary
- [x] Update other NPM packages
- [x] Update i18n keys

Fixes:
- [x] Fix Heroic not finding custom wine/proton when there is no default wine installed.
- [x] Fix #905 

---


electron/legendary/user.ts
```diff
@@ -22,12 +23,20 @@ export class LegendaryUser {
       child.stderr.on('data', (data) => {
         console.log(`stderr: ${data}`)
```
Nocccer, line 24:
```suggestion
        logError(`stderr: ${data}`)
```

electron/legendary/user.ts
```diff
@@ -22,12 +23,20 @@ export class LegendaryUser {
...
           return res('error')
         }
       })
       child.stdout.on('data', (data) => {
         console.log(`stderr: ${data}`)
         if (`${data}`.includes('ERROR')) {
+          dialog.showErrorBox(
+            i18next.t('box.error.login.tittle', 'Cannot Login!'),
+            `${data}`
+          )
```
Nocccer, line 39:
```suggestion
      child.stdout.on('data', (data) => {
        if (`${data}`.includes('ERROR')) {
          logError(`stderr: ${data}`)
          dialog.showErrorBox(
            i18next.t('box.error.login.tittle', 'Cannot Login!'),
            `${data}`
          )
```

Nocccer (changes requested):
Left some suggestions. What do you think ?

electron/legendary/user.ts
```diff
@@ -20,19 +21,29 @@ export class LegendaryUser {
...
       child.stderr.on('data', (data) => {
-        console.log(`stderr: ${data}`)
+        logInfo(`stderr: ${data}`)
         if (`${data}`.includes('ERROR')) {
+          logError(`${data}`)
+          dialog.showErrorBox(
+            i18next.t('box.error.login.tittle', 'Cannot Login!'),
+            `${data}`
+          )
           return res('error')
```
Nocccer, line 31: I don't like the logInfo on top. If an error occurs data will be logged twice. Once as info and once as error. Just log one of them like so:

```suggestion
        if (`${data}`.includes('ERROR')) {
          logError(`${data}`, LogPrefix.Legendary)
          dialog.showErrorBox(
            i18next.t('box.error.login.tittle', 'Cannot Login!'),
            `${data}`
          )
          return res('error')
       } else {
          logInfo(`stderr: ${data}`, LogPrefix.Legendary)
       }
```

electron/legendary/user.ts
```diff
@@ -20,19 +21,29 @@ export class LegendaryUser {
...
         }
```
Nocccer, line 43: Same as above. Also in my eyes stderr should be stdout in logInfo.

```suggestion
        if (`${data}`.includes('ERROR')) {
          logError(`${data}`, LogPrefix.Legendary)
          dialog.showErrorBox(
            i18next.t('box.error.login.tittle', 'Cannot Login!'),
            `${data}`
          )
          return res('error')
       } else {
          logInfo(`stdout: ${data}`, LogPrefix.Legendary)
       }
```

electron/legendary/user.ts
```diff
@@ -20,19 +21,29 @@ export class LegendaryUser {
...
       })
       child.on('close', () => {
-        console.log('finished login')
+        logInfo('finished login')
```
Nocccer, line 46: You can also skip the prefix for now and i will take care of it in another pr :).

```suggestion
        logInfo('finished login', LogPrefix.Legendary)
```

Nocccer (changes requested):
Sorry. Still found some stuff. Left some suggestion how i would implement it. I think easy fix and then we can merge really this time.

PR #908, opened 2022-01-15
[General] Check epic status and show warning messages

flavioislima:
This PR checks Epic Status API and if the Status is Outage show some warning messages to the user on Install, Import and Update games. And others on the console for Sync Saves, etc.


electron/main.ts
```diff
@@ -788,6 +791,18 @@ ipcMain.handle('install', async (event, params) => {
...
+    dialog.showErrorBox(
+      i18next.t('box.warning.title', 'Warning'),
+      i18next.t(
+        'box.warning.epic.install',
+        'Epic Servers are having major outage right now, the game cannot be installed!'
+      )
+    )
+    return { status: 'error' }
+  }
+
```
Nocccer, line 805: You adding alot of these with just different i18n keys. Maybe make it a function and reuse it. Would make the code more adjustable and readable

Nocccer (changes requested):
Beside the comment i left lgtm. If you don't like the note i left, hit me up so i can approve it or just merge it.

electron/main.ts
flavioislima: Yes, might be but i18next has some issues with encapsulated or keys that are under some kind of check.
Like: `varTrue ? 'key.one' : 'key.two'`
Once we run the command to generate the new keys, it just ignore this and none are created.

flavioislima: the same with:
`box.warning.epic.{action'}`
it doesn't created the `actions` keys

PR #909, opened 2022-01-15
Refactor gamepads layotus and add some more

arielj:
This PR is an updated to the layouts handling for different gamepads. Changes are:

- Reorganized the layouts in a gamepad_layouts folder grouping layouts by brand
- Added Genius MaxFire, a N64-USB clone, ps3 clone through usb dongle
- Prevent exceptions when checking buttons to break the gamepad loop

I was able to test the Genius layout and will ask Leopard to test the ps3 one.

---


src/helpers/gamepad.ts
```diff
@@ -217,191 +226,35 @@ export const initGamepad = () => {
...
+        } else if (controller.id.match(/0583.*a009/i)) {
+          checkGenius1(buttons, axes, index, checkAction)
+        } else {
+          // if not specific, fallback to the xbox layout, seems
+          // to be the most common for now and if not exact it seems
+          // to cover at least the left stick and the main 2 buttons
+          checkXbox(buttons, axes, index, checkAction)
+        }
+      } catch (error) {
+        console.log(`Gamepad error: ${error}`)
```
Nocccer, line 251: Please use our log functions. (logger.ts).

You could also write a wrapper around it.
e.g.:

```typescript
const gamepadLog = (type: 'info' | 'warning' | 'error', message: string) => {
  const extMessage = `Gamepad: ${message}`
  switch(type) {
    case 'info':
      logInfo(extMessage)
      break
   case 'warning':
      logWarning(extMessage)
      break
   case 'error':
     logError(extMessage)
     break
   default:
     break
  }
}
```

src/helpers/gamepad.ts
```diff
@@ -414,8 +267,10 @@ export const initGamepad = () => {
         console.log(`button ${button} pressed ${buttons[button].value}`)
     }
     for (const axis in axes) {
-      if (axes[axis] < -0.5) console.log(`axis ${axis} activated negative`)
-      if (axes[axis] > 0.5) console.log(`axis ${axis} activated positive`)
+      if (axes[axis] < -0.1 && axes[axis] >= -1)
+        console.log(`axis ${axis} activated negative`)
```
Nocccer, line 271: use our logger or a wrapper i mentioned above

src/helpers/gamepad.ts
```diff
@@ -414,8 +267,10 @@ export const initGamepad = () => {
...
+      if (axes[axis] > 0.1 && axes[axis] <= 1)
+        console.log(`axis ${axis} activated positive`)
```
Nocccer, line 273: use our logger or a wrapper i mentioned above

Nocccer (changes requested):
Left some suggestions there for better logging.

flavioislima (approved):
Looks good!

src/helpers/gamepad.ts
flavioislima: This is actually on the frontend, will they work there as well? I think they are on the electron folder so cannot be imported here.

Nocccer: True. So if we want this message on the terminal we should use ipcRedender('LogInfo' ...), else we keep it like that.

Nocccer (approved):

arielj:
thanks! I'll update the logger calls after that's merged on the main branch :+1: (I noticed it got merged but something went wrong there during the merge)

flavioislima (approved):
LGTM! 👍🏽

PR #911, opened 2022-01-15
Add prefix parameter for logger

Nocccer:
With more and more different things which are already implemented or are coming up soon, it is good to add a prefix to log messages. This helps to read logs better.

- [x]  Added new parameter prefix for every logger function
- [x] adapt logger functions to still allow string and string[]
- [x] update tests

---


flavioislima (approved):
Nice!

Issue #893, opened 2022-01-11
Remember window size at close


arielj:
**Is your feature request related to a problem? Please describe.**

Currently, if you have a preference for where Heroic is shown in the screen, you have to do that every time you open the app because the position that a user sets is not restored the next time the app opens.

**Describe the solution you'd like**

It's good for user experience to remember the size and position of the window set by the user, the app should store the size and position of the window in a config file when the app is closed (or when resized?) and restore that value when opening the app.

**Describe alternatives you've considered**

Some alternative would be to have a toggle to start the app maximized maybe? but that only accounts for users that prefer heroic fullsrcreen, though it might be a good addition for the future in the steam deck for example.

PR #917, opened 2022-01-21
[General] Remember Window Size

Nocccer:
Implementation for #893.

Just simple store the window bounds on closing and reuse them before the creation of main window.

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -105,9 +105,9 @@ async function createWindow(): Promise<BrowserWindow> {
   if (store.has('window-props')) {
     const tmpWindowProps = store.get('window-props') as Electron.Rectangle
     if (
-      tmpWindowProps !== undefined &&
-      tmpWindowProps.width !== undefined &&
-      tmpWindowProps.height !== undefined &&
+      tmpWindowProps &&
+      tmpWindowProps.width &&
+      tmpWindowProps.height &&
       tmpWindowProps.y !== undefined &&
       tmpWindowProps.x !== undefined
     ) {
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -94,12 +94,31 @@ const tsStore = new Store({
 async function createWindow(): Promise<BrowserWindow> {
   const { exitToTray, startInTray } = await GlobalConfig.get().getSettings()
 
+  let windowProps: Electron.Rectangle = {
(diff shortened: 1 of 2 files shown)


electron/main.ts
```diff
@@ -94,12 +94,31 @@ const tsStore = new Store({
...
+    const tmpWindowProps = store.get('window-props') as Electron.Rectangle
+    if (
+      tmpWindowProps !== undefined &&
+      tmpWindowProps.width !== undefined &&
+      tmpWindowProps.height !== undefined &&
+      tmpWindowProps.y !== undefined &&
+      tmpWindowProps.x !== undefined
+    ) {
+      windowProps = tmpWindowProps
+    }
```
Nocccer, line 114: If someone has a better way of checking this, let me know.

electron/main.ts
```diff
@@ -94,12 +94,31 @@ const tsStore = new Store({
...
+  let windowProps: Electron.Rectangle = {
+    height: 690,
+    width: 1200,
+    x: 0,
+    y: 0
+  }
+
+  if (store.has('window-props')) {
+    const tmpWindowProps = store.get('window-props') as Electron.Rectangle
+    if (
```
flavioislima, line 106:
```suggestion
    if (
      tmpWindowProps &&
      tmpWindowProps.width &&
      tmpWindowProps.heigh &&
      tmpWindowProps.y !== undefined &&
      tmpWindowProps.x !== undefined
    ) {
      windowProps = tmpWindowProps
    }
```

in Javascript we don't need to check if it's different from `undefined` since `undefined` is a `falsy` value. Unless in some specific situations when the value can be falsy as well. So unless some of those values can be 0, for instance, then you can change this suggestion. I guess perhaps y and x could be zero but not width and height.

flavioislima (commented):
Looks good, just left a comment

electron/main.ts
adityaruplaha: Imo we can keep it this way for consistency.

Nocccer:
@flavioislima check again. Insert your comment.

adityaruplaha (approved):
LGTM.

Also nice branch name. ;)

flavioislima (approved):
Tested here, works perfectly 👍🏽

Issue #924, opened 2022-01-26
Python 3.8 error


remivamp:
### Is this a game issue? If YES please use our Discord to get help.
no

### Try to get help on our Wiki first: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Troubleshooting

**Describe the bug**
With python 3.8 installed, get "Heroic requires python 3.8 or newer" before the launcher opens

**Add Logs**
INFO: [Legendary]: Location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary
[5930:0125/175604.840807:ERROR:sandbox_linux.cc(374)] InitializeSandbox() called with multiple threads in process gpu-process.
INFO: [Backend]: 
  Heroic Version: 2.1.0 Rayleigh
  Legendary Version:  0.20.25 Our Benefactors
  OS: LinuxMint KERNEL: 5.10.0-0.bpo.9-amd64 ARCH: x64
  CPU: AMD Ryzen 5 3600 6-Core Processor @3.6 GOVERNOR: schedutil
  RAM: Total: 16.7 GB Available: 5.63 GB
  GRAPHICS: GPU0: Ellesmere [Radeon RX 470/480] VRAM: 256MB DRIVER:  
  PROTOCOL: x11
  
INFO: [Backend]: User Not Found, removing it from Store
WARNING: [Backend]: Protocol already registered.
[5898:0125/175605.117480:ERROR:browser_main_loop.cc(269)] Gtk: gtk_widget_add_accelerator: assertion 'GTK_IS_ACCEL_GROUP (accel_group)' failed
[5898:0125/175605.117528:ERROR:browser_main_loop.cc(269)] Gtk: gtk_widget_add_accelerator: assertion 'GTK_IS_ACCEL_GROUP (accel_group)' failed
[5898:0125/175605.117572:ERROR:browser_main_loop.cc(269)] Gtk: gtk_widget_add_accelerator: assertion 'GTK_IS_ACCEL_GROUP (accel_group)' failed
WARNING: [Backend]: python command not found
WARNING: [Backend]: Command python3 version '3.7.3' not supported.
ERROR: [Backend]: Heroic requires Python 3.8 or newer.

[...]

PR #925, opened 2022-01-26
[Bugfix] Removed python check.

Nocccer:
The app was closed if the python version doesn't match the requirement (>=3.8).
Removed python check since legendary executable does not need it.
#924 

- [x] Removed python version check
- [x] Installed, started and removed a game with wine-ge
- [x] Login/Logout worked

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -38,7 +38,6 @@ import { GlobalConfig } from './config'
 import { LegendaryLibrary } from './legendary/library'
 import { LegendaryUser } from './legendary/user'
 import {
-  checkCommandVersion,
   checkForUpdates,
   clearCache,
   errorHandler,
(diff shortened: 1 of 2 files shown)


flavioislima:
can you test if this popup is necessary? because Rodney said that Legendary doesn't need python installed at all. Can you test it on a VM or something? so we can totally remove this check.

Nocccer:
I can test it via a script where i remove the path to python. Will do it later.

flavioislima:
> I can test it via a script where i remove the path to python. Will do it later.

yes that's right. Thanks

Nocccer:
I wonder if wine-ge needs a specific python version ?

flavioislima:
I don't think so, and still, that would not interfere with Heroic basics functionalities, like it is today.

flavioislima (approved):

PR #932, opened 2022-01-28
[Bugfix] Fixed epic service status check.

Nocccer:
Current epic service check checked for all services (Store, Fortnite, etc.). If one of them is down, we marked epic service as offline. 

- [x] Fixed the check to check for the Epic Games Store instead for all services. 
- [x] Adapt check, which can now check different services independently.
- [x] Fixed logError where error was not formated correctly which leads to javascript tracebacks.

---

### DIFF
--- a/electron/utils.ts
+++ b/electron/utils.ts
@@ -58,8 +58,9 @@ async function isEpicServiceOffline(
       `Failed to get epic service status with ${error}`,
       LogPrefix.Backend
     )
-    return true
   }
+
+  return true
 }
 
 export const getLegendaryVersion = async () => {
--- a/electron/utils.ts
+++ b/electron/utils.ts
@@ -43,8 +43,13 @@ async function isEpicServiceOffline(
   const epicStatusApi = 'https://status.epicgames.com/api/v2/components.json'
   const notification = new Notification({
     title: `${type} ${t('epic.offline-notification-title', 'offline')}`,
-    body: t('offline-notification-body', 'Heroic will not work probably!'),
-    urgency: 'critical'
+    body: t(
+      'epic.offline-notification-body',
+      'Heroic will maybe not work probably!'
+    ),
+    urgency: 'low',
+    timeoutType: 'default',
+    silent: false
   })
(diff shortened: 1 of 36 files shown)


electron/utils.ts
```diff
@@ -37,13 +37,29 @@ async function isOnline() {
...
+      if (name === type) {
+        return indicator === 'major'
+      }
+    }
+  } catch (error) {
+    logError(
+      `Failed to get epic service status with ${error}`,
+      LogPrefix.Backend
+    )
+    return true
```
flavioislima: I wonder if something goes wrong in this request should we return that there is an error on Epic side?
I think perhaps we should return false, since other things could impact this request and the epic servers could be up.
But I am not 100% on that tbh.

Nocccer: If we return false, we could remove the function completly.
The try catch is there, if the page is not reachable.

What we could do is just a frontend notification that epic services might be down?

flavioislima: I mean, the return false should be on the `catch` not at the end. 
Since you are doing a try/catch, the return should be inside the block for both.

Nocccer: Look again

electron/utils.ts
```diff
@@ -41,6 +41,11 @@ async function isEpicServiceOffline(
   type: 'Epic Games Store' | 'Fortnite' | 'Rocket League' = 'Epic Games Store'
 ) {
   const epicStatusApi = 'https://status.epicgames.com/api/v2/components.json'
+  const notification = new Notification({
```
flavioislima, line 44: Perhaps add a timeout for it with a bigger duration than the default.
Also, check if the i18n keys will be generated when using it inside a string literal like you are doing. If not, you need to add the keys as a comment in the code.
run `yarn i18n` and check if they keys were created.

Nocccer: String are generated. Can't set timeout time.

flavioislima (approved):
ok, looks good! 👍🏽

PR #939, opened 2022-01-30
[General] Add timestamp and relined logs

Nocccer:
- Added timestamp for each logged message
- Lined logs so it looks similiar to a table

image

---

### DIFF
--- a/electron/legendary/library.ts
+++ b/electron/legendary/library.ts
@@ -98,7 +98,7 @@ class LegendaryLibrary {
       const child = spawn(legendaryBin, ['list', getUeAssets])
       child.stderr.on('data', (data) => {
         if (`${data}`.includes('ERROR')) {
-          logError(`${data}`, LogPrefix.Legendary)
+          logError(`${data}`.trim(), LogPrefix.Legendary)
         } else {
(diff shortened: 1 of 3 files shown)


electron/legendary/library.ts
```diff
@@ -100,7 +100,7 @@ class LegendaryLibrary {
         if (`${data}`.includes('ERROR')) {
           logError(`${data}`, LogPrefix.Legendary)
```
flavioislima, line 101: Perhaps add trim() to this one would be good as well :)

electron/logger/logger.ts
```diff
@@ -100,10 +135,23 @@ export function logInfo(
   prefix: LogPrefix = LogPrefix.General,
   skipLogToFile = false
 ) {
-  const prefixString = prefix !== LogPrefix.General ? `[${prefix}]: ` : ''
-  const extendText = `INFO: ${prefixString}${convertToStringArray(text).join(
-    ' '
-  )}`
+  // time
+  const ts = new Date()
+  const timeString = `(${[
```
flavioislima, line 140: if this is repeated so much, perhaps create another function that returns this result.

flavioislima (approved):
Nice! Looks good 👍🏽

PR #940, opened 2022-01-30
[General] Added checkbox to uninstall modal to remove prefix aswell

Nocccer:
Added checkbox to uninstall modal to choose if prefix should be removed aswell
image

---

### DIFF
--- a/src/helpers/library.ts
+++ b/src/helpers/library.ts
@@ -143,14 +143,22 @@ async function uninstall({ appName, handleGameStatus, t }: UninstallArgs) {
     buttons: [t('box.yes'), t('box.no')],
     message: t('gamepage:box.uninstall.message'),
     title: t('gamepage:box.uninstall.title'),
-    type: 'warning'
+    type: 'warning',
+    checkboxLabel: t(
+      'gamepage:box.uninstall.checkbox',
+      'Uninstall prefix aswell'
+    ),
+    checkboxChecked: true
   }
 
-  const { response } = await ipcRenderer.invoke('openMessageBox', args)
+  const { response, checkboxChecked } = await ipcRenderer.invoke(
+    'openMessageBox',
+    args
+  )
 
   if (response === 0) {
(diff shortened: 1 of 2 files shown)


flavioislima:
Looks good. But I think Uninstall Prefix is not the right way of saying it. Perhaps Removing the prefix instead.
Like:
**Would you like to remove the Prefix: `path/to/prefix` as well? This can't be undone.**
what do you think?

Nocccer:
> Looks good. But I think Uninstall Prefix is not the right way of saying it. Perhaps Removing the prefix instead. Like: **Would you like to remove the Prefix: `path/to/prefix` as well? This can't be undone.** what do you think?

Sure. "Uninstall prefix aswell" is only the default value for i18n translation

flavioislima:
> Sure. "Uninstall prefix aswell" is only the default value for i18n translation

Yes but in this case, all translations will be based on that.

Nocccer:
@flavioislima added the suggested message

src/helpers/library.ts
```diff
@@ -146,11 +146,33 @@ async function uninstall({ appName, handleGameStatus, t }: UninstallArgs) {
...
+        t(
+          'gamepage:box.uninstall.checkbox',
+          "Would you like to remove the prefix aswell? This can't be undone."
+        ),
+        `${t(
+          'gamepage:box.uninstall.checkbox_prefix',
+          'Prefix'
+        )}: ${wineprefix}`
+      ].join('\n'),
+      checkboxChecked: true
```
flavioislima, line 164: I think should be false by default.
```suggestion
      checkboxChecked: false
```

flavioislima (changes requested):
I think only one last thing that I missed on first review. Leave that checkbox unchecked by default because some people have one prefix for several games. I do that sometimes and keep specific prefixes only for specific games.

flavioislima (approved):

PR #950, opened 2022-02-03
[Linux] Fix DXVK/VKD3D installation

flavioislima:
- [x] Changed from lutris to official repos for DXVK and VKD3D
- [x] Split DXVK and VKD3D settings
- [x] Add ZSTD as dependency on package.json
- [x] Add Dialog with an error message if ZSTD was not found.

---


package.json
```diff
@@ -77,21 +77,24 @@
       "packageCategory": "games",
       "depends": [
         "gawk",
-        "curl"
```
Nocccer, line 80: maybe add tar as dependency, too?

Nocccer:
LGTM. Maybe we should add tar as dependency, too? What do you think ?

Also i would love to have all the tool stuff (dxvk, vkd3d, ...) in a new folder in the backend. 
This folder contains also the ipc functions, so main is getting more cleaned up and readable. But we can do that in a later pr.

Nocccer (changes requested):

package.json
flavioislima: I think its impossible for a distro do not have TAR installed since its used to extract basically everything. If this was rar or zip, perhaps we would need unzip or unrar but not tar.

Nocccer: Yes i agree but then curl should also not be on the list :D. I mean it does not hurt if with add tar to the list ? If someone does not have it for some reason he get an error message, else not.

flavioislima: Its complicated to have dependencies here because this will ask for installing when installing the package.
some time ago I added wine as a dependency here and it broke several installations because some distros have different name for wine. So this is last resource.
I will keep as it is for now and lets see if someone will appear with a missing tar some day.

PR #963, opened 2022-02-07
Increase touch input

wiryfuture:
Significantly increases the size of the touch input on checkboxes by including the label text as part of the touch input.
Regularly, this would already be happening, but the tags were in a weird order before this which prevented it from actually happening.
This makes it much easier to toggle checkboxes with more coarse input (like with the touchscreen on a steam deck 🤫 )

Before: hovering and clicking on text does nothing; had to click directly on checkbox:
Screenshot_20220206_235715
Screenshot_20220206_235811

Now: hovering and clicking on text area also counts as clicking on checkbox:
Screenshot_20220207_000024
After clicking; toggles checkbox:
Screenshot_20220207_000049

❗ **WARNING** ❗ 
I am getting loads of errors and i legitimately cannot tell why. I think my typescript language server is broken, because everything seems to be working fine and i see this on main as well. Please can someone check with their own code client? cheers.
Screenshot_20220207_000305
Screenshot_20220207_000454


wiryfuture:
wtf why is there so many commits- screw this, im making patchfiles and making a new fork

PR #964, opened 2022-02-07
[General] Improve navigation on touchscreen and gamepad and revert some colours

wiryfuture:
Makes clickable area for checkboxes far bigger so it's easier to toggle on touchscreens like the steam deck (and just in general)

before: hover and click does nothing, have to go on checkbox
Screenshot_20220206_235715
Screenshot_20220206_235811

after: click text and it works (well, technically, the entire 'row' beside the checkbox) 
Screenshot_20220207_000024
Screenshot_20220207_000049


flavioislima:
Tested here, looks good. The only missing color now is the Download icon on both the Game Card and the Wine Manager.
they need to be `--seconday` as well.

wiryfuture:
> Tested here, looks good. The only missing color now is the Download icon on both the Game Card and the Wine Manager. they need to be `--seconday` as well.

done!

src/screens/Settings/components/OtherSettings/index.tsx
```diff
@@ -306,23 +321,23 @@ export default function OtherSettings({
               onChange={handleLauncherArgs}
             />
           </span>
+          <InfoBox text="infobox.help">
```
arielj, line 324: this InfoBox code is duplicated, maybe you can store the element in a variable before the conditionals and then do `{infoBox}`

src/screens/Settings/index.tsx
```diff
@@ -343,39 +343,53 @@ function Settings() {
   return (
     <>
       <div className="Settings">
-        <div className="settingsNavbar">
+        <nav role={'list'} className="settingsNavbar">
```
arielj, line 346: you can do `role='list'` without the curly braces

src/components/UI/LanguageSelector/index.tsx
```diff
@@ -99,7 +99,7 @@ export default function LanguageSelector({
     <select
       data-testid="languageSelector"
       onChange={(event) => handleLanguageChange(event.target.value)}
-      className={className}
+      className={className + ' is-drop-down'}
```
arielj, line 102: are these `is-drop-down` classes needed? I see they are added to select tags which are, by default, drop downs, isn't the `.settingSelect` selector enough?

arielj (commented):
I haven't tested it locally, but I added some code style comments

src/components/UI/LanguageSelector/index.tsx
wiryfuture: Needed for the cursor to change appropriately

flavioislima, line 102:
```suggestion
      className={`${className} is-drop-down`}
```

flavioislima (dismissed):
Looks good overall. Left a small comment only but I believe is good to be merged.

src/components/UI/LanguageSelector/index.tsx
wiryfuture: bruh package to join classes is that like isOdd and isEven

wiryfuture: > 

done

flavioislima (approved):
Great job!

Issue #978, opened 2022-02-17
Check for Updates On Startup does not toggle off


redromnon:
### Try to get help on our Wiki first: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Troubleshooting

**Describe the bug**
The checkbox _Check for Updates on Startup_, under General (Global) Settings, always toggles itself on despite turning it off. When I reopen the app or move to another tab/window and come back to the General Settings window, the checkbox is turned on automatically. I believe the toggle's state doesn't get saved and thus leads to this issue.

**Add Logs**
None

**To Reproduce**
1. Go to Settings
2. Click General
3. Under Global Settings 
4. Uncheck the _Check for Updates on Startup_ checkbox

**Expected behavior**
The checkbox should be always unchecked when done so. 

**Screenshots**
![Check_for_Updates_Heroic](https://user-images.githubusercontent.com/74495920/154425388-2eda4876-1a80-4d44-9d88-762ddc24db6e.png)

**Desktop (please complete the following information):**

- OS: Pop_OS 20.04
- Heroic Version - 2.1.1 Rayleigh

**Additional context**
None

Nocccer:
Can reproduce it. Will try to fix it.

flavioislima:
> Can reproduce it. Will try to fix it.

I believe the issue is on this line:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/5d72a3860e427ba68267b7d5906b11ba0c3d27f1/src/screens/Settings/index.tsx#L94

Because this settings'  default is `true`, but on this line, it starts as `false`. So there is some conflict.
changing here to true should fix the issue.

Nocccer:
@flavioislima already fixed it for ~2h

PR #980, opened 2022-02-17
[General] Fix option check for update on startup

Nocccer:
#978 
---


src/screens/Settings/index.tsx
```diff
@@ -227,7 +227,11 @@ function Settings() {
       setAddDesktopShortcuts(config.addDesktopShortcuts || false)
       setAddGamesToStartMenu(config.addStartMenuShortcuts || false)
       setCustomWinePaths(config.customWinePaths || [])
-      setCheckForUpdatesOnStartup(config.checkForUpdatesOnStartup || true)
+      setCheckForUpdatesOnStartup(
```
flavioislima, line 230: I believe this doesnt fix the issue because this check does the same as before. The issue is on line:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/5d72a3860e427ba68267b7d5906b11ba0c3d27f1/src/screens/Settings/index.tsx#L94

Nocccer: I will check again. I mean for me it fixed it.
For my Understanding:
undefined || true = true
false || true = true
-> Actually if the value is false you wanna set it to false and not true!

So maybe we need to fix both places ?

Nocccer: I tested the option alot. set it, close the app and open again. Stores the value now. Before it was set always to true because of that ||

Nocccer: In my eyes it is pretty simple logical stuff there. Doing || with a fixed boolean value is never good!

Nocccer: Tested System Tray options and everything seems fine after the removing

src/screens/Settings/index.tsx
```diff
@@ -194,43 +194,47 @@ function Settings() {
...
-      setEnableEsync(config.enableEsync || false)
-      setEnableFsync(config.enableFsync || false)
-      setEnableFSR(config.enableFSR || false)
+      setExitToTray(config.exitToTray)
+      setStartInTray(config.startInTray)
+      setDarkTrayIcon(config.darkTrayIcon)
+      setDiscordRPC(config.discordRPC)
+      setAutoInstallDxvk(config.autoInstallDxvk)
+      setAutoInstallVkd3d(config.autoInstallVkd3d)
+      setEnableEsync(config.enableEsync)
```
flavioislima, line 218: With options like that. that are not created on the config file by default,  It will `setEnableesync` with which value?
It starts as false, but then it will set as `undefined` no?

Nocccer: Need to check this, once more.

From my understanding, because you told me alot of times, undefined == false casted on a bollean.
So if the functions set... expect a boolean then undefined leeds to false.

flavioislima: yes, but when you check a condition, then it is a `falsy` value. Here it is setting the value to undefined. 
If you leave it like that it will store the initial value as undefined and might break something in the settings page.

flavioislima (approved):
Tested here, looks good! 👍🏽

PR #991, opened 2022-02-21
Better Wine Installation management, better DXVK installation, some small fixes here and there

CommandMC:
This improves a few things:
- [X] The log message about Proton is formatted properly again
- [X] The log message about the Heroic Protocol already being registered is now an Info message instead of a Warning (since it happens on every launch)
- [X] In code, it's now easier to tell if a Wine Version is Wine/Proton/Crossover. This is due to a new `type` property on WineInstallation
- [X] A new `runWineCommand` function now exists, to make it easy to run a Wine command with the context of the current game
- [X] Wine prefix creation was reworked: It now properly creates prefixes when the dir already exists, and Proton prefix creation also works now
- [ ]   DXVK is no longer installed with the `setup_dxvk` script, which makes it possible to install it on some Wine-GE versions again

---

### DIFF
--- a/electron/config.ts
+++ b/electron/config.ts
@@ -109,23 +109,22 @@ abstract class GlobalConfig {
    * @returns Promise<WineInstallation>
    */
   public async getDefaultWine(): Promise<WineInstallation> {
+    const defaultWine: WineInstallation = {
+      bin: '',
+      name: 'Default Wine - Not Found',
+      type: 'wine'
+    }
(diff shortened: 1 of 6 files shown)


CommandMC:
With GOG support merged now, there are a lot of things here that need to be modified again. I'll close this, rebase on main and then make some smaller PRs instead. This has gotten too large either way

PR #1012, opened 2022-02-24
Remove unnecessary spaces from launch commands

CommandMC:
Before, all necessary launch options were inserted into a string with pre-defined spaces. This caused issues, as when an option wasn't set, a space was still inserted.
Now, the "command parts" are collected in a list, which is then filtered to remove empty entries, and then joined together with spaces.

---


flavioislima (approved):
Great job!

PR #1024, opened 2022-02-27
Added missing alternate exe parameter for GOG Linux

redromnon:
I've added the missing alternate executable variable that is included in the launch parameters for GOG Linux games. Since it was a very tiny addition, I didn't build and test it.
---

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -195,7 +195,7 @@ async function launch(
       }
       command = `${envVars} ${gogdlBin} launch "${
         gameInfo.install.install_path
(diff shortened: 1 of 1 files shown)


redromnon:
Well, I've found out that gogdl doesn't support an alternative executable yet. Since it is already implemented in the code for GOG Windows games, I guess you can consider this for GOG Linux whenever the support arrives.

flavioislima:
Let us know when this is ready-for-review by using the Label.

redromnon:
It's Ready For Review.

PR #1040, opened 2022-03-04
Wine-GE-Proton Compatibility

nezd5553:
This fixes an issue caused by the recent renaming of [Wine-GE to Wine-GE-Proton](https://github.com/GloriousEggroll/wine-ge-custom/releases/tag/GE-Proton7-1). Heroic would try to use it as Proton instead of as Wine, so games would not launch.

I also noticed an issue while testing where Wine versions couldn't be selected if Crossover was not installed, which has also been fixed.

### DIFF
--- a/electron/config.ts
+++ b/electron/config.ts
@@ -173,21 +173,22 @@ abstract class GlobalConfig {
     } else if (!isWindows) {
       // Linux
       const crossoverWineBin = '/opt/cxoffice/bin/wine'
-      if (existsSync(crossoverWineBin)) {
-        const crossoverVersion = (
-          await execAsync(crossoverWineBin + ' --version')
-        ).stdout
-          .toString()
-          .split('\n')[2]
-          .split(':')[1]
-          .trim()
-        crossover.add({
-          bin: `'${crossoverWineBin}'`,
-          name: `CrossOver - ${crossoverVersion}`,
-          type: 'crossover',
-          ...this.getWineExecs(crossoverWineBin)
-        })
+      if (!existsSync(crossoverWineBin)) {
+        return crossover
       }
+      const crossoverVersion = (
+        await execAsync(crossoverWineBin + ' --version')
+      ).stdout
+        .toString()
+        .split('\n')[2]
+        .split(':')[1]
+        .trim()
+      crossover.add({
+        bin: `'${crossoverWineBin}'`,
+        name: `CrossOver - ${crossoverVersion}`,
+        type: 'crossover',
+        ...this.getWineExecs(crossoverWineBin)
+      })
     }
 
     return crossover
--- a/electron/config.ts
+++ b/electron/config.ts
@@ -181,14 +181,12 @@ abstract class GlobalConfig {
           .split('\n')[2]
           .split(':')[1]
           .trim()
-        if (existsSync(crossoverWineBin)) {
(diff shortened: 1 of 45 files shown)


electron/gog/setup.ts
```diff
@@ -41,8 +41,9 @@ async function setup(appName: string): Promise<void> {
   const crossoverEnv =
     isCrossover && crossoverBottle ? `CX_BOTTLE=${crossoverBottle}` : ''
   const isProton =
-    gameSettings.wineVersion.name.includes('Proton') ||
-    gameSettings.wineVersion.name.includes('Steam')
+    (gameSettings.wineVersion.name.includes('Proton') ||
+      gameSettings.wineVersion.name.includes('Steam')) &&
+    !gameSettings.wineVersion.name.includes('lutris')
```
CommandMC, line 46: The whole approach to differentiating between Proton, Wine and CX by checking the name/path is hacky.  
Instead of adding onto those hacks, we should modify the frontend to now respect `WineInstallation.type`, which will almost always be correct (one edge case exists when adding custom Wine versions, I'm planning on fixing that up soon)

electron/config.ts
```diff
@@ -173,20 +173,22 @@ abstract class GlobalConfig {
...
-        .split(':')[1]
-        .trim()
       if (existsSync(crossoverWineBin)) {
-        crossover.add({
-          bin: `'${crossoverWineBin}'`,
-          name: `CrossOver - ${crossoverVersion}`,
-          type: 'crossover',
-          ...this.getWineExecs(crossoverWineBin)
-        })
+        const crossoverVersion = (
```
CommandMC, line 177: Arg, yup, this one's my fault. Thanks for catching that

electron/config.ts
```diff
@@ -173,20 +173,22 @@ abstract class GlobalConfig {
...
+          await execAsync(crossoverWineBin + ' --version')
+        ).stdout
+          .toString()
+          .split('\n')[2]
+          .split(':')[1]
+          .trim()
+        if (existsSync(crossoverWineBin)) {
```
CommandMC, line 184: This condition will always be true (since you've already checked for that above). You can just add the version directly

electron/config.ts
flavioislima: also, just a note: when using `await` is good to put it inside a try/catch block.

electron/config.ts
nezd5553: Oh yeah, I missed that.

electron/gog/setup.ts
nezd5553: It looks like using `WineInstallation.type` in `electron/launcher.ts` and `electron/gog/setup.ts` works, but the `WineInstallation` in `src/types.sh` doesn't have `type`.

CommandMC: > but the `WineInstallation` in `src/types.ts` doesn't have type.

This is just caused by me forgetting to update that, you can replace the definition with the one from `electron/types.ts`

electron/config.ts
CommandMC: This is now superseded by https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1046

nezd5553:
I'm happy I was able to contribute to such a fantastic project!

PR #1045, opened 2022-03-05
[All] Truly use `path.join()` everywhere (Fix "Show log file in folder" and wrong game settings path)

CommandMC:
This PR addresses two things:
- Paths are now combined with [`path.join()`](https://nodejs.org/api/path.html#pathjoinpaths) instead of just combining them together as strings
- On Windows, a workaround is used to display files in a folder. This fixes the "Show log file in folder" button not doing anything

**It is very likely that I missed a path when moving to join**. Please double- and triple-check everywhere and let me know if you find any issues.

---

### DIFF
--- a/electron/config.ts
+++ b/electron/config.ts
@@ -17,6 +17,7 @@ import {
   heroicInstallPath,
   heroicToolsPath,
   home,
+  isFlatpak,
   isMac,
   isWindows
 } from './constants'
@@ -475,7 +476,7 @@ class GlobalConfigV0 extends GlobalConfig {
       addStartMenuShortcuts: false,
       autoInstallDxvk: false,
       autoInstallVkd3d: false,
-      checkForUpdatesOnStartup: true,
+      checkForUpdatesOnStartup: !isFlatpak,
       customWinePaths: isWindows ? null : [],
       defaultInstallPath: heroicInstallPath,
       defaultWinePrefix: `${home}/Games/Heroic/Prefixes`,
--- a/electron/config.ts
+++ b/electron/config.ts
@@ -173,7 +173,7 @@ abstract class GlobalConfig {
       })
     } else if (!isWindows) {
       // Linux
-      const crossoverWineBin = join('opt', 'cxoffice', 'bin', 'wine')
+      const crossoverWineBin = '/opt/cxoffice/bin/wine'
       if (!existsSync(crossoverWineBin)) {
         return crossover
       }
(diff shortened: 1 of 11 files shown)


CommandMC:
Resolves #1062

electron/config.ts
```diff
@@ -242,9 +242,17 @@ abstract class GlobalConfig {
     // Known places where Steam might be found.
     // Just add a new string here in case another path is found on another distro.
     const steamPaths = [
-      `${home}/.steam`,
-      `${home}/.var/app/com.valvesoftware.Steam/.local/share/Steam`,
-      '/usr/share/steam'
+      join(home, '.steam'),
```
flavioislima, line 245: I believe this is overkill. Because we wont use this on Windows or Mac so the path will always use `slashes`.
`join` is nice to use when we have a path that is used across platforms.
Doing this everywhere gives no benefit and makes the code less legible imo.

electron/utils.ts
```diff
@@ -326,7 +328,11 @@ function resetHeroic() {
 function showItemInFolder(item: string) {
   if (existsSync(item)) {
     try {
-      shell.showItemInFolder(item)
+      if (isWindows) {
+        execAsync(`explorer /Select,"${item}"`)
```
flavioislima, line 332: `shell.showItemInFolder` doesnt work on windows?

electron/config.ts
CommandMC: Yes, agreed. I'll remove it once I get home

electron/utils.ts
CommandMC: It didn't work for me, and @Laria (can't tag properly on mobile) also had the same issue

flavioislima: Hm ok then.
I forgot to test this yesterday when I was on windows.

flavioislima:
I also need a favor because I wont have time to open a PR today with some fixes for the flatpak.
Can you change this line:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/e2642e258e8e403ef0d51aabe853c1189504e01c/electron/config.ts#L470
to be `false` when the package `isFlatpak`? Since their updates are automatically makes no sense for us to show any message about it.

Also add `required: false` to the build workflow, I don't think we need to force all builds to succeed for merging PRs because sometimes they fail because some error in github action or on npmjs:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/e2642e258e8e403ef0d51aabe853c1189504e01c/.github/workflows/build-prs.yml#L7

flavioislima (approved):
tested here and looks good 👍🏽

PR #1060, opened 2022-03-08
[Fix] error when gamesdb returns 404

imLinguin:
Should fix errors with Cyberpunk 2077 Goodies, and other products that return 404 from gamesdb endpoint

### DIFF
--- a/electron/gog/library.ts
+++ b/electron/gog/library.ts
@@ -325,7 +325,7 @@ export class GOGLibrary {
     let verticalCover: string
     let horizontalCover: string
     let description: string
-    if (gamesdbData.game) {
+    if (gamesdbData?.game) {
       const developers: Array<string> = []
       for (const developer of gamesdbData.game.developers) {
         developers.push(developer.name)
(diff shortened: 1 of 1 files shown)


flavioislima (approved):

Oaklight:
some games still returns 404:
- Hundred Days - Winemaking Simulator
- The Lion's Song

imLinguin:
Well, that's expected in some cases. However at this moment it should fallback to data from GOG API if available

Oaklight:
Why GOG? I should be more specific. These are Epic games

imLinguin:
Could you post the log proving that? GamesDB is only used for GOG at this moment

Oaklight:
Oh, I see. Should I create another issue then? I only meant to report 404 issue for some of my Epic games when I tried to go to their Epic store page in the Tool page. This is the only 404-related open issue page I found so...

imLinguin:
Separate issue would be great. If it's causing serious issues then create one

Oaklight:
so far I was able to run them without error. The issue now is just the 404 webpage from Tool page, while I can search and view directly from the Store page.

imLinguin:
Then it's a different type of 404 than was improved on this PR

PR #1064, opened 2022-03-09
Add more UI effects, fixes w/ user-select, other tweaks.

8Bitz0:
### Changes
- [x] Make the login screen more transparent.
- [x] Make "gamecards" have more depth.
- [x] Remove variables that should not be used to end note and change dependents.
- [x] Make blur amounts consistent with a defined variable.

### Fixes
- [x] Could not highlight and copy URLs, logs, game description and requirements.

### Screenshots
Screen Shot 2022-03-14 at 09 02 33

Screen Shot 2022-03-14 at 09 03 11
---


flavioislima:
I will take a look later today.
Meanwhile, can you make both cards on login the same size?

8Bitz0:
The blur might slow down low-end devices. An option to remove blur might be necessary. I will test this later today.

8Bitz0:
The login screen is a little sluggish on slower devices, but it is still usable.

flavioislima:
Could not test this one, I believe it is because it's not updated with main now that I merged the others and have a few conflicts.
But I am going through the code while that 👍🏽

flavioislima:
Ok, tested here, and looks good.
My only concern is that the Continue button might get a bit hard to find for some people, especially people some kind of visual restraint. So we should have a background on it.
And the blur could also use the background color we have so it would be more consistent with the rest of the look.
But in general, looks good.

flavioislima (approved):

flavioislima:
This one looks good now, I will merge after fixing the conflicts 👍

8Bitz0:
Oops! I didn't mean to push that to the main branch and dismiss the review. Sorry about that!

flavioislima (approved):

PR #1073, opened 2022-03-10
[FIX] Issues with Epic Login and refresh Library

flavioislima:
- Add function to deal with spaces on all platforms
- Add shell to `spawn` methods so binaries with spaces can work normally 
- change how we deal with legendary commands even if the first function to deal with spaces is not reliable

---


Nocccer (approved):
Lgtm

electron/constants.ts
```diff
@@ -43,8 +79,13 @@ function getGOGdlBin() {
         isWindows ? 'gogdl.exe' : 'gogdl'
       )
   )
+
+  // dont fix windows issues
+  if (bin.includes(' ')) {
+    return `"${bin}"`
```
imLinguin, line 85: Based on that, we should make sure gogdl is not wrapped twice in quotes when actually executing it
`gog/games.ts`, `gog/library.ts`

CommandMC (dismissed):
Well, it's a band-aid fix that's surely gonna bite us in the butt in the future  
But it's still a fix

Works on Linux and my Win VM

imLinguin (approved):

flavioislima:
> Well, it's a band-aid fix that's surely gonna bite us in the butt in the future But it's still a fix
> 
> Works on Linux and my Win VM

Ahahaha
What are you worried about? something in specific?

I was thinking that we should make a global function that could be used across the legendary and the gogdl parts. So we could receive the full path for the binary and do the necessary treatments/checks for it to run.

But one thing I learned today was this `process.chdir` this is very useful and maybe you can use that @imLinguin on the gogdl side. to avoid using that workaround for windows with the  `&` I tested and worked fine but I didn't want to commit since I would need more time testing.

electron/constants.ts
flavioislima: so I will return to the old code then. Since GOG is working normally on 2.2.4

flavioislima:
@CommandMC yes, I believe using `chdir` for legendary and gogdl only would be good. 
Unfortunately, nodeJS spawn doesn't work with name with spaces because it expects only one command and not the whole path. So that's the only way I could find.
If you find another way of fixing this, especially on windows, we can talk about it.

PR #1092, opened 2022-03-11
Refactor Legendary/GOGDL handling

CommandMC:
This PR unifies every instance of running/invoking Legendary/GOGDL to use one function each (`runLegendaryCommand`/`runGogdlCommand`).It improves a few things in the process:
 - `process.chdir()` is no longer necessary
 - Adding directories to `%PATH%`/`$PATH` is no longer necessary
 - Paths with spaces are handled properly (by using the `cwd` option of `spawn`)
 - Every Legendary/GOGDL command can now be logged, all it takes is one argument
 - Alternative binaries do not require a restart
 - Some misc fixes/changes
   - Old & unused API removed
   - Download progress is formatted nicer on Linux
   - `ExecResult` now has an optional `fullCommand` argument to indicate the exact command that was ran

---

Note: Testing was, as usual, done on Linux and Windows. I don't have a Mac, if anyone could test on there, that would be great


adityaruplaha:
Someone please add this to Project Beans. :P 

Thanks for this PR, I wanted to get this done for a while but didn't have the time.

CommandMC:
There's an issue with Linux-native games not installing properly right now. I'm working on it

CommandMC:
It seems like everything is working now, the only issue is that installing GOG games doesn't get registered properly so you have to press the "Refresh Library" button once.  
I honestly can't say why this happens since I haven't really changed that code

electron/config.ts
```diff
@@ -223,7 +224,7 @@ abstract class GlobalConfig {
       })
     })
 
-    const lutrisPath = `${homedir()}/.local/share/lutris`
+    const lutrisPath = `${home}/.local/share/lutris`
```
flavioislima, line 227: I needed to change this to `homedir` because on flatpak it was looking inside `flatpakHome`, that's goes to the sandboxed home. But actually steam and lutris are installed outside this home, on the standard home.
So perhaps we can use the `homedir` everywhere except if it relies on the sandboxed home of flatpak

electron/utils.ts
```diff
@@ -337,6 +344,37 @@ function showItemInFolder(item: string) {
...
 
+function splitPathAndName(fullPath: string): { dir: string; bin: string } {
+  const dir = dirname(fullPath)
+  let bin = basename(fullPath)
+  // On Windows, you can just launch executables that are in the current working directory
+  // On Linux, you have to add a ./
+  if (!isWindows) {
+    bin = './' + bin
+  }
+  return { dir: dir, bin: bin }
```
flavioislima, line 355: if key and value have the same name, you can just do it like this:
`return {dir, bin}`

ofc the way you are doing its not wrong as well 👍🏽

electron/legendary/library.ts
```diff
@@ -550,4 +545,87 @@ class LegendaryLibrary {
   }
 }
 
-export { LegendaryLibrary }
+/**
+ * Runs legendary with the given command
+ * @param commandParts The command to run, e. g. 'list', 'egl-sync'...
+ */
+// TODO: This isn't the right place for this, but I don't know where else to put it. 'launcher.ts' maybe?
+export async function runLegendaryCommand(
```
flavioislima, line 553: Pretty good refactor, this avoid repetition by a lot 👍🏽

electron/legendary/games.ts
```diff
@@ -193,66 +193,68 @@ class LegendaryGame extends Game {
...
+    const oldInstallPath = (await this.getGameInfo()).install.install_path
 
-    if (isWindows) {
-      newInstallPath += '\\' + install_path.split('\\').slice(-1)[0]
-    } else {
-      newInstallPath += '/' + install_path.split('/').slice(-1)[0]
-    }
+    newInstallPath = join(newInstallPath, basename(oldInstallPath))
+
+    const command = `mv -f '${oldInstallPath}' '${newInstallPath}'`
```
flavioislima, line 200: Just a reminder to test this on windows because might not be working.

flavioislima (dismissed):
Tested here for a bit and could not find any bug or issue so looks good.
I believe testing on other platforms and on Flatpak would be good as well. I will try to test on a Mac tomorrow.
Also, @imLinguin would be nice if you could take a look at it as well, especially the GOG part.

But pretty good job with the refactor.
The code is cleaner and more efficient now. Also, more easy to scalate ⚔️

electron/legendary/games.ts
CommandMC: I haven't changed the command, I just moved it to up here to display it in the log

electron/gog/games.ts
```diff
@@ -124,75 +129,79 @@ class GOGGame extends Game {
...
+      installedWithDLCs: installDlcs,
+      language: installLanguage,
+      versionEtag: isLinuxNative ? '' : installInfo.manifest.versionEtag,
+      buildId: isLinuxNative ? '' : installInfo.game.buildId
+    }
+    const array: Array<InstalledInfo> =
+      (installedGamesStore.get('installed') as Array<InstalledInfo>) || []
+    array.push(installedData)
+    installedGamesStore.set('installed', array)
+    GOGLibrary.get().refreshInstalled()
```
imLinguin, line 186: About bug you mentioned. This might be a race condition we are looking for, since refreshInstalled reads from the same config. Maybe a function to update installed games directly in memory (instead of reading from the same file again) will solve the issue.

imLinguin (dismissed):
I looked at GOG changes. Looks good to me

electron/main.ts
```diff
@@ -1164,36 +1160,24 @@ ipcMain.handle(
...
-    if (stderr.includes('ERROR')) {
-      logError(`${stderr}`, LogPrefix.Legendary)
-      return 'Error'
-    }
-    return `${stdout} - ${stderr}`
-  } catch (error) {
-    logError(`${error}`, LogPrefix.Legendary)
-    return 'Error'
+  let legendaryArgs = Array<string>()
+  if (args != 'unlink') {
```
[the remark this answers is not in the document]
CommandMC: Is it really necessary to check the type here though? Can anything other than a string be equal to `'unlink'`?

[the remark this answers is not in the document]
flavioislima: Yes, it's not a good practice as well. I thought we had a eslint rule for that, would be good to add.

electron/legendary/library.ts
flavioislima: Yes, pretty good especially when we add a `Heroic Runner` in the future and also others like Amazon and so on.
We can follow the same pattern and will be a lot easier to grow from here :)

electron/legendary/games.ts
flavioislima: Yes, its a reminder for me tbh because I forgot to check if was working on windows and I believe someone said on discord that it was broken since we should be using the `Move-Item` instead, but that's another issue not related with this PR.

electron/gog/games.ts
flavioislima: Maybe we could do it like this from now on. Perhaps on another PR:
After refreshing the library on the backend, we send an event to the frontend to refresh the state.
This way we could solve the racing conditions.

electron/config.ts
```diff
@@ -243,8 +244,8 @@ abstract class GlobalConfig {
     // Known places where Steam might be found.
     // Just add a new string here in case another path is found on another distro.
     const steamPaths = [
-      join(homedir(), '.steam'),
-      join(homedir(), '.var/app/com.valvesoftware.Steam/.local/share/Steam'),
+      join(home, '.steam'),
```
flavioislima, line 247: Here we should also use `homedir()` because on Flatpak it won't find Steam proton, especially on the Deck.
Thats why perhaps we can kill the `home` variable, or change it to be `homedir` and if we need something specific for the flatpak home we can use `flatpakHome`.

flavioislima:
Tested these changes on a Mac and looks good as well. 
The only issue I could find was that after logout the status of the Login Card still shows as logged in. Although I am not sure if this is Mac-specific or if it is related to these changes.

electron/legendary/games.ts
CommandMC: `mv` is an alias for `Move-Item` on Windows. Although I don't know how well it handles that `-f` parameter, it should be `-force`

electron/config.ts
CommandMC: Alright, I'll get on that once I'm home

electron/gog/library.ts
```diff
@@ -591,3 +608,89 @@ export class GOGLibrary {
...
+
+  return new Promise((res, rej) => {
+    const child = spawn(bin, commandParts, { cwd: dir, env: env })
+
+    const stdout = new Array<string>()
+    const stderr = new Array<string>()
+
+    if (logFile) {
+      child.stdout.on('data', (data: Buffer) => {
+        appendFileSync(logFile, data.toString())
```
dawidgarus, line 642: Kudos for all the changes, but maybe you shouldn't use sync method. Also, new file handle is created for each chunk. This could lead to blocking of app.

I suggest opening a file once and appending to it.
It would be also nice to use node's readline module. This would enable further changes to fix the atrocity that is the `requestGameProgress` handler: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/electron/main.ts#L1100

CommandMC: I agree that this definitely needs improvement, but I'm afraid that I don't know enough about the language yet to make these changes + they don't *really* fit into this PR. I'd say we merge this for now & then we can look into improving it.

dawidgarus: Sure, it can be added later.

flavioislima (approved):

PR #1100, opened 2022-03-12
[Windows] Fix "Show log file in folder" once and for all

CommandMC:
- Remove the workaround for showing a file, as `shell.showItemInFolder` actually works fine (it just didn't work back when we mixed / and \ in paths
- Properly remove forbidden characters from log filename

---


Nocccer (approved):
LGTM. If shell does not work on windows also with mixed / \ then we need to open a issue on electron side. They should handle such things too

PR #1105, opened 2022-03-13
[General] add .editorconfig, align text and icons in sidebar, fix sidebar shadow, sidebar css clean up

dawidgarus:
- Fixed drop shadow of the sidebar so it won't cover foreground elements in the main content
- Aligned icons and labels in the sidebar. All labels start the same amount of pixels from the left now
- Clean up of sidebar code, mostly CSS
- Added .editorconfig - I figured that instead of changing my IDE configuration it's better to create .editorconfig as bonus for every contributor

---


flavioislima:
Can you add screenshot? Would be good to remove some spacing from the submenu itens.

dawidgarus:
@flavioislima I've already changed spacing in the submenu items. Here's a screenshot:

Screenshot from 2022-03-14 17-45-31

I was also considering aligning submenu item labels with top level items but it would cause wrapping of the text for longer entries like unreal marketplace

bilianemoreira:
Is possible to give more margin-left or padding-left to Epic Store and GOG Store? They are submenus of Stores and is a UI convention to have submenus like this. An example:  
image

src/components/UI/Sidebar/components/SidebarUtils/index.tsx
```diff
@@ -1,30 +1,41 @@
...
   cwd: 'gog_store'
 })
-import { handleQuit } from 'src/helpers'
-import './index.css'
-import { openDiscordLink } from 'src/helpers'
-import { useTranslation } from 'react-i18next'
-import ContextProvider from 'src/state/ContextProvider'
-import { NavLink, useHistory } from 'react-router-dom'
+
+const QuitButton: React.FC = () => {
```
flavioislima, line 28: Maybe we could add this component to its own file to avoid having more than one component declared in the same file 👍🏽

dawidgarus: Moved

PR #1106, opened 2022-03-13
[General] update .gitignore, UX improvements in webview

dawidgarus:
Various UX improvements in WebView component:
- Disable back/forward button when operation is not possible
- Added tooltips to buttons
- Added a button to open current page in user's web browser
- Changed icon of reload button
- Added colors on hover and indication when user clicked on a button
- User can now select URL
- Changed color of URL to red when schema is not https
- Changed size of the toolbar, gaps and added accent on the bottom for more pleasing look
- Added .idea folder to .gitignore

Preview:
Screenshot from 2022-03-13 02-03-23

---


public/locales/en/translation.json
```diff
@@ -323,6 +323,14 @@
         "manageaccounts": "Manage Accounts",
         "quit": "Quit"
     },
+    "webview": {
```
flavioislima, line 326: it is important that you actually run `yarn i18n` so it will create this key with the default value for all language files. :)

flavioislima (dismissed):
Looks good. I believe you just need to run `yarn i18n` to fix the language files and we can merge it.

public/locales/en/translation.json
dawidgarus: Merged main and generated translations

flavioislima (approved):
Pretty good! ⚔️

PR #1112, opened 2022-03-13
[General] UX improvements in loading screens

dawidgarus:
Various UX improvements in loading screens:
- Default "Loading" message can now be translated
- Loading message is now displayed in the foreground when opening install popups
- Loading indicator is now displayed inside the container with logs to prevent "jumping" when switching between current and last log
- Loading indicator in the library is now vertically centered


src/components/UI/ToggleSwitch/index.tsx
```diff
@@ -28,7 +34,7 @@ export default function ToggleSwitch(props: Props) {
         aria-label={title}
       />
 
-      <span {...props} className="checkmark" />
+      <span {...checkmarkProps} className="checkmark" />
```
flavioislima, line 37: What is this change doing exactly?
I believe that if we pass a prop that is not listed, it will be ignored.
so we need to check the props we use when calling this component :)

src/screens/Game/GamePage/index.tsx
```diff
@@ -422,12 +422,12 @@ export default function GamePage(): JSX.Element | null {
             </div>
           </>
         ) : (
-          <UpdateComponent />
+          <UpdateComponent message={t('loading.default')} />
```
flavioislima, line 425: Perhaps we could use the default message directly into the component in case `message` is undefined. And make the message a optional prop (I believe might be already)

src/components/UI/ToggleSwitch/index.tsx
dawidgarus:
```javascript
const {
    handleChange,
    value,
    disabled,
    title,
    dataTestId = 'toggleSwitch',
    ...rest
  } = props
```
are by any chance thinking of something like `rest` property? Because destructuring an object doesn't remove destrustured properties.
anyway, I removed properties which caused errors: `handleChange` and `disabled`, but I kept the rest just in case the attributes are used somewhere and I didn't want to break anything.

src/components/UI/UpdateComponent/index.tsx
```diff
@@ -1,18 +1,73 @@
...
 }
 
-export default function UpdateComponent({ message = 'Loading' }: Props) {
+export default function UpdateComponent({
+  message,
+  inline = false
+}: UpdateComponentProps) {
+  const componentRef = useRef<HTMLDivElement>(null)
+  const wrapperRef = useRef<HTMLDivElement>(null)
+  const observer = useMemo(() => {
```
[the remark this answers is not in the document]
dawidgarus: `.UpdateComponent` element has always height of 100%. Then we change how much of the `.UpdateComponent` is visible on the screen (ratio) and then change `.UpdateComponent__wrapper`'s height to the ratio.

src/screens/Game/GamePage/index.tsx
dawidgarus: My bad, it's actually is possible. Commited the change

src/components/UI/UpdateComponent/index.tsx
```diff
@@ -1,18 +1,84 @@
...
+      </div>
     </div>
   )
 }
+export default function UpdateComponent({
+  message,
+  ...rest
+}: UpdateComponentProps) {
+  const { t } = useTranslation()
+  if (message === undefined) {
```
flavioislima, line 80: I believe a better check would be:
`if (!message) { `

because then it would use the default also if we send an empty string, null or any other falsy value.

dawidgarus: That's deliberate:
a) It's inline with current behavior because default value only work if the property is missing (`undefined`). You can pass empty string to the component and it won't display default message.
b) I can't see any reason to restrict ability to pass empty string and display icon without any message
c) You can't pass any falsy value, because type checking won't allow it. You can either pass no value, which will cause displaying default value or empty string which will display no message

flavioislima: Ah, I see. Makes sense to load only the icon.
Good then.

flavioislima (approved):
Thanks for that as well

PR #1127, opened 2022-03-17
[General] Improvements in language selector when installing game

dawidgarus:
On game installation dialog
- Available languages are now displayed as human readable names instead of codes
- Attempt to select by default language that matches the language of the application interface

Preview:
Screenshot from 2022-03-17 01-18-39

---

### DIFF
--- a/src/screens/Library/components/InstallModal/index.tsx
+++ b/src/screens/Library/components/InstallModal/index.tsx
@@ -6,7 +6,7 @@ import {
   install,
   writeConfig
 } from 'src/helpers'
-import React, { useContext, useEffect, useState } from 'react'
+import React, { useContext, useEffect, useMemo, useState } from 'react'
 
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
Looks good. Will test later! 
@imLinguin take a look as well if you have time 👍🏽

imLinguin:
Have you tested how it behaves for older games that use old depot system (V1)?

dawidgarus:
> Have you tested how it behaves for older games that use old depot system (V1)? Those are mostly human readable by default.

I didn't test it with anything other than GOG games, because I don't own any Epic game that prompts for selecting language.
But I wrote the code to fail gracefully. If translation can't be find by `Intl` for any reason, it will fallback to original value.
As for default selected value, it will also fallback to selecting first on list, with one caveat - it may in some edge cases default "random" language.

imLinguin:
Then it's going to work for V1 games too.

Good job ⚔️

PR #1129, opened 2022-03-17
[General] Improvements in install dialog

dawidgarus:
Improvements in install dialog:
- Refactor handling of SDLs
- Using browser's `<dialog>` element for handling dialogs which comes with some advantages, like focus trap and esc to close
- Long paths are no longer hidden under select button
- Updated look
- Restoring changes in scrollbar that were broken after a MR

Preview:
Screenshot from 2022-03-17 21-40-10
Screenshot from 2022-03-17 21-40-38
Screenshot from 2022-03-17 21-41-00
Screenshot from 2022-03-17 21-41-18

---

### DIFF
--- a/src/components/UI/Dialog/index.css
+++ b/src/components/UI/Dialog/index.css
@@ -1,0 +1,91 @@
+.Dialog {
+  padding: 0;
+  text-align: left;
+
+  --dialog-margin-horizontal: 32px;
+  --dialog-margin-vertical: 24px;
+  --dialog-gap: 24px;
+}
+
+.Dialog__element {
+  top: 0;
+  z-index: 8;
+  display: flex;
+  flex-direction: column;
+  padding: 0;
+  height: fit-content;
+  min-width: fit-content;
+  overflow: auto;
+  border: solid 1px var(--modal-border);
+  border-radius: 10px;
+  background: var(--modal-background);
+  color: var(--text-default);
+  opacity: 0;
+  transform: translateY(50px);
+  transition: opacity 500ms, transform 500ms;
+}
+
+.Dialog__element[open] {
+  opacity: 1;
+  transform: translateY(0);
+}
+
+.Dialog__element::backdrop {
+  /* var() doesn't work here */
+  background: rgba(13, 15, 28, 0.8);
+}
+
+.Dialog__header {
+  display: flex;
+}
+
+.Dialog__headerTitle {
+  flex: 100% 1 1;
+  padding: var(--dialog-margin-vertical) 16px 0 var(--dialog-margin-horizontal);
+  font-size: 24px;
+}
+
+.Dialog__headerClose {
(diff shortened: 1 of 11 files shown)


src/components/UI/Dialog/index.css
```diff
@@ -0,0 +1,91 @@
+.Dialog {
+  padding: 0;
+  text-align: left;
+
+  --dialog-margin-horizontal: 32px;
+  --dialog-margin-vertical: 24px;
+  --dialog-gap: 24px;
+}
+
+.Dialog__element {
```
flavioislima, line 10: Anything against `camelCase`? ahahah
we use camel case on the other variables and CSS classnames would be good to be consistent :)

flavioislima, line 10: Anything against `camelCase`? ahahah
we use camel case on the other variables and CSS classnames would be good to be consistent :)

src/screens/Library/components/InstallModal/index.tsx
```diff
@@ -24,10 +38,14 @@ import {
   Path,
   Runner
 } from 'src/types'
+import Dialog from '../../../../components/UI/Dialog'
```
flavioislima, line 41: We have relative imports configured on Heroic so you can do it like:
`import Dialog from 'src/components/UI/Dialog'` and so on.

flavioislima, line 41: We have relative imports configured on Heroic so you can do it like:
`import Dialog from 'src/components/UI/Dialog'` and so on.

src/screens/Library/components/InstallModal/index.tsx
```diff
@@ -24,10 +38,14 @@ import {
...
+import DialogContent from '../../../../components/UI/Dialog/DialogContent'
```
flavioislima, line 42: I believe you could have an index.ts file that imports all `Dialog` children's (like we are doing with sidebar, for instance) and then just import all of them on the same line:
`import {DialogContent, DialogFooter, DialogHeader} from 'src/components/UI/Dialog/components'`

flavioislima, line 42: I believe you could have an index.ts file that imports all `Dialog` children's (like we are doing with sidebar, for instance) and then just import all of them on the same line:
`import {DialogContent, DialogFooter, DialogHeader} from 'src/components/UI/Dialog/components'`

flavioislima:
Tested a bit here and looks good.
One improvement that we are lacking on this is actually the ability to choose the wine/proton version when installing.
that would be a good thing to add to the PR if you have the time. Otherwise we can do next :)

src/screens/Library/components/InstallModal/index.tsx
dawidgarus: Is it okay, now?

dawidgarus:
Added selection of wine version:
Screenshot from 2022-03-18 23-03-41

PR #1130, opened 2022-03-18
[Tech] Restore missing css variables

dawidgarus:
Recent PR removed some css variables which were used:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1064
I changed code to not to used them in some places and restored variables that were still missing

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -27,7 +27,7 @@ html,
   height: 100%;
   overflow-y: auto;
   overflow-x: hidden;
-  background: var(--background-shadow);
+  background: var(--gradient-body-background);
 }
 
 .App .content .UpdateComponent--topLevel {
@@ -88,7 +88,7 @@ html,
 }
 
 .button.is-tertiary {
-  background: var(--background-color);
+  background: var(--body-background);
   color: white;
 }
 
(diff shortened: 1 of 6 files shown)


8Bitz0:
I assumed because the variables had a comment saying that they should be avoided and did not follow the same naming, they should be removed altogether.

If something needed to use the variables, shouldn't they be switched to a new one? (Or get a new variable?)

Or am I missing something else here?

dawidgarus:
> I assumed because the variables had a comment saying that they should be avoided and did not follow the same naming, they should be removed altogether.

They should if they are not used. But some of them were. It's probably because changes were made in main branch before the branch was merged.

They only way to ensure things like this won't happen is to manually pull changes from main branch and test before merging PR into main branch. Which ofc may be problematic if contributor and maintainer don't work on the same schedule.

> If something needed to use the variables, shouldn't they be switched to a new one? (Or get a new variable?)

Some of them were changed to new ones. Some of them I kept for now because there weren't any variable with a name that would match yhe intended purpose. I didn't want to dive to deep into it here, only fix what was broken.

flavioislima (approved):
Good catch. I thought they were replaced in the code but I believe not all of them were.

PR #1132, opened 2022-03-18
[Tech] Disable eslint during development, added options to editorconfig

dawidgarus:
Added addition options to .editorconfig
Disable eslint validation during development

Rationale:
The reason to disable it is that it's infuriating to develop anything with eslint enabled.
You comment single line for then you get bombarded with unused variables and import which must also be commented which will cause new errors.
Moving code around is impossible without getting perfect indents.
I don't see any value with checking that during development, especially since prettier will take care of it anyway.
They'll still be checked during build and pipeline execution.

---

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,17 +1,19 @@
 root = true
 
 [*.css]
-indent_style = space
+charset = utf-8
+end_of_line = lf
 indent_size = 2
+indent_style = space
+insert_final_newline = true
+max_line_length = 80
+trim_trailing_whitespace = true
 
-[*.js]
-indent_style = space
+[{*.ts, *.tsx, *.js}]
+charset = utf-8
+end_of_line = lf
 indent_size = 2
-
-[*.ts]
 indent_style = space
-indent_size = 2
-
-[*.tsx]
-indent_style = space
-indent_size = 2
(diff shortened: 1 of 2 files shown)


flavioislima:
I have to say that I don't agree with this one.
Although might be boring, when developing with so many different people that come and go all the time, it's important to enforce at least some basic rules.
I don't know if your editor has this feature but at least on VSCode I set to format (using the current eslint rules) on save, so if I have some ident issue or something like this, when I save the file it gets fixed automatically.

Before adding those rules the code started to become a mess in terms of formatting. At the time we added a lot of rules, some of them were bad for productivity then I removed several, and today we have only the basic set of rules, the recommended ones.
so I believe we need to pass this PR.

dawidgarus:
@flavioislima 
It's not about formatting, that's not important and eslint don't force any formatting rules that are taken care by prettier.
The most important thing are rules regarding unused things, as I said:
> You comment single line for then you get bombarded with unused variables and import which must also be commented which will cause new errors.
It's only for development. Build command will enforce eslint rules, gitlab pipeline will enforce eslint rules and git hooks will also enforce eslint rules. In other words, it's still impossible to merge into this repo code that doesn't pass eslint rule.

PR #1133, opened 2022-03-18
[Fix] Override exe file for GOG games, launch exe on prefix now sets working directory to the location of exe file

dawidgarus:
Tested on HoMM 3 with HD laucher which requires to be located in the same directory as game and searches for executables in the working directory

--

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -96,7 +96,8 @@ async function launch(
       return
     }
   }
-  const exe = targetExe && isLegendary ? `--override-exe ${targetExe}` : ''
+  const exe =
+    targetExe && (isLegendary || isGOG) ? `--override-exe ${targetExe}` : ''
   const isMacNative = gameInfo.is_mac_native
   // const isLinuxNative = gameInfo.is_linux_native
   const mangohud = showMangohud ? 'mangohud --dlsym' : ''
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1,3 +1,5 @@
+import { ExecOptions } from 'child_process'
+import { dirname } from 'path'
 import {
   InstallParams,
   LaunchResult,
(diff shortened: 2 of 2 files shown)


electron/launcher.ts
```diff
@@ -96,7 +96,8 @@ async function launch(
       return
     }
   }
-  const exe = targetExe && isLegendary ? `--override-exe ${targetExe}` : ''
+  const exe =
```
flavioislima, line 99: is this possible already with `gogdl`, @imLinguin ?
There is any risk of breaking GOGDL or we can keep it?

flavioislima, line 99: is this possible already with `gogdl`, @imLinguin ?
There is any risk of breaking GOGDL or we can keep it?

dawidgarus: It's possible:
https://github.com/Heroic-Games-Launcher/heroic-gogdl/commit/bf0ab9d5b39734e9f364a5b947f5f03b29f45308
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/commit/d29ac86a64fc1335d1c49a5d2422aa7ce55dc510

It's unlikely that it will break anything.
I tested it on two games with and without custom exe.
It wasn't previously supported.
But now that I think of it, it's possible that user have set custom exe before, it didn't work so they ignored it and now after an update they will be surprised when game doesn't launch but run some random exe they have set earlier.

imLinguin: Yes it's completely ready on `gogdl` side, I forgot to change that if statement :shrug: 

As a sidenote: It'll use the provided exe's directory as working dir

dawidgarus: > As a sidenote: It'll use the provided exe's directory as working dir
So there shouldn't be a reason for heroic spawn to also be in that directory

changes for setting working directory is not for alternative exe to run, but run exe on prefix

flavioislima (approved):
Pretty good!
Thanks! ⚔️

PR #1137, opened 2022-03-19
[Fix] Cancel/pause downloading of GOG games

dawidgarus:

---

### DIFF
--- a/electron/gog/games.ts
+++ b/electron/gog/games.ts
@@ -44,6 +44,14 @@ const installedGamesStore = new Store({
   name: 'installed'
 })
 
+function verifyProgress(stderr: string): boolean {
+  let index = stderr.lastIndexOf('\n')
+  index = stderr.lastIndexOf('\n', index - 1)
+  const status = stderr.substring(index)
+  const match = status.match(/Progress: ([0-9.]+) ([0-9]+)\/([0-9]+)/)
+  return match != null && 100 === Number(match[1]) && match[2] === match[3]
+}
+
 class GOGGame extends Game {
   public appName: string
   public window = BrowserWindow.getAllWindows()[0]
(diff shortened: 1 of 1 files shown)


electron/gog/games.ts
```diff
@@ -44,6 +44,14 @@ const installedGamesStore = new Store({
   name: 'installed'
 })
 
+function verifyProgress(stderr: string): boolean {
+  let index = stderr.lastIndexOf('\n')
+  index = stderr.lastIndexOf('\n', index - 1)
+  const status = stderr.substring(index)
+  const match = status.match(/Progress: ([0-9.]+) ([0-9]+)\/([0-9]+)/)
+  return match != null && 100 === Number(match[1]) && match[2] === match[3]
```
flavioislima, line 52: Small thing here but let's use `!==` just to keep the pattern. 👍🏽

flavioislima: Also, can you add a few comments to this function to explain what it is doing? In case we need to do maintenance in the future.

dawidgarus: actually `!= null` is idiomatic javascript way of checking if value is not null or undefined

flavioislima: hmm, `match` will return an array or `null`
So this line could maybe be written like:
`return match && Number(match.join('')) === 100`
If a `match` is `null` it will return `null` which is ok since its a `falsy` value.
What do you think?
I believe its easier to read as well.

flavioislima:
Btw, since you are working a lot on the UI/UX improvements, we have a figma project where we discuss the designs :D
https://www.figma.com/file/Zv89aBtLJ8yN4BwFS0AwNb/Heroic-Games?node-id=0%3A1

electron/gog/games.ts
dawidgarus: Changed to `!==`. I still prefer always checking with `!=` because standard library is inconsistent with returning null or undefined. I'm not changing to just `match` though, because falsy is not boolean and I like to be strict about it

flavioislima: Ok, that works as well.
Using `Boolean(match)` would convert the null to false in this case.
but it's ok to keep it that way imo.

flavioislima (approved):

PR #1149, opened 2022-03-22
Open game store page in webview and refactor

### LINKED ISSUE #863: Store Page button opens the link in a new window instead of the webview
**Describe the bug**

When clicking the Store Page link of a game, it opens in a new window. It's weird since the Store link on the sidebar opens the store in the webview inside the main Heroic window.

~~the Check Compatibility link behaves the same.~~

**To Reproduce**

1. Go to the Tools tag of any game
2. Click on Store Page or Check Compatibility
3. The page will open in a new window

**Expected behavior**

I would expect at least the store page to open in the webview, the same way the Store link of the sidebar works.

EDIT: this issue now applies only to the Store Page button, not the compatibility button

- OS: Linux Mint 20.02
- Heroic Version: Main branch
- Game: any
- Wine/Proton version: -
- Hardware specs (RAM, CPU, GPU): -

arielj:
This PR fixes #863 by making the stores open in the webview instead of a popup.

I did a small refactor:
- moved the epic url generation at the loading of the file so `gameInfo.store_url` works for both gog and epic
- we can open the generic page with no language code and the epic store will use the user's language
- I cleaned a little bit of the webview url handling for the `login/:runner` routes

---

### DIFF
--- a/electron/legendary/library.ts
+++ b/electron/legendary/library.ts
@@ -18,7 +18,12 @@ import {
 } from '../types'
 import { LegendaryGame } from './games'
 import { LegendaryUser } from './user'
-import { getLegendaryBin, isEpicServiceOffline, isOnline } from '../utils'
+import {
+  formatEpicStoreUrl,
+  getLegendaryBin,
+  isEpicServiceOffline,
+  isOnline
+} from '../utils'
 import {
   fallBackImage,
   installed,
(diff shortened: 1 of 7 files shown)


flavioislima:
Looks good. 
One thing I noticed though is that it doesn't highlight the right store when opening it.
So for epic games, the epic games submenu should be marked as active. Should be a simple fix I believe.

arielj:
@flavioislima this is ready for a second review, I added the highlight on the sidebar links when it applies

flavioislima (approved):
Pretty good! Thanks for that! 👍🏽

PR #1151, opened 2022-03-22
[Fix] Dialogs hiding behind app window

dawidgarus:
Change dialogs (message boxes and file selection) to be modal (attached to window) so they will block application window until user closes dialog.

Sadly it doesn't work for file selection dialogs because of bug in electron: https://github.com/electron/electron/issues/32857 but I expect it will work after they fix it.

---

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -1,6 +1,5 @@
 // This handles launching games, prefix creation etc..
 
-import { dialog } from 'electron'
 import makeClient from 'discord-rich-presence-typescript'
 import i18next from 'i18next'
 import { existsSync, mkdirSync } from 'graceful-fs'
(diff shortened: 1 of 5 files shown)


flavioislima:
Hi! thanks for another contribution!

My primary idea about those native dialogs would be to remove them all and use native `<dialog>` instead. So we could get the error from the backend and then show the message, or the question, on the frontend, using the current design, since native dialogs don't follow the app design.
so we could get a generic React component that could accept properties to show buttons for OK, YES, NO, etc. just like the natives are doing now.
Ofc the only exception would be the file dialog, but that's not an issue.

Although, since this solves an issue, we can merge this now and I will open a Github Issue for the future. Since I believe this would be a better solution to keep the design consistency. :)

flavioislima (approved):

flavioislima:
Feature request:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1152

Issue #1136, opened 2022-03-19
Recent commit #1092 broke download canceling for GOG


dawidgarus:
### Describe the bug

When you try to pause/cancel downloading game from GOG, it will report as installed

### Add logs

(02:43:57) DEBUG: [Legendary]: Running Legendary command: /home/user/Projects/github/my/HeroicGamesLauncher/public/bin/linux/legendary --version

(02:43:57) INFO: [Legendary]: Legendary location: /home/user/Projects/github/my/HeroicGamesLauncher/public/bin/linux/legendary

(02:43:57) INFO: [Gog]: GOGDL location: /home/user/Projects/github/my/HeroicGamesLauncher/public/bin/linux/gogdl

(02:43:57) INFO: [Backend]:

Heroic Version: 2.2.6 Oden

Legendary Version: 0.20.25 Our Benefactors

OS: Arch KERNEL: 5.16.15-arch1-1 ARCH: x64

CPU: AMD Ryzen 5 3600 6-Core Processor @3.6

RAM: Total: 33.7 GB Available: 23 GB

GRAPHICS: GPU0: NVIDIA GeForce RTX 2060 SUPER VRAM: 8192MB DRIVER: 510.54

PROTOCOL: x11

(02:43:57) INFO: [Gog]: Getting data about the user

(02:43:57) ERROR: [Backend]: Failed to register protocol with OS.

(02:43:58) INFO: [Gog]: Saved user data to config

(02:43:58) INFO: [Frontend]: Refreshing Library

(02:43:58) INFO: [Gog]: Getting GOG library

(02:43:58) INFO: [Gog]: Number of library pages: 1

(02:43:58) INFO: [Gog]: Saved games data

(02:43:58) INFO: [Legendary]: Refreshing library...

(02:43:58) INFO: [Legendary]: Refreshing Epic Games...

(02:43:58) INFO: [Legendary]: Refreshing Epic Games...

(02:43:59) DEBUG: [Legendary]: Running Legendary command: /home/user/Projects/github/my/HeroicGamesLauncher/public/bin/linux/legendary list --include-ue

(02:43:59) INFO: [Legendary]: Updating game list

(02:43:59) INFO: [Legendary]: Game List Updated

(02:44:00) INFO: [Legendary]: Checking for game updates: /home/user/Projects/github/my/HeroicGamesLauncher/public/bin/linux/legendary list-installed --check-updates --tsv

[...]

PR #1157, opened 2022-03-24
[Tech] Refactoring of calling commands and handling installation progress

dawidgarus:
- Create abstraction that can be used with both gogdl and lengendary
- Change how installation progress is handled to push progress from BE instead of pulling
- Fix issue when if you continue download for gog game, it will report bad progress %
- Fix reporting ETA for gog (added independent calculation)
- Replace call of `mv` command for moving gog games (`mv` is not available in Windows, didn't test it on Windows though)
- Fix passing workers argument
- Revert changes from https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1136 - it doesn't always work. Gogdl doesn't always report the progress at 100% before the installation is finished. I think it's needs to be implemented inside gogdl to report confirmation after successful installation to stdout.

---


electron/gog/games.ts
```diff
@@ -45,11 +45,7 @@ const installedGamesStore = new Store({
 })
 
 function verifyProgress(stderr: string): boolean {
```
flavioislima, line 47: Since this function will always return `true`, it means that the places we have a check for it, like line 169 and so on, should be refactored as well?

flavioislima (approved):
I believe this is a pretty good refactoring. 
Tooks some time to test on a Mac and on Linux and looks like everything on the `progress` side is working. Tested all other basics like installing, moving, uninstalling, etc. and everything is working as well.
The only thing I believe would be fine is that some parts of the code could be more documented in case we need to do some maintenance in the future. At least using JSDOC or some comments here and there. If you could do that I appreciate it.

src/screens/Library/components/GameCard/index.tsx
```diff
@@ -19,8 +19,8 @@ import ContextProvider from 'src/state/ContextProvider'
 import fallbackImage from 'src/assets/fallback-image.jpg'
 import { uninstall, updateGame } from 'src/helpers/library'
 import { SvgButton } from 'src/components/UI'
+import { useInstallProgress } from '../../../../hooks'
```
flavioislima, line 22: Remember to use relative imports.

src/screens/Game/GamePage/index.tsx
```diff
@@ -455,22 +426,41 @@ export default function GamePage(): JSX.Element | null {
...
+      percent && bytes
+        ? `${percent}% [${bytes}]${displayEta && ` | ETA: ${displayEta}`}`
+        : '...'
     }`
 
     if (isUpdating && is_installed) {
-      if (eta && eta.includes('verifying')) {
-        return `${t('status.reparing')}: ${percent} [${bytes}]`
-      }
+      // TODO what's that?
```
flavioislima, line 460: this should show the repair progress but :)

src/hooks/useInstallProgress.ts
```diff
@@ -0,0 +1,61 @@
+import { IpcRenderer, IpcRendererEvent } from 'electron'
+import { useEffect, useRef, useState } from 'react'
+import { InstallProgress, Runner } from '../types'
+
+const { ipcRenderer } = window.require('electron') as {
+  ipcRenderer: IpcRenderer
+}
+
+export function useInstallProgress(
```
flavioislima, line 9: Great job with this hook. will be useful once we have a queue or a downloads screen

dawidgarus:
Added check for signal code to determine if process was killed. along with https://github.com/Heroic-Games-Launcher/heroic-gogdl/pull/4 this should fix the issue with canceling download

flavioislima (approved):
Looks pretty good! ⚔️ 
Btw, we are thinking about refactoring the BE to use a Functional instead of OOP. 
I will open a discussion here on GH since you don't use Discord, so you can give your opinion on that :)

dawidgarus:
@flavioislima 
What's exactly the issue so I can fix it?

flavioislima:
@dawidgarus first of all, sorry because I forgot to answer you, my fault 🤦🏽 
so, the issue on Mac was that the verify progress wasn't working, it was thinking that the download was falling since both `legendary` and `gogdl` was not reporting 100% all the time, most of the time there was a delay in sending the final percent, so it broke. 

the other reason was that, even though the progress was working fine for both installing, updating and repairing, the solution with the collectors and so on, without comments, made some of us having a hard time understanding how to scalate and add new features to it. Or even to find where the bug was.

but sorry again for the delay in answering that.
All the other contributions are still on the code and were released on the beta and will enter the stable soon ofc. 

One thing we miss is a way to communicate faster than github, because right now we are a team of 4 or 5 regular contributors and we discuss a lot on how we are going to develop stuff and also to ask for help when we need to. But we are using Discord. I don't know if there is any other tool, maybe matrix, that we can connect with it and then you can join. 
But let me know if you are interested and we can set this up!

:)

PR #1167, opened 2022-03-27
[Tech] User credentials are no longer stored in plain text

dawidgarus:
As title says.
Unfortunately, legendary still saves tokens as a plain text internally, but we don't need to create copy of it.

WARNING:
Switching to this branch will cause migration and your credentials for GOG will be encrypted.
To switch back to other branches you will need to restore old configuration files.
---


JakobDev:
Just a Question: Does this work on the Flatpak version of Heroic? The doc says that it uses a password management tool, which is not available in a Flatpak, if you don't use [Portals](https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Secret). I can't find any information, if the safe-storage API supports that. It would not good, if this change destroys the Flatpak version of Heroic.

flavioislima:
@JakobDev thanks for bringing this up. That's quite a good concern and we will need to find an answer for that before merging this.

flavioislima:
@dawidgarus can we disable this for Heroic Flatpak for now? until we find a solution or answers for that?
We have a global constant to check if it is flatpak or not so you can use that.

dawidgarus:
@flavioislima I'll investigate it. I didn't check it with flatpak build, but I assume it will fail gracefully thanks to `isEncryptionAvailable()`.
I am however concered with: https://github.com/electron/electron/issues/32206
I don't like depending on existance of `BrowserWindow` to read credentials. This should be done 100% on the backend.
I might look into alternative like: https://github.com/atom/node-keytar

dawidgarus:
So yeah, flatpak doesn't work. And it turns out that the encryption doesn't really do much except obscuring the data. All programs have access to system's keyring, so it doesn't prevent malicious programs from stealing your account, only makes it a little bit harder.
So I reduced the scope of this MR to just removing copy legendary credentials that heroic keeps and is not necessary.

flavioislima (approved):
Looks good! 👍🏽

PR #1172, opened 2022-03-29
Remove MimeType from Shortcut

JakobDev:
Heroic add the MimeType=x-scheme-handler/heroic to all shortcuts. The MimeType line means, that the App, that belongs to this .desktop file can open this MimeType. In this case, it says basically, that Game xy can open heric:// links,what is not true. Only Heroic itself can do that.


JakobDev:
Ping @flavioislima

flavioislima:
Ah, good catch, I believe this was leading to some bugs with shortcuts. Thanks for that.

flavioislima (approved):

PR #1174, opened 2022-03-29
[Gog] Setup improvements

imLinguin:
This improves compatibility with most older games.
- Prefix creation now uses `wineboot` and `wineserver` provided with WineVersion.
- Setup relies on root directory mounted in `Z:` of prefix to point to game install directory.
- Installation also checks for `INFO: Done` keword in installation logs do determine if installation was successful. This will be introduced properly with next version of gogdl


electron/gog/setup.ts
```diff
@@ -63,19 +71,54 @@ async function setup(appName: string): Promise<void> {
...
+          const valueType = actionArguments?.valueType
           // If deleteSubkeys is true remove path first
-          if (actionArguments.deleteSubkeys) {
-            const command = `${commandPrefix} reg delete "${registryPath}" /f`
-            logInfo(
-              ['Setup: Deleting a registry key', registryPath],
-              LogPrefix.Gog
-            )
-            await execAsync(command)
+          // if (actionArguments?.deleteSubkeys) {
```
flavioislima, line 84: any reason to keep this commented code here?

electron/gog/setup.ts
```diff
@@ -101,15 +144,21 @@ async function setup(appName: string): Promise<void> {
...
-          const executablePath = path.join(
+
+          const supportPath = `"${path.join(
             gameInfo.install.install_path,
             'support',
-            appName,
-            executableName
+            appName
+          )}"`
+          const workingDir = actionArguments?.workingDir
```
flavioislima, line 153: maybe checking if `workingDir` exists before continuing would be safer since I wonder if this `replace` will break the code if we get `undefined` here.

electron/gog/setup.ts
```diff
@@ -137,17 +182,33 @@ async function setup(appName: string): Promise<void> {
             gameInfo.install.install_path
           )
           const type = actionArguments.type
+          const sourcePath = actionArguments?.source
```
flavioislima, line 185: same here with the `source` since it depends on it to get going.

flavioislima (dismissed):
I think looks good, left just a few comments about small concerns 👍🏽

electron/gog/setup.ts
imLinguin: I tested both cases you mentioned. sourcePath will just be undefined. And since replace is not a memeber of undefined it won't get executed.

Nocccer:
Will see over  the pr today after i slepped

electron/gog/setup.ts
Nocccer: will not crash with message replace is not a function of undefined ?

electron/gog/setup.ts
```diff
@@ -217,6 +270,7 @@ async function setup(appName: string): Promise<void> {
             }
         ],
     */
+    //TODO
```
Nocccer, line 273: What needs to be done here ?

Nocccer (changes requested):

electron/gog/setup.ts
imLinguin: Execute instruction based on example above.

flavioislima (approved):

PR #1176, opened 2022-03-30
Update & Standardize CHANGELOG.md

Flonky:
This PR updated the CHANGELOG.md up to Version 2.2.6.

It also Introduces an standardization in the Changelog as following:

- Changes should be marked with either [General], [Linux], [Windows], [MacOS], [Bugfix] or [i18n] and sorted in the same order
- If a bugfix is Platform Specific the Platform should be added behind a `|`
- Are there multiple platforms for one change they should be seperated by a `/` 

---

I've tried to apply these standardizations retrospectively, but errors might have occured during this process, so please correct me if you spot anything.

---

Additionally, the prettier check doesn't seem to have passed, help with fixing that would be greatly appreciated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,261 +1,367 @@
 # Changelog
 
-## 2.0.0
 
-### General
+## 2.2.6 Oden - Hotfix 5
 
-- New Layout with a Sidebar instead of a Navbar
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
Looks good! Thanks for that!

PR #1196, opened 2022-04-09
Keep installing if the pause/stop dialog is closed

Aikufurr:
Currently, the confirmation dialog for stopping a download will default to stopping the download and removing all files if the dialog is closed. Adding `cancelId: 0` will instead default to resuming the download.

---


arielj (approved):
I tested this and it works as expected (closing the dialog with the ESC key works like clicking `Keep Installing`)

Also all buttons work as expected with this change too.

flavioislima:
@Aikufurr can you provide more info on how to replicate this?

Aikufurr:
@flavioislima So currently, if you start a download and then press the `PAUSE/CANCEL` button it brings up the dialog, if at that point you press `ESC` or close it via pressing the `X` the download is cancelled and doesn't save the files. With this commit, repeating those actions now continues the download instead of pausing/cancelling.

flavioislima:
@Aikufurr thanks for fixing that 👍🏽

PR #1221, opened 2022-04-18
Order games by status

### LINKED ISSUE #878: Make installing games appear at the top
When installing a game youd like to check on its progress on a fast and easy way, right now you have to use the search bar or scroll down to find the game youre installing, both only work for 1 game at the time which consume more time.
my suggestion is to make games that are installing appear at the top of the list until they are installed.

Make the game appear at the top of the list when they are installing

make a dedicated download page

### LINKED ISSUE #977: [Feature request] Show installed games first
Perhaps installed games could always appear on top. That way it would not be necessary to constantly switch between "Ready" and "Not ready" (Actually, it could be removed entirely). My idea is to have all games installed permanently at the top of the library, with those not intalled below it, and then accessing the uninstalled games could be done just by scrolling down.

I would basically make the "Played recently" section, the "Installed games" section; A permanent section always availabe. The "Played recently" section could be easily toggled with an "Show recently played games" option within the settings for those who want it, since many users of the launcher have asked for an option to hide them.

Example of what I mean: https://imgur.com/a/SH7Sd6g

arielj:
This PR changes the order of the games in the library to display installed games first, then games that are being installed, and then games that are not installed nor installing.

This also removes the selector of All / Ready / Not Ready / Downloading to simplify the UI.

Closes #977 
Closes #878

---

### DIFF
--- a/src/components/UI/Header/index.tsx
+++ b/src/components/UI/Header/index.tsx
@@ -1,7 +1,7 @@
 import { faApple, faLinux, faWindows } from '@fortawesome/free-brands-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import cx from 'classnames'
-import React, { useContext, useMemo } from 'react'
+import React, { useContext } from 'react'
 import { useTranslation } from 'react-i18next'
 import { SearchBar } from 'src/components/UI'
 import ContextProvider from 'src/state/ContextProvider'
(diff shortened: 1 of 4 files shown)


flavioislima (changes requested):
Looks good!
one thing though: we need to set the filter to `all` now otherwise if someone upgrade Heroic and has a filter on `uninstalled` or another one, it will hide the games. this happened to me while testing.
Ideally, just remove the filter part on `GlobalState` so it won't update the library based on that.

arielj:
> Looks good! one thing though: we need to set the filter to `all` now otherwise if someone upgrade Heroic and has a filter on `uninstalled` or another one, it will hide the games. this happened to me while testing. Ideally, just remove the filter part on `GlobalState` so it won't update the library based on that.

thanks for the test and finding that, I'll update the PR later!

arielj:
@flavioislima PR is updated!

flavioislima (approved):
Great! :D

PR #1228, opened 2022-04-23
Add option to hide games from the library

### LINKED ISSUE #948: Option to hide games
Game list becomes too cluttered with games the user just claimed out of habit

Ability to hide games from the game list

arielj:
This PR adds a the ability to mark games from the library as `hidden` so they don't show up by default. This can be toggled on/off with the new checkbox.

Games can be hid/unhid using the right click context menu.

You can see it working in this video with the game A Short Hike (second to last, bottom right)

https://user-images.githubusercontent.com/188464/164943074-053065f1-bd62-4cd4-bc65-a77fa2344db6.mp4

(The total games counter changes by more than 1 because I have other hidden games outside the viewport)

This I tried and was not happy with (but accept any feedback):
- using an icon for the `Show hidden` toggle instead of text, it didn't look clear at least to me
- adding an icon in the game card to toggle the hide/unhide, it added noise for an action that's probably done only once or twice ever for a game

Haven't tested with Unreal Market (I don't use it, I don't know if this is expected to work there too? I understand it works but I don't know if we should ignore this feature there).

Closes #948

---

### DIFF
--- a/public/locales/bg/translation.json
+++ b/public/locales/bg/translation.json
@@ -114,6 +114,7 @@
     "GOG": "GOG",
     "gog-store": "Магазин на GOG",
     "header": {
+        "ignore_hidden": "Ignore Hidden",
         "platform": "Платформа",
         "show_hidden": "Show Hidden"
     },
--- a/public/locales/bg/translation.json
+++ b/public/locales/bg/translation.json
@@ -105,20 +105,17 @@
         "syncing": "Синхронизиране",
         "unsync": "Десинхронизиране"
     },
-    "Downloading": "Сваляне",
     "epic": {
         "offline-notification-body": "Услугите, зависещи от Интернет, може да не работят изцяло, тъй като сървърите на Epic Games не са достъпни!",
         "offline-notification-title": "offline"
     },
     "Epic Games": "Epic Games",
-    "filter": {
-        "noFilter": "No Filter"
-    },
     "globalSettings": "Глобални настройки",
     "GOG": "GOG",
     "gog-store": "Магазин на GOG",
     "header": {
-        "platform": "Платформа"
+        "platform": "Платформа",
+        "show_hidden": "Show Hidden"
     },
     "help": {
         "general": "Синхронизирайте с EGS, в случай че вече имате работеща инсталация на Epic Games Launcher и искате да внесете игрите си от там, за да избегнете повторното им сваляне.",
@@ -170,7 +167,6 @@
         "unsync": "Десинхронизацията е завършена"
     },
     "nogames": "Няма намерени игри",
-    "Not Ready": "Не е готово",
     "notify": {
         "error": {
             "move": "Грешка при преместването на играта",
@@ -221,7 +217,6 @@
     },
     "Plugins": "Добавки",
     "Projects": "Проекти",
-    "Ready": "Готово",
     "Recent": "Играно наскоро",
     "search": "Въведете името на играта...",
     "setting": {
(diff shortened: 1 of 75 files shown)


public/locales/bg/translation.json
```diff
@@ -105,20 +105,17 @@
         "syncing": "Синхронизиране",
         "unsync": "Десинхронизиране"
     },
-    "Downloading": "Сваляне",
```
arielj, line 108: There are a lot of things being removed here by the `yarn i18next` command

arielj: Oh I think it's because I removed that filter on a previous PR and didn't know I had to run `yarn i18next` again, those keys are not used anymore

src/helpers/library.ts
```diff
@@ -297,17 +297,17 @@ type RecentGame = {
...
-    const recentGames: Array<RecentGame> =
-      (configStore.get('games.recent') as Array<RecentGame>) || []
-    const recentGamesList = recentGames.map((a) => a.appName) as string[]
-    return recentGamesList.includes(game.app_name)
-  })
+  const recentGames =
+    (configStore.get('games.recent') as Array<RecentGame>) || []
+  const recentGamesList = recentGames.map((a) => a.appName) as string[]
+
+  return library.filter((game) => recentGamesList.includes(game.app_name))
```
arielj, line 310: While coding this feature initially I used a similar method like this one so I included this refactor. The refactor is a small improvement (we don't need to initialize the store and query it for every callback function call) but I can remove it from this PR if it adds noise, since it's not really needed for this feature at the end.

flavioislima:
Personally, I don't like the idea of having more text on the header.
I wonder if we could have another form control as we have for the platform filter and add the `Visibility` Icon there and change it to `VisibilityOFF` when the games are hidden.

something like this:
image

Using:
https://mui.com/material-ui/material-icons/?query=visibility&selected=Visibility
and
https://mui.com/material-ui/material-icons/?query=visibility&selected=VisibilityOff

what do you think?

public/locales/bg/translation.json
flavioislima: Yes

arielj:
> Personally, I don't like the idea of having more text on the header. I wonder if we could have another form control as we have for the platform filter and add the `Visibility` Icon there and change it to `VisibilityOFF` when the games are hidden.
> 
> something like this: image
> 
> Using: https://mui.com/material-ui/material-icons/?query=visibility&selected=Visibility and https://mui.com/material-ui/material-icons/?query=visibility&selected=VisibilityOff
> 
> what do you think?
> 
> with that approach, we can add a Star or Heart to show/hide favorites as well.

I can change that to try, what I'm worried about is that an eye icon doesn't make it obvious that it will show/hide hidden games. I imagine a user clicking the icon to see what happens, if they didn't hide any game nothing will happen.

I guess the tooltip would give a hint if the user leaves the mouse hovering the icon long enough, but tooltips don't show up when using the controller to navigate heroic.

Maybe I'm overthinking it xD

flavioislima:
Yes. Let's collect some feedback on the beta. 😀

arielj:
I moved this right next to the Total Games label, I imagine that would give more context that the icon is related to how many games you can see

image

I also fixed a bug, the selection of the filter was not persisted.

src/components/UI/FormControl/index.css
```diff
@@ -93,6 +93,10 @@
   border-bottom-right-radius: 0;
 }
 
+.FormControl--segmented > button > svg {
+  display: block;
+}
```
arielj, line 98: the svg icons where not properly sentered (svg images are inline by default and it was adding a few empty pixels under it)

flavioislima (approved):
Nice! :D

PR #1236, opened 2022-04-24
Don't return value in useEffect call in Wine Manager

### LINKED ISSUE #1235: Going from Wine Manager to any other screen leaves app empty
### Describe the bug

After opening the Wine Manager screen, any other link leaves the app completely empty (just a blue background).

### Add logs

I can't see anything in the heroic logs (is there a way to see dev tool console logs in the releases?)

I've notice during development that there's an error in the devtools console when opening the Wine Manager, I'll update this later with the exact error.

EDIT:

```
Warning: useEffect must not return anything besides a function, which is used for clean-up.

It looks like you wrote useEffect(async () => ...) or returned a Promise. Instead, write the async function inside your effect and call it immediately:

useEffect(() => {
  async function fetchData() {
    // You can await here
    const response = await MyAPI.getData(someId);
    // ...
  }
  fetchData();
}, [someId]); // Or [] if effect doesn't need props or state

Learn more about data fetching with Hooks: https://reactjs.org/link/hooks-data-fetching
    at WineManager (http://localhost:3000/static/js/src_screens_WineManager_index_tsx.chunk.js:41:68)
    at Route (http://localhost:3000/static/js/bundle.js:74703:29)
    at Switch (http://localhost:3000/static/js/bundle.js:74905:29)
    at main
    at Router (http://localhost:3000/static/js/bundle.js:74334:30)
    at HashRouter (http://localhost:3000/static/js/bundle.js:74000:35)
    at div
    at App (http://localhost:3000/static/js/bundle.js:173:68)
    at GlobalState (http://localhost:3000/static/js/bundle.js:6452:5)
    at I18nextWithTranslation (http://localhost:3000/static/js/bundle.js:72865:31)
    at Suspense
    at I18nextProvider (http://localhost:3000/static/js/bundle.js:72366:19)
```

### Steps to reproduce

1. Go to Wine Manager
2. Wait a moment
3. Click Library, or Settings
4. Screen is empty (no sidebar, no content, just the blue background)

### Expected behavior

Heroic should show the screen that clicked on.

### Screenshots

_No response_

### System Information

- OS [e. g. "Ubuntu"]: Linux Mint
- Heroic Version [e. g. 2.1.1]: 2.3.0-beta.1

### Additional information

_No response_

arielj:
This PR fixes #1235. The `useEffect` call was returning a value and that was triggering that React error shown in the issue description.

After the error, the application was broken when changing screens.

---

### DIFF
--- a/src/screens/WineManager/index.tsx
+++ b/src/screens/WineManager/index.tsx
@@ -17,7 +17,7 @@ export default function WineManager(): JSX.Element | null {
     useContext(ContextProvider)
 
   useEffect(() => {
-    return refreshWineVersionInfo(true)
+    refreshWineVersionInfo(true)
   }, [])
 
   if (refreshing) {


flavioislima (approved):
Looks good. That was a basic mistake since using the return there was acting like a cleanup function for the `useEffect` so that was causing the issue. Good catch!

PR #1239, opened 2022-04-25
fix: undefined install path passed to install func

wiryfuture:
The install button wasn't passing the user selected path to the install function and would just crash and die.

+Install button properly passes install path to the function, and uses the default in empty/undef

- [Y] Tested the feature and it's working on a current and clean install.
- [Y] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [N] Created / Updated Tests (If necessary)
- [N] Created / Updated documentation (If necessary)

### DIFF
--- a/src/screens/Library/components/InstallModal/index.tsx
+++ b/src/screens/Library/components/InstallModal/index.tsx
@@ -159,6 +159,8 @@ export default function InstallModal({
   }
 
   async function handleInstall(path?: string) {
+    // If path provided by user is empty, use default path (as shown in placeholder; $HOME/Games/Heroic)
+    if (!path || path == '') path = defaultPath
     backdropClick()
 
     // Write Default game config with prefix on linux
@@ -542,7 +544,7 @@ export default function InstallModal({
                 {t('button.import')}
               </button>
               <button
-                onClick={() => handleInstall()}
+                onClick={() => handleInstall(installPath)}
                 className={`button is-secondary`}
               >
                 {previousProgress.folder === installPath


src/screens/Library/components/InstallModal/index.tsx
```diff
@@ -159,6 +159,8 @@ export default function InstallModal({
   }
 
   async function handleInstall(path?: string) {
+    // If path provided by user is empty, use default path (as shown in placeholder; $HOME/Games/Heroic)
+    if (!path || path == '') path = defaultPath
```
flavioislima, line 163: why path would come undefined?
Maybe the issue is on another place, because the path should never come empty. 🤔 
i believe to proper solve this we need to check the path when opening the install modal and then check if the path is coming correct, if not then we need to get from the config.

I mean, if the path is not coming here, the issue might be that the path is empty on the settings and maybe we could avoid that.

wiryfuture: When you open install modal the path is filled with the default, but if you delete the content then the placeholder of the default path still shows but the actual value of the path is empty, which gets passed to this function as undefined.
Also, it appears that the function call doesn't currently actually get the path passed to it... which should have made it always broken 🤔 
Nevertheless, I tested this with installing to a couple locations and importing a game as well so it has worked.

arielj: I don't think we should use the default path if the path is empty, if the user deletes the content of the input we should tell them it's an invalid value and ask them to put a valid path. Falling back to the default hides some behavior to the user and can be confusing. Maybe they expect an empty path to mean something different, or maybe it was deleted by mistake and they want another folder.

If the user removed the content of the input (which is what happened to me when I reported this, I deleted it by mistake) we should just warn them that something is not right.

flavioislima: Yes, we could show an error message below the box.

wiryfuture: will do

flavioislima:
I think we can close this one since I came up with a more global approach on #1336.

PR #1245, opened 2022-04-26
Fetch libraries in the background after login

arielj:
This PR makes the app fetch the library immediately after loging in a user in either epic or gog without the need of clicking `Continue` in the Log in screen.

The library shows the loading animation while this is happening.

I replaced the `Continue` button with a `Go to Library` button

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -710,8 +710,10 @@ if (existsSync(installed)) {
 }
 
 ipcMain.handle('refreshLibrary', async (e, fullRefresh) => {
-  await GOGLibrary.get().sync()
-  return await LegendaryLibrary.get().getGames('info', fullRefresh)
+  await Promise.allSettled([
+    GOGLibrary.get().sync(),
+    LegendaryLibrary.get().getGames('info', fullRefresh)
+  ])
 })
 
 ipcMain.on('logError', (e, err) => logError(`${err}`, LogPrefix.Frontend))
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -99,7 +99,7 @@
         "yes": "YES"
     },
     "button": {
-        "continue": "Continue",
+        "go_to_library": "Go to Library",
         "login": "Log in",
         "sync": "Sync",
         "syncing": "Syncing",
@@ -141,6 +141,9 @@
         }
     },
     "info": {
+        "heroic": {
+            "version": "Heroic Version"
+        },
         "settings": "Settings are saved automatically"
     },
     "infobox": {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,12 +2,9 @@ import React, { lazy } from 'react'
 
 import './App.css'
 import { HashRouter, Route, Switch } from 'react-router-dom'
-import ElectronStore from 'electron-store'
 import Sidebar from 'src/components/UI/Sidebar'
 import Login from './screens/Login'
 import WebView from './screens/WebView'
-
-const Store = window.require('electron-store')
 
 const Library = lazy(() => import('./screens/Library'))
 const Settings = lazy(() => import('./screens/Settings'))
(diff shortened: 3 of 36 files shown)


electron/main.ts
```diff
@@ -710,8 +710,10 @@ if (existsSync(installed)) {
 }
 
 ipcMain.handle('refreshLibrary', async (e, fullRefresh) => {
-  await GOGLibrary.get().sync()
-  return await LegendaryLibrary.get().getGames('info', fullRefresh)
+  await Promise.allSettled([
+    GOGLibrary.get().sync(),
+    LegendaryLibrary.get().getGames('info', fullRefresh)
```
arielj, line 715: we can refresh both libraries in parallel

src/App.tsx
```diff
@@ -2,37 +2,23 @@ import React, { lazy } from 'react'
...
   return (
     <div className="App">
       <HashRouter>
         <Sidebar />
         <main className="content">
           <Switch>
-            <Route exact path="/">
-              {user && <Library />}
-            </Route>
+            <Route exact path="/" component={Library} />
```
arielj, line 21: we don't need to check for the user here, the SidebarLinks component is redirecting to the login if no user logged in (though that shouldn't happen in the SidebarLinks either, it makes more sense to do that check in the GlobalState component)

public/locales/en/translation.json
```diff
@@ -141,6 +141,9 @@
         }
     },
     "info": {
+        "heroic": {
```
wiryfuture, line 144: Why is this key being added in this pull?
I don't see any instances of it being used in your code.
Can this go into a different pull?

wiryfuture (changes requested):
lgtm but why is `info.heroic.version` being added to the translation files here?
I don't see any uses of it in your commits elsewhere, so it appears unrelated, can this go in a different pull?

public/locales/en/translation.json
arielj: it was done automatically by the `yarn i18next` command, I think this string was added in another PR but the `yarn i18next` command was not run so this was not included at the right time

I guess I could leave this out and let somebody else make another PR, but since translation files are just a sideeffect of the actual code, removing from this PR and adding another looks like extra work (and this will show up again in any PR that needs to add strings anyway until somebody makes such PR)

flavioislima: I think I forgot to run the command on my previous commit.

flavioislima (commented):
Looks good overall. Will do some testing on it before approval. 👍🏽

flavioislima:
Tested here and works fine.
One thing though, that is a bug before your PR, but sometimes the status of the card doesn't change to logged in. I wonder if we are missing something or its some kind of racing condition and the frontend doesn't get the status right away. we would need to find a way to wait for the backend before changing the status

arielj:
> Yes, I believe its something with this line that is not always working:
> 
> https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/77fbf7147d1cd075d1db413c83d44e24ce5dd3ea/src/screens/Login/index.tsx#L49

If it's what I think, I think it's because of this timeout https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/77fbf7147d1cd075d1db413c83d44e24ce5dd3ea/src/screens/Login/index.tsx#L45, we wait 1 second actually setting the current user there, I didn't know the reason for that haha

arielj:
> Tested here and works fine. One thing though, that is a bug before your PR, but sometimes the status of the card doesn't change to logged in. I wonder if we are missing something or its some kind of racing condition and the frontend doesn't get the status right away. we would need to find a way to wait for the backend before changing the status

I would prefer to handle that in a different PR, I want to refactor part of that to have the current user/s set only in the global context, so we only handle that in one place and then react would take care of reflecting that to other components.

flavioislima (approved):

Issue #1251, opened 2022-04-28
Cancelling GOG games installation show game as installed


arielj:
### Describe the bug

When cancelling the installation of a GOG game, the installation is canceled, but the game status is updated as installed. It can be "uninstalled" sometimes but sometimes uninstalling the game fails because it can't find files to uninstall.

### Add logs

Not relevant, the log shows that the installation process got killed as expected.

### Steps to reproduce

1. Open the GOG library
2. Start installing a game
3. After a moment, cancel the installation
4. Game shows as installed
5. Try to uninstall it, sometimes it will work and the game status changes to not installed, but sometimes it fails to uninstall

### Expected behavior

Cancelling the installation should work like the epic library, it should NOT add the game in the installed games store.

### System Information

- OS [e. g. "Ubuntu"]: Not relevant
- Heroic Version [e. g. 2.1.1]: Main branch

PR #1252, opened 2022-04-28
Refactor/handle game status update

arielj:
This PR is a refactor of the `handleGameStatus` function in the global state:
- there was a lot of repeated code, so I grouped the different conditions to remove conditionals and branching
- I moved the `removeShortcuts` call to the backend so the shortcuts are removed as soon as the backend finishes uninstalling a game instead of calling that method from the frontend
- removed all the setTimeout calls
- removed changing `filter` in different cases (it's useless since the filter select was removed), this allowed it to group setState calls

Testing these changes I found a bug when cancelling the installation of GOG games, the bug is present in the main branch before this refactor so I reported that here #1251, I'm not familiar with the GOG code yet to be able to fix that.

Tested that everything behaves the same as it did before the refactor, would appreciate any feedback by testing these.

---

### DIFF
--- a/electron/gog/games.ts
+++ b/electron/gog/games.ts
@@ -332,6 +332,7 @@ class GOGGame extends Game {
     }
     installedGamesStore.set('installed', array)
     GOGLibrary.get().refreshInstalled()
+    removeShortcuts(this.appName, 'gog')
     return res
   }
 
--- a/electron/legendary/games.ts
+++ b/electron/legendary/games.ts
@@ -341,6 +341,8 @@ class LegendaryGame extends Game {
         ['Failed to uninstall', `${this.appName}:`, res.error],
         LogPrefix.Legendary
       )
+    } else {
+      removeShortcuts(this.appName, 'legendary')
     }
     return res
   }
--- a/src/state/GlobalState.tsx
+++ b/src/state/GlobalState.tsx
@@ -266,114 +266,59 @@ export class GlobalState extends PureComponent<Props> {
     runner
   }: GameStatus) => {
     const { libraryStatus, gameUpdates } = this.state
-    const currentApp =
-      libraryStatus.filter((game) => game.appName === appName)[0] || {}
+    const currentApp = libraryStatus.filter(
+      (game) => game.appName === appName
+    )[0]
 
-    if (currentApp && currentApp.status === status) {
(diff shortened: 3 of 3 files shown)


flavioislima:
I can still reproduce the issue on this branch. I clicked install and then clicked cancel. The folder doesn't exist but the game show as installed.
image

@imLinguin maybe you have an idea why this is happening?

imLinguin:
Is that native game?

flavioislima:
Looking at the code I can see we are saving the installed info before even starting the installation, we should move to another place and return only if the install was successful.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/54a9e50ac6d151635c6b34f85ddc73f5b26b8807/electron/gog/games.ts#L177

@imLinguin tested both, the issue is like I said above. 
Maybe we forgot something on some refactor.

imLinguin:
I believe it was introduced by some refactor.
Tested 2.2.6 no issue there

imLinguin:
Here, the command is ran, but after that it doesn't check if installation was cancelled. It only checks for errors of command itself

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/54a9e50ac6d151635c6b34f85ddc73f5b26b8807/electron/gog/games.ts#L142=

flavioislima:
@imLinguin yes it was some refactor.
I believe if we add a try/catch here:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/54a9e50ac6d151635c6b34f85ddc73f5b26b8807/electron/gog/games.ts#L142

might fix, I will try.

imLinguin:
how did it work before then, there wasn't any check either, it just somehow cancelled it.

flavioislima:
I don't know. But adding the try catch fixes it.
Actually would be good to have a eslint rule to force try/catch when using await because it is the correct way to do it and we forget from time to time.

flavioislima:
Tbh, the line `if (res.error) {` should be avoiding the writing on the installed.json file so the issue might be somewhere else, like on the `res` itself that its not returning an error maybe? 🤔

flavioislima:
@arielj yes I added the fix in my PR #1254

flavioislima (approved):
Tested on Linux and Mac. Everything works fine!
codewise looks good 👍🏽

PR #1253, opened 2022-04-29
Add favourites feature and library top section config

arielj:
This PR uses a similar pattern used for the feature to hide games but to implement the feature of marking games as Favourites. As discussed in Discord, I also added a configuration in the General Settings to configure what is displayed at the top of the Library (nothing / recently played / favourites).

https://user-images.githubusercontent.com/188464/165879423-c48f2664-5e63-4858-8aab-42150d127bfe.mp4

---

### DIFF
--- a/public/locales/bg/gamepage.json
+++ b/public/locales/bg/gamepage.json
@@ -36,11 +36,13 @@
         "yes": "ДА"
     },
     "button": {
+        "add_to_favourites": "Add To Favourites",
         "cancel": "Пауза/Отказ",
         "continue": "Продължаване на свалянето",
         "hide_game": "Hide Game",
         "import": "Внасяне на игра",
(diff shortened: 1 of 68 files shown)


flavioislima (changes requested):
Tested here and everything works pretty good. ⚔️ 
Two things that I believe could be better:
On the settings, put the new setting above the Max Workers one
Put the Recently played as the default setting since it's the current behavior.

arielj:
> Tested here and everything works pretty good. crossed_swords Two things that I believe could be better: On the settings, put the new setting above the Max Workers one Put the Recently played as the default setting since it's the current behavior.

done, I move the option, changed the default and also added translatable strings for the options that I forgot that

flavioislima (approved):

PR #1256, opened 2022-04-30
Refactor: Push progress updates for install/update from the backend

arielj:
This PR is a refactor on how we get updates when games are beings installed/updated.

Instead of adding a timeout that fetchs data from the backend, this PR updates the runLegendaryCommand and runGogdlCommand functions to report progress updated if matching specific log data. Then, the install and update methods will push the update to the frontend. This removes the need of using `tail` or `cat` to process the game install log.

In the frontend, instead of adding an interval, we can simply listen to the setGameStatus event from the backend and display the info.

I would really appreciate some tests and feedback (I tested this only on linux!!). EDIT: I was finally able to test this in a Mac and progress works fine for GOG and Epic and also when resuming a cancelled installation! I don't have a windows machine to test this

One concern:
- for some reason, the log from GOGdl and Legendary, both are pushed to the stderr (and not to the stdout), not sure if this is a linux specific thing

---


electron/legendary/games.ts
```diff
@@ -216,6 +216,61 @@ class LegendaryGame extends Game {
...
+    // parse log for game download progress
+    const etaMatch = data.match(/ETA: (\d\d:\d\d:\d\d)/m)
+    const bytesMatch = data.match(/Downloaded: (\S+.) MiB/m)
+    if (etaMatch && bytesMatch) {
+      const eta = etaMatch[1]
+      const bytes = bytesMatch[1]
+
+      // original is in bytes, convert to MiB with 2 decimals
+      totalDownloadSize =
+        Math.round((totalDownloadSize / 1024 / 1024) * 100) / 100
```
flavioislima, line 244: why dont you use Pretty-Bytes here? its the library we are already using for Byts convertion.

flavioislima (dismissed):
Looks good! I will try to test on Windows but I believe will be fine since you are not using shell specific commands to get the progress like we did before.

electron/legendary/games.ts
arielj: I noticed pretty bytes shows a different value than Legendary

I was testing with Absolute Drift for example and in the install dialog it shows 158 (seems to be dividing by 1000 using MB) but then Legendary reports 150.71 (seems to be dividing by 1024 using MiB) so the calculations were off (maybe I'm just missing some option for the prettyBytes function to set the unit)

Also, pretty bytes returns a string and then I would have to convert it back to a number for the calculations

one comemnt though, maybe we can remove the pretty bytes dependency and move this code to a helper? this type of calculation seems simple enough to not need an external dependency since we don't need that much flexibility, should be a simple refactor

flavioislima: Yes, could be better then. So we could use the same approach for all and get rid of a library.

arielj: I can do the refactor in another PR, it's used in many places and it's kinda unrelated to this change, I'll keep a note and refactor that in another moment to keep this PR simpler

flavioislima:
Everything looks good. We can merge it after fixing the conflicts :)

arielj:
@flavioislima fixing the conflicts removed your approval so I can't merge this myself

flavioislima (approved):

PR #1265, opened 2022-05-01
[UI] Add basic themes support

arielj:
This PR adds basic themes support with a few hardcoded themes.

https://user-images.githubusercontent.com/188464/166167440-d7586a23-e194-4bde-afc3-89b408f8ac5b.mp4

I removed a few css variables there were not being used, I didn't do a refactor on the variables (it could be nice to improve how we use the variables but it was not really needed for this PR)

---


src/components/UI/Dialog/index.css
```diff
@@ -28,11 +28,7 @@
 .Dialog__element[open] {
   opacity: 1;
   transform: translateY(0);
-}
-
-.Dialog__element::backdrop {
-  /* var() doesn't work here */
-  background: rgba(13, 15, 28, 0.8);
+  box-shadow: 0px 0px 0px 100vmax var(--modal-backdrop);
```
arielj, line 31: since modal backdrop can't use css variables, this is a workaround using a box-shadow that covers the whole screen

flavioislima:
Tested here and looks pretty great 💯 
The only thing is that we need to check with Billie if the colors on the game page (install, settings, etc) are ok.
I think there is one mistake on our current theme that we are using the same colors for Import and Install, I asked her to fix that.
And for the new themes, the hover colors I believe are not 100% ok as well. But lets wait until she updates the design in Figma :)

arielj:
This is ready for review, I updated the themes based on figma (though the design don't completely match so things can be a bit off since I had to do some guessing). I also added `classic` versions of all with the variant on the sidebar links color since I know it's a somewhat controversial topic.

flavioislima:
Tested it and looks fine. If we need to update something we can do it later.

PR #1267, opened 2022-05-03
A lot of rendering refactors related to libraries (and a few extras)

arielj:
This PR includes a lot of refactoring to improve the calculations done in each render:

- moved all the filter and sort logic from GlobalState to Library (we were doing all the filtering in every render, and even when the library was not shown) and memoize it so it's only calculated when a dependency changes
- simplified the filter and sort logic to only filter/sort the currently displayed library (we where doing all that for both library every time)
- moved the `numberOfGames` calculation from GlobalState to Header (same as before, calculated in every render, even when not needed) and memoize it
- load current language from browser's storage as soon as possible (removes an issue where english was shown while the component was rendering the first time)
- memoize the calculation of favourites (we were recalculating them in every render of the library when showing them)
- set initial state from localstorage to prevent flickering with the wrong state during the initial render
- move the fetching of recent games into the componentDidMount callback and store it in the state (we were doing that calculation in every render of the global state)
- prevent refreshing the library while the library is refreshing

I'll add more comments on the diff of the PR

---


electron/legendary/library.ts
```diff
@@ -143,7 +143,6 @@ export class LegendaryLibrary {
 
     if (fullRefresh) {
       try {
-        logInfo('Refreshing Epic Games...', LogPrefix.Legendary)
```
arielj, line 146: The `refresh` function called right after this line also prints this to the logs, it gave the impression that the epic library was getting refreshed twice when it didn't.

src/index.tsx
```diff
@@ -32,7 +34,7 @@ i18next
     interpolation: {
       escapeValue: false
     },
-    lng: 'en',
+    lng: storage.getItem('language') || 'en',
```
arielj, line 37: when selecting a language other than `en`, the app was loaded in english for a moment and then changed to the user's language, which looked weird and triggered an extra re-render

src/screens/Library/index.tsx
```diff
@@ -184,9 +281,9 @@ export default function Library(): JSX.Element {
 
         <h3 className="libraryHeader">{titleWithIcons()}</h3>
 
-        {refreshing && <UpdateComponent inline />}
+        {refreshing && !refreshingInTheBackground && <UpdateComponent inline />}
```
arielj, line 284: I added this `refreshingInTheBackground` to be able to differenciate when the library was being refreshed in the background and in the foreground

src/state/GlobalState.tsx
```diff
@@ -79,7 +81,7 @@ export class GlobalState extends PureComponent<Props> {
     return games
   }
   state: StateProps = {
-    category: 'epic',
+    category: storage.getItem('category') || 'epic',
```
arielj, line 84: we can load all these local storage data here instead of in the componentDidMount callback

with the previous code, if you close heroic when in the gog library, when you open heroic it will open it in the epic library for a moment and then change to gog, which looked glitchy

flavioislima:
This is pretty good for performance. As much stuff we can remove from GlobalState the better. Especially to refactor it as a functional component.

flavioislima:
Tested here and looks good.
Ready to Merge ⚔️

arielj:
just to confirm, can I merge this? I know you want to make a new release soon and maybe you prefer to leave this to after the release since it's a refactor, I'm happy to get this merged but I understand if you prefer to wait

arielj:
just to confirm, can I merge this? I know you want to make a new release soon and maybe you prefer to leave this to after the release since it's a refactor, I'm happy to get this merged but I understand if you prefer to wait

flavioislima:
Yes, perhaps we should wait for the release then and then put this on on a minor release.
Btw, I just found a small bug: the number of games is not being updated based on the Hidden status.

arielj:
> Yes, perhaps we should wait for the release then and then put this on on a minor release. Btw, I just found a small bug: the number of games is not being updated based on the Hidden status.

thanks, just pushed a fix for this, but let's wait for after the release and then this can be tested more as part of the next betas :+1:, I plan to do other refactors too

flavioislima:
I think this one can be merged :)

arielj:
merging this then, so I can see if it adds conflicts to other stuff

PR #1270, opened 2022-05-03
[UI/UX] Add Platform selection on Install dialog

flavioislima:
On Linux and Mac, adds ability to select if the user wants to install the native game or the windows version of it.

---


CoolDog22YT (approved):

CoolDog22YT:
Can someone please add this in

imLinguin:
Work in progress, it will be available in next release 😄

electron/gog/library.ts
```diff
@@ -667,6 +675,7 @@ export async function runGogdlCommand(
       })
     })
     child.on('error', (error) => {
+      console.log({ error })
```
Nocccer, line 678: Please use our log function

electron/types.ts
```diff
@@ -258,6 +258,7 @@ export interface InstallParams {
   sdlList?: Array<string>
   installLanguage?: string
   runner: Runner
+  platformToInstall: 'Windows' | 'Mac' | 'Linux'
```
Nocccer, line 261: You calling everywhere toLowerCase. 
Why not declare everything here in lower case and get rid of the toLowerCase calls.

Maybe i missed something, why you doing this ?

src/types.ts
```diff
@@ -330,3 +330,4 @@ export interface GamepadActionStatus {
 }
 
 export type Runner = 'legendary' | 'gog' | 'heroic'
+export type PlatformToInstall = 'Windows' | 'Mac' | 'Linux' | ''
```
Nocccer, line 333: Same here. i would go with lower cases and get rid of the toLowerCase calls.

Nocccer (changes requested):
Found some small issues. Beside that LGTM.

electron/types.ts
flavioislima: I will check if legendary accepts it with all lower case, because on their docs says that it is CamelCase.
So then I start from this and then need to lowercase because this is how gogdl works.

flavioislima: Yes, checking here and Legendary only accepts Camel case:
``` ./public/bin/linux/legendary install 19927295d6e3467887d4e830d8c85963 --platform mac
[cli] WARNING: Platform "mac" may be invalid. Valid ones are: Windows, Win32, Mac.
```

PR #1281, opened 2022-05-08
[UI/UX] Add Accessibility Screen

arielj:
This PR adds a few accessibility features and includes some CSS fixes for when zooming in:

- Allow changing zoom level (from 80% to 200%)
- Allow changing font family for content and buttons with any font found in the system
- The theme selector is both in the General Settings and in the Accessibility screen so it's easier to find

I'll add some comments on the PR's diff on why I did some of the CSS changes.

Here's a recording of the feature:

https://user-images.githubusercontent.com/188464/167309975-3a2aaa00-0d21-46f3-82cd-0ae5422c41e2.mp4

(the diff has a looot of files, but most are translation files and changes in the css to css variables instead of hardcoded font family values)

---


electron/main.ts
```diff
@@ -350,6 +351,10 @@ if (!gotTheLock) {
       handleProtocol(mainWindow, url)
     }
 
+    const zoomFactor =
+      parseFloat((configStore.get('zoomPercent') as string) || '100') / 100
+    mainWindow.webContents.setZoomFactor(zoomFactor)
```
arielj, line 356: set zoom factor when heroic boots

package.json
```diff
@@ -143,6 +143,7 @@
     "electron-store": "^8.0.1",
     "electron-updater": "^5.0.1",
     "filesize": "^8.0.7",
+    "font-list": "^1.4.5",
```
arielj, line 146: needed to get the system's fonts, it says it works in all Windows, Mac and Linux systems https://github.com/oldj/node-font-list

src/screens/Accessibility/index.css
```diff
@@ -0,0 +1,11 @@
+input[type='range'] {
+  width: 100%;
+  max-width: 513px;
+}
+
+.Accessibility label,
+.Accessibility span,
+.Accessibility select,
+.Accessibility option {
+  font-size: 1.5rem;
```
arielj, line 10: setting a bigger font size for the accessibility screen by default in case a user has troubles finding the zoom option with the original font size

src/screens/Settings/components/AdvancedSettings/index.tsx
```diff
@@ -135,7 +135,7 @@ export const AdvancedSettings = ({
             'Choose an Alternative Legendary Binary  (needs restart)to use'
           )}
         </span>
-        <span>
+        <span className="settingInputWithButton">
```
arielj, line 138: needed to fix some responsiveness issues with high zoom levels

src/screens/Settings/components/SyncSaves/index.tsx
```diff
@@ -141,7 +141,8 @@ export default function SyncSaves({
           style={{
             display: 'flex',
             justifyContent: 'space-between',
-            width: '513px'
+            width: '100%',
+            maxWidth: '513px'
```
arielj, line 145: needed to fix some responsiveness issues with high zoom levels

src/screens/Settings/components/WineSettings/index.tsx
```diff
@@ -222,9 +221,7 @@ export default function WineSettings({
                   <RemoveCircleIcon
                     data-testid="removeWinePath"
                     style={{
-                      color: selectedPath
-                        ? 'var(--danger)'
-                        : 'var(--background-darker)',
+                      color: selectedPath ? 'var(--danger)' : 'black',
```
arielj, line 224: the `-` icon was invisible in some themes

src/screens/Settings/index.css
```diff
@@ -151,14 +151,16 @@
...
+  max-width: 513px;
   white-space: break-spaces;
   padding-top: 4px;
 }
 
 .smallLink {
   font-size: 12px;
-  width: 513px;
+  width: 100%;
+  max-width: 513px;
```
arielj, line 163: all the changes in this file are needed to fix some responsiveness issues with high zoom levels

flavioislima:
Looks  pretty good. I will try to find time for testing tomorrow.
Two things to have in mind: using the lighthouse to check the accessibility, it complained about the cards button that have no title. Easy fix.
And the second, on the classic theme I noticed that the color for the play button is almost invisible. Another easy fix though. So if you have time send a fix on this pr since should be just a few lines 🙂

package.json
ChlorideCull: I'll quickly chime in that `font-list` has some peculiar behavior on Windows:
* It prefers font names in Chinese
* It will invoke powershell on every call (which will then also be intercepted by defender), so caching them and not constantly invoking the method will be mandatory to everything grinding to a halt
* It genuinely lists all the fonts, which is probably not what you want - a significant amount of fonts only provide glyphs for certain character sets, which makes the list balloon. `Microsoft Tai Le` is not a useful font if you run US English, for example.

arielj: for 1, what do you mean? like, fonts with chinese characters are listed first? maybe it's a sorting thing?

for 2, I can check if we can cache the list and then add a `refresh` button to refresh the cache

for 3, I don't think I can filter out stuff from the list of fonts, I wouldn't know what to safely remove since I don't know which font somebody would need and which names those fonts could have (what if the font name for some reason matches something we are ignoring?) I'll see if I can change the type of the element to allow the user to type the name + suggestions

arielj: ok I changed it to cache the list of fonts and added a reload/refresh option, now it should only get the list once and then only when the action is triggered by the user

I now order all fonts using the user's locale, so hopefully it generates a list with a more relevant order of each user

I don't really know how to filter fonts out, I'm worried that a list of things to ignore is hard to maintain and I don't want to have false positives that will hide fonts by mistake

I tried to use a text input + datalist element, but it was impossible to style and acted really glitchy while typing

arielj:
> Looks pretty good. I will try to find time for testing tomorrow. Two things to have in mind: using the lighthouse to check the accessibility, it complained about the cards button that have no title. Easy fix. And the second, on the classic theme I noticed that the color for the play button is almost invisible. Another easy fix though. So if you have time send a fix on this pr since should be just a few lines slightly_smiling_face

I fixed most screens to have 95+ accessibility using lighthouse, there are some contrast issue in a few places but I don't want to add random values to meet the requirements since eventually the figma files will have better definitions, I understand you already fixed the issue with the play buttons being too dark, right?

package.json
ChlorideCull: For 1, I mean that it tries to get the name in Chinese, and then in English if it can't get it in Chinese. The names can be localized, but the package only tries those two.

For 3, on Windows you can get a list of applicable fonts based on charset with a GDI function, but I don't know how calling out to C is done in node projects.

Honestly, it's probably not something you can fix without writing your own package, I couldn't find any alternative package that does it properly.

flavioislima:
I am testing here and on Mac the Zoom transition is weird, look at the video:

https://user-images.githubusercontent.com/26871415/168242793-38bab403-af1a-4902-bca8-1180e82756d6.mov

It is fine if I just click at it, but if I slide it isn't.
So I wonder if would solve if we have a delay of 500ms in between selecting and changing the zoom.
Or add some kind of global transition effect, if that could fix it.

src/state/GlobalState.tsx
```diff
@@ -103,14 +106,35 @@ export class GlobalState extends PureComponent<Props> {
...
+      (configStore.get('actionsFontFamily') as string) || "'Rubik', sans-serif"
   }
 
   setTheme = (newThemeName: string) => {
     configStore.set('theme', newThemeName)
     this.setState({ theme: newThemeName })
   }
 
+  setZoomPercent = (newZoomPercent: number) => {
+    configStore.set('zoomPercent', newZoomPercent)
```
flavioislima, line 123: Adding these changes inside a `setTimeout` with 300ms set makes the UI better to use IMO, because then it doesn't change abruptly.

flavioislima (approved):
Did a lot of test here and works pretty good. :) 
Really good job on this one ⚔️

src/state/GlobalState.tsx
arielj: I'll try this, I noticed the weird jumps but I didn't really know what to try here, I'll check with a delay

I don't think I can transition zoom, because it's not css zoom, is chrome's zoom and it doesn't have transition (unless I implement a gradual change by setting the intermediate values in some loop but I know that would be bad for people that prefers reduced motion)

arielj: just sent a commit with the delay, looks way better :muscle:

flavioislima: NIce, so we can merge like it is now and we can improve that later. I think it is stable enough for the next release.

flavioislima:
I will merge this one since it will add new i18n keys then the translators can take a look at it.

wouter1984:
tested the PR and love the zoom function. 

Only thing I noticed is I miss filter options in the Library view. Normally my default view is "Ready" filter. I am missing the option to switch between "Ready" and " No Filter"

flavioislima:
@wouter1984 this was replaced by a button that order the games by installed status. Just in the side of the one that order alphabetically.

wouter1984:
@flavioislima Ok I was already thinking that I was  missing something.

Keep up the good work!! Eagerly awaiting the next stable (deb) release.

PR #1304, opened 2022-05-14
[UI/UX] Settings UI adjustments

flavioislima:
Fixes some styling, and icons, and improved consistency across the settings.

image

---

### DIFF
--- a/src/components/UI/ThemeSelector/index.tsx
+++ b/src/components/UI/ThemeSelector/index.tsx
@@ -21,7 +21,7 @@ export const ThemeSelector = () => {
         value={theme}
         className="settingSelect is-drop-down"
       >
-        <option value="">Default</option>
+        <option value="Default">Default</option>
         <option value="classic">Classic</option>
         <option value="old-school">Old School Heroic</option>
         <option value="dracula">Dracula</option>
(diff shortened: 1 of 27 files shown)


src/components/UI/ThemeSelector/index.tsx
```diff
@@ -21,7 +21,7 @@ export const ThemeSelector = () => {
         value={theme}
         className="settingSelect is-drop-down"
       >
-        <option value="">Default</option>
+        <option value="Default">Default</option>
```
arielj, line 24: maybe use lowercase `default` to match the case of the other values?

src/index.tsx
```diff
@@ -73,8 +73,8 @@ i18next
     ]
   })
 
-const themeClass = (configStore.get('theme') as string) || ''
-document.body.classList.add(themeClass)
+const themeClass = configStore.get('theme') as string
```
arielj, line 76: let's keep a ` || 'default'` here, so the class matches the value in the select tag of ThemeSelector, so it's selected properly

src/state/ContextProvider.tsx
```diff
@@ -41,7 +41,7 @@ const initialContext: ContextType = {
     add: () => null,
     remove: () => null
   },
-  theme: '',
+  theme: 'Default',
```
arielj, line 44: same with the lowercase `default` here

we should also change the fallback to `default` in GlobalState.tsx:115

arielj (commented):
added some suggestions regarding the theme default fallback

PR #1306, opened 2022-05-15
Refactor: Handle login in a single place

arielj:
This PR is a refactor around the login/logut actions to address a few issues:
- logged in user affects all the app, makes more sense that it's something in the global context so we can use react's reactivity to propagate login changes
- components that needed the current user/s were reading the user from the electron stores each time
- the login flow was using a refresh function with a setTimeout
- the SidebarLinks component was handling redirection to the login (this was a problem because it makes more sense to handle that at the router level and also because it was preventing access to the Accessibility section, which users might need before log in)
- the find-in-page event was triggering multiple times

Now, the login for both epic and gog is handled by the GlobalState component and the invoke calls wait for the response to set the current username and now react's reactiveness takes care of updating whatever is needed.

I'll add more comments in the diff.

Any thought about or against this refactor? I think adding those new `epic` and `gog` objects in the context grouping stuff makes things more readable.

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -626,12 +626,6 @@ ipcMain.on('resetHeroic', async () => {
   }
 })
 
-ipcMain.handle('authGOG', async (event, code) =>
-  GOGUser.login(code).then(() =>
-    mainWindow.webContents.send('updateLoginState')
-  )
-)
-
 ipcMain.on('createNewWindow', async (e, url) =>
   new BrowserWindow({ height: 700, width: 1200 }).loadURL(url)
 )
@@ -685,13 +679,8 @@ ipcMain.handle('getUserInfo', async () => LegendaryUser.getUserInfo())
 // Checks if the user have logged in with Legendary already
 ipcMain.handle('isLoggedIn', async () => LegendaryUser.isLoggedIn())
 
-ipcMain.handle('login', async (event, sid) =>
-  LegendaryUser.login(sid).then((value) => {
-    mainWindow.webContents.send('updateLoginState')
-    return value
-  })
-)
-
+ipcMain.handle('login', async (event, sid) => LegendaryUser.login(sid))
+ipcMain.handle('authGOG', async (event, code) => GOGUser.login(code))
 ipcMain.handle('logoutLegendary', async () => LegendaryUser.logout())
 ipcMain.handle('logoutGOG', async () => GOGUser.logout())
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,12 +13,15 @@ import Accessibility from './screens/Accessibility'
 import ContextProvider from './state/ContextProvider'
 
 function App() {
-  const { contentFontFamily, actionsFontFamily } = useContext(ContextProvider)
+  const { epic, gog, contentFontFamily, actionsFontFamily } =
+    useContext(ContextProvider)
 
   const style = {
     '--content-font-family': contentFontFamily,
     '--actions-font-family': actionsFontFamily
   } as React.CSSProperties
+
+  const loggedIn = epic.username || gog.username
 
   return (
     <div className="App" style={style}>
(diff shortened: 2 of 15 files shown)


electron/main.ts
```diff
@@ -685,13 +679,8 @@ ipcMain.handle('getUserInfo', async () => LegendaryUser.getUserInfo())
 // Checks if the user have logged in with Legendary already
 ipcMain.handle('isLoggedIn', async () => LegendaryUser.isLoggedIn())
 
-ipcMain.handle('login', async (event, sid) =>
-  LegendaryUser.login(sid).then((value) => {
-    mainWindow.webContents.send('updateLoginState')
```
arielj, line 690: this event was not being used in the frontend

src/App.tsx
```diff
@@ -13,20 +13,23 @@ import Accessibility from './screens/Accessibility'
...
+  const loggedIn = epic.username || gog.username
+
   return (
     <div className="App" style={style}>
       <HashRouter>
         <Sidebar />
         <main className="content">
           <Routes>
-            <Route path="/" element={<Library />} />
+            <Route path="/" element={loggedIn ? <Library /> : <Login />} />
```
arielj, line 32: handle the required login here instead of in SidebarLinks, not sure if we want to require user login for other urls, maybe Settings should have the same?

src/components/UI/Sidebar/components/SidebarLinks/index.tsx
```diff
@@ -79,32 +73,43 @@ export default function SidebarLinks() {
...
+        >
+          <>
+            <div className="Sidebar__itemIcon">
+              <FontAwesomeIcon icon={faGamepad} />
+            </div>
+            {t('Library')}
+          </>
+        </NavLink>
+      )}
+      {!loggedIn && (
```
arielj, line 95: split this into 2 different links to make it more clear

src/screens/Login/components/Runner/index.tsx
```diff
@@ -15,22 +15,21 @@ interface RunnerProps {
...
 }
 
 export default function Runner(props: RunnerProps) {
   const { t } = useTranslation()
   async function handleLogout() {
     await props.logoutAction()
-    window.localStorage.clear()
-    props.refresh()
+    // FIXME: only delete local storage relate to one store, or only delete if logged out from both
+    //window.localStorage.clear()
```
arielj, line 25: we have a bug here, but it's outside of the scope of this refactor: when we log out from one store this is clearing the localStorage but we are still logged in with the other store so this shouldn't happend

maybe we have to only delete the data specific to the store that's logged out only?

src/screens/WebView/index.tsx
```diff
@@ -95,36 +87,48 @@ export default function WebView() {
...
-            })
+            if (code) {
+              gog.login(code).then(() => {
+                handleSuccessfulLogin()
+              })
+            }
+          }
+        } else {
+          if (
+            webview.getURL() === 'https://www.epicgames.com/id/api/redirect'
```
arielj, line 98: only check for the sid if in this url

src/screens/WebView/index.tsx
```diff
@@ -95,36 +87,48 @@ export default function WebView() {
...
+                      message: t('status.logging', 'Logging In...')
+                    })
+                    await epic.login(sid)
+                    handleSuccessfulLogin()
+                  } catch (error) {
+                    console.error(error)
+                    ipcRenderer.send('logError', error)
+                  }
+                },
+                { once: true }
```
arielj, line 127: listen for the event only once, it was getting triggered more than once

electron/main.ts
arielj, line 690: this updateLoginState event was not being used in the frontend

flavioislima (commented):
Looks good, Let's just wait to see if it fixes the issue there and test on all platforms.

flavioislima (approved):
Ok this one looks good to merge.
What a good timing since this fixed that weird issue on login.
Thanks for that 😁

PR #1307, opened 2022-05-15
Modularization of themes

CleoMenezesJr:
Anticipating that the list of themes will grow, it is interesting that there is a modularization of themes for better organization and so that in the future there will not be a giant archive of themes.
---


flavioislima:
Looks good but I noticed that the Marine theme is not working:

image

Other than that looks fine 👍🏽

flavioislima (changes requested):
Also you need to run `yarn prettier` to fix the linting. :)

Issue #1301, opened 2022-05-13
[UI] option to show all game tiles in color instead of black and white


jianpan:
### Feature description

Can we add an option to show all tiles in library in color? Currently only installed games are in color. I have a relatively large list. Black and white tiles make games a little bit difficult to distinguish. Thanks.
![heroic](https://user-images.githubusercontent.com/1950818/168378475-ce842c10-bcba-4ea4-862a-17d2aac560dc.png)

arielj:
This would be a good fit in the new Accessibility settings :+1:

PR #1311, opened 2022-05-15
[UI] Add accessibility option to show all game tiles in color

connercsbn:
Patch to address feature request https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1301

This adds an option in accessibility settings that, when enabled, adds the `allTilesInColor` class to the GameCard component, making all tiles show in color. 

image
image

Just wanted to get this out there and get some feedback. Will do checklist soon!

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -1,7 +1,9 @@
 {
     "accessibility": {
         "actions_font_family": "Actions Font Family (Default: \"Rubik\")",
+        "all_tiles_in_color": "Show all game tiles in color",
         "content_font_family": "Content Font Family (Default: \"Cabin\")",
+        "fonts": "Fonts",
         "title": "Accessibility",
         "zoom": "Zoom"
     },
@@ -197,6 +199,9 @@
             "startInstall": "Installation Started"
         },
         "moved": "Finished Moving",
+        "refresh": {
+            "error": "Couldn't fetch releases from upstream, maybe because of Github API restrictions! Try again later."
+        },
         "uninstalled": "Uninstalled",
         "update": {
             "canceled": "Updated Canceled",
@@ -253,7 +258,6 @@
         "enableFSRHack": "Enable FSR Hack (Wine version needs to support it)",
         "esync": "Enable Esync",
         "exit-to-tray": "Exit to System Tray",
-        "FsrSharpnessStrenght": "FSR Sharpness Strength",
         "fsync": "Enable Fsync",
         "gamemode": "Use GameMode (Feral Game Mode needs to be installed)",
         "language": "Choose App Language",
@@ -282,7 +286,6 @@
             "title": "Manual Sync Saves",
             "upload": "Upload"
         },
-        "maxRecentGames": "Played Recently to Show",
         "maxworkers": "Maximum Number of Workers when downloading",
         "minimize-on-launch": "Minimize Heroic After Game Launch",
         "offlinemode": "Run Game Offline",
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -258,6 +258,7 @@
         "enableFSRHack": "Enable FSR Hack (Wine version needs to support it)",
         "esync": "Enable Esync",
         "exit-to-tray": "Exit to System Tray",
+        "FsrSharpnessStrenght": "FSR Sharpness Strength",
         "fsync": "Enable Fsync",
         "gamemode": "Use GameMode (Feral Game Mode needs to be installed)",
         "language": "Choose App Language",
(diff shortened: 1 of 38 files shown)


src/screens/Accessibility/index.tsx
```diff
@@ -162,6 +165,20 @@ export default function Accessibility() {
...
         <ThemeSelector />
+        <span className="setting">
+          <label className={classNames('toggleWrapper', { isRTL: isRTL })}>
+            <ToggleSwitch
+              value={allTilesInColor}
+              handleChange={() => {
+                setAllTilesInColor(!allTilesInColor)
+              }}
+              title={t(
+                'accessibility.allTilesInColor',
```
arielj, line 176: Please use camel_case here `accessibility.all_titles_in_color`.

Also, please run `yarn i18next` after changing that so it updates all the locales for all languages.

connercsbn:
Did something go wrong when I ran yarn i18next?

public/locales/fa/translation.json
```diff
@@ -160,8 +181,20 @@
...
+        "saves": {
+            "syncing": ""
+        }
+    },
+    "launch": {
+        "options": "Launch Options..."
+    },
     "library": {
-        "gridView": "نمای شبکه‌ای",
+        "gridView": "نمای شبکه\u200cای",
```
imLinguin, line 197: Could you remove this `\u200c` manually? this is i18n bug with languages like this

connercsbn: 👍 https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1311/commits/3842ba3b58a7fcb42003ad4f47ffd56e356c8826

public/locales/bg/gamepage.json
```diff
@@ -66,14 +64,6 @@
         "moving": "Преместване",
         "repairing": "Поправяне"
     },
-    "info": {
```
flavioislima, line 69: those removals doesnt seems right

flavioislima:
Need to check why so many translations were deleted on this PR. I cannot find any change related to that, so we need to check `main` and run the command there to see whats happening because it is deleting a lot of keys from settings, gamepage, etc.

public/locales/bg/gamepage.json
connercsbn: that's what I was thinking, but that's what happens when I run yarn i18next, even on the up-to-date main branch

flavioislima: `yarn i18next` or `yarn i18n`? should be the same, I guess. But I will try to investigate this later.

flavioislima:
@arielj I believe that the missing keys are related to one of the refactors on the UI.
Some keys moved to a different file and then lost the translation.
Example: `status.moving` was on the `gamepage.json` but moved to `translations.json` and lost the keys.
Maybe was when some components were moved. I dont know if this is breaking some language right now, but might be.
We should investigate that in another PR and try to move the keys around files, or maybe start using one single file for all translations.

arielj:
Agree on using one single file for all the translations, since we don't really need to modify those files manually anyway, should simplify some code and allow re-usability

flavioislima:
@arielj yes, that will be a lot of work, we have 64 translations file now, but will work best and it is better than losing all translations.

flavioislima:
@connercsbn @arielj 
Found the issue and it is simple to solve.
Just edit this line:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/cb7b47560c1e244e38355fa3d757a64920e39573/src/screens/Game/GamePage/index.tsx#L48

to put it like this:
`const { t: t2 } = useTranslation()`

and then run `yarn i18n` again. In case it doesnt fix it, try removing the folder `locales` and then run the command again. Should be fine.

The thing is that for the default file, that is `translations.json` we dont need to pass an argument, because it is the default.
I mean, the i18next should be able to catch that but it doesnt.

public/locales/en/translation.json
```diff
@@ -282,7 +286,6 @@
             "title": "Manual Sync Saves",
             "upload": "Upload"
         },
-        "maxRecentGames": "Played Recently to Show",
```
flavioislima, line 285: weird that it is still removing a few keys not related with the PR. 🤔

connercsbn: Ah, looks like it's caused by a couple deleted lines from here: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/commit/421ae61f5a1db85c83d6acac2f8601e515e9e452

flavioislima: Yes, you are right. Let me try to fix those

flavioislima:
ok, fixed it in main. Try to update your branch now.
also tested the feature and it is working fine. :)

connercsbn (commented):
Seems like last thing I need to do is revert the deletion of these lines, since they're being generated in English for these translation files. I'll see if I can grab the files from the main branch and rerun i18n

flavioislima (approved):
Perfect! Thanks a lot for that! :D

PR #1315, opened 2022-05-15
Refactor: Form elements

arielj:
This PR is a refactor around how we use selects, text inputs, and the toggle switch tags.

There were some issues:
- we had to copy some extra tags with specific classes in specific order for things to work properly (like to add a text field we had to add a `span.setting`, then a label with a specific code for RTL, then another element and so on), now each tag has it's own component so to add a select field (with a label and some extra stuff) we can use SelectField, similar with TextInputField, also each component handles the RTL variable
- the css for those elements was a big hacky and elements were not all the same width (which was driving me crazy xD)
- responsiveness was not asking properly in some cases
- for the ToggleSwitch, I removed some elements to make it simpler relying a big more in CSS instead of code
- The FormControl element is doing to much and still requires specific html, these new components are more focused

I'll add more comments in the diff

---


src/components/UI/LanguageSelector/index.tsx
```diff
@@ -7,85 +16,94 @@ export enum FlagPosition {
...
 
 interface Props {
-  className?: string
-  currentLanguage?: string
   flagPossition?: FlagPosition
-  handleLanguageChange: (language: string) => void
+  showWeblateLink?: boolean
+}
+
+const languageLabels: { [key: string]: string } = {
```
arielj, line 23: moved these 2 constants outside the component so they are not initialized in every call to the function

src/components/UI/ToggleSwitch/index.tsx
```diff
@@ -1,44 +1,34 @@
...
 export default function ToggleSwitch(props: Props) {
-  const {
-    handleChange,
-    value,
-    disabled,
-    title,
-    dataTestId = 'toggleSwitch'
-  } = props
-  // TODO fixes errors in the console, but the props may not be necessary at all
-  const checkmarkProps = {
```
arielj, line 22: I was able to remove the need of the `.checkmark` element completely so we don't need this anymore

src/components/UI/ToggleSwitch/index.css
```diff
@@ -1,82 +1,54 @@
...
+  font-size: 1.5rem;
+  line-height: 1.2rem;
+  font-weight: 900;
+  font-family: Arial, Helvetica, sans-serif;
 }
 
-.switch input:checked ~ .checkmark:after {
-  display: block;
+.toggleSwitchWrapper.checked:before {
+  content: '\2713';
```
arielj, line 35: replaced the hacky check done using css borders and rotation with a real `check` character

src/screens/Accessibility/index.css
```diff
@@ -3,29 +3,36 @@ input[type='range'] {
   max-width: 513px;
 }
 
-.Accessibility > .settingsWrapper > .setting > .is-drop-down {
-  background-position: calc(100% - 20px) calc(1em + -4px),
-    calc(100% - 15px) calc(1em + -4px), calc(101% - 2.5em) calc(0.5em + -8px);
```
arielj, line 8: changed the main background properties so it works in smaller and bigger selects so this is not needed

src/screens/Login/index.tsx
```diff
@@ -13,14 +13,9 @@ import { configStore, gogConfigStore } from 'src/helpers/electronStores'
...
 const { ipcRenderer } = window.require('electron')
 
-const storage: Storage = window.localStorage
 export default function NewLogin() {
-  const { t, i18n } = useTranslation()
-  const currentLanguage = i18n.language
-  const handleChangeLanguage = (language: string) => {
-    storage.setItem('language', language)
-    i18n.changeLanguage(language)
-  }
```
arielj, line 23: language changes are handled by the LanguageSelector component directly

src/App.css
```diff
@@ -185,17 +185,10 @@ html,
 }
 
 .material-icons.settings.folder {
```
flavioislima, line 187: One thing, not really related with the PR.
Have you noticed that we are using two different folder icons? one on the settings and another on the InstallModal.
I think we should pick one, maybe the one that we are using on the InstallModal looks a bit better imo.
But we can do that in another task if you think this could increase the scope of this PR.

flavioislima (approved):
Looks pretty good! 
I can see all settings finally are well aligned :)

just a question, the ToggleSwitch checkmark background should not be using transparent, or the same color as the setting's background? 🤔 
Screenshot 2022-05-17 at 09 31 06

flavioislima:
Another bug I just found was tha the settings are not working for RTL anymore, a param for that would be necessary to change the layout for those languages.
Screenshot 2022-05-17 at 09 54 07

arielj:
> Looks pretty good! I can see all settings finally are well aligned :)
> 
> just a question, the ToggleSwitch checkmark background should not be using transparent, or the same color as the setting's background? thinking Screenshot 2022-05-17 at 09 31 06

I have to check this, I had an issue with hidding the original input and I was covering it setting the background of the check box, but I fixed that problem and forgot to remove that extra background

arielj:
> Another bug I just found was tha the settings are not working for RTL anymore, a param for that would be necessary to change the layout for those languages. Screenshot 2022-05-17 at 09 54 07

weird, the compontents itself now gets the `isRTL` from the context so the prop is not needed, maybe I'm not setting that properly, I'll check that in a moment

src/App.css
arielj: yeah, I didn't know the reason so I just skipped that, I agree that a single icon in all palces would look better

I'm also thinking about adding another component `PathInputField` that would take care of the input with the folder icon, we use that in a lot of places, I think it's a good opportunity to simplify that

arielj:
Fixed the RTL language style (I also improved some elements to do full RTL alignment and not just the label

image

arielj:
Also added an extra `TextFieldWithIconField` component so we don't need to pass specific markup for the icon, we just pass the icon and the click handler and it will create the correct html.

Finally, I replaced all uses of a Folder icon with the one in the InstallDialog component

src/App.css
arielj: I extracted a TextInputWithIconField component instead, maybe we can extract a PathInputField in the future, but not all path inputs behave the same so decided to go with a more generic input with icon

flavioislima (approved):
I think that looks great now!
Ready to merge! ⚔️

PR #1324, opened 2022-05-16
[UX] Open main window on tray double click

arielj:
This PR adds the feature of opening the main window when double clicking the icon in the Tray/Taskbar.

This DOES NOT work on linux, the events that we need to listen to only work on window and mac as stated in the Electron docs https://www.electronjs.org/docs/latest/api/tray/#event-double-click-macos-windows

There is no possible workaround for Linux until electron adds support for it.

---


flavioislima (approved):

flavioislima:
Looks good!

PR #1331, opened 2022-05-17
[Feature] Option: Add games as none steam game

Nocccer:
- [ ] Create a backend function which adds a game to the shortcuts.vdf, so steam picks it up has none steam game
- [ ] Create a function to remove a game from the shortcuts.vdf
- [ ] Handle flatpak and none flatpak 
- [ ] Handle os paths

---

### DIFF
--- a/electron/gog/games.ts
+++ b/electron/gog/games.ts
@@ -38,7 +38,7 @@ import {
   setupEnvVars,
   setupWrappers
 } from '../launcher'
-import { addShortcuts, removeShortcuts } from '../shortcuts'
+import { addShortcuts, removeShortcuts } from '../shortcuts/shortcuts'
 import setup from './setup'
 import { getGogdlCommand, runGogdlCommand } from './library'
 
--- a/electron/legendary/games.ts
+++ b/electron/legendary/games.ts
@@ -27,7 +27,7 @@ import {
   setupWrappers,
   launchCleanup
 } from '../launcher'
-import { addShortcuts, removeShortcuts } from '../shortcuts'
+import { addShortcuts, removeShortcuts } from '../shortcuts/shortcuts'
 import { basename, join } from 'path'
 import { runLegendaryCommand } from './library'
 import { gameInfoStore } from './electronStores'
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1007,7 +1007,7 @@ ipcMain.handle('uninstall', async (event, args) => {
     .uninstall()
     .then(() => {
       if (shouldRemovePrefix) {
-        logInfo(`Removing prefix ${winePrefix}`)
+        logInfo(`Removing prefix ${winePrefix}`, LogPrefix.Backend)
         if (existsSync(winePrefix)) {
           // remove prefix if exists
           rmSync(winePrefix, { recursive: true })
@@ -1142,7 +1142,7 @@ ipcMain.handle('updateGame', async (e, game, runner) => {
       logInfo('finished updating', LogPrefix.Backend)
     })
     .catch((err) => {
-      logError(err, LogPrefix.Backend)
+      logError(`${err}`, LogPrefix.Backend)
       notify({ title, body: i18next.t('notify.update.canceled') })
       return err
     })
--- a/electron/shortcuts/nonesteamgame.ts
+++ b/electron/shortcuts/nonesteamgame.ts
@@ -1,0 +1,116 @@
+import {
+  existsSync,
+  readdirSync,
+  readFileSync,
+  writeFileSync
+} from 'graceful-fs'
+import { homedir } from 'os'
+import { join } from 'path'
+import { GameInfo } from '../types'
+import { getIcon } from './utils'
+
+async function addNoneSteamGame(gameInfo: GameInfo): Promise<string> {
+  return new Promise((resolve, reject) => {
+    // Todo handle none linux paths
+    const userdataDir = join(homedir(), '.steam', 'steam', 'userdata')
+
+    if (!existsSync(userdataDir)) {
+      reject(`${userdataDir} does not exist. Can't add game to steam!`)
+    }
+
+    // get all folders in userdata folder
+    const folders = readdirSync(userdataDir, {
+      withFileTypes: true
+    })
+      .filter((dirent) => dirent.isDirectory())
+      .map((dirent) => dirent.name)
+
+    if (folders.length <= 0) {
+      reject(`${userdataDir} does not contain a single directory!`)
+    }
+
+    let added = false
+
+    // special character used by steam
+    // see https://www.utf8-chartable.de/unicode-utf8-table.pl?number=128&names=2&utf8=string-literal
+    const nullChar = '\x00'
+    const backspace = '\x08'
+    const startOfHeading = '\x01'
+    const startOfText = '\x02'
+
+    for (const folder of folders) {
+      const shortcutsFile = join(folder, 'config', 'shortcuts.vdf')
+
+      if (!existsSync(shortcutsFile)) {
+        continue
+      }
+
+      // read file and remove last two backspaces
+      let content = readFileSync(shortcutsFile).toString().slice(0, -2)
+
+      // structure is from:
+      // https://github.com/CorporalQuesadilla/Steam-Shortcut-Manager/wiki/Steam-Shortcuts-Documentation#shortcut-entry-structure
+      const appName = `${startOfHeading}AppName${nullChar}${gameInfo.app_name}${nullChar}`
+
+      // TODO: handle none flatpak
+      const exe = `${startOfHeading}Exe${nullChar}"flatpak run com.heroicgameslauncher.hgl heroic://launch/${gameInfo.app_name}"${nullChar}`
+      const startDir = `${startOfHeading}StartDir${nullChar}"${process.cwd()}"${nullChar}`
+      const icon = `${startOfHeading}icon${nullChar}"${getIcon(
+        gameInfo.app_name,
+        gameInfo
+      )}"${nullChar}`
+      const shortcutPath = `${startOfHeading}ShortcutPath${nullChar}${nullChar}`
+      const launchOptions = `${startOfHeading}LaunchOptions${nullChar}${nullChar}`
+      const isHidden = `${startOfText}IsHidden${nullChar}${nullChar}${nullChar}${nullChar}${nullChar}`
+      const allowDesktopConfig = `${startOfText}AllowDesktopConfig${nullChar}${startOfHeading}${nullChar}${nullChar}${nullChar}`
+      const allowOverlay = `${startOfText}AllowOverlay${nullChar}${startOfHeading}${nullChar}${nullChar}${nullChar}`
+      const openVR = `${startOfText}OpenVR${nullChar}${nullChar}${nullChar}${nullChar}${nullChar}`
+      const devkit = `${startOfText}Devkit${nullChar}${nullChar}${nullChar}${nullChar}${nullChar}`
+      const devkitGameID = `${startOfHeading}DevkitGameID${nullChar}${nullChar}`
+      const devkitOverrideAppID = `${startOfText}DevkitOverrideAppID${nullChar}${nullChar}${nullChar}${nullChar}${nullChar}`
+      const lastPlayTime = `${startOfText}LastPlayTime${nullChar}${nullChar}${nullChar}${nullChar}${nullChar}`
+      const tags = `${nullChar}tags${nullChar}`
+      const end = `${backspace}${backspace}`
+
+      content = [
+        content,
+        appName,
+        exe,
+        startDir,
+        icon,
+        shortcutPath,
+        launchOptions,
+        isHidden,
+        allowDesktopConfig,
+        allowOverlay,
+        openVR,
+        devkit,
+        devkitGameID,
+        devkitOverrideAppID,
+        lastPlayTime,
+        tags,
+        end
+      ].join()
+
+      try {
+        writeFileSync(shortcutPath, content)
+      } catch (error) {
+        reject(
+          `Adding ${gameInfo.app_name} to ${shortcutPath} failed with: ${error}`
+        )
+      }
+
+      added = true
+    }
+
+    if (added) {
+      resolve(`${gameInfo.app_name} was succesfully added to steam.`)
+    } else {
+      reject(
+        "Game was not added, because couldn't find a shortcuts.vdf in one of the userdata/:uid/config folders!"
+      )
+    }
+  })
+}
+
+export { addNoneSteamGame }
--- a/electron/shortcuts/shortcuts.ts
+++ b/electron/shortcuts/shortcuts.ts
@@ -1,12 +1,13 @@
 import { app, shell } from 'electron'
-import { existsSync, mkdirSync, unlink, writeFile } from 'graceful-fs'
-import { logError, logInfo, LogPrefix } from './logger/logger'
-import { GlobalConfig } from './config'
-import { execAsync, removeSpecialcharacters } from './utils'
-import { Game } from './games'
-import { Runner, GameInfo } from './types'
-import { heroicIconFolder, userHome } from './constants'
-import { GOGLibrary } from './gog/library'
+import { unlink, writeFile } from 'graceful-fs'
+import { logError, logInfo, LogPrefix } from '../logger/logger'
+import { GlobalConfig } from '../config'
+import { removeSpecialcharacters } from '../utils'
+import { Game } from '../games'
+import { Runner, GameInfo } from '../types'
+import { userHome } from '../constants'
+import { GOGLibrary } from '../gog/library'
+import { getIcon } from './utils'
 
 /**
  * Adds a desktop shortcut to $HOME/Desktop and to /usr/share/applications
@@ -125,33 +126,6 @@ function shortcutFiles(gameTitle: string) {
   }
 
   return [desktopFile, menuFile]
-}
-
-async function getIcon(appName: string, gameInfo: GameInfo) {
-  if (!existsSync(heroicIconFolder)) {
-    mkdirSync(heroicIconFolder)
-  }
-  if (gameInfo.runner === 'legendary') {
-    const image = gameInfo.art_square.replaceAll(' ', '%20')
-    let ext = image.split('.').reverse()[0]
-    if (ext !== 'jpg' && ext !== 'png') {
-      ext = 'jpg'
-    }
-    const icon = `${heroicIconFolder}/${appName}.${ext}`
-    if (!existsSync(icon)) {
-      await execAsync(`curl '${image}' --output ${icon}`)
-    }
-    return icon
-  } else if (gameInfo.runner === 'gog') {
-    const apiData = await GOGLibrary.get().getGamesData(appName)
-    let iconUrl = apiData?._links?.icon.href
-    iconUrl = iconUrl.replace('{ext}', 'png')
-    const icon = `${heroicIconFolder}/${appName}.png`
-    if (!existsSync(icon)) {
-      await execAsync(`curl '${iconUrl}' --output ${icon}`)
-    }
-    return icon
-  }
 }
 
 export { removeShortcuts, addShortcuts }
--- a/electron/shortcuts/utils.ts
+++ b/electron/shortcuts/utils.ts
@@ -1,0 +1,34 @@
+import { existsSync, mkdirSync } from 'graceful-fs'
+import { execAsync } from '../utils'
+import { GOGLibrary } from '../gog/library'
+import { heroicIconFolder } from '../constants'
+import { GameInfo } from '../types'
+
+async function getIcon(appName: string, gameInfo: GameInfo) {
+  if (!existsSync(heroicIconFolder)) {
+    mkdirSync(heroicIconFolder)
+  }
+  if (gameInfo.runner === 'legendary') {
+    const image = gameInfo.art_square.replaceAll(' ', '%20')
+    let ext = image.split('.').reverse()[0]
+    if (ext !== 'jpg' && ext !== 'png') {
+      ext = 'jpg'
+    }
+    const icon = `${heroicIconFolder}/${appName}.${ext}`
+    if (!existsSync(icon)) {
+      await execAsync(`curl '${image}' --output ${icon}`)
+    }
+    return icon
+  } else if (gameInfo.runner === 'gog') {
+    const apiData = await GOGLibrary.get().getGamesData(appName)
+    let iconUrl = apiData?._links?.icon.href
+    iconUrl = iconUrl.replace('{ext}', 'png')
+    const icon = `${heroicIconFolder}/${appName}.png`
+    if (!existsSync(icon)) {
+      await execAsync(`curl '${iconUrl}' --output ${icon}`)
+    }
+    return icon
+  }
+}
+
+export { getIcon }
--- a/electron/utils.ts
+++ b/electron/utils.ts
@@ -410,7 +410,7 @@ async function searchForExecutableOnPath(executable: string): Promise<string> {
         return stdout.split('\n')[0]
       })
       .catch((error) => {
-        logError(error, LogPrefix.Backend)
+        logError(`${error}`, LogPrefix.Backend)
         return ''
       })
   }
--- a/electron/wine-manager/ipc_handler.ts
+++ b/electron/wine-manager/ipc_handler.ts
@@ -40,7 +40,7 @@ ipcMain.handle('refreshWineVersionInfo', async (e, fetch) => {
     const releases = await updateWineVersionInfos(fetch)
     return releases
   } catch (error) {
-    logError(String(error), LogPrefix.WineDownloader)
+    logError(`${error}`, LogPrefix.WineDownloader)
     throw error
   }
 })


CommandMC:
This is very nitpicky and I'm sorry, but it's "Non-Steam game", not "None-Steam game"

CommandMC (commented):
Will give it a test on Win and Mac in a bit (provided Steam runs on my VMs)

electron/constants.ts
```diff
@@ -107,11 +107,26 @@ function fixAsarPath(origin: string): string {
   return origin
 }
 
-function getSteamCompatFolder() {
-  if (existsSync(`${userHome}/.var/app/com.valvesoftware.Steam/.steam/steam`)) {
-    return `${userHome}/.var/app/com.valvesoftware.Steam/.steam/steam`
+export function getSteamCompatFolder() {
+  // Paths are from https://savelocation.net/steam-game-folder
+  if (isWindows) {
+    return join('C:', 'Program Files (x86)', 'Steam')
```
arielj, line 113: maybe you can try running the `where steam` command in the system instead of hardcoding this path? I don't have windows but I found it's kinda the equivalent of `which steam` in linux. If a user changes their installation folder this will break

arielj: oh nevermind, I see this looks for the compat folder, not the steam binary folder

src/screens/Game/GameSubMenu/index.tsx
```diff
@@ -112,6 +112,10 @@ export default function GamesSubmenu({
     ipcRenderer.send('addShortcut', appName, runner, true)
   }
 
+  function handleAddToSteam() {
+    ipcRenderer.send('addToSteam', appName, runner, true)
```
Nocccer, line 116:
```suggestion
    ipcRenderer.send('addToSteam', appName, runner)
```

electron/shortcuts/nonsteamgame.ts
```diff
@@ -0,0 +1,371 @@
...
+        gameTitle: props.gameInfo.title,
+        error: errors.join('\n'),
+        adding: true
+      })
+    }
+
+    if (errors.length === 0) {
+      const message = `${props.gameInfo.title} was successfully added to steam.`
+      logInfo(message, LogPrefix.Shortcuts)
+      notifyFrontend({ message, adding: true })
```
arielj, line 296: We should add the message to the translations since we are displaying it in the frontend.

electron/shortcuts/nonsteamgame.ts
```diff
@@ -0,0 +1,371 @@
...
+          `${props.gameInfo.app_name}.lastPlayed`
+        ) as Date
+      } else {
+        newEntry.LastPlayTime = new Date()
+      }
+
+      content.shortcuts.push(newEntry)
+
+      // rewrite shortcuts.vdf
+      writeShortcutFile(shortcutsFile, content)
```
arielj, line 280: maybe wrap this into a try/catch? and only set added = true if no error, I'm always worried about writing files

electron/shortcuts/nonsteamgame.ts
```diff
@@ -0,0 +1,371 @@
...
+ * @returns none
+ */
+async function addNonSteamGame(props: {
+  steamUserdataDir: string
+  gameInfo: GameInfo
+}): Promise<void> {
+  const { folders, error } = checkSteamUserDataDir(props.steamUserdataDir)
+
+  if (error) {
+    showErrorInFrontend({
```
arielj, line 196: Maybe write this error in the logs too?

src/screens/Game/GameSubMenu/index.tsx
```diff
@@ -191,6 +195,12 @@ export default function GamesSubmenu({
                 {t('submenu.addShortcut', 'Add shortcut')}
               </button>
             )}
+            <button
```
arielj, line 198: can we detect if steam is actually installed? since we are looking for a steam folder, maybe we can hide this button if not

arielj (commented):
Hey! I added some comments, mainly related to translations but 2 or 3 other small things.

One general comment, I would say `Steam` with capital `S` everywhere.

src/screens/Game/GameSubMenu/index.tsx
Nocccer: We could only on mac and linux i think. With `which steam` we can atleast check that the executable is there.

electron/shortcuts/nonsteamgame.ts
```diff
@@ -0,0 +1,420 @@
...
+  const { folders, error } = checkSteamUserDataDir(props.steamUserdataDir)
+
+  if (error) {
+    logError(error, LogPrefix.Shortcuts)
+    showErrorInFrontend({
+      gameTitle: props.gameInfo.title,
+      error,
+      adding: false
+    })
+  } else {
```
CommandMC, line 354: Another early-return opportunity here

CommandMC (commented):
Some small changes here and there, but looks great overall. Will give it some tests on Linux and Windows now

electron/constants.ts
```diff
@@ -107,11 +108,22 @@ function fixAsarPath(origin: string): string {
   return origin
 }
 
-function getSteamCompatFolder() {
-  if (existsSync(`${userHome}/.var/app/com.valvesoftware.Steam/.steam/steam`)) {
-    return `${userHome}/.var/app/com.valvesoftware.Steam/.steam/steam`
+export function getSteamCompatFolder() {
+  // Paths are from https://savelocation.net/steam-game-folder
+  if (isWindows) {
+    return join('C:/Program Files (x86)/Steam')
```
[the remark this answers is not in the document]
Nocccer: Atually we need to use normalize from path. Let path care about it.

electron/gog/games.ts
```diff
@@ -550,7 +552,9 @@ class GOGGame extends Game {
     }
     installedGamesStore.set('installed', array)
     GOGLibrary.get().refreshInstalled()
-    removeShortcuts(this.appName, 'gog')
+    await removeShortcuts(this.appName, 'gog')
+    const gameInfo = await this.getGameInfo()
+    await removeNonSteamGame({ steamUserdataDir, gameInfo })
```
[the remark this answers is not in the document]
Nocccer: getIcon() needs the gameInfo aswell. Also we need the app_name. This is the best option so far.

electron/gog/library.ts
```diff
@@ -129,7 +129,22 @@ export class GOGLibrary {
...
+    return this.library.get(slug) || this.getInstallAndGameInfo(slug) || null
+  }
+
+  public getInstallAndGameInfo(slug: string): GameInfo {
+    const game = (libraryStore.get('games') as GameInfo[]).find(
+      (value) => value.app_name === slug
+    )
+
+    if (!game) return null
+    const installedInfo = this.installedGames.get(String(game.app_name))
```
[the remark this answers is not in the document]
Nocccer: Actually no difference. game.app_name is already a string so unessecary String call there

electron/constants.ts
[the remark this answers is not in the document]
CommandMC: Are you sure normalize does anything for us here?
```
> path.normalize('C:/Program Files (x86)/Steam')
'C:/Program Files (x86)/Steam'
```

electron/shortcuts/nonsteamgame.ts
```diff
@@ -0,0 +1,420 @@
...
+    }
+  }
+
+  const folders = readdirSync(steamUserdataDir, {
+    withFileTypes: true
+  })
+    .filter((dirent) => dirent.isDirectory())
+    .map((dirent) => dirent.name)
+
+  if (folders.length <= 0) {
```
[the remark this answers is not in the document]
Nocccer: It can't have -1 but we are not in C where you can define that your number is usigned, so it is better to check for negative values, too.

electron/gog/games.ts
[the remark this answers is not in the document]
CommandMC: `removeNonSteamGame` only uses the title, at least if my usage search isn't missing anything. Although I can agree with wanting to keep its arguments consistent compared to `addNonSteamGame`

electron/shortcuts/nonsteamgame.ts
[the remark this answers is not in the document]
CommandMC: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length)
> The value is an unsigned, 32-bit integer

electron/shortcuts/nonsteamgame.ts
```diff
@@ -0,0 +1,420 @@
...
+      })
+    }
+
+    if (errors.length === 0) {
+      logInfo(
+        `${props.gameInfo.title} was successfully added to Steam.`,
+        LogPrefix.Shortcuts
+      )
+
+      const message = i18next.t('notify.finished.add.steam.body1', {
```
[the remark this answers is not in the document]
Nocccer: The problem is that both have the finished ('success') value and both somehow where succesfully. In one case not all shortcuts.vdf could be adapted.

electron/shortcuts/nonsteamgame.ts
```diff
@@ -207,6 +196,7 @@ async function addNonSteamGame(props: {
       error,
       adding: true
     })
+    return
   } else {
```
CommandMC, line 200: Well now you have to actually get rid of the `else` to get rid of the indentation

electron/gog/games.ts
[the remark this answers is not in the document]
Nocccer: Ok yes removeNonSteamGame could only need the title. But you still need to get the gameInfo in the ipc_handler.ts to get the title. Does not really matter in my eyes if passing gameInfo or title.

electron/constants.ts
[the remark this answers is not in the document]
Nocccer: Yes exactly. It is the same as in Python. We should start using the node Path object instead of strings.
In Python i learned to use always Path type instead of strings.

CommandMC (approved):
Looks good, works great as well

One neat thing (that can totally be added in another PR) would be if we'd look at the game metadata (`<legendary config folder>/metadata/AppName.json`) and would add more images to Steam based on the ones listed there (`keyImages` key).  
`DieselGameBox` could be used for the wide-screen image, while `DieselGameBoxTall` could be the vertical one

Nocccer:
> Looks good, works great as well
> 
> One neat thing (that can totally be added in another PR) would be if we'd look at the game metadata (`<legendary config folder>/metadata/AppName.json`) and would add more images to Steam based on the ones listed there (`keyImages` key). `DieselGameBoxTall` could be used for the wide-screen image, while `DieselGameBoxTall` could be the vertical one

Added. Works great on my end. Can you check on your machine if you have time ?

Nocccer:
> Hm, not sure if it's supposed to be like this, but the vertical image is not working for me: image
> 
> Also, if possible, maybe hide the title text if the logo is already in the image (which I believe is always the case for the wide image?). This doesn't look _that_ great: image

Fixed the code so now vertical and hover picture is added. 
About the title i can't hide it. What we would need is a game banner from legendary and gog without a title in it.

imLinguin:
You may want to checkout static function exported by GOGLibrary class `getGamesdbData` to get background without a title.

Nocccer:
> getGamesdbData

Would be good to have the same functionality for gog and legendary else it will be a big if, else cluster.

imLinguin:
You can actually just get the data from that endpoint for both Epic and GOG (check function arguments and jsdoc for that function)

Nocccer:
To get the new background banner for gog games you need to reset game infos to fetch new art_banner.

Nocccer:
Added a little circle to frontend.
@arielj if you have time can you help me making the cirlce the size of parent ? I was really lost there.
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/add-none-steam-game/src/screens/Game/GameSubMenu/index.tsx#L202-L213

felipecrs:
image

Not sure if I'm doing something wrong. The game was installed by Epic Games but recognized by Heroic.

Nocccer:
> image
> 
> Not sure if I'm doing something wrong. The game was installed by Epic Games but recognized by Heroic.

You added a game without appName,
Can you give me your shortcuts.vdf under <path/to/steam>/userdata/<userid>/config/shortcuts.vdf

felipecrs:
[shortcuts.vdf.txt](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/files/9024626/shortcuts.vdf.txt)

There it is. Rename back from txt to vdf.

Nocccer:
Ok i got the mistake. appname is small laters. I check AppName!
Will fix this.
Can you delete this shortcuts.vdf and add fall Guys via heroic to steam ?

felipecrs:
Sure, I did, then it added it to Steam! :D

Unfortunately, running it does not work because the Heroic.exe seems to be located in a Temporary location because I'm using the Portable version. Every time I close the Heroic, it creates another location, then the shortcut stops working (I guess this issue would not happen when not using the portable version).

Nocccer:
> Sure, I did, then it added it to Steam! :D
> 
> Unfortunately, running it does not work because the Heroic.exe seems to be located in a Temporary location because I'm using the Portable version. Every time I close the Heroic, it creates another location, then the shortcut stops working (I guess this issue would not happen when not using the portable version).

Maybe, I don't run it on windows. Does this happens also if you close the heroic and start the game from steam ?

felipecrs:
But if I keep the Heroic open (thus with the temporary `.exe` still there) before I start the game, then it works fine!

> Does this happens also if you close the heroic and start the game from steam ?

That's specifically when it does not work. But don't worry about it, I'm sure this is because I'm using the portable version and not the installable one.

Another thing though, Steam does not seem to recognize the game as playing:

https://user-images.githubusercontent.com/29582865/176813160-0e04ccba-d41f-42ba-a074-1960a6732049.mp4

felipecrs:
> I know it does not work on flatpak if heroic is already running. So closing heroic and then start from stea works for me on linux.

Hm.. Maybe that would help with Steam recognizing the game as running.

Nocccer:
> But if I keep the Heroic open (thus with the temporary `.exe` still there) before I start the game, then it works fine!
> 
> > Does this happens also if you close the heroic and start the game from steam ?
> 
> That's specifically when it does not work. But don't worry about it, I'm sure this is because I'm using the portable version and not the installable one.
> 
> Another thing though, Steam does not seem to recognize the game as playing:
> 
>  steam_FnBPgMDfFk.mp4

With AppImage on Linux everything works. Windows ... 
The thought the windows portable works like appimage on linux.
I don't have a machine to test it at home...

Nocccer:
Btw i will also repair the shortcuts soon. Then you can launch games via Desktop icons.

felipecrs:
Oh, I found a way to bypass the Portable issue:

1. Run the Portable (it will extract the Heroic Games Launcher files to a temporary location and open it)
2. Go to that location (right click the Heroic icon, Properties)
3. Copy everything from there
4. Paste (replacing files) in the place where you have the normal Heroic installed

By doing this, I basically updated my installed Heroic to this new version manually.

---

So now I can launch games with Heroic closed. And yay, everything seems to work just fine this way! Steam recognizes the game as playing (and also when it stops):

https://user-images.githubusercontent.com/29582865/176815097-8690eb20-8401-4182-b571-aac907be2163.mp4

Thanks a million, you made my day! :D Now I can play Fall Guys on Steam Link in peace.

So, I guess it is worth noting that for the shortcut to function well, Heroic must not be open already. I wonder if there is anything that can be made to solve this more gracefully.

felipecrs:
Another small feedback: with Steam Big Picture mode, the horizontal image does not seem to match the aspect ratio that Steam expects.







I wonder if it is worth to programatically perform a horizontal cut or a stretch in the image provided by Epic to make it fit Steam's constraints.

Nocccer:
@felipecrs can you send me the property page of fall guys in steam? Should look like this:
image

I will cut or resize the image for big picture. For steam header it was a bit ugly. Example:
image

felipecrs:
This is why the Portable version does not work:

https://user-images.githubusercontent.com/29582865/176983843-987487eb-47b8-4d7d-891d-aa2b502af1a1.mp4

The shortcut which it creates on Steam points to Heroic.exe at, let's say, `C:\Users\felip\AppData\Local\Temp\2BMVnqaSmuJ9UGV88RpEWjLtPtM\Heroic.exe`. But it should actually point to `C:\Games\Heroic 2.4.0-beta.exe`, which is where I put my portable Heroic.

I tried to manually set the shortcut on Steam to `C:\Games\Heroic 2.4.0-beta.exe` and everything seems to work fine after doing so.

---

The property page:


https://user-images.githubusercontent.com/29582865/176984136-ba836d61-40ca-414e-9a43-023d440abd12.mp4

(I had to record a video because the whole page does not show in a single picture)

```
--no-gui --no-sandbox "heroic://launch/0a2d9f6403244d12969e11da6713137b"
```

---

Removing the `--no-sandbox` for me worked just fine as with it. Removing this option does not seem to cause any issue.

---

I tested the new version with the big picture image fixed:

image

It lookgs great!

felipecrs:
But I noticed you are stretching the image. Have you tried to cut/zoom it instead to see if it looks better, at least for your Control example?

Nocccer:
@felipecrs i found a enviroment variable wish should point now to the protable executable. I couldn't test because i don't have windows machine at home. Can you send me what it adds in steam on the property page ?

Added a button to remove games and images from steam. It is underneath the add button.

Zooming and croping does nothing, because steam does it already. Only resizing (which strech the image) is the best option so far.

felipecrs:
Testing it now.

Nocccer:
> Testing it now.

In case you couldn't find it
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/suites/7190996070/artifacts/287211254

felipecrs:
This new environment variable does not seem to work:

image

felipecrs:
I fixed it:

https://github.com/felipecrs/HeroicGamesLauncher/commit/f4525156c6c754208722720ce514bd1922e8e5ff

The following command copies the commit to your branch:

```console
git fetch https://github.com/felipecrs/HeroicGamesLauncher add-none-steam-game && git cherry-pick FETCH_HEAD
```

Nocccer:
> I fixed it:
> 
> [felipecrs@f452515](https://github.com/felipecrs/HeroicGamesLauncher/commit/f4525156c6c754208722720ce514bd1922e8e5ff)
> 
> The following command copies the commit to your branch:
> 
> ```
> git fetch https://github.com/felipecrs/HeroicGamesLauncher f4525156c6c754208722720ce514bd1922e8e5ff && git cherry-pick FETCH_HEAD
> ```
> 
> image

Thanks i wanted to try this env next. Do you know if it still works, if someone moves the portable executable and remove/add the game to steam again? I think it should work

felipecrs:
Yes, it works fine! I tried to rename it to felipe.exe, and removing it from steam (and adding it back) worked as expected.

But it does not work if I click on "Add to Steam" when the game is already there. It will continue with the old .exe name.

public/locales/pt_BR/translation.json
```diff
@@ -221,6 +232,20 @@
             "reparing": "Erro ao reparar"
         },
         "finished": {
+            "add": {
+                "steam": {
+                    "corrupt": "{{game}} could not be added to all found Steam users. See logs for more info.",
```
felipecrs, line 237:
```suggestion
                    "corrupt": "{{game}} não pôde ser adicionado a todos os usuários encontrados do Steam. Veja os logs para mais informações.",
```

public/locales/pt_BR/translation.json
```diff
@@ -221,6 +232,20 @@
...
+                    "success": "{{game}} was successfully added to Steam.",
```
felipecrs, line 238:
```suggestion
                    "success": "{{game}} foi adicionado com sucesso ao Steam.",
```

public/locales/pt_BR/translation.json
```diff
@@ -221,6 +232,20 @@
...
+                    "title": "Added to Steam"
```
felipecrs, line 239:
```suggestion
                    "title": "Adicionado ao Steam"
```

public/locales/pt_BR/translation.json
```diff
@@ -221,6 +232,20 @@
...
+                }
+            },
+            "remove": {
+                "steam": {
+                    "body1": "{{game}} was successfully removed to Steam.",
```
felipecrs, line 244:
```suggestion
                    "body1": "{{game}} foi removido com sucesso do Steam.",
```

public/locales/pt_BR/translation.json
```diff
@@ -221,6 +232,20 @@
...
+                    "body2": "{{game}}  could not be removed from all found Steam users. See logs for more info.",
```
felipecrs, line 245:
```suggestion
                    "body2": "{{game}} não pôde ser removido de todos os usuários encontrados do Steam. Veja os logs para mais informações.",
```

public/locales/pt_BR/translation.json
```diff
@@ -221,6 +232,20 @@
...
+                    "title": "Removed from Steam"
```
felipecrs, line 246:
```suggestion
                    "title": "Removido do Steam"
```

public/locales/en/gamepage.json
```diff
@@ -123,10 +123,12 @@
     },
     "submenu": {
         "addShortcut": "Add shortcut",
+        "addToSteam": "Add to steam",
```
felipecrs, line 126:
```suggestion
        "addToSteam": "Add to Steam",
```

But I think it should be changed in all files.

public/locales/en/translation.json
```diff
@@ -221,6 +232,20 @@
...
+                "steam": {
+                    "corrupt": "{{game}} could not be added to all found Steam users. See logs for more info.",
+                    "success": "{{game}} was successfully added to Steam.",
+                    "title": "Added to Steam"
+                }
+            },
+            "remove": {
+                "steam": {
+                    "body1": "{{game}} was successfully removed to Steam.",
+                    "body2": "{{game}}  could not be removed from all found Steam users. See logs for more info.",
```
felipecrs, line 245:
```suggestion
                    "body2": "{{game}} could not be removed from all found Steam users. See logs for more info.",
```

I think it should be fixed in all files.

felipecrs (changes requested):

public/locales/pt_BR/gamepage.json
```diff
@@ -123,10 +123,12 @@
     },
     "submenu": {
         "addShortcut": "Adicionar Atalho",
+        "addToSteam": "Add to steam",
```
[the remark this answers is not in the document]
Nocccer: Translation via webplate please:
https://hosted.weblate.org/projects/heroic-games-launcher/

felipecrs:
BTW there is a third-party library for getting the Steam installation path which works on Windows:

https://www.npmjs.com/package/steam-game-path

On Windows, the Steam installation path can be reliably found by looking at the registry keys:

https://stackoverflow.com/a/34091380/12156188

But probably it's not worth to implement.

Nocccer:
> implemen

It is actually worth.

Nocccer:
@CommandMC @imLinguin @arielj @flavioislima a final review would be good.
I think i covered alot now and it is pretty stable.

electron/constants.ts
```diff
@@ -112,7 +111,16 @@ function fixAsarPath(origin: string): string {
 export function getSteamCompatFolder() {
   // Paths are from https://savelocation.net/steam-game-folder
   if (isWindows) {
-    return normalize(join('C:/Program Files (x86)/Steam'))
+    const defaultWinPath = normalize('C:/Program Files (x86)/Steam')
```
felipecrs, line 114: You can also use:

```js
// 64-bit systems
process.env["PROGRAMFILES(X86)"] // 'C:\\Program Files (x86)'

// 32-bit
process.env["PROGRAMFILES"] // 'C:\\Program Files'
```

flavioislima:
I will try to take a look at it tomorrow.

electron/constants.ts
```diff
@@ -107,15 +107,41 @@ function fixAsarPath(origin: string): string {
   return origin
 }
 
-function getSteamCompatFolder() {
-  if (existsSync(`${userHome}/.var/app/com.valvesoftware.Steam/.steam/steam`)) {
-    return `${userHome}/.var/app/com.valvesoftware.Steam/.steam/steam`
+export function getSteamCompatFolder() {
+  // Paths are from https://savelocation.net/steam-game-folder
+  if (isWindows) {
+    const defaultWinPath = join('C:', process.env['PROGRAMFILES(X86)'], 'Steam')
```
felipecrs, line 113:
```suggestion
    const defaultWinPath = join(process.env['PROGRAMFILES(X86)'], 'Steam')
```

src/screens/Game/GameSubMenu/index.tsx
```diff
@@ -112,6 +115,26 @@ export default function GamesSubmenu({
...
+    await ipcRenderer.invoke('removeFromSteam', appName, runner)
+    isAddedToSteam()
+    setSteamRefresh(false)
+  }
+
+  const isAddedToSteam = () => {
+    ipcRenderer.invoke('isAddedToSteam', appName, runner).then((added) => {
+      setAddedToSteam(added)
+    })
+  }
```
arielj, line 136: Maybe do:

```ts
  async function isAddedToSteam() {
    const added = await ipcRenderer.invoke('isAddedToSteam', appName, runner)
    setAddedToSteam(added)
  }
```

for consistency with the previous 2 functions?

Nocccer: Good point. I rewrote everything in frontend to a more consistent and clean way

CommandMC:
One last small nitpick:
I assume there's no way to fix the aspect ration/stretching issue on the game's page...
image
...without breaking the widescreen image here?
image

felipecrs:
Hm, I just thought on something: how about adding a continuation blur on both sides rather than stretching the image? Not sure if there is a npm library for that, and not sure if it would look good.

felipecrs:
Perhaps we should report this issue to Steam. The same way they handle the aspect ratio mismatch in the library page (they zoom in the image), they should do it for the big picture mode.

Otherwise, we have to manually fix the image, which would cause either one to not look as good.

Nocccer:
> One last small nitpick: I assume there's no way to fix the aspect ration/stretching issue on the game's page... image ...without breaking the widescreen image here? image

I tried everything, but the best solution so far is to resize it for the header, else for your example the 20XX would be half cutoff.
Let's think about a solution in another PR. We could for example fetch data from SteamGridDB, if the game title matches a entry there.

Nocccer:
> Perhaps we should report this issue to Steam. The same way they handle the aspect ratio mismatch in the library page (they zoom in the image), they should do it for the big picture mode.
> 
> Otherwise, we have to manually fix the image, which would cause either one to not look as good.

I don't know if steam will fix this. If you open a ticket you should not mark heroic. Just tell that adding an image for a manually added game, does not fit well. But i think they will point you to SteamGridDB.

felipecrs:
> If you open a ticket you should not mark heroic

I can reference this discussion to make a better case still, right?

Nocccer:
> Lol. Got it. Well, I think they are open-minded, but ok.

Yes, but i think without talking about epic, maybe it is faster fixed.

Nocccer:
> One last small nitpick: I assume there's no way to fix the aspect ration/stretching issue on the game's page... image ...without breaking the widescreen image here? image

@felipecrs i added steamgriddb images if provided. I just look for matching game name and for matching sizes. 
The only think we need is a API key from steagriddb. This key can be greated if logged in with a steam account. Actually i don't want to provide my private steam acc for that. If you have a test steam acc, would be appreciated if you can create the API key for us. Else i need to create a test steam account.

Nocccer:
> @Nocccer I just got the API key using one of my test accounts. Do you still want me to send? (I noticed you reverted the commit)

Yes because the owner of the API said we can use it only if the user provide there API keys.
One of our devs will write a fucntion to create a fitting image with blurred sides.

Nocccer:
> Okay, got it. I kind of prefer the images from Epic as well, for consistency.
> 
> And I agree, if the user had to provide their API keys, it would clutter the UI. We may consider implement this as a feature in the future, if users ask for it.

@arielj got it to work. Images are now blurred at the side.

flavioislima (approved):
Pretty good stuff! ⚔️

felipecrs:
The right one is from Steam, the left one is the one from GOG added through Heroic. Astonishing, this is such a nice feature!

felipecrs:
@Nocccer I'm trying this out on Ubuntu 22.04, with Steam installed through .deb (because the Flatpak version warns about lack of `steam-devices`), and while Heroic says it added to Steam, it does not appear on Steam.

After some investigation, it seems that the correct dir is:

```
/home/felipecrs/.steam/debian-installation/userdata/<my-id>/config
```

Which I think Heroic isn't looking on.

Nocccer:
> @Nocccer I'm trying this out on Ubuntu 22.04, with Steam installed through .deb (because the Flatpak version warns about lack of `steam-devices`), and while Heroic says it added to Steam, it does not appear on Steam.
> 
> After some investigation, it seems that the correct dir is:
> 
> ```
> /home/felipecrs/.steam/debian-installation/userdata/<my-id>/config
> ```
> 
> Which I think Heroic isn't looking on.

You can set the folder in general settings > others if i remember correctly.

felipecrs:
Ok... I'm just not sure what I should write there. I tried `/home/felipecrs/.steam/debian-installation` and it didn't work.

felipecrs:
Ok, the issue is, no matter what I set in the Steam path, it always revert to how it was before (i.e. my setting does not apply).

felipecrs:
Editing the config.json manually worked for me. I wrote:

/home/felipecrs/.steam/debian-installation

Nocccer:
> Ok, the issue is, no matter what I set in the Steam path, it always revert to how it was before (i.e. my setting does not apply).

ok will fix this. But actually i think the default path is correct, because on Pop_OS! it is also this path `/home/niklas/.steam/steam/userdata`

felipecrs:
But I would advise you to check this folder as well. It is probably the most common folder where Steam could be found, considering that Ubuntu is one of the main used distros over there.

felipecrs:
> > Ok, the issue is, no matter what I set in the Steam path, it always revert to how it was before (i.e. my setting does not apply).
> 
> ok will fix this. But actually i think the default path is correct, because on Pop_OS! it is also this path `/home/niklas/.steam/steam/userdata`

For me this is what was in the config.json:

/home/felipecrs/.var/app/com.valvesoftware.Steam/.steam/steam

Could it be because I previously had the Steam Flatpak installed? Perhaps the Steam install dir should be re-evaluated on every start?

Nocccer:
> But I would advise you to check this folder as well. It is probably the most common folder where Steam could be found, considering that Ubuntu is one of the main used distros over there.

POP_OS is based on ubuntu. And if you press on debian-installation it is just a shortcut to `/home/niklas/.steam/steam`

Nocccer:
> > > Ok, the issue is, no matter what I set in the Steam path, it always revert to how it was before (i.e. my setting does not apply).
> > 
> > 
> > ok will fix this. But actually i think the default path is correct, because on Pop_OS! it is also this path `/home/niklas/.steam/steam/userdata`
> 
> For me this is what was in the config.json:
> 
> /home/felipecrs/.var/app/com.valvesoftware.Steam/.steam/steam
> 
> Could it be because I previously had the Steam Flatpak installed? Perhaps the Steam install dir should be re-evaluated on every start?

delete steam flatpak if you don't want to use it. Else add-to-steam will add it to steam flatpak.

felipecrs:
For me it does not seem to be a shortcut:

```
 ls -la .steam
total 24
drwxrwxr-x  3 felipecrs felipecrs 4096 jul 21 13:55 .
drwxr-x--- 33 felipecrs felipecrs 4096 jul 21 13:58 ..
lrwxrwxrwx  1 felipecrs felipecrs   28 jul 21 13:55 bin -> /home/felipecrs/.steam/bin32
lrwxrwxrwx  1 felipecrs felipecrs   54 jul 21 13:55 bin32 -> /home/felipecrs/.steam/debian-installation/ubuntu12_32
lrwxrwxrwx  1 felipecrs felipecrs   54 jul 21 13:55 bin64 -> /home/felipecrs/.steam/debian-installation/ubuntu12_64
drwxrwxr-x 26 felipecrs felipecrs 4096 jul 21 13:57 debian-installation
-rwxrwxr-x  1 felipecrs felipecrs 1149 jul 21 13:55 registry.vdf
lrwxrwxrwx  1 felipecrs felipecrs   42 jul 21 13:55 root -> /home/felipecrs/.steam/debian-installation
lrwxrwxrwx  1 felipecrs felipecrs   50 jul 21 13:55 sdk32 -> /home/felipecrs/.steam/debian-installation/linux32
lrwxrwxrwx  1 felipecrs felipecrs   50 jul 21 13:55 sdk64 -> /home/felipecrs/.steam/debian-installation/linux64
lrwxrwxrwx  1 felipecrs felipecrs   42 jul 21 13:55 steam -> /home/felipecrs/.steam/debian-installation
-rw-rw-r--  1 felipecrs felipecrs    6 jul 21 13:55 steam.pid
prw-------  1 felipecrs felipecrs    0 jul 21 13:54 steam.pipe
-r--------  1 felipecrs felipecrs   16 jul 21 13:55 steam.token
```

I think PopOS packages their own Steam deb.

felipecrs:
> delete steam flatpak if you don't want to use it. Else add-to-steam will add it to steam flatpak.

Of course I did it... still, Heroic was looking on .var.

Nocccer:
Please first delete steam-flatpak and his folders. There are  folders in .var left still for steam flatpak.

felipecrs:
Got it. I didn't delete the folder indeed.

Nocccer:
> Got it. I didn't delete the folder indeed.

Not the default should work. Because i assume Ubuntu - steam is different then Pop_OS - steam installation

Nocccer:
> Ok, I confirmed that `/home/felipecrs/.steam/steam` also works.
> 
> But I had to write it manually in the config.json, even after deleting the Flatpak folder in .var.

Btw better create a github issue instead writing here on a closed pr.

PR #1335, opened 2022-05-18
[Fix] Check for game updates before launch

flavioislima:
This fixes the issue we are having now where it is not possible to skip an update when launching a game.
I changed the logic to check the update on frontend and show the message before even launching the game instead of waiting for an error from Legendary.

---

### DIFF
--- a/electron/legendary/games.ts
+++ b/electron/legendary/games.ts
@@ -522,8 +522,6 @@ class LegendaryGame extends Game {
   }
 
   public async launch(launchArguments: string): Promise<LaunchResult> {
-    console.log('launch', { launchArguments })
-
     const gameSettings =
       GameConfig.get(this.appName).config ||
       (await GameConfig.get(this.appName).getSettings())
--- a/electron/legendary/games.ts
+++ b/electron/legendary/games.ts
@@ -521,7 +521,9 @@ class LegendaryGame extends Game {
     return res
   }
 
-  public async launch(launchArguments = ''): Promise<LaunchResult> {
+  public async launch(launchArguments: string): Promise<LaunchResult> {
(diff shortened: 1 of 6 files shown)


src/helpers/library.ts
```diff
@@ -249,43 +249,45 @@ type LaunchOptions = {
...
+    const { response } = await ipcRenderer.invoke('openMessageBox', args)
+
+    if (response === 0) {
+      return updateGame(appName, runner)
+    }
+
+    return ipcRenderer.invoke('launch', {
+      appName,
+      runner,
+      launchArguments: runner === 'legendary' ? '--skip-version-check' : ''
```
CommandMC, line 278: I think this should be done in `launch` directly instead of appending it here

electron/legendary/games.ts
```diff
@@ -521,7 +521,9 @@ class LegendaryGame extends Game {
     return res
   }
 
-  public async launch(launchArguments = ''): Promise<LaunchResult> {
+  public async launch(launchArguments: string): Promise<LaunchResult> {
+    console.log('launch', { launchArguments })
+
```
CommandMC, line 526:
```suggestion
```

CommandMC (changes requested):

src/helpers/library.ts
flavioislima: you mean the check? because the backend needs to receive the argument from here so then it can pass it to legendary.
Honestly, I just don't know how GOGDL react to this cli, otherwise we could pass it to both.

flavioislima: Ok, gogdl ignores the flag so I can remove that check

CommandMC (approved):

PR #1336, opened 2022-05-18
[UI] Show if there is enough space available on Install

flavioislima:
I fixed the error handler on the backend but the main check now is on the frontend.

I will check if there enough space and if the path is writable, if not, then the Install Button will be disabled.

Linux/Mac
image
image
image

Windows:
image
image

---


electron/main.ts
```diff
@@ -461,9 +461,20 @@ ipcMain.handle('kill', async (event, appName, runner) => {
   return Game.get(appName, runner).stop()
 })
 
-ipcMain.handle('checkDiskSpace', async (event, folder) => {
-  const { free, size: diskSize } = await checkDiskSpace(folder)
-  return `${getFileSize(free)}/${getFileSize(diskSize)}`
+ipcMain.handle('checkDiskSpace', async (event, folder: string) => {
+  if (folder.startsWith('/')) {
```
arielj, line 465: just wondering, does this work on windows? looks like unix specific

src/screens/Library/components/InstallModal/index.tsx
```diff
@@ -237,18 +240,21 @@ export default function InstallModal({
           setInstallPath(config.defaultInstallPath)
         }
         if (installPath !== 'default') {
```
arielj, line 242: maybe we should do this check also if installPath is the default?

flavioislima: Tbh checking only on the backend should do it.

electron/main.ts
flavioislima: yes, won't work. I will figure out another way of checking that.

src/screens/Library/components/InstallModal/index.tsx
flavioislima: And actually, we are changing the install path just before checking the space so that would never trigger.

electron/main.ts
flavioislima: So actualy the function `checkDiskSpace` checks if the path is valid based on the platform so we dont need to do it ourselves.
So I just changed it to a try catch now.

PR #1346, opened 2022-05-20
[Fix] Check only for remaining download size when continuing the download

### LINKED ISSUE #1344: Resuming downloads checks for whole game size not remaining
### Describe the bug

When resuming a download the launcher checks for sufficient disk space, but does so with the whole game size instead of the remaining.

### Add logs

Irrelevant

### Steps to reproduce

1. Download a game for a bit, so that your remaining disk space is less than the size of the whole game.
2. Stop the download (and keep the files)
3. Resume the download

### Expected behavior

Check for the remaining size (maybe by calculating it with the remaining percentage). 

### Screenshots

grafik

### System Information

- OS: Pop!_OS 
- Heroic Version: 2.3.2

### Additional information

_No response_

flavioislima:
This fixes #1344 
Also for the Download size, shows only the remaining size as well.

image

---

### DIFF
--- a/src/screens/Library/components/InstallModal/index.tsx
+++ b/src/screens/Library/components/InstallModal/index.tsx
@@ -256,10 +256,22 @@ export default function InstallModal({
           installPath
         )
         if (gameInstallInfo?.manifest?.disk_size) {
-          const notEnoughDiskSpace = free < gameInstallInfo.manifest.disk_size
-          const spaceLeftAfter = size(
+          let notEnoughDiskSpace = free < gameInstallInfo.manifest.disk_size
+          let spaceLeftAfter = size(
             free - Number(gameInstallInfo.manifest.disk_size)
           )
+          if (previousProgress.folder === installPath) {
+            const progress = 100 - getProgress(previousProgress)
+            notEnoughDiskSpace =
+              free <
+              (progress / 100) * Number(gameInstallInfo.manifest.disk_size)
+
+            spaceLeftAfter = size(
+              free -
+                (progress / 100) * Number(gameInstallInfo.manifest.disk_size)
+            )
+          }
+
           setSpaceLeft({
             message,
             notEnoughDiskSpace,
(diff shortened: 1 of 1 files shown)


electron/launcher.ts
```diff
@@ -281,12 +281,15 @@ async function verifyWinePrefix(
...
+  if (wineVersion.type === 'crossover') {
     return { res: { stdout: '', stderr: '' }, updated: false }
   }
 
   if (!existsSync(winePrefix)) {
     mkdirSync(winePrefix, { recursive: true })
+    if (wineVersion.type === 'proton') {
+      return { res: { stdout: '', stderr: '' }, updated: true }
+    }
   }
```
CommandMC, line 294:
```suggestion
if (wineVersion.type === 'proton') {
  return { res: { stdout: '', stderr: ''}, updated: false }
}
```
Otherwise, the `wineboot` command will be ran if the folder already exists

electron/main.ts
```diff
@@ -552,48 +553,73 @@ interface Tools {
...
+    'dist/bin/wine64'
+  )
+
+  const protonWinePath = existsSync(newProtonWinePath.replaceAll("'", ''))
+    ? newProtonWinePath
+    : oldProtonWinePath
+
+  const wineBin = isProton ? protonWinePath : wine
+
+  let winePrefix = prefix.replace('~', userHome)
```
CommandMC, line 577: Having a non-absolute prefix path causes problems elsewhere (`launch` doesn't do any checks for it, for example).

I'd say instead of having this `.replace` everywhere, we should make sure it's absolute when saving the settings and replace things like ~ there. But that can of course be a different PR

electron/main.ts
```diff
@@ -552,48 +553,73 @@ interface Tools {
...
       ...execOptions
     }
 
-    // existsSync is weird because it returns false always if the path has single-quotes in it
-    const protonWinePath = existsSync(newProtonWinePath.replaceAll("'", ''))
-      ? newProtonWinePath
-      : oldProtonWinePath
-    let wineBin = isProton ? `'${protonWinePath}'` : wine
-    let winePrefix = prefix.replace('~', userHome)
+    const wineBin = isProton ? `'${wine}' run` : wine
```
CommandMC, line 611: Ideally all of this should use `runWineCommand`, to avoid having this logic twice.

Problem with that is that we don't know the AppName of the game we're running this on. If we did, we could make this a lot simpler. Not sure how easy that is to implement in the frontend

electron/launcher.ts
flavioislima: ok

electron/main.ts
flavioislima: yes, we need to look into that. Maybe if we expose the userHome on frontend we could get the right folder.

electron/main.ts
flavioislima: shoulds be one more argument only, I will test a few things

electron/main.ts
flavioislima: Done

CommandMC (approved):

PR #1352, opened 2022-05-21
Run GOG setup again + don't skip Wineboot for Proton

CommandMC:
GOG Setup instructions were not ran due to a missing `await`
I've also reverted the changes about not running Wineboot since otherwise, a Proton prefix would only count as "updated" if the root folder was created by Heroic (so if the user deletes all files inside it, it would still not run the setup instructions)

---


imLinguin (approved):
LGTM

flavioislima (approved):
Looks good! 👍

PR #1355, opened 2022-05-22
Set Steam App ID variables when using the Runtime on Proton + fixup setting user-defined env vars on native games

CommandMC:
- When the Steam Runtime is used for Proton, `STEAM_COMPAT_APP_ID` and `SteamAppId` are set to 0. This avoids a ProtonFix crash for GE-Proton
- User-defined environment variables are now set in `setupEnvVars` (they were in `setupWineEnvVars` before). Not sure why I did it that way back then. Because of that, setting these variables on native games works again now.

---

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -266,8 +266,12 @@ function setupWineEnvVars(gameSettings: GameSettings, gameId = '0') {
     // This sets the name of the log file given when setting PROTON_LOG=1
     ret.SteamGameId = `heroic-${gameId}`
     ret.PROTON_LOG_DIR = flatPakHome
-    // Stop Proton from overriding WINEDEBUG; this prevents logs growing to a few GB for some games
-    ret.WINEDEBUG = 'timestamp'
+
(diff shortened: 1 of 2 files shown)


imLinguin (approved):

CommandMC:
With the latest commit, Proton's logging feature will now also work. Adding `PROTON_LOG=1` to the environment variables will now give you a `steam-0.log` file in your home directory with the usual Wine log
Maybe we should even override the log folder (`PROTON_LOG_DIR` env var) and then append the log to our own game log automatically?

CommandMC:
I'd say this is now feature-complete. I have not opted for automatically setting `PROTON_LOG=1` since the log can get quite big, and it's not necessary to have it 90% of the time  
If the user still needs the log, setting the variable and then navigating to the correct directory (`/home/$USER/` normally, `/home/$USER/.var/app/com.heroicgameslauncher.hgl/` for Flatpak) is fine

flavioislima (approved):
Looks nice 👍🏽

Issue #1320, opened 2022-05-16
Refresh GOG and Epic libraries individually


arielj:
### Problem description

Currently, refreshing the library for one of the stores triggers a refresh in both. If your intention is to refresh one library and the other library is slower (and maybe you already know there's nothing new), the refresh will take longer doing more work than needed waiting for the slower one to finish.

### Feature description

Clicking the Refresh icon in Library > GOG should only trigger a refresh in GOG. Clicking the Refresh icon in Library > Epic should only trigger a refresh in Epic. Same when logging in.

I think it's fine that both libraries are refreshed at the same time in the background when the app starts.

### Alternatives

Continue with the current refresh behavior. It doesn't generate any error and it works, it's just not as efficient.

PR #1356, opened 2022-05-22
Allow for separate library refreshes

connercsbn:
This patch, addressing https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1320, adds an additional string[] parameter to GlobalState.refresh and checkGameUpdates, allowing you to update just one library at a time. 

Launching still refreshes both libraries, but logging in or clicking the refresh button in a library only refreshes the corresponding library.

The way I did it, there's some repeated code. There's probably a better way of doing it, especially as more libraries are added, but for now this was the best I could do. Let me know if you have a better idea and I'll happily rewrite it.

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -597,9 +597,13 @@ ipcMain.handle(
 
 /// IPC handlers begin here.
 
-ipcMain.handle('checkGameUpdates', async () => {
-  const legendaryUpdates = await LegendaryLibrary.get().listUpdateableGames()
-  const gogUpdates = await GOGLibrary.get().listUpdateableGames()
+ipcMain.handle('checkGameUpdates', async (event, libraries) => {
+  const legendaryUpdates = libraries.includes('epic')
+    ? await LegendaryLibrary.get().listUpdateableGames()
+    : []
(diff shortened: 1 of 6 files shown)


electron/main.ts
```diff
@@ -597,9 +597,13 @@ ipcMain.handle(
 
 /// IPC handlers begin here.
 
-ipcMain.handle('checkGameUpdates', async () => {
-  const legendaryUpdates = await LegendaryLibrary.get().listUpdateableGames()
-  const gogUpdates = await GOGLibrary.get().listUpdateableGames()
+ipcMain.handle('checkGameUpdates', async (event, libraries) => {
```
flavioislima, line 600: I believe you should receive a `Runner` here and then depending on the Runner you call the right one.
Use a `switch` to make it more scalable. :)

flavioislima: could also have a full parameter to refresh all libraries

flavioislima (approved):
Looks pretty good. Thanks for that! ⚔️ 
tested and is everything working fine.

PR #1358, opened 2022-05-22
Allow specifying wrappers in the env var section again

CommandMC:
With this patch, you can now enter wrappers into the environment variable section of the game settings again & they will be properly executed instead of being added as empty variables

---


CommandMC:
I've tested it with gamescope now & just by adding `echo` before, works fine. People on the issue are also reporting that it's working. So I guess I'll merge?

flavioislima:
Yes, ready to merge! :D

PR #1369, opened 2022-05-24
[Linux] Search for Proton on all steam libraries

flavioislima:
Added a helper function to read from `libraryfolder.vdf` and add the libraries to the `steampaths` 

---

### DIFF
--- a/electron/config.ts
+++ b/electron/config.ts
@@ -20,7 +20,8 @@ import {
   userHome,
   isFlatpak,
   isMac,
-  isWindows
+  isWindows,
+  getSteamLibraries
 } from './constants'
 import { execAsync } from './utils'
 import { logError, logInfo, LogPrefix } from './logger/logger'
@@ -239,15 +240,14 @@ abstract class GlobalConfig {
       })
     }
 
+    const steamLibraries = getSteamLibraries()
     const protonPaths = [`${heroicToolsPath}/proton/`]
 
     // Known places where Steam might be found.
     // Just add a new string here in case another path is found on another distro.
-    const steamPaths = [
-      join(userHome, '.steam'),
-      join(userHome, '.var/app/com.valvesoftware.Steam/.local/share/Steam'),
-      '/usr/share/steam'
-    ].filter((path) => existsSync(path))
+    const steamPaths = [...steamLibraries, '/usr/share/steam'].filter((path) =>
+      existsSync(path)
+    )
 
     steamPaths.forEach((path) => {
       protonPaths.push(`${path}/steam/steamapps/common`)
@@ -267,12 +267,15 @@ abstract class GlobalConfig {
             name.startsWith('proton') || name.startsWith('ge-proton')
           if (hasProtonName && !name.includes('runtime')) {
             const protonBin = join(path, version, 'proton')
-            proton.add({
-              bin: protonBin,
-              name: `Proton - ${version}`,
-              type: 'proton'
-              // No need to run this.getWineExecs here since Proton ships neither Wineboot nor Wineserver
-            })
+            // check if bin exists to avoid false positives
+            if (existsSync(protonBin)) {
+              proton.add({
+                bin: protonBin,
+                name: `Proton - ${version}`,
+                type: 'proton'
+                // No need to run this.getWineExecs here since Proton ships neither Wineboot nor Wineserver
+              })
+            }
           }
         })
       }
--- a/electron/config.ts
+++ b/electron/config.ts
@@ -240,16 +240,9 @@ abstract class GlobalConfig {
       })
     }
 
-    const steamLibraries = getSteamLibraries()
     const protonPaths = [`${heroicToolsPath}/proton/`]
 
-    // Known places where Steam might be found.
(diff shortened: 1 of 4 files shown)


electron/config.ts
```diff
@@ -239,15 +240,14 @@ abstract class GlobalConfig {
...
     const protonPaths = [`${heroicToolsPath}/proton/`]
 
     // Known places where Steam might be found.
     // Just add a new string here in case another path is found on another distro.
-    const steamPaths = [
-      join(userHome, '.steam'),
-      join(userHome, '.var/app/com.valvesoftware.Steam/.local/share/Steam'),
-      '/usr/share/steam'
-    ].filter((path) => existsSync(path))
+    const steamPaths = [...steamLibraries, '/usr/share/steam'].filter((path) =>
```
CommandMC, line 248: Why not just call `getSteamLibraries()` directly here?

electron/constants.ts
```diff
@@ -109,6 +114,18 @@ function getSteamCompatFolder() {
...
+export function getSteamLibraries() {
+  const vdfFile = join(steamCompatFolder, 'steamapps', 'libraryfolders.vdf')
+  if (existsSync(vdfFile)) {
+    const json = parse(readFileSync(vdfFile, 'utf-8'))
+    const folders = Object.values(json.libraryfolders) as Array<{
+      path: string
+    }>
+    return folders.map((folder) => folder.path)
+  }
+  logDebug('No other steam libraries found', LogPrefix.Backend)
```
CommandMC, line 126: This log could be clearer, something like `Unable to load Steam Libraries, libraryfolders.vdf not found`

CommandMC, line 126: Might want to return an empty list if nothing is found & then specify the return type of the function

electron/config.ts
flavioislima: Yes, still doesn't make any difference in practice :P

CommandMC: Well, one less variable to worry about

flavioislima: Did a refactor to put most of the work on the helper function now, check if it looks better.

electron/config.ts
```diff
@@ -267,12 +267,15 @@ abstract class GlobalConfig {
...
           if (hasProtonName && !name.includes('runtime')) {
             const protonBin = join(path, version, 'proton')
-            proton.add({
-              bin: protonBin,
-              name: `Proton - ${version}`,
-              type: 'proton'
-              // No need to run this.getWineExecs here since Proton ships neither Wineboot nor Wineserver
-            })
+            // check if bin exists to avoid false positives
+            if (existsSync(protonBin)) {
```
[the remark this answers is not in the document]
flavioislima: done

electron/constants.ts
flavioislima: done

flavioislima: done

electron/config.ts
```diff
@@ -240,16 +240,9 @@ abstract class GlobalConfig {
...
 
-    // Known places where Steam might be found.
-    // Just add a new string here in case another path is found on another distro.
-    const steamPaths = [...steamLibraries, '/usr/share/steam'].filter((path) =>
-      existsSync(path)
-    )
-
-    steamPaths.forEach((path) => {
+    getSteamLibraries().forEach((path) => {
       protonPaths.push(`${path}/steam/steamapps/common`)
```
CommandMC, line 246: I think these paths can be updated now. In my case (with two Steam Library folders), I have the following:
- Main folder at `~/.local/share/Steam/`, contains a `compatibilitytools.d` folder for custom Proton and `steamapps/common` for installations
- 2nd one on external drive with just `steamapps/common`

So the first and 3rd path can be dropped

electron/constants.ts
```diff
@@ -114,16 +114,24 @@ function getSteamCompatFolder() {
   return `${userHome}/.steam/steam`
 }
 
-export function getSteamLibraries() {
+export function getSteamLibraries(): string[] {
   const vdfFile = join(steamCompatFolder, 'steamapps', 'libraryfolders.vdf')
+  const libraries = ['/usr/share/steam']
```
CommandMC, line 119: Under which conditions does this `/usr/share/steam` library get created? It doesn't exist for me

CommandMC (approved):
Looks good in general, left some comments but they're not crucial

electron/constants.ts
flavioislima: I believe if you install proton-ge from AUR.

electron/config.ts
flavioislima: But do you remember that the steam folder on ubuntu and derivatives are not the same as Arch? So I think I will keep as it is, for now, to avoid breaking some distro. Its better to be wrong putting too much than too little :P

CommandMC: The Steam folder is in another location, that's true. But the library folders point to the same thing (at least they do on my PopOS testing VM, and I'd be *very* surprised if they would store those differently depending on the distro)
But you're right, doesn't really matter

Issue #1153, opened 2022-03-23
[General UI/UX] Update Login Screen with new Designs


flavioislima:
### Problem description

Right now we have a login screen that suffices for two stores, but as soon as Heroic starts adding support for more stores, would be important to have a design that could fit them.

### Feature description

We should have the design as on the screenshots below (Amazon Games is an example but could be other stores as well):

![image](https://user-images.githubusercontent.com/26871415/159680300-c02418f0-a0cc-4a62-b137-089d4d2d4ca5.png)
![image](https://user-images.githubusercontent.com/26871415/159680348-9161b16f-bbc4-40fb-b7a6-a31f34834bf0.png)

tildemarky:
I'd love to work on this if it's still available!

flavioislima:
@tildemarky 
Yes, no one is working on it. Feel free to grab it 👍🏽

flavioislima:
@tildemarky you can check for the design specs here
https://www.figma.com/file/Zv89aBtLJ8yN4BwFS0AwNb/Heroic-Games?node-id=0%3A1

I forgot to share before, sorry.

tildemarky:
> @tildemarky you can check for the design specs here https://www.figma.com/file/Zv89aBtLJ8yN4BwFS0AwNb/Heroic-Games?node-id=0%3A1
> 
> I forgot to share before, sorry.

No problem! Thank you that is super useful 🙌

flavioislima:
done

PR #1376, opened 2022-05-26
[UI] Update login with new designs

tildemarky:
Added the new vertical design to the login screen without adding additional components, leaving spacing for additional stores to be added as their own Runners. 

Open to making more design changes to more closely fit the design image posted in the reference issue #1153 

Screen Shot 2022-05-26 at 3 50 04 PM

---


flavioislima:
Thanks for the contribution! 👍🏽 
It would be better to stick to the design on that issue because, as it looks on the screenshot, it's not much of a change.
That design was made to support more stores in a optimized space.

tildemarky:
> That design was made to support more stores in a optimized space.

Good point! I'll make some revisions so that it looks closer to the original design

tildemarky:
Updated screenshot with current changes: 
Screen Shot 2022-05-27 at 1 28 20 PM

tildemarky:
Changed the styling to include the relevant feedback from the team. It's ready for a bunch of logins to be added as well. 

The language selector is mostly left unstyled because it's also connected to the settings page, maybe that should be the next phase?

Screen Shot 2022-05-29 at 10 01 20 AM

flavioislima:
@tildemarky are you still gonna work on this one? I think you have a few conflicts and the design might need some adjustments but it is pretty close. Sorry for forgetting about this one as well.

@biliesilva can you take a look at this latest screenshot and see what needs to be improved for the final design once you have the time?

biliesilva:
I suggest using these spaces to give more breath to the elements:
image

And use these paddings and gaps here to the language selector: 
image

tildemarky:
> @tildemarky are you still gonna work on this one? I think you have a few conflicts and the design might need some adjustments but it is pretty close. Sorry for forgetting about this one as well.

Hey sorry for the late reply, things got very hectic. I'll push an update with suggested changes within the coming week if nobody else has already stepped forward with this or a different concept.

flavioislima:
> > @tildemarky are you still gonna work on this one? I think you have a few conflicts and the design might need some adjustments but it is pretty close. Sorry for forgetting about this one as well.
> 
> Hey sorry for the late reply @flavioislima, things got very hectic. I'll push an update with suggested changes within the coming week if nobody else has already stepped forward with this or a different concept.

no worries. 
I believe no one is working on it, feel free to continue :)

tildemarky:
Still working on conflicts 😅

tildemarky:
Screenshot 2022-10-15 at 6 11 29 PM

The current look with the suggested changes! The language selector is mostly left as is, because the design changes would probably require changing the strings to not include the flag so it can be separately styled away from the language label, so that should maybe be its own PR. Not entirely sure. 

Let me know if anything else should be adjusted @flavioislima & @biliesilva

biliesilva:
I have some suggestions below, let me know if you need more examples:

image

tildemarky:
> I have some suggestions below, let me know if you need more examples:

That example helped a lot thank you! Here's what it looks like with those changes

Screenshot 2022-10-15 at 7 16 58 PM

biliesilva (approved):

arielj:
@tildemarky can you use `Epic  Alternative Login Method` instead of `Epic Games External Login` for that button? we discussed this on discord and we think it's more clear to the users than `External Login`.

arielj:
The library button looks wrong when logged in:
image

src/frontend/screens/Login/components/Runner/index.tsx
```diff
@@ -26,40 +26,54 @@ export default function Runner(props: RunnerProps) {
...
           </div>
         )}
+        <div className="runnerButtons">
+          {!props.isLoggedIn ? (
+            <Link to={props.loginUrl}>
+              <div className="runnerLogin">{`${
+                props.class === 'epic'
+                  ? props.class + ' Games Login'
+                  : props.class + ' Login'
+              }`}</div>
```
arielj, line 47: let's make this a more explicit if/else, this is not translation friendly:

```
let buttonText = ''

if (props.class === 'epic') {
  buttonText = t('epic_login', 'Epic Games Login')
} else {
  buttonText = t('gog_login', 'GOG Login')
}
```

And then `<Link to...>{buttonText}</Link>`

Then running `yarn i18next` will update the locale files.

src/frontend/screens/Login/components/Runner/index.tsx
```diff
@@ -26,40 +26,54 @@ export default function Runner(props: RunnerProps) {
...
+            <span>
+              {String(props.user).slice(0, maxNameLength) +
+                (String(props.user).length > maxNameLength ? '...' : '')}
+            </span>
           </div>
         )}
+        <div className="runnerButtons">
+          {!props.isLoggedIn ? (
+            <Link to={props.loginUrl}>
+              <div className="runnerLogin">{`${
```
arielj, line 43: we don't really need this extra div here, we can apply the same style to the `.runnerButtons a` selector instead and we can remove the div

src/frontend/screens/Login/index.tsx
```diff
@@ -47,33 +53,47 @@ export default function NewLogin() {
...
+              <LanguageSelector
+                flagPossition={FlagPosition.PREPEND}
+                showWeblateLink={true}
+              />
+            )}
+          </div>
+
+          <p className="runnerMessage">{loginMessage}</p>
+
+          <div className="runnerGroup">
```
arielj, line 74: We don't need  this `runnerGroup` wrapper div, having the runners directly after the `p` works fine (they just need `margin-bottom: var(--login-padding)`).

src/frontend/screens/Login/index.css
```diff
@@ -16,61 +18,186 @@
...
+  background-color: var(--background);
+  padding: var(--wrapper-padding);
+  border-radius: 10px;
+  margin-top: 20px;
+  min-width: 300px;
+}
+
+.languageSelector {
+  position: absolute;
+  top: -75px;
```
arielj, line 49: maybe use `bottom: calc(100% + 10px)` here? with -75px is easy for changes to create issues, this `help translate Heroic` was fixed (it was missing) and it overlaps now for example:
image

src/frontend/screens/Login/index.tsx
arielj: hmmm checking the CSS I think this is added here for the future if we add more stores?

src/frontend/screens/Login/components/Runner/index.tsx
tildemarky: Good call, I've just moved the classname to the Link and removed the added div

src/frontend/screens/Login/index.tsx
tildemarky: > hmmm checking the CSS I think this is added here for the future if we add more stores?

Yeah that's the goal, it should look something like pic attached (just duplicated gog login for demo). Although it can be removed for another PR down the line it's kind of not useful right now. 
Screenshot 2022-10-22 at 4 50 09 PM

src/frontend/screens/Login/index.css
tildemarky: Here's what that looks like with those fixes implemented: 
Screenshot 2022-10-22 at 5 18 07 PM

tildemarky:
Do I also commit the translation files once they're picked up? I've never used i18next before so the procedure is a bit new to me and I'm not sure if I'm the one to do that or if it gets automatically picked up during a later build.

arielj:
> Do I also commit the translation files once they're picked up? I've never used i18next before so the procedure is a bit new to me and I'm not sure if I'm the one to do that or if it gets automatically picked up during a later build.

Yes, commit the modified files. For more details check https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Translations#how-to-add-new-translatable-strings

src/frontend/screens/Login/index.css
```diff
@@ -16,61 +18,171 @@
...
-  background-color: var(--background-darker);
+  font-weight: 700;
+  color: var(--background);
+  background-color: var(--accent);
   border: none;
-  font-size: var(--text-xl);
-  padding: var(--space-md);
+  padding: 10px;
+  padding-left: 40px;
+  padding-right: 40px;
```
arielj, line 68: We can use the shorthand padding to set all of them at once:

```suggestion
  padding: 10px 40px;
```

src/frontend/screens/Login/index.css
```diff
@@ -16,61 +18,171 @@
...
+  border-radius: 10px;
+  font-size: var(--text-default);
   border-radius: var(--space-2xs);
   cursor: pointer;
   transition: 0.2s;
   width: fit-content;
   backdrop-filter: blur(var(--blur-light));
+  margin-left: auto;
+  margin-top: 20px;
+  margin-bottom: 10px;
```
arielj, line 78: we can use the shorthand here:
```
margin: 20px auto 10px 0px;
```

arielj (approved):
Looks good!

Issue #1375, opened 2022-05-26
[macOS] Games not installing in directory’s with spaces in them


duvbolone:
### Describe the bug

Every time I try to install a game it stops immediately installing, the install was trying to get performed on an external disk.

Renaming the disk so that it hasn't a space fixes the bug and I can install perfectly

### Add logs

**log from settings**

```
(19:43:15) DEBUG:   [Legendary]:       Running Legendary command: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary --version
(19:43:20) INFO:    [Legendary]:       Legendary location: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary
(19:43:20) INFO:    [Gog]:             GOGDL location: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/gogdl
(19:43:20) INFO:    [Backend]:         
  Heroic Version: 2.3.3 Brook
  Legendary Version:  0.20.25 Our Benefactors
  OS: macOS KERNEL: 21.3.0 ARCH: x64
  CPU: Apple M1 @2.4 
  RAM: Total: 8 GiB Available: 6.8 GiB
  GRAPHICS: GPU0: Apple M1 VRAM: nullMB DRIVER:  
  
  
(19:43:21) WARNING: [Backend]:         Protocol already registered.
(19:43:28) INFO:    [Frontend]:        Refreshing Library
(19:43:28) INFO:    [Legendary]:       Refreshing library...
(19:43:28) INFO:    [Legendary]:       Refreshing Epic Games...
(19:43:29) DEBUG:   [Legendary]:       Running Legendary command: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary list
(19:43:36) DEBUG:   [Legendary]:       Using cached install info
(19:43:36) DEBUG:   [Legendary]:       Using cached install info
(19:43:36) WARNING: [Backend]:         default is not writable
(19:43:37) WARNING: [Backend]:         /Users/duvbobruni/Games/Heroic is not writable
(19:43:37) INFO:    [Legendary]:       Updating game list
(19:43:37) INFO:    [Legendary]:       Game List Updated
(19:43:37) WARNING: [Backend]:         /Users/duvbobruni/Games/Heroic is not writable
[...]

PR #1379, opened 2022-05-27
[macOS & Linux] Wrap game install path in single quotes

redromnon:
This PR references #1375. The install path may have spaces in between and this leads to errors (the space splits the actual path into separate paths). Thus, it's better to wrap the path in single quotes. This only applies to Linux and macOS.

---

### DIFF
--- a/electron/legendary/games.ts
+++ b/electron/legendary/games.ts
@@ -392,6 +392,14 @@ class LegendaryGame extends Game {
       ...workers,
       '-y'
     ]
+
+    //Wrap the path in singles quotes for macOS and Linux to avoid errors due to spaces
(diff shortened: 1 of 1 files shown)


flavioislima:
This is not actually needed since we are using `spawn` to launch the process. The thing is that we found a bug on macOS that we need to investigate on why `spawn` is not behaving correctly there.

Also, even if that was the case we would need to fix that when launching the command and not `games.ts`, and that `platformToInstall` is the version of the game that will be downloaded, not actually the `platform` you are running heroic :)

redromnon:
I see. I misunderstood what `games.ts` meant. I'll close this PR then.

PR #1383, opened 2022-05-27
[FIX] Use ELECTRON_IS_DEV for dev environments with devtools

YakoYakoYokuYoku:
When using HGL as an unpackaged Electron application (much like `code` from Arch or `zulip-desktop-electron` from AUR) the dev environment is loaded by default and it fails to import `electron-devtools-installer` if a production build was used.
So instead of adding `electron-is-dev` again let's just add `ELECTRON_IS_DEV` with a default value of `0`. Otherwise, if the value is `1` then the dev environment is loaded as desired.

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -179,7 +179,8 @@ async function createWindow(): Promise<BrowserWindow> {
     handleExit(mainWindow)
   })
 
-  if (!app.isPackaged) {
+  const isDevEnvSet = (process.env?.ELECTRON_IS_DEV ?? '0') === '1'
+  if (isDevEnvSet && !app.isPackaged) {
     /* eslint-disable @typescript-eslint/ban-ts-comment */
     //@ts-ignore
     import('electron-devtools-installer').then((devtools) => {


electron/main.ts
```diff
@@ -179,7 +179,8 @@ async function createWindow(): Promise<BrowserWindow> {
     handleExit(mainWindow)
   })
 
-  if (!app.isPackaged) {
+  const isDevEnvSet = (process.env?.ELECTRON_IS_DEV ?? '0') === '1'
```
flavioislima, line 182: Maybe we could have a default env variable with the dev env disabled but we could enable it using an env var when calling Heroic, so we could enable the DEV env with a packaged app as well in case we want to ask someone to debug. Makes sense?

So we could remove the `isPackaged` and defined the var to be 0 by default. On the npm script `electron` and `start` we could pass it as 1. Then, after we package the app we can call:
` ELECTRON_IS_DEV=1 heroic` and have access to more debugging like the react dev tools and any other we want.

Maybe I am overthinking this 😅

YakoYakoYokuYoku:
On the contrary, I was thinking about that too and it doesn't seem to be a bad idea, so I'll do it.

electron/main.ts
YakoYakoYokuYoku: On the contrary, I was thinking about that too and it doesn't seem to be a bad idea, so I'll do it.

flavioislima (commented):
Looks good. Will make some tests when I have the time.

flavioislima:
In theory could work but actually in practice it tries to load localhost with the debug enabled, so if fails

image


So maybe get back to previous idea :)

arielj:
closing this, there's no sign from the author since May

PR #1385, opened 2022-05-27
[Linux - Fix] Winetricks on Flatpak

flavioislima:
Changed the `WINE` env var to use `PATH` as suggested by @CommandMC.

---

### DIFF
--- a/electron/tools.ts
+++ b/electron/tools.ts
@@ -210,7 +210,11 @@ export const Winetricks = {
 
     const { winePrefix, wineBin } = getWineFromProton(wine, isProton, prefix)
 
-    const command = `WINEPREFIX='${winePrefix}' WINE='${wineBin}' ${winetricks} -q`
+    // use wine instead of wine64 since it breaks on flatpak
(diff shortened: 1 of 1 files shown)


electron/tools.ts
```diff
@@ -210,7 +210,10 @@ export const Winetricks = {
 
     const { winePrefix, wineBin } = getWineFromProton(wine, isProton, prefix)
 
-    const command = `WINEPREFIX='${winePrefix}' WINE='${wineBin}' ${winetricks} -q`
+    const winepath = dirname(wineBin)
+
+    // use wine instead of wine64 since it breaks on flatpak
+    const command = `WINEPREFIX='${winePrefix}' PATH=$PATH:'${winepath}' ${winetricks} -q`
```
CommandMC, line 216:
```suggestion
    const command = `WINEPREFIX='${winePrefix}' PATH='${winepath}':$PATH ${winetricks} -q`
```
Otherwise Winetricks will find the system wine first and use that instead

flavioislima: That's true

CommandMC (approved):

PR #1386, opened 2022-05-27
Always update Winetricks if we are online

CommandMC:
Before, we were not downloading Winetricks if the file already existed. This resulted in very outdated Winetricks versions (for example, I was unknowingly running a version from Feb. 2021)

---

### DIFF
--- a/electron/tools.ts
+++ b/electron/tools.ts
@@ -192,7 +192,7 @@ export const Winetricks = {
     const path = `${heroicToolsPath}/winetricks`
     const downloadCommand = `curl -L ${url} -o ${path} --create-dirs`
 
-    if (existsSync(path)) {
+    if (!isOnline()) {
       return
     }
 


flavioislima:
Would not be better to call `winetricks  --self-update` when calling it? instead of updating all the time even if its the same version?
Because winetricks doesn't have releases too often afaik.

CommandMC:
It looks like --self-update does the same thing

flavioislima:
> It looks like --self-update does the same thing. It also just downloads the newest file and replaces the current script with it, even if it's already at the newest version https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L1258

Amazing...

flavioislima (approved):
⚔️

CommandMC:
In the end of the day, winetricks isn't even a MB big, I'd say downloading that on startup is fine

PR #1387, opened 2022-05-27
[Fix] Proper Check if a game is native or not

flavioislima:
Fix how we chack `isNative` on GOG and Legendary.
---


CommandMC:
The same check is also present for Legendary, could you fix it there as well?

Nocccer:
If it is needed for both a util/generic function would be good

flavioislima:
> The same check is also present for Legendary, could you fix it there as well?

I sent the fix but now I will need to test, because last time I checked `gameinfo.platform.install` was coming undefined from legendary.

CommandMC (approved):

imLinguin (approved):

PR #1389, opened 2022-05-27
[UX] Use better paths on open dialog

flavioislima:
This is a small quality of life improvement. 
When opening dialogs, the starting folder will be more specific to the setting instead of being generic and opening the recent files.

---

### DIFF
--- a/src/screens/Settings/components/Tools/index.tsx
+++ b/src/screens/Settings/components/Tools/index.tsx
@@ -5,6 +5,7 @@ import React, { useState } from 'react'
 import { IpcRenderer } from 'electron'
 import { WineInstallation } from 'src/types'
 import { useTranslation } from 'react-i18next'
+import { getGameInfo } from 'src/helpers'
 
 const { ipcRenderer } = window.require('electron') as {
   ipcRenderer: IpcRenderer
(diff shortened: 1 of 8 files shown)


CommandMC (commented):
In general, a great PR. ~~Left a comment about the one change I don't like that much~~ Huh, seems like that didn't save. One sec

src/screens/Settings/components/Tools/index.tsx
```diff
@@ -43,10 +44,12 @@ export default function Tools({ wineVersion, winePrefix, appName }: Props) {
 
   const handleRunExe = async () => {
     let exe = ''
+    const gameinfo = await getGameInfo(appName)
     const { path } = await ipcRenderer.invoke('openDialog', {
       buttonLabel: t('box.select.button', 'Select'),
       properties: ['openFile'],
-      title: t('box.runexe.title')
+      title: t('box.runexe.title'),
+      defaultPath: gameinfo.install.install_path
```
CommandMC, line 52: This might not be helpful, usually (at least when I use this feature) I want to run an EXE that I got from somewhere (DirectX installers, .NET, things like that)
I think it also sends the wrong message. If you want to run a different executable as the game, you should use the alternative EXE option and not this.

flavioislima: Yes, wasn't sure about this one but I remember that most games comes with a redist folder with the dependencies so then people might actually look there first. If not they can choose a different folder just like it's today.
What do you think?

CommandMC: You're right, doesn't hurt to start there. If they really need another EXE, the home folder is one click away (while getting to the game folder from Home is a bit more effort)

CommandMC (approved):

PR #1392, opened 2022-05-28
Try to fix macOS not launching CX

CommandMC:
---

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -475,11 +475,6 @@ async function runLegendaryOrGogdlCommand(
     bin = runner.bin
   }
 
-  // Quote command parts on Mac since we still have to use the shell there
-  if (isMac) {
-    commandParts = commandParts.map(quoteIfNecessary)
-  }
-
   return new Promise((res, rej) => {
     const child = spawn(bin, commandParts, {
       cwd: runner.dir,
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -478,7 +478,7 @@ async function runLegendaryOrGogdlCommand(
   return new Promise((res, rej) => {
     const child = spawn(bin, commandParts, {
       cwd: runner.dir,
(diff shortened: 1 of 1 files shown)


flavioislima:
Did you confirm that this is still an issue on macOS? 
I will do some tests disabling the shell again to see if it will break.

CommandMC:
There are a bunch of people in the macOS channel having this issue, so I'd assume that it's still present
Sadly I nuked my macOS VM yesterday by accident, will need some time to get that setup again to test

flavioislima:
I mean, I am testing here by using the shell on mac again to try to reproduce the first issue we had.
Because after that we had several refactors that might could have fixed the issue.

flavioislima:
So far, removing the `shell: isMac` when running the command and this works:
- Install Epic game
- Install GOG game
- play windows games
- play native games
- play windows version of native games
- update a game (could test only on GOGDL since I don't have games to update on epic) is broken, even with latest test version of gogdl. But the command is being called so heroic is doing its part.
- Uninstall a game
- Login for both platform

So I believe it is 100% safe to remove the shell on macOS now. Could not find one single issue with the command call.

CommandMC:
Alright, shell's removed now

With the last commit, I believe I've also fixed a bug: When you had an environment variable defined in both your process's env (so for example in .bashrc) and in Heroic's Settings, the process one would override Heroic's settings. That's now the other way around, which I think makes more sense

flavioislima (approved):
Looks good! 👍🏽

flavioislima:
> Alright, shell's removed now
> 
> With the last commit, I believe I've also fixed a bug: When you had an environment variable defined in both your process's env (so for example in .bashrc) and in Heroic's Settings, the process one would override Heroic's settings. That's now the other way around, which I think makes more sense

yes, exactly, the latest override the first inside an object.

PR #1394, opened 2022-05-28
[Fix] Styling and error Dialogs 

flavioislima:
- Remove error dialog for strings with ERROR and CRITICAL since it will create false positives
- Add small fix to relative paths on wineprefix for old configs
- Fixed some styling issues on wine settings and save-sync settings

---

### DIFF
--- a/electron/gog/games.ts
+++ b/electron/gog/games.ts
@@ -23,7 +23,8 @@ import {
   isWindows,
   execOptions,
   isMac,
-  isLinux
+  isLinux,
+  userHome
 } from '../constants'
 import { configStore, installedGamesStore } from '../gog/electronStores'
 import { logError, logInfo, LogPrefix, logWarning } from '../logger/logger'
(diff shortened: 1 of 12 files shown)


electron/gog/games.ts
```diff
@@ -364,7 +365,9 @@ class GOGGame extends Game {
       const { wineVersion, winePrefix, launcherArgs, useSteamRuntime } =
         gameSettings
       let wineFlag = ['--wine', wineVersion.bin]
-      let winePrefixFlag = ['--wine-prefix', winePrefix]
+
+      // avoid breaking on old configs when path is not absolute
+      let winePrefixFlag = ['--wine-prefix', winePrefix.replace('~/', userHome)]
```
CommandMC, line 370: This is something I've been avoiding on purpose. IMO, we should replace the ~ when the settings are loaded from the JSON, instead of doing it everywhere it's used (bonus points if you also manage to save the modified settings, so we "update" them along the way as well)

flavioislima: Yes, this would be only a palliative.
I will try a more complete solution for that.

flavioislima: Ok, changed it now to work like this:
- If the user goes to Global or Game Settings, Heroic will auto override the `~` with the user home. It will change on frontend and save on config file.
- If the user start the game and the path is relative, heroic will fix it when reading the config, BUT not save the config, because there is no way to save the settings when playing the game. This is something we need to deal with in the future, a global method to save settings instead of using `writeConfig`. But lets do that once we decide to refactor the settings.

CommandMC (approved):

PR #1402, opened 2022-05-29
[Fix] Sync saves not working on different languages than english

flavioislima:
This fixes an old bug where the Force Upload and Force Download were not working on languages other than English.
Also improves the logic since the previous approach was too convoluted

---


imLinguin (approved):
why was it like that in the first place lmao

flavioislima:
> why was it like that in the first place lmao

Man, I have no idea. And this was on my list of things to refactor for a long time.

PR #1405, opened 2022-05-29
No longer log commands twice

CommandMC:
- `getLegendaryOrGogdlCommand` and its variants are no longer exposed/no longer exist
- A new `purpose` attribute was added to `runLegendaryOrGogdlCommand`'s options, which makes it possible to define what the command we're running is supposed to do
- Instances of the "getCommand, log it with custom message, runCommand" workflow have been converted to using `purpose`

I think this is a good compromise between being verbose on what we're doing vs avoiding log spam

---


Nocccer:
I don't know if we need the purpose attribute. I will see tomorrow how i would implement it. Maybe i miss something and we need the purpose, but i want todo it with less arguments

CommandMC:
Superseded by #1415

PR #1409, opened 2022-05-30
[Linux] Added option to display Wine executable path

Delta2Force:
I added this due to myself often getting confused by multiple Proton versions with the same name but different paths.

---


src/screens/Settings/components/WineSettings/index.tsx
```diff
@@ -262,23 +262,26 @@ export default function WineSettings({
...
-                <li>~/.var/app/com.valvesoftware.Steam (Steam Flatpak)</li>
-                <li>/usr/share/steam</li>
-                <li>Everywhere on the system (CrossOver Mac)</li>
-                <li>/opt/cxoffice (CrossOver Linux)</li>
-              </i>
-            </ul>
-            <span>{t('help.wine.part2')}</span>
-          </InfoBox>
+          <>
+            <InfoBox text="infobox.wine-path">{wineVersion.bin}</InfoBox>
```
flavioislima, line 266: Question: why did you prefer to add it to the help box?
I wonder if adding it like we are doing with the Legendary alternative binary or on the language selector, that small text below the input/select. I think its easier to see.

flavioislima (approved):
Looks good overall, just wondering if the info could be more accessible.

src/screens/Settings/components/WineSettings/index.tsx
Delta2Force: While it would be easier to see, I thought this was something not necessary normally (as normally one would have only one installation of everything) and therefore wanted to hide it away only for the users who are confused with multiple versions of the same name.

PR #1411, opened 2022-05-30
[Fix] Library not showing and Corrupted config dialog

flavioislima:
- Disable most verbosity that I added on previous PRs
- Fix wrong runner when refreshing the library that was setting the category wrong.
- Now Category will be same as Runner: `legendary` or `gog`, instead of `epic`
- Disabled auto-update on macOS

---

### DIFF
--- a/src/state/GlobalState.tsx
+++ b/src/state/GlobalState.tsx
@@ -142,8 +142,7 @@ export class GlobalState extends PureComponent<Props> {
       (configStore.get('contentFontFamily') as string) || "'Cabin', sans-serif",
     actionsFontFamily:
       (configStore.get('actionsFontFamily') as string) || "'Rubik', sans-serif",
-    allTilesInColor:
-      (configStore.get('allTilesInColor', false) as boolean) || false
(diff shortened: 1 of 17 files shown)


src/state/GlobalState.tsx
```diff
@@ -100,7 +101,7 @@ export class GlobalState extends PureComponent<Props> {
     return games
   }
   state: StateProps = {
-    category: storage.getItem('category') || 'epic',
+    category: (storage.getItem('category') as Category) || 'epic',
```
CommandMC, line 104: Shouldn't this be
```suggestion
    category: (storage.getItem('category') as Category) || 'legendary',
```
?

flavioislima: yes, noticed afterwards. Noticed that if you were using heroic and it was epic it was showing nothing as well. Sent the fix on latest commits

flavioislima:
finished testing on Windows, macOS and Linux.
Was able to reproduce the issues on all platforms.
cannot reproduce anymore after the fixes.

PR #1424, opened 2022-06-01
[Fix] Fix run exe on prefix when having spaces + updated SDL list + updated legendary

flavioislima:
Add SDL list for `Suphalak`

Cannot test this one since I don't own this game, actually have no idea which game is that.
Got the list from legendary API

It seems the game is total war warhammer 2

- Also removed an unnecessary check that was causing login to now always work (I believe)
- Fixed an issue for `runExe` with spaces on the path.

---


flavioislima:
I am also updating legendary to latest version.

PR #1426, opened 2022-06-02
[Fix] Missing Percent on download + fuzzy search

flavioislima:
When I fixed the issue with the dialog breaking on continuing downloads, I introduced this other bug.
Basically was returning the wrong variable for new downloads.

---


imLinguin (approved):
Looks good, we really need to extend time we spend on testing in the future

PR #1442, opened 2022-06-06
[General] Fix: No longer replace token/sid with redacted

CommandMC:
It seems like arrays are passed by reference in JS, which means the getRunnerCallWithoutCredentials function was altering the actual commandParts passed to Legendary/GOGDL. By copying the array before passing it, this no longer happens

---

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -597,6 +597,15 @@ async function callRunner(
     })
 }
 
+/**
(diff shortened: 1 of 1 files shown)


Nocccer (approved):
LGTM. Maybe add function documentation via tsdoc so people know that they should provide a copy of commandParts.

flavioislima (approved):
LGTM

PR #1447, opened 2022-06-07
Start adding frontend tests

arielj:
In this PR I'm adding configuration needed to start adding tests for the frontend.

Some relevant things this is adding:
- a `renderWithState` helper method that will initialize i18next and allows passing an object to mock specific ContextProvider values (usefull to set values extracted from the context)
- i18next is initialized in english, I don't think it makes sense to test internationalization (the i18next library takes care of that)
- some mocking of the ipcRenderer can be done in `src/setupTests.ts` if the component uses `window.require('ipcRenderer')`, but can't be customized for each test if needed, we'll need a better solution in the future (like passing the ipcRenderer around in the app context instead of importing it everywhere, that way we can stub it with the renderWithState helper)

---


package.json
```diff
@@ -201,7 +201,7 @@
   "devDependencies": {
     "@testing-library/dom": "^7.31.0",
     "@testing-library/jest-dom": "^5.16.4",
-    "@testing-library/react": "^13.1.1",
+    "@testing-library/react": "12.1.5",
```
arielj, line 204: Had to downgrade this because of https://stackoverflow.com/a/72183519/1430810

src/__tests__/screens/Accessibility.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+import Accessibility from 'src/screens/Accessibility'
+
+test('can change the content font', async () => {
+  const setFont = jest.fn()
+
+  await act(async () => {
+    renderWithState(<Accessibility />, {
+      contentFontFamily: 'Arial',
+      setContentFontFamily: setFont
+    })
```
arielj, line 16: the renderWithState helper is used like this, the `await act` wrapper is not always needed, it's just for this component that triggers a `setFont` during the initial render, other components may not need it

src/setupTests.tsx
```diff
@@ -0,0 +1,15 @@
+const ipcRenderer = {
+  invoke: (method: string) => {
+    if (method === 'getFonts') {
```
arielj, line 3: we can mock ipcRenderer here

src/state/ContextProvider.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 
 import { ContextType } from 'src/types'
 
-const initialContext: ContextType = {
+export const initialContext: ContextType = {
```
arielj, line 5: export so we can use as default for tests

src/screens/Accessibility/index.tsx
```diff
@@ -86,7 +86,7 @@ export default function Accessibility() {
         </h1>
 
         <span className="rangeWrapper Field">
-          <label className={classNames({ isRTL: isRTL })}>
+          <label className={classNames({ isRTL: isRTL })} htmlFor="zoom-range">
```
arielj, line 89: added so the label and the input are associated, to test it properly

flavioislima (approved):
LGTM as a good restart of FE tests. 
Perhaps take a look at how we were mocking the electron stuff a few months ago, @Nocccer did the implementation.
Maybe we could resurrect some code from before :)

arielj:
Closing this for now, this is adding less tests and the mocking is not as good as it was before, I'll keep thinking of a better way to test these things (I feel like we need end-to-end tests instead of backend/frontend tests but I still don't understand how to do that)

PR #1449, opened 2022-06-08
[UI/UX] Library improvements

flavioislima:
First, sorry for the big PR. 😅 

Some things might still change but I believe this is 99% done.

### Changed
- Add the current Download/Update(s) on the sidebar;
- Show a unified library + add a new Store Filter. 
- Add the ability to collapse the sidebar.
- Updated the Sidebar with a new design;
- Add a Favorites button to show only the favorites;
- Add Auto-Complete to the Search Bar;
- Simplify Game List for a cleaner look;
- Use this new game list for the wine manager as well;
- Add scale effect on GameCard on mouse hover. 
- Change the Design of buttons, selects, and inputs to match the current Figma Design.
- Fine-tune the fuzzy search since it was showing a lot of items not related to the search.
- Several styles fixes to improve consistency;
- Add the Heroic version to the sidebar;

### Other
- Added a global export for the `ipcRenderer` on the frontend. So we should import it from `src/helpers` now instead of `window.require('electron')`
- Added a notification when starting moving a game

### Screenshots

image
image
image
image

---


Etaash-mathamsetty:
is it possible to get a download queue implemented in this pr too?

flavioislima:
> is it possible to get a download queue implemented in this pr too?

No, will be on a different one.

src/components/UI/SearchBar/index.css
```diff
@@ -3,3 +3,27 @@
   width: 100%;
   position: relative;
 }
+
+.autoComplete {
+  position: absolute;
+  top: 41px;
+  width: 100%;
+  height: 200px;
```
arielj, line 11: maybe we can use max-height here? it looks weird with fea items

image

image

src/App.css
```diff
@@ -13,6 +13,13 @@
 
 * {
   box-sizing: border-box;
+  /* Measurements */
+}
+
+@media screen and (max-width: 730px) {
+  * {
+    --sidebar-width: calc(var(--sidebar-width) - 1em);
```
arielj, line 21: I'm not sure what this is doing, maybe it should be applied only to the body? I think it's more expensive for the browser to apply this css variable to every element

apart from that I'm not sure what is this used for

src/components/UI/Sidebar/index.css
```diff
@@ -60,3 +68,104 @@
...
+}
+
+.collapseIcon {
+  position: fixed;
+  left: calc(var(--sidebar-width) - 8px);
+  bottom: 55vh;
+  align-items: center;
+  cursor: pointer;
+  opacity: 0;
+  transition: 500ms;
```
arielj, line 98: definitely my idea about the icon fadeout delay thing was a bad idea xD

src/screens/Library/components/GameCard/index.tsx
```diff
@@ -309,6 +323,11 @@ const GameCard = ({
               { '--installing-effect': installingGrayscale } as CSSProperties
             }
           >
+            {runner === 'legendary' ? (
+              <EpicLogo className="store-icon" />
+            ) : (
+              <GOGLogo className="store-icon" />
+            )}
```
arielj, line 330: maybe show the icon only if there's more than one store logged in? it's kinda useless for people using only one store and it clutters the grid with a repeated icon

src/screens/Library/components/GameCard/index.css
```diff
@@ -82,89 +80,132 @@
...
 
-.gameCard:hover > .icons,
-.gameCard:focus-within > .icons {
-  opacity: 1;
+.gameCard > .icons > .svg-button {
+  justify-self: center;
 }
 
-.gameCard:hover > .gameTitle,
-.gameCard:focus-within > .gameTitle {
```
arielj, line 115: maybe this change can be an option in the accessibility tab? if the reason of the change is accessibility, because the library is kinda cluttered now by default and there's no way back if one wants to use a cleaner layout

arielj (commented):
I added a few comments, I have to admit it's pretty hard to review all these changes in one big PR so I couldn't review it as much as I would wanted to, there's a lot going on between refactors, bugfixes, restyling, new features

src/App.css
flavioislima: This is used when the sidebar is not collapsed and the screen is too small, it makes the uncollapsed sidebar a bit smaller in width.
I think it's kind of useless now because if the view is too small makes more sense to just collapse the sidebar.

src/components/UI/Sidebar/index.css
flavioislima: But do you we should remove it?

src/screens/Library/components/GameCard/index.tsx
flavioislima: Yes, makes sense!

src/screens/Library/components/GameCard/index.css
flavioislima: Do you mean without showing the title all the time?
I don't know, this is how it looks in the current design in Figma. 
On Twitter at least we got good feedback from people, let's release like this on the beta and check what people will say. 👍🏽

src/components/UI/SearchBar/index.css
flavioislima: done

src/components/UI/Sidebar/index.css
arielj: after playing a bit with the CSS, I think we can use this:

change that line to be `transition: opacity 500ms 500ms;` so we only transition the opacity and not the position, with a small delay

then replace all the other styles for the transition:

```css
.Sidebar:hover .collapseIcon {
  transition: opacity 500ms;
  opacity: 1;
}

.Sidebar.collapsed:hover .collapseIcon {
  transition: opacity 3s;
  opacity: 1s;
}

.Sidebar.collapsed .collapseIcon {
  transition: opacity 1s;
}
```

with only this one:

```css
.Sidebar:hover .collapseIcon {
  opacity: 1;
}
```

with those changes there's a delay, then it fades in and it follows the sidebar change immediately when expanded/collapsed

flavioislima: done! 
Thanks for the help 👍🏽

PR #1464, opened 2022-06-13
[Fix] multiple settings writes

### LINKED ISSUE #1463: Settings are saved multiple times just by opening the settings screen
### Describe the bug

When opening any settings screen (either Heroic's or a game), configuration is saved to disk multiple times.

Opening Heroic's settings the first time it writes the config 13 times. After that, changing to another sub-section (General, Wine, Other, etc), write the settings 3 more times. That's without doing any change.

Opening a game's settings the first time it writes the game's config 8 times for one game, 15 times for another, 16 times for another. After that, changing sub-sections writes the config 3 times for every section.

Going to another screen and back to settings starts with all the writes again.

### Add logs

Nothing shows up on the logs, I put a `console.log('writing config')` in this `useEffect` in the settings index

```ts
  useEffect(() => {
    console.log('writing config')
    writeConfig([appName, settingsToSave])
  }, [GlobalSettings, GameSettings, appName])
```

This is just opening the settings for Magic The Gathering Arena (no changes at all)

image

### Steps to reproduce

1. Add the code shared above in the settings index.tsx line 329
2. Open DevTools in Heroic and go to the console tab
3. Enter the settings for heroic or any game, move around
4. Multiple `writing config` lines are logged in the console

### Expected behavior

Configuration should only be written when it changes, and only once. It should also NOT be written when opening settings or changing between sections.

### Screenshots

_No response_

### System Information

- Heroic Version [e. g. 2.1.1]: current beta branch, but it's a long running issue since older versions

### Additional information

I'm already debugging this trying to refactor it (I assigned this to me already), I'm creating the issue to keep track of it.

arielj:
This PR fixes #1463 by only writting settings when they change, and only after they are completely loaded. This applies for both Heroic's settings and games' settings.

This should improve the consistency of settings, we were storing things so many times while things where loading (I've seen up to 16 writes during the initial render!!) that it probably created some issues with race conditions having that many writes to the same files all at once.

**How to test/QA**

I added a call to `logInfo` that prints information on the logs every time call to writeConfig is done. You can play around with the settings while checking at the logs to see this is working as expected.

Note that the diff in the Files changed tab of the PR is a mess, it doesn't really reflect the changes properly, github gets really confused because I moved things around.

---

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -604,7 +604,7 @@ async function callRunner(
  * @param env Enviroment variables to use
  * @param wrappers Wrappers to use (gamemode, steam runtime, etc.)
  * @param runnerPath The full path to the runner executable
- * @returns 
+ * @returns
  */
 function getRunnerCallWithoutCredentials(
   commandParts: string[],
--- a/src/screens/Settings/components/AdvancedSettings/index.tsx
+++ b/src/screens/Settings/components/AdvancedSettings/index.tsx
@@ -216,7 +216,7 @@ export const AdvancedSettings = ({
             isSuccess: isCopiedToClipboard
           })}
           onClick={() => {
-            clipboard.writeText(JSON.stringify({ ...settingsToSave }))
+            clipboard.writeText(JSON.stringify({ ...settingsToSave }, null, 2))
             setCopiedToClipboard(true)
           }}
         >
--- a/src/screens/Settings/index.tsx
+++ b/src/screens/Settings/index.tsx
@@ -181,6 +181,9 @@ function Settings() {
   const [autoSyncSaves, setAutoSyncSaves] = useState(false)
   const [altWine, setAltWine] = useState([] as WineInstallation[])
 
+  const [configLoaded, setConfigLoaded] = useState(false)
+  const [settingsToSave, setSettingsToSave] = useState({})
+
   const { appName = '', type = '' } = useParams()
   const isDefault = appName === 'default'
   const isGeneralSettings = type === 'general'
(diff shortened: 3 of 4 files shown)


src/screens/Settings/components/AdvancedSettings/index.tsx
```diff
@@ -216,7 +216,7 @@ export const AdvancedSettings = ({
             isSuccess: isCopiedToClipboard
           })}
           onClick={() => {
-            clipboard.writeText(JSON.stringify({ ...settingsToSave }))
+            clipboard.writeText(JSON.stringify({ ...settingsToSave }, null, 2))
```
arielj, line 219: added this so the copied settings in the clipboard are pretty printed, easier to read

electron/launcher.ts
```diff
@@ -604,7 +604,7 @@ async function callRunner(
  * @param env Enviroment variables to use
  * @param wrappers Wrappers to use (gamemode, steam runtime, etc.)
  * @param runnerPath The full path to the runner executable
- * @returns 
+ * @returns
```
arielj, line 607: the linter was complaining about this so I had to fix it :man_shrugging:

flavioislima (approved):
Looks good in general.
In the future, I believe that could be fixed by using a React Hook that could work to write and read the settings.
But for now that is doing a good job :)

src/screens/Settings/index.tsx
```diff
@@ -478,7 +543,7 @@ function Settings() {
               setAltLegendaryBin={setAltLegendaryBin}
               altGogdlBin={altGogdlBin}
               setAltGogdlBin={setAltGogdlBin}
-              settingsToSave={settingsToSave}
+              settingsToSave={settingsToSave as AppSettings}
```
flavioislima, line 546: One small thing though, why do we need this type conversion?
Have you check if the buttons to copy the settings are working fine?

flavioislima: Tested here, they are working fine

arielj: the problem is that I set the default state value as `{}` and typescript (or the linter?) complains that it doesn't have the expected properties fron an AppSettings, so forcing it solves the warning, another option is to say allow 2 types of data in the AdvancedSettings component I guess? I don't know if typescript likes `{}` in a type

PR #1467, opened 2022-06-13
Store only game setting keys in game config file during installation

### LINKED ISSUE #1465: All Heroic settings are stored as game settings during game installation
### Describe the bug

When installing a game, all Heroic settings are stored for the game's settings, even settings that don't apply to games like `maxWorkers` or `darkTrayIcon` for example.

### Add logs

Nothing shows up in logs, I checked the config file for a game after installation.

### Steps to reproduce

1. Install a game that you have never configured before (or remove the config file it it was installed before)
2. Look for its config file in ~/.config/Heroic/GameConfig and open it in a text editor

### Expected behavior

Config file should only include the settings can be applied to games.

### Screenshots

_No response_

### System Information

- Heroic Version [e. g. 2.1.1]: beta branch, but the problem is not new

### Additional information

I'm assigning it to myself, I'm already debugging this, adding this to keep track of it.

arielj:
This PR fixes #1465 by using keys that apply only for games when creating the config file for a game during installation instead of copying the complete object.

I understand this should also fix that issue that was reported where the solution was to go to the settings and make one change so the settings got updated with correct data.

This also fixes an issue where all games were created with the current language code in the `language` setting which was unnecessary and broke the language override feature (the `language` setting should be empty by default to fallback to Heroic's setting).

**How to test/QA**

Install a game that you have never configured before (if you want to try with a game that you used, delete the config file first at `~/.config/heroic/GameConfig/`).

Compare the JSON file created before and after this fix, the file will contain only game settings and not all heroic settings.

---

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -604,7 +604,7 @@ async function callRunner(
  * @param env Enviroment variables to use
  * @param wrappers Wrappers to use (gamemode, steam runtime, etc.)
  * @param runnerPath The full path to the runner executable
- * @returns 
+ * @returns
  */
 function getRunnerCallWithoutCredentials(
   commandParts: string[],
--- a/src/screens/Library/components/InstallModal/index.tsx
+++ b/src/screens/Library/components/InstallModal/index.tsx
@@ -382,12 +382,14 @@ export default function InstallModal({
         )
         if (Array.isArray(newWineList)) {
           setWineVersionList(newWineList)
-          if (
-            !newWineList.some(
-              (newWine) => wineVersion && newWine.bin === wineVersion.bin
-            )
-          ) {
-            setWineVersion(undefined)
+          if (wineVersion?.bin) {
+            if (
+              !newWineList.some(
+                (newWine) => wineVersion && newWine.bin === wineVersion.bin
+              )
+            ) {
+              setWineVersion(undefined)
+            }
           }
         }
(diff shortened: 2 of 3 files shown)


electron/launcher.ts
```diff
@@ -604,7 +604,7 @@ async function callRunner(
  * @param env Enviroment variables to use
  * @param wrappers Wrappers to use (gamemode, steam runtime, etc.)
  * @param runnerPath The full path to the runner executable
- * @returns 
+ * @returns
```
arielj, line 607: linter complaining about this file

src/screens/Library/components/InstallModal/index.tsx
```diff
@@ -382,12 +382,14 @@ export default function InstallModal({
         )
         if (Array.isArray(newWineList)) {
           setWineVersionList(newWineList)
-          if (
-            !newWineList.some(
-              (newWine) => wineVersion && newWine.bin === wineVersion.bin
-            )
-          ) {
-            setWineVersion(undefined)
+          if (wineVersion?.bin) {
```
arielj, line 385: only do this check if we actually have a wineVersion, it was creating some issues with race conditions and sometimes the wineVersion setting got lost showing the wrong wineVersion preselected

flavioislima (approved):
LGTM 👍🏽

Issue #1491, opened 2022-06-17
Game config files on windows and mac have wine and wineprefix information


arielj:
### Describe the bug

When we see logs from windows issues, there are settings for wine and winePrefix, this is not a big problem since it's not used but it adds confusion.

There is a problem though, when creating this extra configuration values, sometimes it fails on windows with this error `TypeError: winePrefix.replace is not a function` which should never happen in windows. This error is related to this bug since it comes from the place the settings are fetch.

### Add logs

This is one of the errors reported:

```
16:43:21) ERROR:   [Frontend]:        Error: Error invoking remote method 'requestSettings': TypeError: winePrefix.replace is not a function
```

### Steps to reproduce

I'm not sure about the specific steps to reproduce this since I don't have windows, but for some reason users may end up with incorrect data.

### Expected behavior

Game config should have references to wine and winePrefix only on linux.
Game config should have references to crossover only on Mac.
Game config shouldn't have references to wine, winePrefix, nor crossover on Windows.

### System Information

- Heroic Version [e. g. 2.1.1]: The error reported points to a line in the current release, but it would happen also on beta

PR #1492, opened 2022-06-17
[Tech] Refactor game settings with fallbacks 

arielj:
This PR should fix the 2 issues mentioned here https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1491 by adding more customization around which settings are stored for a game depending on the platform. I think there's still room for improvement there (like... gamemode, mongohud, audio fix, and more can be moved to only linux), but I wanted to fix the issue related to the wine config first before moving more things around and have more tests.

The issues I noticed in the original code:
- we were defaulting to the complete global settings if no config file for a game, this is not necesarilly bad but we just dumped everything there without any sanity check
- we were trying to do a `winePrefix.replace` even on windows on mac, which is not only not needed but also failed in some cases when `winePrefix` was not a string
- we were storing extra information for all platforms, which makes reading logs kinda confusing sometimes

---

### DIFF
--- a/electron/game_config.ts
+++ b/electron/game_config.ts
@@ -6,6 +6,8 @@ import {
   currentGameConfigVersion,
   heroicConfigPath,
   heroicGamesConfigPath,
+  isLinux,
+  isMac,
   userHome
 } from './constants'
 import { logError, logInfo, LogPrefix } from './logger/logger'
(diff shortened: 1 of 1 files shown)


electron/game_config.ts
```diff
@@ -252,19 +246,36 @@ class GameConfigV0 extends GameConfig {
...
-    return settingsToReturn
+    // set specific keys depending on the platform
+    if (isMac) {
+      defaultSettings.wineCrossoverBottle = wineCrossoverBottle
+    } else if (isLinux) {
+      defaultSettings.wineVersion = wineVersion
+      defaultSettings.winePrefix = winePrefix || `${userHome}/.wine`
+
+      // fix winePrefix if needed
+      if (gameSettings.winePrefix) {
```
flavioislima, line 269:
```suggestion
      if (gameSettings.winePrefix?.includes('~')) {
```
so maybe we can avoid doing replace always?

flavioislima (approved):
Looks good, just one small comment.

arielj:
I was testing it more on a Mac and I broke something, I won't merge this for now, I can't ignore `wineVersion` on mac like I was doing here

flavioislima (approved):
LGTM

Issue #1484, opened 2022-06-16
Recently played games cannot be launched through the tray icon menu on Linux.


rabidcopy:
### Describe the bug

When attempting to launch recently played games listed in the right-click menu of the tray icon, games do not launch. With logging enabled, it puts out "Check failed: chdir(current_directory) == 0". May be related to #968.

### Add logs

Check failed: chdir(current_directory) == 0

### Steps to reproduce

1. Right click the tray icon with a recently played game listed.
2. Left click the game you wish to launch.
3. Nothing happens

### Expected behavior

Expected behavior is for the game to be launched when the menu entry is selected.

### Screenshots

![image](https://user-images.githubusercontent.com/8052832/173976187-2fa3891e-b0c6-4881-9d07-a138b1be9080.png)

### System Information

Heroic Version: 2.3.7 Brook

Legendary Version:  0.20.26 Dark Energy

OS: EndeavourOS KERNEL: 5.18.1-259-tkg-bmq ARCH: x64

CPU: AMD Ryzen 5 2600 Six-Core Processor @3.4 

RAM: Total: 15.64 GiB Available: 11.21 GiB

GRAPHICS: GPU0: Ellesmere [Radeon RX 470/480/570/570X/580/580X/590] VRAM: 256MB DRIVER:  

PROTOCOL: x11

arielj:
How did you install Heroic? flatpak? appimage? repos? I'm wondering if that has something to do and if it works if you try a different installation option.

rabidcopy:
> How did you install Heroic? flatpak? appimage? repos? I'm wondering if that has something to do and if it works if you try a different installation option.
> 
> EDIT: nevermind, I'm able to reproduce the error using the source and I think I have a fix for it

Just to be thorough, I installed Heroic through the AUR repository. I compiled #1493 myself and can confirm games launch from the tray icon menu now. Thank you!

PR #1493, opened 2022-06-17
Run game from tray icon directly without using shell

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1484 by changing the action executed when a tray icon game is clicked.

Instead of opening the heroic URI using `shell`, now we run the `handleProtocol` function directly, removing that middle man that was not really needed (it had to go to the shell to open that URI and wait for the OS to handle it and pass it back to Heroic).

---


flavioislima (approved):
Looks good!

PR #1500, opened 2022-06-19
Add Hover Description for Toggles

redromnon:
Simple hover boxes to describe what toggles/checkboxes do using a help icon.
This applies to Wine & Other Settings.

---


Etaash-mathamsetty:
I wouldn't say improves improves performance for vkd3d, I would say, "enables support for d3d12 games" or something like that

redromnon:
@Etaash-mathamsetty You're right. The built-in in WineD3D doesn't support Direct3D 12 and needs VKD3D.

src/screens/Settings/components/WineSettings/index.tsx
```diff
@@ -307,7 +307,10 @@ export default function WineSettings({
             ])
             return toggleAutoInstallDxvk()
           }}
-          title={t('setting.autodxvk', 'Auto Install/Update DXVK on Prefix')}
+          title={t(
+            'setting.autodxvk',
+            'Auto Install/Update DXVK on Prefix (Improves performance for DirectX 9, 10 and 11 games)'
```
flavioislima, line 312: Changing the key here won't make a difference since this is just the fallback in case the key is not translated for a specific language.
You need to change the english json file as well on the public folder.

redromnon: Understood.

redromnon:
Yeah, that sounds cool.

redromnon:
Is it fine to use the help icon button that is currently being used for Wine Version, Advanced Options and Game Arguments?

flavioislima:
> Is it fine to use the help icon button that is currently being used for Wine Version, Advanced Options and Game Arguments?

Yes, would be nice to add the info icon on the side of the setting name, then putting the mouse over it, could show a description of the feature.

But my two cents here: Both DX9, 10, 11 and 12 works on Wine without DXVK and VKD3D.
so what DXVK and VKD3D do are actually try to increase the compatibility with some games. NOT all games work fine when using them, especially old games, that will work better with plain wine.

So I believe the description of the setting should be like:
DXVK: `May increase compatibility with (DX9, 10 and 11) games. Might case issues with old games`
VKD3D: `May increase compatibility with (DX12) games. Has no effect on DX9, 10 or 11 games.`
Wine Prefix: `This is a folder where all your games saves configurations are. Removing this folder will erase your progress and game settings.`

redromnon:
I believe the default WineD3D won't be able to run those DX12 games and thus requires enabling VKD3D.

flavioislima:
Hmm, I dont think so.
Doing some search you can see that since wine 4.0 they support dx12. I found this changelog from version 7.4 as well where they changed something on their api:
https://www.winehq.org/announce/7.4

In practice, you are right, VKD3D will be necessary for probably 99.9% of games.

flavioislima:
> > Hmm, I dont think so. Doing some search you can see that since wine 4.0 they support dx12. I found this changelog from version 7.4 as well where they changed something on their api: https://www.winehq.org/announce/7.4
> > 
> > In practice, you are right, VKD3D will be necessary for probably 99.9% of games.
> 
> that release just bundled vkd3d with wine lmao
> but it's the nerfed version of vkd3d, the one that the wine devs maintain, the good version in vkd3d-proton, which is way better

Yes, I know. Like I said, wine has d3d12 support. If it is good or not or work or not doesn't matter. But we cannot say that it doesn't, just that the proton version is more mature and stable 😜

redromnon:
From what I've learnt, WineD3D was specifically made for DX 1-11 games and vkd3d is another project. So Wine, as a whole, uses vkd3d for DX 12 games. For more clarity, [here's the wiki page link](https://emulation.gametechwiki.com/index.php/Wrappers).

public/locales/en/translation.json
```diff
@@ -159,7 +159,9 @@
         },
         "wine": {
             "part1": "Heroic searches for versions of Wine, Proton, and CrossOver in the following folders:",
-            "part2": "For other places, use a symbolic link to one of these folders"
+            "part2": "For other places, use a symbolic link to one of these folders",
+            "dxvk": "DXVK is a Vulkan-based translational layer for DirectX 9, 10 and 11 games. Enabling may improve compatibility. Might cause issues especially for older DirectX 1-8 games.",
```
redromnon, line 163: How do I use these values I've created in `index.tsx`?

redromnon:
https://user-images.githubusercontent.com/74495920/174737748-c72f0c58-54a8-475a-b52d-5bb4b6a90b30.mp4

So basically here's how it looks.

redromnon:
I've add hover box descriptions for more toggles under Wine & Other Settings.
I'll update the language keys once finalized.

redromnon:
Here's how it finally look -

Heroic_New_Hover

README.md
```diff
@@ -1,6 +1,6 @@
...
...cms+PC9yZGY6UkRGPjwvbWV0YWRhdGE+PHBhdGggZD0iTTguMzIgMjEuOTdhLjU0Ni41NDYgMCAwIDEtLjI2LS4zMmMtLjAzLS4xNS0uMDYuMTEuNi00LjA5Yy42LTMuOC41OS0zLjc0LjY3LTMuODVjLjEzLS4xNy4xMS0uMTcgMS42MS0uMThjMS4zMi0uMDMgMS42LS4wMyAyLjE5LS4xMmMzLjI1LS40NSA1LjI2LTIuMzYgNS45Ni01LjY2Yy4wNC0uMjIuMDgtLjQxLjA5LS40MWMwLS4wMS4wNy4wNC4xNS4xYzEuMDMuNzggMS4zOCAyLjIyLjk5IDQuMTRjLS40NiAyLjI5LTEuNjggMy44MS0zLjU4IDQuNDZjLS44MS4yOC0xLjQ5LjM5LTIuNjkuNDJjLS44LjA0LS44Mi4wNC0xLjA1LjE5Yy0uMTcuMTctLjE2LjE0LS41NSAyLjU1Yy0uMjcgMS43LS4zNyAyLjI1LS40MSAyLjM1Yy0uMDcuMTYtLjIxLjMtLjM3LjM4bC0uMTEuMDdIMTBjLTEuMjkgMC0xLjYyIDAtMS42OC0uMDNtLTQuNS0yLjIzYy0uMTktLjEtLjMyLS4yNy0uMzItLjQ3QzMuNSAxOSA2LjExIDIuNjggNi4xOCAyLjVjLjA5LS4xOC4zMi0uMzcuNS0uNDRMNi44MyAyaDMuNTNjMy45MSAwIDMuNzYgMCA0LjY0LjJjMi42Mi41NSAzLjgyIDIuMyAzLjM3IDQuOTNjLS41IDIuOTMtMS45OCA0LjY3LTQuNSA1LjNjLS44Ny4yMS0xLjQ4LjI3LTMuMTQuMjdjLTEuMzEgMC0xLjQxLjAxLTEuNjcuMTVjLS4yNi4xNS0uNDcuNDItLjU2Ljc1Yy0uMDQuMDctLjI3IDEuNDctLjUzIDMuMWEyNDEuMyAyNDEuMyAwIDAgMC0uNDcgMy4wMmwtLjAzLjA2SDUuNjljLTEuNTggMC0xLjggMC0xLjg3LS4wNHoiIGZpbGw9IiNmZmYiLz48L3N2Zz4=)](https://www.paypal.me/heroicgl) [![kofi](https://img.shields.io/badge/Ko--Fi-Donate-orange?style=flat&logo=ko-fi)](https://ko-fi.com/heroicgames)
```
flavioislima, line 3: What changed here?

.github/workflows/draft-release-linux.yml
```diff
@@ -2,6 +2,7 @@ name: Draft Release Linux
 
 on:
   push:
+    branches: [main]
```
Nocccer, line 5: I think we want a draft on every push, not only on main.

.github/workflows/draft-release-mac.yml
```diff
@@ -2,6 +2,7 @@ name: Draft Release MacOSX
 
 on:
   push:
+    branches: [main]
```
Nocccer, line 5: same

.github/workflows/draft-release-win.yml
```diff
@@ -2,6 +2,7 @@ name: Draft Release Windows
 
 on:
   push:
+    branches: [main]
```
Nocccer, line 5: same

electron/legendary/games.ts
```diff
@@ -293,14 +293,12 @@ class LegendaryGame extends Game {
...
     })
-    const { maxWorkers, downloadNoHttps } =
-      await GlobalConfig.get().getSettings()
+    const { maxWorkers } = await GlobalConfig.get().getSettings()
     const info = await Game.get(this.appName, 'legendary').getInstallInfo()
     const workers = maxWorkers ? ['--max-workers', `${maxWorkers}`] : []
-    const noHttps = downloadNoHttps ? ['--no-https'] : []
     const logPath = join(heroicGamesConfigPath, this.appName + '.log')
 
-    const commandParts = ['update', this.appName, ...workers, ...noHttps, '-y']
```
Nocccer, line 303: Something is broken. You did not merged correctly. You reverting alot of code, which was added in another pr.

Nocccer (changes requested):
There is a wrong merge going on.

Issue #723, opened 2021-10-31
Support adding game shortcuts on Mac


arielj:
**Is your feature request related to a problem? Please describe.**

Currently, the Add shortcuts action is not present in Mac

**Describe the solution you'd like**

The app should show the Add shortcuts button on Mac and should add the icons in the correct places

arielj:
Hmm I can't make games work on MacOS. Seems like Legendary does not officialy support MacOS https://github.com/derrod/legendary/discussions/169 and after following that discussion's instructions I still can't get any game to run so I can't really test if a shortcut works if I manage to create one (which seems to be also really complicated on MacOS).

Is there any documentation on how to get this running on Mac? maybe I'm missing something

flavioislima:
Legendary doesn't support Mac yet but we have a legendary binary compiled for MacOSX that works on Heroic.
You need to install Crossover and create a new bottle called Heroic. Then on the game settings, after installing it, just select the crossover app under Wine, and then the crossover bottle you write Heroic if it doesn't detect automatically.

flavioislima:
About how Shortcuts works on Mac, I have no idea, you will need to do some research. You also need to check if the protocol is being registered since Heroic is using the `heroic://` protocol on Mac but I don't know how to call it. On the logs, when running heroic from the terminal, or when running `yarn electron` on the terminal in dev mode, you can check if the protocol is being registered. I think you will need to start there, checking the protocol and then checking how to call this protocol under mac and then generating the shortcuts.

arielj:
Oh I'll check with crossover, I was trying with wine and any game I tried failed to start.

About the protocol, I tried and it works with `open heroic://launch/...`, I have to do some more research on how to create a shortcut after I get some games working on mac

dragonDScript:
https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/AccessingFilesandDirectories/AccessingFilesandDirectories.html#//apple_ref/doc/uid/TP40010672-CH3-SW10

arielj:
I was doing some research for this but I can't find a way to create a desktop icon that will execute `open heroic://launch/...`. There seems to be no way to even do that manually on OSX.

I can do a symlink pointing to the executable file maybe, but skipping heroic when opening the game sounds like a bad idea so I don't want to try it if it's useless.

dragonDScript:
[...]

PR #1501, opened 2022-06-19
Refactor/fix: Wait until frontend is ready before processing protocol/args

arielj:
With the current code, we process the protocol/args 3s after Heroic starts, this value is sometimes too much (if heroic starts really fast) and sometimes not enough (in slow computers like my Mac it takes more than those 3 second for the frontend to be ready).

Instead of having that delay, with this PR I added a message from the frontend to the backend to tell it it's ready, and then we process the protocol.

This will also help with issue I had with the desktop shortcut feature on Mac, since I think this was causing the problem described in my comment (my mac being slow taking more than 3 seconds to have the frontend ready) https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/723

**How to Test/QA:**

Change the `electron` script in package.json to be something like:
```
"electron": "yarn build-electron && electron . heroic://launch/Fortnite --trace-warnings",
```

Or any installed game, run `yarn electron`.

You can set a really small timeout to show what happens if that runs before the frontend is ready in the code without this fix.

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -375,8 +375,6 @@ if (!gotTheLock) {
       mainWindow.show()
     }
 
-    handleProtocol(mainWindow, process.argv)
-
     // set initial zoom level after a moment, if set in sync the value stays as 1
     setTimeout(() => {
       const zoomFactor =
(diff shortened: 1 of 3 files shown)


flavioislima (approved):
LGTM 👍🏽

PR #1504, opened 2022-06-20
[Linux] Add anticheat information when possible

arielj:
This PR adds anticheat information when possible using the information from AreWeAntiCheatYet.com.

It downloads the games.json from https://github.com/Starz0r/AreWeAntiCheatYet/blob/master/games.json at boot in the background only on Linux, to keep the information up to date (it's less than 100kb).

I left this message `It may not work due to denied or broken anticheat support` on purpose without interpolation of the actual status to help translations (also, I don't know how to handle translations based on a variable).

I went over all the games in the json and looked for them in the epic store to create the map of names for games that have differences but I might have missed something, we can always add more or update when needed.

Denied:
Screen Shot 2022-06-20 at 7 26 33 PM

Supported:
Screen Shot 2022-06-20 at 7 26 59 PM

Broken:
Screen Shot 2022-06-20 at 7 27 14 PM

Denied in install modal:
Screen Shot 2022-06-20 at 7 26 44 PM

---


electron/anticheat/utils.ts
```diff
@@ -0,0 +1,37 @@
+import { heroicAnticheatDataPath, isLinux } from '../constants'
+import * as axios from 'axios'
+import { logInfo, LogPrefix, logWarning } from '../logger/logger'
+import { readFileSync, writeFileSync } from 'graceful-fs'
+import { AntiCheatInfo } from '../types'
+import { epicTitleToAnticheat } from './titles_map'
+
+async function downloadAntiCheatData() {
+  if (!isLinux) return
```
flavioislima, line 9: Does this mean we only show the info on Linux? Because I guess most of those titles are not available for macOS natively so we should show them on mac as well.

src/components/UI/Anticheat/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <p>
+          {t(
+            'anticheat.may_not_work',
+            'It may not work due to denied or broken anticheat support.'
+          )}
+        </p>
+      )}
+      <span>
+        <b>{t('anticheat.anticheats', 'Anticheats')}:</b>&nbsp;
+        {anticheatInfo.anticheats.length
```
flavioislima, line 48: I wonder if `anticheats` can be something else other than an array or string, this could break.
Maybe do `anticheatInfo.anticheats?.length`

Nocccer (approved):
Besides what Flavio found, looks good to me. We can actually implement tests for the backend functions. But maybe in another pr.

Just an idea: Maybe a new logprefix `Anticheat` ?

JakobDev:
It might be good to include informations like the reference link notes and updates from the JSON file if present.

electron/anticheat/utils.ts
arielj: hmmmm maybe we should have a different source if on a Mac? the site says `A comprehensive and crowd-sourced list of games using anti-cheats and their compatibility with GNU/Linux or Wine/Proton.` so I thought it was linux specific, like... I understand Fortnite works on a Mac so this would report an incorrect alert in a case like that?

src/components/UI/Anticheat/index.tsx
arielj: I could add that just to be safe (doesn't hurt), but they do `anticheats.map` directly in the website without a guard https://github.com/Starz0r/AreWeAntiCheatYet/blob/master/components/GamesList.tsx#L85 and in the games.json I only saw arrays of strings (or empty array in the case of Far Cry 5)

arielj:
> It might be good to include informations like the reference link notes and updates from the JSON file if present.

I tried, but the information varies between games and it was really difficult to consider all the cases in a consistent design (you have games with multiple updates with different references, the main reference may not even match the last update reference so I can't use the main reference always without the updates) and without making it overwhelming. It felt like too much information, the average user might not care about all the details.

I did add the link to open areweanticheatyet to see those details for the people that do want to know.

electron/anticheat/utils.ts
arielj: maybe I have to only check this if not a native mac version of a game when we detect it's a mac os?

I'm still not sure if areweanticheatyet considers mac games on the status though, the list is built with linux in mind so mac might dispaly false positives/negatives?

electron/anticheat/titles_map.ts
```diff
@@ -0,0 +1,27 @@
+// this hash contains a mapping of titles as displayed in Epic to titles as
+// displayed in areweanticheatyet.com https://github.com/Starz0r/AreWeAntiCheatYet/blob/master/games.json
+//
+// comparison is case insensitive, so no need to add cases like 'VALORANT': 'Valorant'
+// if the epic title is contained, so no need to add cases like 'Fall Guts': 'Fall Guys: Ultimate Knockout'
+//
+// left is the epic title, right is the areweanticheatyet title
+const TITLES_MAP: { [key: string]: string } = {
+  'Battlefield™ 2042': 'Battlefield 2042',
```
felipecrs, line 9: You could also consider stripping out any ™ characters from the comparison

electron/anticheat/titles_map.ts
```diff
@@ -0,0 +1,27 @@
...
+  'Back 4 Blood: Standard Edition': 'Back 4 Blood',
+  'Rainbow Six Siege': 'Rainbow Six: Siege',
```
felipecrs, line 11: And perhaps even the colon signs

electron/anticheat/titles_map.ts
```diff
@@ -0,0 +1,27 @@
+// this hash contains a mapping of titles as displayed in Epic to titles as
+// displayed in areweanticheatyet.com https://github.com/Starz0r/AreWeAntiCheatYet/blob/master/games.json
+//
+// comparison is case insensitive, so no need to add cases like 'VALORANT': 'Valorant'
+// if the epic title is contained, so no need to add cases like 'Fall Guts': 'Fall Guys: Ultimate Knockout'
+//
+// left is the epic title, right is the areweanticheatyet title
+const TITLES_MAP: { [key: string]: string } = {
+  'Battlefield™ 2042': 'Battlefield 2042',
+  'Back 4 Blood: Standard Edition': 'Back 4 Blood',
```
felipecrs, line 10: > if the epic title is contained, so no need to add cases like 'Fall Guts': 'Fall Guys: Ultimate Knockout'

Why not the opposite too?

electron/anticheat/titles_map.ts
```diff
@@ -0,0 +1,27 @@
...
+  'Rainbow Six Siege': 'Rainbow Six: Siege',
+  'Phantasy Star Online 2 New Genesis': 'Phantasy Star Online 2',
+  'Knockout City™': 'Knockout City',
+  'Heroes & Generals WWII': 'Heroes & Generals',
+  'Crysis Remastered': 'Crysis',
+  'STAR WARS™ Battlefront™ Ultimate Edition': 'Star Wars: Battlefront (2015)',
+  'STAR WARS™ Battlefront™ II: Celebration Edition':
+    'Star Wars: Battlefront II (2017)',
+  'The Division': "Tom Clancy's The Division", // added this to prevent false match with The Division 2
+  'Battlefield V Definitive Edition': 'Battlefield V'
```
felipecrs, line 20: But it would be great if areweanticheatyet could provide store ids for the games listed there. Like:

```jsonc
{
        "url": "",
        "name": "Battlefield 2042",
        "logo": "",
        "native": false,
        "status": "Broken",
        "reference": "https://www.protondb.com/app/1517290#atmpx3XrXL",
        "anticheats": [
            "Easy Anti-Cheat"
        ],
        "notes": [],
        "updates": []
        "ids": {
          "steam": "1517290",
          "epic": "<how-do-I-find-the-epic-game-id?>"
        }
},
```

arielj: I'm ok with that, if they are ok with that it would make the check easier, but it's up to them (I understand it's out of the scope of areweanticheatyet and it would be another thing to maintain for them that they don't really use).

I don't know how to get Epic's id for arbitrary games, but I know that we can get information about a game in the metadata legendary downloads (`~/.config/Legendary/metadata/....json`). In that file we have the appName (what we use internatlly everywhere in heroic) and there's an `id` also in the metadata property but we don't really use it and I don't know if it is a game id (I don't know why Legendary uses the appName for everything either).

I also tought about something like `name_in_stores` instead of `ids` so it's easier to find out. But again, I think it might be outside of the scope of that project.

electron/anticheat/titles_map.ts
arielj: I thought about that, but I didn't want to start stripping things out because I don't know if that can generate any issue in the future.

So I went with a more strict approach for now, since we still need to cover cases where stripping characters is not enough like `'STAR WARS™ Battlefront™ Ultimate Edition'` and `'Star Wars: Battlefront (2015)'`. If, in the future, we see it's too hard to maintain this list maybe we can find the right things to remove.

electron/anticheat/titles_map.ts
arielj: Same for this, I agree that removing some `:` would remove this lines with the current information, but we still need this mapping for other cases.

electron/anticheat/titles_map.ts
felipecrs: > But again, I think it might be outside of the scope of that project.

Well, it's not as if they had to add this information. It's on best-effort basis. If they don't add it, we can add it.

Also, something to consider is that the map you created in this repository would only be evaluated on updates of Heroic. This could get out of sync with areweanticheatyet and would not allow to automatically discover the maps as we auto discover the areweanticheatyet.

electron/anticheat/titles_map.ts
arielj: hmmm honestly? I don't remember now (I did this over 2 weeks about on a late night and manually checked the +200 games in the games.json file one by one in the epic store xD)

I think I thought about that and for some reason I decided to keep it simple and only in this direction

Maybe it works just fine doing the check the other way around, but I know that for example I had to add this entry `'The Division': "Tom Clancy's The Division"` to prevent a false match. I suspect that some of the game names are also false matches if you compare them the other way around (I can't confirm if tthis was the case thought, and I wouldn't go over all the games again by hand to check this claim haha)

If somebody is willing to check them all again I'm ok with that change, but it would be just to remove some entries from this mapping and maybe add some if needed, I don't know if it's worth it.

electron/anticheat/titles_map.ts
arielj: Yes, I was aware of the dependency of new heroic updates neede to update the map, but for now it's the best we can do I guess, or at least it's good enough for now, there's always room for improvements.

My main reason to add this feature was actually just fortnite, since we get A LOT of questions of users trying to install fortnite on linux.

Curve: Hey, maintainer of AWACY here, I added the storeIds property to the list.  
Personally I'd like to store the `ID` (or the `appName` because you said it's used internally anyway) along with the `name_in_store` so that we can benefit from this as well ^^.

I'd also offer you that I can add all the missing `name_in_store` properties for Pull-Requests that want to add just the id.

electron/anticheat/utils.ts
Curve: > maybe I have to only check this if not a native mac version of a game when we detect it's a mac os?
> 
> I'm still not sure if areweanticheatyet considers mac games on the status though, the list is built with linux in mind so mac might dispaly false positives/negatives?

As far as I'm aware if the game is not native on Mac Cross-Over is currently the only option - However I don't think that the EAC/BattlEye runtime will work with that - So it's basically: Every game that doesn't work natively on MacOS and includes intrusive Anti-Cheat like EAC / BattlEye won't work.

Some games that don't use strong client side Anti-Cheat or solely Server-Side AntiCheat, for example Star Wars Battlefront (2) actually run in Cross-Over I believe

So AWACY can theoretically be used to determine if a non native game could work on MacOS, but intrusive Anti-Cheats will most likely not work.

Curve:
> > It might be good to include informations like the reference link notes and updates from the JSON file if present.
> 
> I tried, but the information varies between games and it was really difficult to consider all the cases in a consistent design (you have games with multiple updates with different references, the main reference may not even match the last update reference so I can't use the main reference always without the updates) and without making it overwhelming. It felt like too much information, the average user might not care about all the details.
> 
> I did add the link to open areweanticheatyet to see those details for the people that do want to know.

Currently, the Reference is mostly used for games where there is no official word from the developers or there have not been more than 1 changes to the status in the past (Mostly the reference is also the most straight to the point).

Updates is what's used for the most recent additions and contains more precise information than the reference link, if I were to show any information in Heroic I'd go for the latest update.

However the reference links are sometimes not really in sync with the updates at all, I will try to improve that and keep things more consistent

Curve:
> 



> > > It might be good to include informations like the reference link notes and updates from the JSON file if present.
> > 
> > 
> > I tried, but the information varies between games and it was really difficult to consider all the cases in a consistent design (you have games with multiple updates with different references, the main reference may not even match the last update reference so I can't use the main reference always without the updates) and without making it overwhelming. It felt like too much information, the average user might not care about all the details.
> > I did add the link to open areweanticheatyet to see those details for the people that do want to know.
> 
> Currently, the Reference is mostly used for games where there is no official word from the developers or there have not been more than 1 changes to the status in the past (Mostly the reference is also the most straight to the point).
> 
> Updates is what's used for the most recent additions and contains more precise information than the reference link, if I were to show any information in Heroic I'd go for the latest update.
> 
> However the reference links are sometimes not really in sync with the updates at all, I will try to improve that and keep things more consistent

Alright, I've update the list.  
If you still want to display relevant information on the Anti-Cheat Status you should:

- use `reference` if set, it should be preferred over the updates
- in case `reference` **is not set** use the last recorded update

electron/anticheat/titles_map.ts
felipecrs: Since AWACY now supports `storeIds.epic.namespace`, we should remove this file `titles_map.ts` and add the `namespace` id to any mismatch to AWACY instead.

`namespace` ids for games can be easily found on https://database.egdata.app/offers/

arielj: @felipecrs are you planning to add the namespace for the games in AWACY? I know it now supports the feature but I would like to have this feature merged and I don't know if I should wait until that's ready to be used or if I can merge this for now and then a refactor can be done using namespace when that's added to all games

felipecrs: - https://github.com/Starz0r/AreWeAntiCheatYet/pull/564#issue-1299592641

electron/anticheat/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  try {
+    const { data } = await axios.default.get(
+      'https://raw.githubusercontent.com/Starz0r/AreWeAntiCheatYet/master/games.json'
```
felipecrs, line 13:
```suggestion
      'https://raw.githubusercontent.com/Starz0r/AreWeAntiCheatYet/HEAD/games.json'
```

To avoid branch renaming from breaking it.

electron/anticheat/titles_map.ts
arielj: thanks!!

arielj:
I updated this to use the namespace and to show a link to the latest reference for a given status:
image

@felipecrs can you test this again with these changes?

Thanks you both @felipecrs and @Curve for the initiative and the quick addition of this info!

arielj:
> Just an idea: Maybe a new logprefix `Anticheat` ?
I was going to add this but then I saw I only use this in one place and the string always says `AreWeAntiCheatYet` on it. I'll skip this, seems too much to add that new logprefix for this single use.

felipecrs:
> I updated this to use the namespace and to show a link to the latest reference for a given status:
> image
> 
> @felipecrs can you test this again with these changes?
> 
> Thanks you both @felipecrs and @Curve for the initiative and the quick addition of this info!

Hey, great to see this. Unfortunately, I don't have an environment to test it, I trust you.

By the way, I think if you display: 

Source: AreWeAntiCheatYet

I think it would look a little better.

arielj:
> > I updated this to use the namespace and to show a link to the latest reference for a given status:
> > image
> > @felipecrs can you test this again with these changes?
> > Thanks you both @felipecrs and @Curve for the initiative and the quick addition of this info!
> 
> Hey, great to see this. Unfortunately, I don't have an environment to test it, I trust you.
> 
> By the way, I think if you display:
> 
> Source: AreWeAntiCheatYet
> 
> I think it would look a little better.

Thanks, I'll change that, I think it will also make translations easier

Curve:
> I updated this to use the namespace and to show a link to the latest reference for a given status:
> image
> 
> @felipecrs can you test this again with these changes?
> 
> Thanks you both @felipecrs and @Curve for the initiative and the quick addition of this info!

No problem, always an honor ^^

PR #1518, opened 2022-06-23
[UI / Linux] Prevent wine manager list overflow

joephela:
Found a small display bug where the row text would overflow it's container.

Before:
Screenshot from 2022-06-22 21-50-03

After:
Screenshot from 2022-06-22 21-58-19


Nocccer:
@arielj added you, because you are the css expert.
@joephela please change the branch from `main` to `beta`

joephela:
> @arielj added you, because you are the css expert. @joephela please change the branch from `main` to `beta`

Looks like this has already been fixed in beta, my bad.

PR #1520, opened 2022-06-23
[General] Add launcherArgs to native games again

CommandMC:

### DIFF
--- a/electron/legendary/games.ts
+++ b/electron/legendary/games.ts
@@ -589,7 +589,8 @@ class LegendaryGame extends Game {
         gameInfo.app_name,
         ...exeOverrideFlag,
         offlineFlag,
-        launchArguments
+        launchArguments,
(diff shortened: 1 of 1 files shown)


electron/legendary/games.ts
```diff
@@ -589,7 +589,8 @@ class LegendaryGame extends Game {
         gameInfo.app_name,
         ...exeOverrideFlag,
         offlineFlag,
-        launchArguments
+        launchArguments,
+        gameSettings.launcherArgs
```
Nocccer, line 593: Does `gameSettings.launcherArgs` not exist under windows?

CommandMC: Not sure what you mean, the `if (!isWindows)` block ends at line 584. These command parts are always set, regardless of the OS we're on

Nocccer (approved):
Checked back. LGTM

PR #1522, opened 2022-06-24
[Steam-Deck] Always maximize Heroic if Steam-Deck gamemode

CommandMC:
This checks the `XDG_CURRENT_DESKTOP=gamescope` environment variable to detect if Heroic is running in Steam-Deck gamemode and starts Heroic in fullscreen.

---


CommandMC:
One thing that needs further testing is if we really have access to `/run/host/etc/os-release` in Flatpak. I was able to access it just fine on my machine, but it would be great if some more people could verify it

electron/main.ts
```diff
@@ -134,6 +135,11 @@ async function createWindow(): Promise<BrowserWindow> {
     }
   }
 
+  if (isSteamDeckInGamingMode()) {
```
Nocccer, line 138: This should be above line 111 `if(configStore.has('window-props') ...` and if should be replaced on line 111 with else if.

Also skip saving the size in GameMode. Saving is in `mainWindow.on('close', async (e) => { ...`

CommandMC:
After further testing, we can indeed not access that file on Flatpak.  
I'm not quite sure what I have to change to make it accessible either. @flavioislima could you help out here?

electron/utils.ts
```diff
@@ -549,6 +557,47 @@ function quoteIfNecessary(stringToQuote: string) {
...
+  }
+
+  let foundDeckVariant = false
+  for (const line of readFileSync(osReleasePath, 'utf-8').split('\n')) {
+    const [key, value] = `${line}`.split('=')
+    if (key === 'VARIANT_ID') {
+      foundDeckVariant = value === 'steamdeck'
+      break
+    }
+  }
```
Nocccer, line 584: Not 100% if syntax correct, but i think this is cleaner.

```suggestion
  const fileContent = readFileSync(osReleasePath).toString() // 'utf-8' is default
  const foundDeckVariant = fileContent.split('\n').find((line: string) => {
    const [key, value] = line.split('=')
    return key === 'VARIANT_ID' && value ==='steamdeck'
  }
```

CommandMC:
On another note, the code does work fine on AppImage. So it's really just that permission issue that's stopping this

Nocccer (changes requested):
2 small issues. One is from my eyes necessary. The other is just a way i would write the code.

Nocccer:
> 

I would go with a window-size command line option, instead detecting if we are on steam deck. This works for everyone everywhere.

CommandMC:
That would be a good addition, but I'd say having a more automatic option like this is also good

flavioislima:
Do you need to access /etc from the flatpak? That's not possible. I believe the only way this might work is by checking the kernel version, might be that Steam OS had something specific in the name. I'm not sure though.

electron/utils.ts
CommandMC: Each day I learn a new thing about JS

electron/utils.ts
```diff
@@ -549,6 +557,44 @@ function quoteIfNecessary(stringToQuote: string) {
...
+  // Just in case someone is *not* running the Flatpak
+  if (isFlatpak) {
+    osReleasePath = '/run/host' + osReleasePath
+  }
+
+  const fileContent = readFileSync(osReleasePath).toString()
+  const foundDeckVariant = fileContent.split('\n').find((line: string) => {
+    const [key, value] = line.split('=')
+    return key === 'VARIANT_ID' && value === 'steamdeck'
+  })
```
Nocccer, line 581: As longer i think about we could go even cleaner

```suggestion
  const fileContent = readFileSync(osReleasePath).toString()
  const foundDeckVariant = fileContent.split('\n').find((line: string) => {
    return line.replaceAll(' ', '').includes('VARIANT_ID=steamdeck'')
  })
```

flavioislima:
> Well, it's possible if we change the file access permissions. I just don't know which exact permission this is & where to change it

Wdym? Which file? 
Flatpaks cannot see the /etc folder even if we give access to the host system. It's a blacklisted directory:
https://docs.flatpak.org/en/latest/sandbox-permissions.html

CommandMC:
Yes, but if we give it access, we can see it at `/run/host/etc`. I know this because the application that originally implemented this Steam Deck check is also distributed via Flatpak

CommandMC:
Hm, actually, it seems like there's just a `/run/host/os-release` file

CommandMC:
image
Looks like it's normal (search for `os-release` [here](https://docs.flatpak.org/en/latest/flatpak-command-reference.html#))

flavioislima:
> image
> Looks like it's normal (search for `os-release` [here](https://docs.flatpak.org/en/latest/flatpak-command-reference.html#)). TIL

So maybe add the permission specifically to this file with full path.

CommandMC:
You don't need a permission to access that file, as far as I can tell
At least I just reset my permissions locally and it's still there + it's there on a clean install

JakobDev:
It looks like /run/host/os-release is not available on older Flatpak versions, so make sure Heroic doesn't crash if it is not available.

Nocccer:
I adapt @CommandMC solution and just maximize heroic if we run heroic from steam.
What do you think about that solution? Should we add a option to disable this behavior, if wanted?

CommandMC:
Had some ideas on how to clean this up
I really like just checking for Steam instead of the Deck, not sure why I wanted to limit it to that

Nocccer:
> Had some ideas on how to clean this up
> I really like just checking for just Steam instead of the Deck, not sure why I wanted to limit it to that

Yes good changes.

Nocccer (approved):
LGTM. I would still consider a option to disable/enable this behavior. Waht do you think ?

electron/main.ts
```diff
@@ -147,6 +148,14 @@ async function createWindow(): Promise<BrowserWindow> {
     }
   })
 
+  if (isRunningInSteam) {
+    logInfo(
+      'Heroic started via Steam. Switching to fullscreen',
+      LogPrefix.Backend
+    )
+    mainWindow.maximize()
```
arielj, line 156: I wonder if fullscreen makes more sense than maximize https://www.electronjs.org/docs/latest/api/browser-window#winsetfullscreenflag

in a desktop it hides the taskbar and window decoration (steam big picture style), no idea how that works on the steamdeck though

electron/constants.ts
```diff
@@ -30,6 +30,7 @@ const fontsStore = new Store({
 const isMac = platform() === 'darwin'
 const isWindows = platform() === 'win32'
 const isLinux = platform() === 'linux'
+const isRunningInSteam = 'SteamEnv' in env
```
felipecrs, line 33: I wonder if this will cause Epic to open maximized/fullscreen even when not in Game Mode or Big Picture, and if it's really desired to open maximized/fullscreen in such situations.

CommandMC: There is sadly no way to just detect Game Mode/Big Picture

Nocccer: If someone wants to run heroic inside steam he is most likely in big picture or game mode. So this few people who don't want the fullscreen should open a Ticket.

felipecrs: Fair enough.

electron/main.ts
flavioislima: That's a good point, maybe @Nocccer could test the difference on the deck.

Nocccer: Pretty hard to test without a dongle on steam deck. I will instead test big picture mode.

electron/constants.ts
felipecrs: FYI:

> At least on SteamOS, when gamescope runs as a system session, you get `XDG_CURRENT_DESKTOP` set to `gamescope`.

https://github.com/Plagman/gamescope/issues/544#issuecomment-1178863974

Nocccer:
I tested the appiamge on steam deck. It's working.

CommandMC:
Well, can't approve since it's my PR as far as GitHub is concerned, but looks good!

flavioislima (approved):
LGTM!

PR #1529, opened 2022-06-26
[Fix] "default not writable"/"Path might not be writable" error

CommandMC:
`installPath` is set to "default" by default. We were calling `setInstallPath` when this was the case, but the rest of the function still continued with the `default` value (according to some googling, this is because states are only updated once per rendering cycle ([SO thread](https://stackoverflow.com/a/58877875)))
This is now fixed by checking `installPath` again right before passing it to 'checkDiskSpace', and swapping in the correct path if it's `default`

---


imLinguin (approved):
Looks good

PR #1533, opened 2022-06-27
[Linux/MacOS] Split Enviroment Variable and Wrapper Option

Nocccer:
For better adding of enviroment variables a new component "TwoColTableInput" (Better name requested) was written. 
- This component creates a two column table with key and value
- Via Textfields a key and value can be added. 
- Via delete button entry can be removed

- Enviroment Variables and Wrapper are now two seperated options.

@arielj Would be very grateful if you could help me with the component. (Design/Improvments)

image

Launch Command:
`Launch Command: MYENV="Hello World" WINEPREFIX=/home/niklas/Games/Heroic/Prefixes/WheelsofAurelia myexe launch --disable-ui /tmp/.mount_HeroicJjlL37/resources/app.asar.unpacked/build/bin/linux/legendary launch Escolar --language de --wine /home/niklas/.config/heroic/tools/wine/Wine-GE-Proton7-16/bin/wine --wine-prefix /home/niklas/Games/Heroic/Prefixes/WheelsofAurelia`

---


electron/launcher.ts
```diff
@@ -240,18 +244,11 @@ function setupEnvVars(gameSettings: GameSettings) {
...
-      .filter((val) => val.indexOf('=') !== -1)
-      .forEach((envKeyAndVar) => {
-        const keyAndValueSplit = envKeyAndVar.split('=')
-        const key = keyAndValueSplit.shift()
-        const value = keyAndValueSplit.join('=')
-        ret[key] = value
-      })
+  if (gameSettings.enviromentOptions) {
+    gameSettings.enviromentOptions.forEach((envEntry: EnviromentVariable) => {
+      ret[envEntry.key] = quoteIfNecessary(envEntry.value)
```
CommandMC, line 249: I don't think you want `quoteIfNecessary` here

Nocccer: Exactly because of this pr i did it:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1541

electron/config.ts
```diff
@@ -461,7 +461,8 @@ class GlobalConfigV0 extends GlobalConfig {
       maxWorkers: 0,
       minimizeOnLaunch: false,
       nvidiaPrime: false,
-      otherOptions: '',
```
arielj, line 464: we need some way to port the old config to the new config, this would break games for people already using the `otherOptions` config, maybe we should still show and support reading this (maybe using readonly on the input field to prevent changing it) and with a `deprecated` label telling user to use the new method?

electron/launcher.ts
```diff
@@ -240,18 +244,11 @@ function setupEnvVars(gameSettings: GameSettings) {
   if (gameSettings.audioFix) {
     ret.PULSE_LATENCY_MSEC = '60'
   }
-  if (gameSettings.otherOptions) {
```
arielj, line 243: then we could use this only in the `else` for `if (gameSettings.envitonmentOptions)` to use it only as a fallback

src/components/UI/TwoColTableInput/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+                </tr>
+              )
+            })}
+        </tbody>
+      </table>
+      <div className="TableInputDiv">
+        <TextInputField
+          label={`${header.key}:`}
+          value={valueInputs.key}
+          htmlId={''}
```
arielj, line 112: we should have an htmlId here so clicking the label focuses the input that it's related, for accessibility

src/components/UI/TwoColTableInput/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+          {!!rowData.length &&
+            rowData.map((row: ColumnProps, key) => {
+              return (
+                <tr key={key}>
+                  <td>{row.key}</td>
+                  <td>{row.value}</td>
+                  <td>
+                    <SvgButton onClick={() => removeRow(row)}>
+                      <RemoveCircleIcon
+                        data-testid="removeWinePath"
```
arielj, line 94: there's currently no tests, but if we add all the icons will have the same `data-testid`, I would remove this attribute

src/components/UI/TwoColTableInput/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+          value={valueInputs.value}
+          htmlId={''}
+          placeholder={inputPlaceHolder.value}
+          onChange={(event) => {
+            setValueInputs({ ...valueInputs, value: event.target.value })
+          }}
+        />
+        <SvgButton onClick={() => addRow(valueInputs)} className={`is-primary`}>
+          <AddBoxIcon
+            data-testid="addWinePath"
```
arielj, line 129: same here, I would remove this attribute, we don't have tests for this (and I think it's better to use accessibility attributes for testing so we can ensure things are accessible at the same time)

src/screens/Settings/components/OtherSettings/index.tsx
```diff
@@ -280,12 +312,43 @@ export default function OtherSettings({
...
+          rows={getEnvironmentVariables()}
+          onChange={handleEnviromentVariables}
+          inputPlaceHolder={{
+            key: t('options.advanced.placeHolderKey', 'ENVIORMENT'),
+            value: t(
+              'options.advanced.placeHolderValue',
+              'E.g.: Path/To/ExtraFiles'
+            )
+          }}
+          afterInput={info}
```
arielj, line 331: I think we don't want this `info` block after the inputs anymore, right? it's shown multiple times and doesn't match the new elements in that screen

src/components/UI/TwoColTableInput/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+            style={{ color: 'var(--success)', cursor: 'pointer' }}
```
arielj, line 130: we don't need the `cursor: pointer` on these icons, the `SvgButton` element already sets this for the wrapper button tag

src/screens/Settings/components/OtherSettings/index.tsx
```diff
@@ -280,12 +312,43 @@ export default function OtherSettings({
...
+        />
+      )}
+      {!isWin && (
+        <TableInput
+          label={t('options.wrapper.title', 'Wrapper additional arguments')}
```
arielj, line 336: I'm confused since I've never used the previous wrapper option, but how does multiple wrappers work? I imagine it is order-dependent so maybe it should be specified somewhere (using the `info` block) that the wrappers are applied in the order defined in the table?

electron/config.ts
Nocccer: i added a mapper in main.ts. Have a look.

src/screens/Settings/components/OtherSettings/index.tsx
Nocccer: I also never used wrapper. I thin if someone uses it it should be clear to them.

electron/main.ts
```diff
@@ -705,11 +707,48 @@ ipcMain.handle('readConfig', async (event, config_class) => {
...
 ipcMain.handle('requestSettings', async (event, appName) => {
+  // To the changes how we handle env and wrappers
+  // otherOptions is deprectaed and needs to be mapped
+  // to new approach.
+  // Can be removed if otherOptions is removed aswell
+  const mapOtherSettings = (config: AppSettings | GameSettings) => {
+    if (config.otherOptions) {
+      if (config.enviromentOptions.length <= 0) {
+        config.otherOptions
+          .split(' ')
```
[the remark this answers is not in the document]
Nocccer: The mapping is the exact way how we splitted the string before. So it should work 100% with already defined otherOptions.
Spaces in enviroment variables never worked before, thatswhy i refactored it.

CommandMC:
I know I'm pretty late to the party here, but some ideas:
- Separating wrapper arguments with semicolons is a little janky, ideally it would be possible to use quotes like normal. Could be done with something like this:
   ```ts
   const wrapperArgs = '--some-arg --another-arg --path "/path/with/a/spa ce/" --some-more-args' // This should be the string we store
  const splitAtQuote = wrapperArgs.split('"')
  // splitAtQuote[0] = `--some-arg --another-arg --path `
  // splitAtQuote[1] = `/path/with/a/spa ce/`
  // splitAtQuote[2] = ` --some-more-args`
  const finalArgs: string[] = []
  for (const [index, value] of splitAtQuote.entries()) {
    // If the index is even, we have to split the string, if it's odd, we don't
    if (index % 2) {
      finalArgs.push(value)
    } else {
      finalArgs.push(...value.split(' '))
    }
  }
  console.log(finalArgs.filter(Boolean))
   ```
- The "Game Arguments" setting right below these could also get the same treatment

Nocccer:
> I know I'm pretty late to the party here, but some ideas:
> 
> * Separating wrapper arguments with semicolons is a little janky, ideally it would be possible to use quotes to "escape" spaces like normal. Converting from string -> string[] could be done like this:
>   ```ts
>   const wrapperArgs = '--some-arg --another-arg --path "/path/with/a/spa ce/" --some-more-args' // This is the string the user types in
>   const splitAtQuote = wrapperArgs.split('"')
>   // splitAtQuote[0] = `--some-arg --another-arg --path `
>   // splitAtQuote[1] = `/path/with/a/spa ce/`
>   // splitAtQuote[2] = ` --some-more-args`
>   const finalArgs: string[] = []
>   for (const [index, value] of splitAtQuote.entries()) {
>    // If the index is even, we have to split the string, if it's odd, we don't
>    if (index % 2) {
>      finalArgs.push(value)
>    } else {
>      finalArgs.push(...value.split(' '))
>    }
>   }
>   console.log(finalArgs.filter(Boolean))
>   ```
> * The "Game Arguments" setting right below these could also get the same treatment

For me this worser. We need to tell the user to escape paths with spaces. This creates more edge cases than a simple seperator.

CommandMC:
Having to use quotes around paths with spaces is standard practice though. Steam does it like that, Lutris (as far as I know) does it like that, even just the shell does it like that. We would be the one doing it differently.
Substituting *some* spaces (again assuming this is a casual user not knowing the difference between "argument separating space" and "space in file path") takes more getting used to IMO

But y'know, just my 2 cents

Nocccer:
> Having to use quotes around paths with spaces is standard practice though. Steam does it like that, Lutris (as far as I know) does it like that, even just the shell does it like that. We would be the one doing it differently. Substituting _some_ spaces (again assuming this is a casual user not knowing the difference between "argument separating space" and "space in file path") takes more getting used to IMO
> 
> But y'know, just my 2 cents
> 
> > If the user does not use the seperator it still can work
> 
> You'd then pass everything as one argument. So far not a single program I tested supports this

Probably quoting envs is also standard. Sometimes i don't understand you, because else we expect that every path with spaces is quoted by the user or none needs to be quoted. Means we should be strict on every thing the user can add (Envs, Wrapper, Launch Arguments).

Btw launchArguments are not seperated and this broken aswell.

CommandMC:
> Probably quoting envs is also standard

Alright, you got me there. Although, then again, the only real standard to compare against in that regard is Lutris, which also doesn't use quotes in environment variables but does use them in arguments

In the end it's personal preference, I just think it's more natural not to use a separator like ;

Nocccer:
> > Probably quoting envs is also standard
> 
> Alright, you got me there. Although, then again, the only real standard to compare against in that regard is Lutris, which also doesn't use quotes in environment variables but does use them in arguments
> 
> In the end it's personal preference, I just think it's more natural not to use a separator like ;

Does Lutris split the arguments via spaces? If so we can go with your solution. Probably a warning is needed then.

Nocccer:
> > > Probably quoting envs is also standard
> > 
> > 
> > Alright, you got me there. Although, then again, the only real standard to compare against in that regard is Lutris, which also doesn't use quotes in environment variables but does use them in arguments
> > In the end it's personal preference, I just think it's more natural not to use a separator like ;
> 
> Does Lutris split the arguments via spaces? If so we can go with your solution. Probably a warning is needed then.

imLinguin:
I assume Lutris does something like this
```python
import shlex

arguments_array = shlex.split(arguments_string)
```

Nocccer:
> ```python
> split
> ```

Ok i will use npm shlex then aswell and i will add a warning. 
https://www.npmjs.com/package/shlex

CommandMC:
It does. Wrote a little thing that prints out `argv` and added that to Lutris:
image
`sys.argv=['/home/commandmc/src/ArgList/dist/arglist', '--some-arg', '--path', '/path/with/spac es/', '--another-arg']`

Nocccer:
> It does. Wrote a little thing that prints out `argv` and added that to Lutris: image `sys.argv=['/home/commandmc/src/ArgList/dist/arglist', '--some-arg', '--path', '/path/with/spac es/', '--another-arg']`

Ok thanks. I will use shlex because it also removes the quotes. This is pretty good library imo.

CommandMC:
Yup sounds good. Had no idea this library existed to be honest, but it sounds like it's perfect for this

Nocccer:
@CommandMC ok i rewrote everything like lutris is doing it. Also fixed shlex, but will be in a future release.
Shlex PR: https://github.com/rgov/node-shlex/pull/22

Nocccer:
> Codewise looks good. Haven't tested yet though but should work fine. Thanks for refactoring that crossed_swords

I added a small fix and a button to edit entries. Can you have another look?

CommandMC:
With the "Old School Heroic" theme, the table edges are rather hard to make out:
image

Also, the "Variable Name"/"Value"/"Wrapper" text feels a little out of place IMO. Does it really have to be a different font size than all other text on the site? Why not use `<h3>` like the settings sub-heading does?

electron/launcher.ts
```diff
@@ -247,7 +247,8 @@ function setupEnvVars(gameSettings: GameSettings) {
   }
   if (gameSettings.enviromentOptions) {
     gameSettings.enviromentOptions.forEach((envEntry: EnviromentVariable) => {
-      ret[envEntry.key] = envEntry.value
+      // replace(/^"+/, '').replace(/"+$/, '') removes quotes around the value
+      ret[envEntry.key] = envEntry.value.replace(/^"+/, '').replace(/"+$/, '')
```
CommandMC, line 251: Off-loading this into a `removeQuotesIfNecessary` function would be neat

Nocccer:
> With the "Old School Heroic" theme, the table edges are rather hard to make out: image
> 
> Also, the "Variable Name"/"Value"/"Wrapper" text feels a little out of place IMO. Does it really have to be a different font size than all other text on the site? Why not use `<h3>` like the settings sub-heading does?

Before i mess something up in CSS, maybe @arielj can fix this?

arielj:
I added the fix in this PR since it was pretty simple, but the theme in general had a problem with the inputs having the same background as the background, I changed that so now it has better contrast and it looks better and more like the other themes
image

GenocideStomper:
How does this wrapper functionality work? Comparing to Lutris, is it supposed to work like the "Pre-launch script" functionality, or "Command prefix"? Or perhaps something completely different?

Basically I want to run `kde-inhibit --colorCorrect` as long as the game is running, sort of like `mangohud` or `gamemoderun` would be if manually executed.
However when I set `kde-inhibit --colorCorrect` as a wrapper, it executes briefly and exits before the game starts.

Issue #1032, opened 2022-03-01
Disable HTTPS for CDNs


jkms:
### Problem description

Heroic does not appear to be working with my LanCache setup.  I suspect (but have not confirmed) that it's defaulting to SSL instead of HTTP.

### Feature description

Please add an option to force HTTP instead of HTTPS so that my LanCache server can cache downloads

### Alternatives

N/A

### Additional information

LanCache: https://lancache.net/

wiryfuture:
I think this might be an upstream issue as well bc i dont think legendary supports it.

arielj:
Actually, Legendary supports the `--no-https` option with this exact description `Download games via plaintext HTTP (like EGS), e.g. for use with a lan cache`, so it's should be fairly easy to implement (just a switch/boolean and adding the flag to the install command if the value is true)

PR #1538, opened 2022-06-29
[General] add optional '--no-https' flag for legendary args for LanCache compatibility

Keksgesicht:
fixes issue #1032 
Epic Games uses also only http for downloads.
This flag is default disabled, can be enabled via the advanced settings and gets applied to the uses of legendary commands "install", "update" and "repair".
New settings option has translations for english and german. Other languages are missing.

---

### DIFF
--- a/electron/legendary/games.ts
+++ b/electron/legendary/games.ts
@@ -293,12 +293,14 @@ class LegendaryGame extends Game {
       runner: 'legendary',
       status: 'updating'
     })
-    const { maxWorkers } = await GlobalConfig.get().getSettings()
+    const { maxWorkers, downloadNoHttps } =
+      await GlobalConfig.get().getSettings()
     const info = await Game.get(this.appName, 'legendary').getInstallInfo()
     const workers = maxWorkers ? ['--max-workers', `${maxWorkers}`] : []
+    const noHttps = downloadNoHttps ? ['--no-https'] : []
     const logPath = join(heroicGamesConfigPath, this.appName + '.log')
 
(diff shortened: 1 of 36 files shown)


Nocccer (approved):
LGTM

arielj:
How did you add the changes in the translation files? check this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Translations#how-to-add-new-translatable-strings

Currently languages other than english and german will show an empty string instead of english. You should only add the key and the fallback in the code and use `yarn i18next` to udpate the files. Then, to add the geman translation, you have to use weblate instead of the code directly.

electron/legendary/games.ts
```diff
@@ -369,10 +371,12 @@ class LegendaryGame extends Game {
     platformToInstall
   }: InstallArgs): Promise<{ status: 'done' | 'error' }> {
     const { maxWorkers } = await GlobalConfig.get().getSettings()
+    const { downloadNoHttps } = await GlobalConfig.get().getSettings()
```
arielj, line 374: You can do this here, so you don't have to request this twice:

```
const { maxWorkers, downloadNoHttps } = await GlobalConfig.get().getSettings()
```

arielj: same in line 296/7

arielj: and 462/3

CommandMC (approved):

flavioislima (approved):
LGTM! 👍🏽

PR #1539, opened 2022-06-30
Make Shortcuts on the Desktop executable

JakobDev:
I don't know why, but KDE shows a Warn symbol if a .desktop file on the Desktop is not executable and it asks if I want to start the Program when i click it. For some reason this is only needed on the Desktop and not the Menu.

Before:
grafik
grafik

After:
grafik

### DIFF
--- a/electron/shortcuts.ts
+++ b/electron/shortcuts.ts
@@ -41,6 +41,7 @@ Categories=Game;
         writeFile(desktopFile, shortcut, () => {
           logInfo(`Shortcut saved on ${desktopFile}`, LogPrefix.Backend)
         })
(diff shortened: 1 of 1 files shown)


arielj:
Do you know if this is ok with other desktop environments? I can try with Cinnamon which is what I use but links worked fine for me before anyway.

Nocccer:
I will try the old approach on pop_os. I also wanna refactor the shortcut backend.

Nocccer (changes requested):
Use https://www.geeksforgeeks.org/node-js-fs-chmod-method/ from graceful-fs
Instead of execAsync

PR #1541, opened 2022-07-01
[Fix] Quote environment variables in generated command

CommandMC:
Small thing, but we weren't quoting environment variables in the getCommand function, leading to commands like this:
```
WINEPREFIX=/home/commandmc/Games/Heroic/Prefixes/Akalabeth - World of Doom WINEFSYNC=1 ...
```
That's now fixed:
```
WINEPREFIX="/home/commandmc/Games/Heroic/Prefixes/Akalabeth - World of Doom" WINEFSYNC=1 ...
```
This is *not* a change relating to how commands are ran, this is just how they're represented

---

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -616,7 +616,7 @@ function getLegendaryOrGogdlCommand(
         continue
       }
     }
-    formattedEnvVars.push(`${key}=${value}`)
+    formattedEnvVars.push(`${key}=${quoteIfNecessary(value)}`)
   }
 
   return [
--- a/electron/utils.ts
+++ b/electron/utils.ts
@@ -543,7 +543,11 @@ const formatEpicStoreUrl = (title: string) => {
 }
 
 function quoteIfNecessary(stringToQuote: string) {
-  if (stringToQuote.includes(' ')) {
+  if (
+    (stringToQuote.charAt(0) !== '"' ||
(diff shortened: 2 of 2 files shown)


Nocccer:
Please fix `quoteIfNecessary` with this 
```typescript
function quoteIfNecessary(stringToQuote: string) {
  if (stringToQuote.includes(' ')) {
  if (
    (stringToQuote.charAt(0) !== '"' ||
      stringToQuote.charAt(stringToQuote.length - 1) !== '"') &&
    stringToQuote.includes(' ')
  ) {
    return `"${stringToQuote}"`
  }
  return stringToQuote
}
```
Else some values will be double quoted.

CommandMC:
Done  
I'm not quite sure *how* you'd get a double-quoted value here though, especially if we combine this with your env var & wrapper changes

Nocccer:
> Done I'm not quite sure _how_ you'd get a double-quoted value here though, especially if we combine this with your env var & wrapper changes

You get this easy if someone is "clever" enough (like me :D) to quote the envs already in the frontend. And that happend in the old solution aswell. You always put quotes around strings with spaces, without checking if there are already quoted.

`string = "some string with sapces"`
`quoteIfNecessary(string)` returns `""some string with sapces""`

Nocccer (approved):

PR #1548, opened 2022-07-04
[General] Logging & launching improvements

CommandMC:
Main changes:
- Logs are now live (meaning new output is written to the file immediately instead of waiting for game exit)
- If an error happens, part of the log is still written (as much info as possible is provided, to help debug the issue)
- Errors themselves are now logged with a stacktrace (not too helpful in production since function & file names are minified, but might help in development)
- The "Error log"/"Game log" (stdout & stderr) split was removed, it's both in "Game log" again now
- The log now contains a timestamp

Other changes:
- `getRunnerCallWithoutCredentials` is exported again since it's needed for the new log behavior
- Some code cleanup, didn't bother to open a separate PR for just that
---

### DIFF
--- a/electron/games.ts
+++ b/electron/games.ts
@@ -23,8 +23,8 @@ abstract class Game {
 
   abstract appName: string
   abstract window: BrowserWindow
-  abstract getExtraInfo(namespace: string): Promise<ExtraInfo>
-  abstract getGameInfo(installPlatform?: string): Promise<GameInfo>
+  abstract getExtraInfo(): Promise<ExtraInfo>
+  abstract getGameInfo(installPlatform?: string): GameInfo
   abstract getInstallInfo(installPlatform?: string): Promise<InstallInfo>
   abstract getSettings(): Promise<GameSettings>
   abstract hasUpdate(): Promise<boolean>
@@ -39,7 +39,7 @@ abstract class Game {
   abstract syncSaves(arg: string, path: string): Promise<ExecResult>
   abstract uninstall(): Promise<ExecResult>
   abstract update(): Promise<{ status: 'done' | 'error' }>
-  abstract isNative(): Promise<boolean>
+  abstract isNative(): boolean
   abstract runWineCommand(
     command: string,
     altWineBin?: string,
--- a/electron/games.ts
+++ b/electron/games.ts
@@ -22,7 +22,7 @@ abstract class Game {
   public get logFileLocation() {
     return join(
       heroicGamesConfigPath,
-      `${this.appName.split(' ')[0]}-lastPlay.log`
+      `${this.appName}-lastPlay.log`
     )
   }
 
--- a/electron/games.ts
+++ b/electron/games.ts
@@ -5,7 +5,6 @@ import {
   GameSettings,
   InstallArgs,
   InstallInfo,
-  LaunchResult,
   Runner
 } from './types'
 
@@ -20,6 +19,12 @@ abstract class Game {
       return GOGGame.get(appName)
     }
   }
+  public get logFileLocation() {
+    return join(
+      heroicGamesConfigPath,
+      `${this.appName.split(' ')[0]}-lastPlay.log`
+    )
+  }
 
   abstract appName: string
   abstract window: BrowserWindow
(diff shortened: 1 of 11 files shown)


electron/games.ts
```diff
@@ -20,26 +19,32 @@ abstract class Game {
       return GOGGame.get(appName)
     }
   }
+  public get logFileLocation() {
+    return join(
+      heroicGamesConfigPath,
+      `${this.appName.split(' ')[0]}-lastPlay.log`
```
Nocccer, line 25:
```suggestion
      `${this.appName}-lastPlay.log`
```

electron/gog/games.ts
```diff
@@ -392,35 +395,43 @@ class GOGGame extends Game {
...
+      wrappers,
+      join(...Object.values(getGOGdlBin()))
     )
+    appendFileSync(
+      this.logFileLocation,
+      `Launch Command: ${fullCommand}
+
+Game Log:
+`
+    )
```
Nocccer, line 410: Maybe this is better readable?
```suggestion
    appendFileSync(
      this.logFileLocation,
      [`LAUNCH COMMAND: ${fullCommand}`, 'GAME LOG:'].join('\n'))
```

electron/legendary/games.ts
```diff
@@ -659,15 +657,29 @@ class LegendaryGame extends Game {
...
+      commandEnv,
+      wrappers,
+      join(...Object.values(getLegendaryBin()))
+    )
+    appendFileSync(
+      this.logFileLocation,
+      `Launch Command: ${fullCommand}
+
+Game Log:
+`
```
Nocccer, line 671: Same here. Maybe we think of a layout for games log aswell? Like general log has it? 
Calculated tabs/spaces etc.

electron/main.ts
```diff
@@ -814,60 +813,50 @@ ipcMain.handle(
...
+    writeFileSync(
+      game.logFileLocation,
+      `System Info:
 ${await getSystemInfo()}
 
-Game Settings: ${JSON.stringify(gameSettings, null, '\t')}
+Game Settings: ${JSON.stringify(await game.getSettings(), null, '\t')}
+
+Game launched at: ${startPlayingDate}
+
```
Nocccer, line 834: Don't like the way of writing this multiline strings. There should be a better way like string arrray etc.

E.g.:

```typescript
['first line', 'second line'].join('\n')
```

Nocccer (changes requested):
Have some suggestions and ideas we need to talk about.

electron/games.ts
CommandMC: Well, this was how the filename was originally generated. But I do agree that the split is useless, AppNames can't have spaces

electron/legendary/games.ts
CommandMC: In terms of a "layout" for these logs: I'd say the current format is *fine* for what it needs to do. No need to over-complicate it

electron/legendary/games.ts
```diff
@@ -665,10 +665,7 @@ class LegendaryGame extends Game {
     )
     appendFileSync(
       this.logFileLocation,
-      `Launch Command: ${fullCommand}
-
-Game Log:
-`
+      `Launch Command: ${fullCommand}\n\nGame Log:\n`
```
Nocccer, line 668:
```suggestion
    `Launch Command: ${fullCommand}\n` + 
          '\n' + 
          'Game Log:\n'
```

Nocccer (changes requested):
I left some suggestions. For me it is important to have readable strings in the source code. A string with 1-2 linebreaks is maybe readable and therefore fine, but multine strings should be fast readable, especially with alot of linebreaks.

electron/legendary/games.ts
CommandMC: Linting automatically changes this to `` `Launch Command: ${fullCommand}\n` + '\n' + 'Game Log:\n' ``

CommandMC:
> What do you think?

I honestly don't really mind. Whatever works works for me, if you want a specific style then I'll use that specific style

Nocccer (approved):
Sorry for this back adn forth.
I think the best way is to write code always in mind someone else needs to understand it possible with the least effort. So readable code is the key.

But LGTM now.

PR #1559, opened 2022-07-09
Fix: Don't 'handleProtocol' if mainWindow doesn't exist yet

arielj:
This fixes another issue I found on my jurney to add shortcuts on MacOS. On this pr https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1501 I was testing opening heroic with arguments, but I didn't test opening heroic by letting the OS handle the `heroic://launch` url.

When I tested that, I notices that the `open-url` event happens before the mainWindow object is even created, and it was trying to `handleProtocol` with no mainWindow object. Then the `process.argv` variable was empty in the other `handleProtocol` in the `frontendReady` callback becuase it was not passed as an argument to the process by the OS.

In this PR I'm checking first if the mainWindow is exists, if it does it just calls `handleProtocol`, it if doesn't then it stores the url somewhere so it can be used in the `frontendReady` callback.

This issue happened only when executing a protocol when heroic was closed.

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -205,6 +205,7 @@ async function createWindow(): Promise<BrowserWindow> {
 // Some APIs can only be used after this event occurs.
 let appIcon: Tray = null
 const gotTheLock = app.requestSingleInstanceLock()
+let openUrlArgument = ''
 
 const contextMenu = () => {
   const recentGames: Array<RecentGame> =
@@ -433,7 +434,7 @@ ipcMain.on('Notify', (event, args) => {
 })
 
 ipcMain.on('frontendReady', () => {
-  handleProtocol(mainWindow, process.argv)
+  handleProtocol(mainWindow, [openUrlArgument, ...process.argv])
 })
 
 // Maybe this can help with white screens
@@ -510,7 +511,11 @@ app.on('window-all-closed', () => {
 
 app.on('open-url', (event, url) => {
   event.preventDefault()
-  handleProtocol(mainWindow, [url])
+  if (mainWindow) {
+    handleProtocol(mainWindow, [url])
+  } else {
+    openUrlArgument = url
+  }
 })
 
 ipcMain.on('openFolder', async (event, folder) => openUrlOrFile(folder))


electron/main.ts
```diff
@@ -433,7 +434,7 @@ ipcMain.on('Notify', (event, args) => {
 })
 
 ipcMain.on('frontendReady', () => {
-  handleProtocol(mainWindow, process.argv)
+  handleProtocol(mainWindow, [openUrlArgument, ...process.argv])
```
arielj, line 437: because of how handleProtocol works, this is fine, since it will go over all the elements of this array until if finds a heroic procotol, so if this is an empty string it's just skipped

flavioislima (approved):
Looks good!

Nocccer (approved):
LGTM

imLinguin (approved):

CommandMC:
According to [the Electron docs](https://www.electronjs.org/docs/latest/api/app#event-will-finish-launching), we should add the `open-url` event handler a little differently than we're doing right now (at least as far as I can tell). Would changing that maybe delay the handling long enough?
Besides, I don't think there's a reason we're calling the Frontend in the protocol, I think that can also be improved

This feels more like a band-aid fix to be honest

arielj:
> According to [the Electron docs](https://www.electronjs.org/docs/latest/api/app#event-will-finish-launching), we should add the `open-url` event handler a little differently than we're doing right now (at least as far as I can tell). Would changing that maybe delay the handling long enough?

Not sure, because we are not only depending on the front end, we are depending on the react app to be ready in the frontend, not just electron thinking the app is ready, I think the `will-finish-launching` event will still fire too early for us.

> Besides, I don't think there's a reason we're calling the Frontend in the protocol, I think that can also be improved. This feels more like a band-aid fix to be honest

I agree, the underlaying issue is a different one and this is a workaround on the wrong thing we are doing so at least it works, it fixes something for now but I agree that a refactor of how we launch games should be done at some point to not have to wait for the frontend

flavioislima:
> Besides, I don't think there's a reason we're calling the Frontend in the protocol, I think that can also be improved

The thing is that if we don't show anything on the frontend, in case the game doesn't run for some reason or we get some error, for instance, during save-sync or so on, the user will get no feedback and then will think that Heroic is broken.

flavioislima (approved):

CommandMC (approved):
> the underlaying issue is a different one and this is a workaround on the wrong thing we are doing so at least it works, it fixes something for now but I agree that a refactor of how we launch games should be done at some point to not have to wait for the frontend

Yup, good to see we're on the same page then  
Still, let's merge this for now

PR #1564, opened 2022-07-11
[Linux] Added option to prefer system libs

Nocccer:
- [x] Added option which set to on, will use system libs, instead wine-ge, wine-lutris libraries.
- [x] Adapt `LD_LIBRARY_PATH` for wine-ge, wine-lutris
- [x] Define `LD_PRELOAD` if undefined.
    - this will prevent Fall Guys crashing with std::logic_error and maybe other games with eac (thanks to [@Diyou](https://github.com/Diyou)) 
    - https://user-images.githubusercontent.com/61798668/178372186-d55df81d-efee-44dc-9190-2839a58bb59b.mp4

---


src/screens/Settings/components/WineSettings/index.tsx
```diff
@@ -362,6 +366,24 @@ export default function WineSettings({
...
+              handleChange={togglePreferSystemLibs}
+              title={t('setting.preferSystemLibs', 'Prefer system librarys')}
+            />
+
+            <FontAwesomeIcon
+              className="helpIcon"
+              icon={faCircleInfo}
+              title={t(
+                'help.preferSystemLibs',
+                'Wine alternatives (Wine-Ge, Wine-Lutris) is shipped with its library dependencies. By enabling this option wine will use the system libraries. Warning! Malfunctions can occur if dependencies are not met.'
```
CommandMC, line 383:
```suggestion
                'Custom Wine versions (Wine-GE, Wine-Lutris) are shipped with their library dependencies. By enabling this option, these shipped libraries will be ignored and Wine will load system libraries instead. Warning! Issues may occur if dependencies are not met.'
```

electron/launcher.ts
```diff
@@ -240,6 +240,30 @@ function setupEnvVars(gameSettings: GameSettings) {
...
+        // append wine libs at the beginning
+        ret.LD_LIBRARY_PATH = [lib32, lib64, process.env.LD_LIBRARY_PATH].join(
+          ':'
+        )
+      } else {
+        logError(
+          [
+            `Couldn't find all library folders of ${gameSettings.wineVersion.name}!`,
+            `Missing ${lib32} or`,
+            `${lib64}!`,
```
CommandMC, line 260:
```suggestion
            `Missing ${lib32} or ${lib64}!`,
```

electron/launcher.ts
```diff
@@ -240,6 +240,30 @@ function setupEnvVars(gameSettings: GameSettings) {
...
+            `Fallback to system librarys!`
```
CommandMC, line 261:
```suggestion
            `Falling back to system libraries!`
```

CommandMC (commented):
In terms of code, looks good to me. Will give this a test on a fresh VM without dependencies tomorrow

Nocccer:
> In terms of code, looks good to me. Will give this a test on a fresh VM without dependencies tomorrow

Nice thank you. What do you think about moving `LD_PRELOAD` to `setupENVVars` ?

CommandMC:
Note that since you ran `yarn i18next`, you might have to update the translation files manually as well
I'd recommend only running the command once you're absolutely sure no translations will change

CommandMC:
> What do you think about moving `LD_PRELOAD` to `setupEnvVars` ?

Yeah I was wondering why you did it in `callRunner` itself, it would definitely fit into `setupEnvVars` better

Nocccer:
> > What do you think about moving `LD_PRELOAD` to `setupEnvVars` ?
> 
> Yeah I was wondering why you did it in `callRunner` itself, it would definitely fit into `setupEnvVars` better

I  wanted to place it as near as possible to spawn for testing.

electron/launcher.ts
```diff
@@ -240,6 +240,29 @@ function setupEnvVars(gameSettings: GameSettings) {
   if (gameSettings.audioFix) {
     ret.PULSE_LATENCY_MSEC = '60'
   }
+  if (
+    !gameSettings.preferSystemLibs &&
+    gameSettings.wineVersion.type === 'wine'
+  ) {
+    if (gameSettings.wineVersion.bin) {
+      const lib32 = join(gameSettings.wineVersion.bin, '../../lib')
```
flavioislima, line 248: another idea for dealing with that is adding the `lib` and `lib32` to the `wineVersion` object when getting the wine list. I think that would make it more flexible as well.
We also need to check the libraries for proton, or not? 🤔

Nocccer: proton-ge is handling it by itself.If you open the proton executable in vscode you will see that is just a script.It sets the LD_LIBRARY_PATH in there.

CommandMC: Actually, I agree. We need that same info in `runWineCommand`

Nocccer:
You need to switch wine version once, else the lib and lib32 are not stored correctly.

CommandMC (approved):

CommandMC:
Yup, looks even better now.
Setting up a testing VM for this right now, gimme a bit

CommandMC:
Hm, the "common" `No such file or directory` error seems to still happen. Log from testing VM:
[Quail-lastPlay.log](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/files/9096195/Quail-lastPlay.log)

Nocccer:
@CommandMC maybe we need to add wine bin folder aswell?

CommandMC:
Hm, I don't think that would help anything

Controversial take here:
In my opinion, we shouldn't support this dependency-less usage of Heroic on anything other than the Flatpak. And luckily, this issue does not happen there.
We should instead add Wine's dependencies to our own (in the repos we're shipping to). [Lutris does it like that](https://archlinux.org/packages/community/any/lutris/#pkgdeps), [Bottles does it like that](https://aur.archlinux.org/packages/bottles#pkgdeps), [Steam does it like that](https://archlinux.org/packages/multilib/x86_64/steam/#pkgdeps) (in fact both Lutris and Bottles just depend on `wine` itself)

Now this doesn't mean this PR is irrelevant. I'd say it doesn't *hurt* to prefer these shipped libraries, and maybe we can one day figure out why this error happens and add *something* that works around it. For now though, doing this shot-in-the-dark fixing when there's a clear solution just doesn't make sense IMO

Nocccer:
> Hm, I don't think that would help anything
> 
> Controversial take here: In my opinion, we shouldn't support this dependency-less usage of Heroic on anything other than the Flatpak. And luckily, this issue does not happen there. We should instead add Wine's dependencies to our own (in the repos we're shipping to). [Lutris does it like that](https://archlinux.org/packages/community/any/lutris/#pkgdeps), [Bottles does it like that](https://aur.archlinux.org/packages/bottles#pkgdeps), [Steam does it like that](https://archlinux.org/packages/multilib/x86_64/steam/#pkgdeps) (in fact both Lutris and Bottles just depend on `wine` itself)
> 
> Now this doesn't mean this PR is irrelevant. I'd say it doesn't _hurt_ to prefer these shipped libraries, and maybe we can one day figure out why this error happens and add _something_ that works around it. For now though, doing this shot-in-the-dark fixing when there's a clear solution just doesn't make sense IMO

I thought wine-ge ships all necessary libs to run it. I don't understand why we need runtimes if wine-ge can include all of them.

CommandMC:
> I thought wine-ge ships all necessary libs to run it. I don't understand why we need runtimes if wine-ge can include all of them.

This is a *completely* uneducated guess here, but it might be the case that just having the files isn't enough.

I don't really know how all of the Wine dependencies interact with the host system, but there's a similar thing where I *do* know how it works: GameMode.  
It's, according to the [official repo](https://github.com/FeralInteractive/gamemode), a `daemon/lib combo`. So there's a library (`libgamemodeauto.so.0`) that then communicates with a daemon running in the background. If this daemon isn't running, even if you have the library, it won't work. Something similar might be going on here

Nocccer:
> > I thought wine-ge ships all necessary libs to run it. I don't understand why we need runtimes if wine-ge can include all of them.
> 
> This is a _completely_ uneducated guess here, but it might be the case that just having the files isn't enough.
> 
> I don't really know how all of the Wine dependencies interact with the host system, but there's a similar thing where I _do_ know how it works: GameMode. It's, according to the [official repo](https://github.com/FeralInteractive/gamemode), a `daemon/lib combo`. So there's a library (`libgamemodeauto.so.0`) that then communicates with a daemon running in the background. If this daemon isn't running, even if you have the library, it won't work. Something similar might be going on here

GE suggest to use the system libs. Strange. But we could add wine as a dependecy for deb, pacman and rpm in the package.json. This will make sure people install wine, if they wanna install heroic.

CommandMC:
> GE suggest to use the system libs

He does? Where? As far as I know, Wine-GE was built to run inside the Lutris Runtime:
image

> But we could add wine as a dependecy for deb, pacman and rpm in the package.json. This will make sure people install wine, if they wanna install heroic via this three options.

I'd say that would be a sensible thing to do, yes

Nocccer:
> > GE suggest to use the system libs
> 
> He does? Where? As far as I know, Wine-GE was built to run inside the Lutris Runtime:
> image
> 
> > But we could add wine as a dependecy for deb, pacman and rpm in the package.json. This will make sure people install wine, if they wanna install heroic via this three options.
> 
> I'd say that would be a sensible thing to do, yes

In the summary section he suggest system libs over runtime libs:
https://www.gloriouseggroll.tv/how-to-get-out-of-wine-dependency-hell/

CommandMC:
I believe that guide is mostly unmaintained by now. Judging by the discord messages, his stance on the matter has changed.
Well, we're not here to gossip about other people.

The Steam Runtime not working correctly is most likely an issue in our implementation, maybe we're not setting a variable or not considering an edge case. I don't really bother with it anymore since Wine-GE is just the better option

Nocccer:
@CommandMC i tested the app image of this PR on my steam deck in desktop mode. I could run WheelsOfAurelia with wine-ge. 
Just wonder what lib is missing on your end? Can you try WheelsOfAurelia aswell? I think the error is game depended, because some games need optional libs, which are not shipped by wine-ge

Nocccer:
Before we merge this with wine as dependency for deb, pacman and rpm, i wanna know if @flavioislima as a argument against it?

Nocccer:
> It's definitely not game-specific, the command that's failing is the `wineboot` one (that's ran for every game to create the prefix). I also cannot even run the `wine` executable on the Terminal (getting the same error there). This is on a fresh EndeavourOS install (which is pretty minimal in terms of pre-installed stuff), if you have the time, feel free to test yourself.
> 
> Although it is good to know that it isn't failing on the Deck. Maybe I'll try to find out which exact dependency it is that makes it work later

I tested on my PC with just setting `LD_LIBRARY_PATH` to only the wine lib folders and remove all system folders.
Game still starts. Maybe it is some utils that are missing on your VM?

CommandMC:
There's the next thing I'm wondering about: On the VM, it just says this:
```
$ ldd ./wine
        not a dynamic executable
```
Everything behaves like the file doesn't even exist, but it's there:
```
$ file ./wine
./wine: ELF 32-bit LSB executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=3f2fbf89cf13086a94e8053b26ddd80c2e003039, stripped
```

Nocccer:
> not a dynamic executable

Do you have a 64 bit VM? I don't know if 32 and 64Bit makes a difference here.

CommandMC:
I don't think you can just *make* a 32-bit VM, it's up to your host CPU to decide the architecture

`lib32-libelf` is the culprit, installing that makes `ldd` work fine & the executable launches

Nocccer:
> I don't think you can just _make_ a 32-bit VM, it's up to your host CPU to decide the architecture
> 
> `lib32-libelf` is the culprit, installing that makes `ldd` work fine & the executable launches

I enabled the wrong Bios option. Now it works. Will test with Ubunutu 22.04 64Bit

CommandMC:
> I don't know if it is possible to add this libs as dependencies for deb, pacman and rpm, because they are 32bit libs?

As far as I know, you can depend on 32-bit packages just fine in the AUR (and in official Arch repos).

A more hacky solution would be to invoke the package manager and see if the package is installed:
```
$ pacman -Q lib32-libelf
lib32-libelf 0.187-1
$ echo $?
0
$ pacman -Q somepackagethatisntinstalled
error: package 'somepackagethatisntinstalled' was not found
$ echo $?
1
```

Nocccer:
> > I don't know if it is possible to add this libs as dependencies for deb, pacman and rpm, because they are 32bit libs?
> 
> As far as I know, you can depend on 32-bit packages just fine in the AUR (and in official Arch repos).
> 
> A more hacky solution would be to invoke the package manager and see if the package is installed:
> ```
> $ pacman -Q lib32-libelf
> lib32-libelf 0.187-1
> $ echo $?
> 0
> $ pacman -Q somepackagethatisntinstalled
> error: package 'somepackagethatisntinstalled' was not found
> $ echo $?
> 1
> ```

You can add every dependency to electron-builder in the package.json. This will install the package alongside heroic. But i wonder what happens if i386 is not enabled.

PR #1566, opened 2022-07-12
[Fix] Not Checking games update on all platforms on refresh

flavioislima:
Small fix that will check all games updates instead of only one platform when clicking the refresh button and on startup.

---

### DIFF
--- a/src/components/UI/ActionIcons/index.tsx
+++ b/src/components/UI/ActionIcons/index.tsx
@@ -16,6 +16,7 @@ import { useTranslation } from 'react-i18next'
 import ContextProvider from 'src/state/ContextProvider'
 import FormControl from '../FormControl'
 import './index.css'
+import { Runner } from 'src/types'
 import { Visibility, VisibilityOff } from '@mui/icons-material'
 import classNames from 'classnames'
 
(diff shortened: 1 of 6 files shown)


CommandMC:
Hm, maybe, instead of always ignoring the runner, we could refresh both libraries when on "All", and only the respective runner when on "Epic"/"Gog"

flavioislima:
> Hm, maybe, instead of always ignoring the runner, we could refresh both libraries when on "All", and only the respective runner when on "Epic"/"Gog"

I just don't see a use case for that now since the libraries are unified. But ok, will do the change 👍🏽

CommandMC (approved):
IMO, when someone is filtering the library to only display Epic games, and then presses refresh, it should also only refresh their Epic games.
At least I'd imagine that's what the user expects

flavioislima:
> IMO, when someone is filtering the library to only display Epic games, and then presses refresh, it should also only refresh their Epic games. At least I'd imagine that's what the user expects

Yes, you are right, I am not looking to the right place here. I just need to fix the list of updatable games we keep on the storage. That's the issue, don't need to touch the `refreshLibrary` method.

flavioislima:
@CommandMC now I kept the old behavior and just fixed the list of games that were overridden when using a filter.

PR #1568, opened 2022-07-12
Clean up `runWineCommand` and other functions in `launcher.ts`

CommandMC:
- `getWineEnvSetup` was removed. Not sure why I split it and `setupWineEnvVars` up like that. This now means that all other game settings are also considered when using the `Run EXE on Prefix` option, which I'd say makes more sense
- `runWineCommand` itself now takes a `Game` instead of just `GameSettings`, this is mainly done to easily get the game install folder name (didn't want to add *another* parameter just for that)
- `runWineCommand` no longer takes an `altWineBin`, this was never used and I don't think it could've been that helpful

---


Nocccer (approved):
LGTM

flavioislima (approved):
LGTM 👍🏽

PR #1569, opened 2022-07-12
[UI] Hide game title when the card is not focused

flavioislima:
This brings the old behavior for the game title. Instead of showing the title all the time, it will be shown only on hover/focus.

---


CommandMC:
Thanks, I think most people (including myself) prefer this design.

I have 4 ideas relating to this:
1. The title is a little bit high up now, seems like it's spanning half the game image now. I'd say that's a little too much:  
   image
2. Not sure if it was introduced with this PR, but just a little oddity I've noticed: The game image is cut off by the settings/play icons:  
   image
3. Would it be possible to hide the settings/play icons as well, so it's just the image?
4. If possible, adding a setting to control this behavior would be great. Maybe even have the distinction between "Hide title"/"Hide title and buttons"/"Hide nothing" there?

flavioislima:
> Thanks, I think most people (including myself) prefer this design.
> 
> I have 4 ideas relating to this:
> 
> 1. The title is a little bit high up now, seems like it's spanning half the game image now. I'd say that's a little too much:
>    image
> 2. Not sure if it was introduced with this PR, but just a little oddity I've noticed: The game image is cut off by the settings/play icons:
>    image
> 3. Would it be possible to hide the settings/play icons as well, so it's just the image?
> 4. If possible, adding a setting to control this behavior would be great. Maybe even have the distinction between "Hide title"/"Hide title and buttons"/"Hide nothing" there?

1. Yeah I can try that
2. I believe it was after I changed the design or after the PR that changed the size of the cards to make them use the whole space.
3. Then it's like we have it on Stable. I think having the icons is a good thing since with it you have the info already that you have an update, or the game is installed or not. these changes were made also to improve the accessibility,
4. I think that's too much, as more settings we add more complicated things become to end-user and for us to maintain. I think we should avoid settings for the UI.

Bottom line is that most people liked the new changes, except for the title being shown all the time, at least this was the feedback we got on twitter and on discord.

CommandMC:
Ah, fair enough. If it's what the people want, do it

Well, as soon as those two issues (1. and 2.) are fixed, I'm happy with this

arielj:
@flavioislima can I push a commit here to fix issues 1 and 2 reported by CommandMC?

flavioislima:
> @flavioislima can I push a commit here to fix issues 1 and 2 reported by CommandMC?

Just saw your message now.
I sent I few fixes, don't know if they are really good.
If you find a better solution feel free to send a commit :D

arielj:
> > @flavioislima can I push a commit here to fix issues 1 and 2 reported by CommandMC?
> 
> Just saw your message now. I sent I few fixes, don't know if they are really good. If you find a better solution feel free to send a commit :D

Done! I removed some fixed numbers and calcs. I also removed the limit of 4 lines of text for the title, the longest title I found was tiny tina's and it was 5 lines, they fit just fine inside the card, I don't think we need that limit

CommandMC (approved):

PR #1580, opened 2022-07-16
[Fix] Delete duplicate ThemeSelector element

redromnon:
I found that ThemeSelector is present in both Accessibility and General Settings screens. Removed the extra one from General Settings.

I can remove it from the Accessibility screen instead. 

---

### DIFF
--- a/src/screens/Settings/components/GeneralSettings/index.tsx
+++ b/src/screens/Settings/components/GeneralSettings/index.tsx
@@ -5,7 +5,6 @@ import { useTranslation } from 'react-i18next'
 import ContextProvider from 'src/state/ContextProvider'
 import { InfoBox, SelectField, ToggleSwitch } from 'src/components/UI'
 import LanguageSelector from 'src/components/UI/LanguageSelector'
-import { ThemeSelector } from 'src/components/UI/ThemeSelector'
 
 import { IpcRenderer } from 'electron'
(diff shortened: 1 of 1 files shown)


arielj:
Hmmm I added it in the Accessibility screen without removing it from General Settings because I didn't know where people would look for it:
- people with accessibility needs may look for it in that screen to change colors
- people just looking for a visual change without any accessibility need may expect it in the general settings

I intentiontionally left it in both places since.

It can be removed if people think it's useless to have it in both places, but I thought it made sense in both places (like we currently show the language selector in both the login screen and the settings)

redromnon:
It perfectly makes sense to place the Language Selector on the Login screen too. As for the Theme selector is concerned, I feel one should be sufficient. But then again, as you mentioned, let's see what the others have to say about this.

arielj:
> It perfectly makes sense to place the Language Selector on the Login screen too. As for the Theme selector is concerned, I feel one should be sufficient. But then again, as you mentioned, let's see what the others have to say about this.

maybe ask this in the UX-UI channel in discord for more visibility? since it's related to user experience

flavioislima:
I think it's fine to have it in two different places like it is today. But if we want to remove one, perhaps the one from settings would be better since we should make it easier to be found for people with disabilities.

flavioislima (approved):

PR #1583, opened 2022-07-16
[Windows] Prompt for Visual C++ Runtime installation when not found

CommandMC:
On Startup, we now check if the 64- and 32-bit-versions of the [Visual C++ Runtime](https://docs.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#visual-studio-2015-2017-2019-and-2022) are installed and display a message if they aren't. These runtimes are required for a few games (Fall Guys for example), but they're expected to already be installed (*I think* the EGL installs them when it's installed?)

Some remarks:
- The way I'm checking for the runtimes can probably have false positives (for example if someone has the actual development environment installed), although I don't think that's too big of an issue since you should still be fine if you have *more* than just the runtimes
- The messages in the dialog boxes could be improved a bit (which is why I didn't run `yarn i18next` yet)
- Displaying the message on each launch might be a little annoying, although I'd say the runtimes are important enough that everyone should have them installed

---

### DIFF
--- a/electron/utils.ts
+++ b/electron/utils.ts
@@ -2,7 +2,7 @@ import { Runner } from './types'
 import { WineInstallation } from './types'
 import * as axios from 'axios'
 import { app, dialog, net, shell, Notification, BrowserWindow } from 'electron'
-import { exec, spawnSync } from 'child_process'
+import { exec, spawn, spawnSync } from 'child_process'
 import { existsSync, rmSync, stat } from 'graceful-fs'
 import { promisify } from 'util'
 import i18next, { t } from 'i18next'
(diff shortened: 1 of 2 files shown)


electron/utils.ts
```diff
@@ -577,6 +577,73 @@ function killPattern(pattern: string) {
...
+ * Only works on Windows of course
+ */
+function detectVCRedist(mainWindow: BrowserWindow) {
+  if (!isWindows) {
+    return
+  }
+
+  const detectedVCRInstallations: string[] = []
+  let stderr = ''
+  const child = spawn('wmic', [
```
flavioislima, line 593: What is this WMIC? is the same as: https://docs.microsoft.com/en-us/windows/win32/wmisdk/wmic ?
BEcause it says that was deprecated on Win10 and perhaps wont work on Win11.

CommandMC: It's still working fine on W11, in fact, the `check-disk-space` library we're using uses it as well  
We could also add [an extra library](https://www.npmjs.com/package/node-wmi) to handle WMI calls, if you want to avoid using wmic

flavioislima: Is up to you, if its still working then I am good. But if using PS is more future proof, then would be better to use it.

flavioislima: So Lets merge it then and after the beta we can rewrite it.

flavioislima (approved):

PR #1587, opened 2022-07-17
make desktop shortcuts executable 

Etaash-mathamsetty:
This is basically a continuation of #1539, to use the graceful-fs chmod function
unfortunately I can't test this because heroic games launcher doesn't even create a desktop shortcut or a start menu shortcut even when I enabled the options
edit: got the desktop shortcuts working, and they are executable, I do think that someone else should also test this


Etaash-mathamsetty:
it's not letting me run yarn lint lol, how do I run it

electron/shortcuts/shortcuts/shortcuts.ts
```diff
@@ -42,6 +42,12 @@ Categories=Game;
         writeFile(desktopFile, shortcut, () => {
           logInfo(`Shortcut saved on ${desktopFile}`, LogPrefix.Backend)
         })
+        chmod(desktopFile, '+x', () => {
```
arielj, line 45: I think it's better call this insdie the `writeFile` callback, because if the write takes longer than expected the chmod command will try to set permissions to a yet non existing file

Etaash-mathamsetty: fixed!

arielj:
I was testing this and I see this error in the console when trying to create a shortcut:

```
(21:25:32) ERROR:   [Backend]:         TypeError: The argument 'mode' must be a 32-bit unsigned integer or an octal string. Received '+x'
```

The shortcut is created, but it's not executable.

If it helps, I use Linux Mint with Cinnamon, maybe the `+x` is not cross-desktop-environment compatible?

Etaash-mathamsetty:
> I was testing this and I see this error in the console when trying to create a shortcut:
> 
> ```
> (21:25:32) ERROR:   [Backend]:         TypeError: The argument 'mode' must be a 32-bit unsigned integer or an octal string. Received '+x'
> ```
> 
> The shortcut is created, but it's not executable.
> 
> If it helps, I use Linux Mint with Cinnamon, maybe the `+x` is not cross-desktop-environment compatible?

hmm what should I set it to then? 700 or 777?

arielj:
> > I was testing this and I see this error in the console when trying to create a shortcut:
> > ```
> > (21:25:32) ERROR:   [Backend]:         TypeError: The argument 'mode' must be a 32-bit unsigned integer or an octal string. Received '+x'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The shortcut is created, but it's not executable.
> > If it helps, I use Linux Mint with Cinnamon, maybe the `+x` is not cross-desktop-environment compatible?
> 
> hmm what should I set it to then? 700 or 777? 777 is more foolproof, I will go with that

Tested with `777` and it's setting this permissions when I check with `ls -l`: `-r----x--t`, I think `777` is not correct.

I tried with `0x777` and it sets this `-r-xrwSrwt` which is a bit better I guess but looks really weird (like my user doesn't have write access?)

I think what you want to pass there is `511` which is this sequence of bits `111111111`. (I would add a comment in the code so it's clear where that 511 comes from)

Etaash-mathamsetty:
> > > I was testing this and I see this error in the console when trying to create a shortcut:
> > > ```
> > > (21:25:32) ERROR:   [Backend]:         TypeError: The argument 'mode' must be a 32-bit unsigned integer or an octal string. Received '+x'
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The shortcut is created, but it's not executable.
> > > If it helps, I use Linux Mint with Cinnamon, maybe the `+x` is not cross-desktop-environment compatible?
> > 
> > 
> > hmm what should I set it to then? 700 or 777? 777 is more foolproof, I will go with that
> 
> Tested with `777` and it's setting this permissions when I check with `ls -l`: `-r----x--t`, I think `777` is not correct.
> 
> I tried with `0x777` and it sets this `-r-xrwSrwt` which is a bit better I guess but looks really weird (like my user doesn't have write access?)
> 
> I think what you want to pass there is `511` which is this sequence of bits `111111111`. (I would add a comment in the code so it's clear where that 511 comes from)

I get -r-x--x--x with 511, is that correct?

arielj:
From heroic, with 511 I get `-rwxrwxrwx`, on a terminal I get `-r-x--x--x`
From herioc, with 777 I get `-r----x--t`, on a terminal I get `-rwxrwxrwx`

Have you tried 511 with Heroic setting that?

Nocccer:
We need a refactor of shortcut. I was planning todo it, but imo stable release is more important. If you want you can refactor the shortcuts by yourself. Make sure to write some tests for it. 80% code coverage is the goal.

Etaash-mathamsetty:
apparently it's base 10 vs base 8
so I switched this code to base 8

arielj:
> We need a refactor of shortcut. I was planning todo it, but imo stable release is more important. If you want you can refactor the shortcuts by yourself. Make sure to write some tests for it. 80% code coverage is the goal.

I think this PR is a good fix to get it merged for the stable and then we can do the refactor (I want to get the mac shorcuts stuff done first so the refactor is done with the whole code there)

arielj (approved):
Looks good with base 8 now, works as expected also on Linux Mint with Cinnamon setting the shortcut as executable.

arielj (changes requested):
oh wait, I just realized we are doing extra work with this, we can already create the file with the correct permissions

You can pass options to the `writeFile` method with the `mode` and the permissions you want to set https://nodejs.org/api/fs.html#fswritefilefile-data-options-callback

Can you try that? should make this code simpler

arielj (approved):
Looks good and clean now :+1:

Issue #1577, opened 2022-07-15
[Feature] Sort Played Recently by game's last played time


redromnon:
### Problem description

Currently, the **Played Recently** game list is not sorted by the last played time. This feature should display the last played game first as the header title suggests.

### Feature description

Sort **Played Recently** by a game's last played time.

arielj:
weird, this should be happening already (we prepend games to the list of recently played when a game is launched https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/1bc88715486f0299c588264f3ec7d53c2ca817aa/electron/main.ts#L816)

one thing that I noticed once is that you have to restart heroic for this changes to replicate in the UI, is that the case here? can you confirm that?

redromnon:
I can confirm it doesn't correct the order after restarting Heroic.

To elaborate, the last game I launched is Symphonia (not Fall Guys) but it is being displayed at the end in the Recently Played order.
![Screenshot from 2022-07-19 14-42-55](https://user-images.githubusercontent.com/74495920/179713931-26177631-4929-4c73-89c7-23904c9805b9.png)

arielj:
Closing this since the mentioned PR fixes the issue.

PR #1603, opened 2022-07-19
Refresh recent games when game is launched + no longer sort them on a per-store basis

CommandMC:
This fixes two things:
- Recent games were fetched in the order "Epic -> GOG", which resulted in Epic games always appearing first
- When a game was launched, the recent game list was not updated
  Note: This is done in a somewhat hacky way that relies on the backend first setting the recent games and then sending the `playing` game status update. I would've liked to use `configStore.onDidChange('games.recent', callback)` here, but the callback is for some reason never called

---

### DIFF
--- a/src/state/GlobalState.tsx
+++ b/src/state/GlobalState.tsx
@@ -460,6 +460,20 @@ export class GlobalState extends PureComponent<Props> {
       (game) => game.appName === appName
     )[0]
 
+    // Update recently played games if the game was just launched
+    // HACK: Ideally we'd use `configStore.onDidChange('games.recent')` here, but it's for some reason not working
+    if (status === 'playing') {
+      const { epic, gog } = this.state
+      const recentGames: GameInfo[] = getRecentGames([
+        ...epic.library,
+        ...gog.library
+      ])
+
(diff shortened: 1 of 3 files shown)


redromnon:
Should solve #1577

src/state/GlobalState.tsx
```diff
@@ -460,6 +460,20 @@ export class GlobalState extends PureComponent<Props> {
       (game) => game.appName === appName
     )[0]
 
+    // Update recently played games if the game was just launched
```
flavioislima, line 463: I think you can remove this comment, this doesn't look hacky to me.
The `configstore.onDidChange` was giving me false positives in the past. I was using it to update the game status on the frontend some time ago.

flavioislima (approved):
LGTM

src/state/GlobalState.tsx
CommandMC: > this doesn't look hacky to me

Changing something (seemingly) unrelated to this (the "setGameStatus" IPC call) will break this functionality. This shouldn't be the case, so it's a hack IMO.

PR #1615, opened 2022-07-22
[Fix] Esync/Fsync toggle buttons for Proton

CommandMC:
Proton enables Esync/Fsync by default. Instead of enabling it with `WINEESYNC`/`WINEFSYNC`, you use `PROTON_NO_ESYNC`/`PROTON_NO_FSYNC` to disable it.
We weren't setting these variables correctly (we just always used `WINEESYNC`/`WINEFSYNC`). That's now fixed

---


Nocccer (approved):

flavioislima (approved):
Looks good!

PR #1620, opened 2022-07-23
[Fix] Runtime toggle logic fix + GameMode message for Flatpak

CommandMC:
This PR addresses two issues:
- It was possible to exit the Wine Extensions page in a way that enabled a runtime that wasn't installed
  Note: Even if this was rather unlikely, the changes also avoid unnecessary settings updates and thus disk writes
- This is more of a feature, but I'd say it's a feature that fixes an issue:
  On Flatpak, GameMode is still required for the EAC runtime to not crash. Messages about this are now displayed when either enabling the runtime while GameMode is disabled, or disabling GameMode while the runtime is enabled

Since this adds new translation keys, I've as always waited on running `yarn i18next` until final approval of the provided defaults

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -92,7 +92,8 @@ import {
   isMac,
   isSteamDeckGameMode,
   isCLIFullscreen,
-  isCLINoGui
+  isCLINoGui,
+  isFlatpak
 } from './constants'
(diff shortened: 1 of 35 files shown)


flavioislima (approved):
LGTM

PR #1622, opened 2022-07-24
Fix stale state of games in Recently Played list

arielj:
This PR fixes a problem where games in the Recently Played header were not updated when a game was uninstalled, leading to a state where an uninstalled game showed up as installed in the list of Recently Played but uninstalled in the list of total games.

Some notes:
- I added a configuration to also be able to ignore the uninstalled games
- moved the code related to recet games in the frontend to it's own component, I was able to remove something from GlobalState which is nice

---


flavioislima (approved):
Good stuff ⚔️

flavioislima:
I think that maybe to fix these conflicts, just accept all incoming changes and then run i18n command again.

PR #1626, opened 2022-07-24
Fix incorrect variable passed to setupWineEnvVars

manueliglesiasgarcia:
This fixes protonfixes in ProtonGE for the GOG integration and allows to create workarounds for games https://github.com/GloriousEggroll/protonfixes

---

### DIFF
--- a/electron/launcher.ts
+++ b/electron/launcher.ts
@@ -229,8 +229,7 @@ async function prepareWineLaunch(game: LegendaryGame | GOGGame): Promise<{
     }
   }
 
-  const { folder_name: installFolderName } = game.getGameInfo()
-  const envVars = setupWineEnvVars(gameSettings, installFolderName)
+  const envVars = setupWineEnvVars(gameSettings, game.appName)
 
   return { success: true, envVars: envVars }
 }
@@ -321,7 +320,7 @@ function setupWineEnvVars(gameSettings: GameSettings, gameId = '0') {
   }
   if (wineVersion.type === 'proton') {
     // If we don't set this, GE-Proton tries to guess the AppID from the prefix path, which doesn't work in our case
-    ret.STEAM_COMPAT_APP_ID = '0'
+    ret.STEAM_COMPAT_APP_ID = `heroic-${gameId}`
     ret.SteamAppId = ret.STEAM_COMPAT_APP_ID
     // This sets the name of the log file given when setting PROTON_LOG=1
     ret.SteamGameId = `heroic-${gameId}`
@@ -440,14 +439,13 @@ async function runWineCommand(
   forceRunInPrefixVerb = false
 ) {
   const gameSettings = await game.getSettings()
-  const { folder_name: installFolderName } = game.getGameInfo()
 
   const { wineVersion } = gameSettings
 
   const env_vars = {
     ...process.env,
     ...setupEnvVars(gameSettings),
-    ...setupWineEnvVars(gameSettings, installFolderName)
+    ...setupWineEnvVars(gameSettings, game.appName)
   }
 
   let additional_command = ''


CommandMC:
Protonfixes are intended for Steam games, GOG's AppIds are a different thing. Protonfixes are also not intended to be ran outside of Steam, we're in a completely different environment here and don't do many of the prerequisites Steam does.

manueliglesiasgarcia:
Sorry I think that I didn't explain myself properly in the description It's true that  works with this changes but only if the Steam Runtime is enabled and I've testing most of the functions that protonfixes has and they're working properly.
I understand that the're different enviroments but with these fex tweaks protonfixes are fully working at least on the GoG side of the implementation

Nocccer:
Does it really work. Maybe i understand protonfixes wrong, but if i look into https://github.com/GloriousEggroll/protonfixes/tree/master/gamefixes this files are all steam_ids and not gog ids. So can you explain me, how protonfixes get from `heroic-${gameId}` the correct `<steamid>.py` ?

manueliglesiasgarcia:
Sure I can explain as an example I've Neverwinter Nights Diamond with ID https://www.gogdb.org/product/1207658890
So with my changes and enabling Steam Runtime I can run this protonfix 

heroic-1207658890.py
```
""" Game fix test for Neverwinter Nights Diamond
"""
#pylint: disable=C0103

from protonfixes import util

def main():
    """ Test
    """

    with open('readme.txt', 'w') as f:
        f.write('Testfeawdawfafs!')
    util.replace_command('nwmain.exe', 'nwconfig.exe')
```

Nocccer:
Ah now i got it. You will write your own fixes. Thanks for the clarification. Does wine-ge have something like that without steam runtime? We could possible provide this fixes for everyone then.

manueliglesiasgarcia:
I'm afraid not ProtonGE runs it automatically because protonfixes, It's part of the proton python script
https://github.com/GloriousEggroll/proton-ge-custom/blob/master/proton#L1692

Nocccer:
Ok. I would then consider adding a option to enable this. Is disabled by default. Because normally we don't suggest proton on heroic (expect steamos) and i would say this is then a advanced option for experienced users

Etaash-mathamsetty:
> Sure I can explain as an example I've Neverwinter Nights Diamond with ID https://www.gogdb.org/product/1207658890 So with my changes and enabling Steam Runtime and I've created this protonfix inside the ProtonGE folder that runs automatically
> 
> heroic-1207658890.py
> 
> ```
> """ Game fix test for Neverwinter Nights Diamond
> """
> #pylint: disable=C0103
> 
> from protonfixes import util
> 
> def main():
>     """ Test
>     """
> 
>     with open('readme.txt', 'w') as f:
>         f.write('Test!')
>     util.replace_command('nwmain.exe', 'nwconfig.exe')
> ```

this is actually pretty interesting, what if we made a fork of wine-ge or something and made custom protonfixes for heroic games launcher
that would be really cool

manueliglesiasgarcia:
I've just test it out with FallGuys seems to work as well with my changes enabling, steam runtime and with this protonfix

heroic-0a2d9f6403244d12969e11da6713137b.py
```
""" Game fix for Fall Guys
"""
#pylint: disable=C0103

import os
import subprocess

def main():
    """ Create symlink of eac so at the right location
    """
    if os.path.exists('FallGuys_client_game_Data/Plugins/x86_64/easyanticheat_x64.so'):
        subprocess.call(['rm', '-rf', 'FallGuys_client_game_Data/Plugins/x86_64/easyanticheat_x64.so'])
    subprocess.call(['ln', '-s', '../../../EasyAntiCheat/easyanticheat_x64.so', 'FallGuys_client_game_Data/Plugins/x86_64/easyanticheat_x64.so'])

    # Fixes the ini file.
    subprocess.call(['sed', '-i', 's/TargetApplicationPath=FallGuysEACLauncher.exe/TargetApplicationPath=FallGuys_client_game.exe/', 'FallGuys_client.ini'])
    
```

CommandMC:
Implementing protonfixes ourselves like this might *work*, yes, although relying on Proton to do these fixes doesn't make sense IMO.

I'm working on automatically applying game-specific workarounds in Heroic itself (actually have been working on that for some time now), which, again in my opinion, will be more understandable from a code perspective & will be more flexible. Protonfixes are somewhat opaque in that you run them and don't know if they actually worked or what was done, with a custom implementation, one could only run a specific workaround, only run it when certain conditions are met, or have optional workarounds if certain additional features are desired

imLinguin:
We can make something like protonfixes but totally separate from wine, and make it rely on totally different environment variable.

Nocccer:
Best thing would be a typescript package with all the util functionality to adapt, remove, move, create files and more.

imLinguin:
We can do that. That would be good base for CommandMCs work.

CommandMC (changes requested):
So, as I said above, I really don't like this approach to the problem. It also won't solve compatibility issues for native GOG games or games added to Heroic directly (whenever that feature might be implemented).  
A built-in "game workarounds" feature would be a lot better IMO, but this is of course way too much to ask for a first time contributor, I get that

manueliglesiasgarcia:
I think that for the moment I'll close this pull request and start to work in the Typescript alternative like @CommandMC suggested

PR #1628, opened 2022-07-25
[Feat] Proper Handler of Adding and Removal of Desktop/Menu Shortcuts

flavioislima:
This PR improves a few things related to shortcuts:
- [x] Add IPC handler to check if the menu or desktop shortcuts exists
- [x] Change the text Add Shortcut to Remove Shortcut if there is any
- [x] Fix Message box not showing up when adding shortcuts
- [x] Add new Message box with info on removed shortcuts

image

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -681,7 +681,6 @@ ipcMain.handle('getGameInfo', async (event, appName, runner) => {
       return null
     }
     info.extra = await game.getExtraInfo()
-    // eslint-disable-next-line @typescript-eslint/return-await
     return info
   } catch (error) {
     logError(`${error}`, LogPrefix.Backend)
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -681,6 +681,7 @@ ipcMain.handle('getGameInfo', async (event, appName, runner) => {
       return null
     }
     info.extra = await game.getExtraInfo()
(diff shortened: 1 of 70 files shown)


electron/main.ts
```diff
@@ -681,6 +681,7 @@ ipcMain.handle('getGameInfo', async (event, appName, runner) => {
       return null
     }
     info.extra = await game.getExtraInfo()
+    // eslint-disable-next-line @typescript-eslint/return-await
```
Nocccer, line 684: Is this needed?

Nocccer (changes requested):
LGTM. I wonder if the eslint disable is really needed? It wasn't there before and worked.

electron/main.ts
flavioislima: yes, weird, for me it is giving an error that makes no sense.
Let me check again.

flavioislima: Yes, removed and the error is gone.
Before it was complaining that I should have an `await` there, but makes no sense and then it was complaining that the `await` was not needed.

Nocccer: I have this error sometimes if i switch branches. I think eslint or vscode is sometimes slow.
I started to use `git clean -dxf` on every branch switch.

Nocccer (approved):

PR #1633, opened 2022-07-26
[Tech] Switch from react-scripts to Vite & clean up everything to work with strict mode

CommandMC:
- Move from [react-scripts/CRA](https://github.com/facebook/create-react-app) to [Vite](https://vitejs.dev/)
- Clean up all (well, most of) the backend code to work with TS' `strict` mode
   - This involved a lot of refactoring and a (almost) completely rewritten LegendaryLibrary
   - Also contains most of the changes made in #1602

Tested everything on Linux so far, will test on Windows as well. Mac testing would be appreciated as always

The way development works now is the following:
- If you want to quickly test new things, you run `yarn dev`. Heroic will open and any change you make in the frontend or backend will reload/restart it.
- If you want to build, you run `yarn dist:linux`/`dist:mac`/`dist:win` as always
- If you want to make sure that no code violates TS rules, you run `yarn codecheck`.
  Note: This is different from how it worked before. In the interest of speed & de-duplication, Vite does not check for violations itself since it's (1) way slower to do so and (2) you should already know about them through *some other tool* (VSCode shows them for example).  
  If you want to be absolutely sure everything's fine, you can run this command to (try to) compile every `.ts[x]` file in the project with `tsc`

---


tsconfig.json
```diff
@@ -22,7 +22,9 @@
     "isolatedModules": true,
     "noEmit": true,
     "noFallthroughCasesInSwitch": true,
-    "baseUrl": "src/"
+    "baseUrl": "src/",
+    "types": ["vite-plugin-svgr/client"],
+    "typeRoots": ["./node-modules/@types", "src/common/typedefs"]
```
CommandMC, line 27: Not sure if this line is necessary. VSCode stopped complaining as soon as I created the file, but according to the TS docs it *should* be necessary

CommandMC:
Since there are *a lot* of changes here, feel free to only review code you've written/you know yourself & leave the parts you don't know that well up to someone else. Last thing I want is this not getting reviewed because of its size

src/backend/launcher.ts
```diff
@@ -66,16 +66,6 @@ async function prepareLaunch(
...
-      showErrorBoxModalAuto(
-        i18next.t(
-          'box.error.no-offline-mode.title',
-          'Offline mode not supported.'
-        ),
-        i18next.t(
-          'box.error.no-offline-mode.message',
-          'Launch aborted! The game requires a internet connection to run it.'
-        )
-      )
```
arielj, line 78: was this error removed on purpose?

CommandMC: Yes, errors are now displayed in the `launch` function itself
Not sure how to do one of those inline code snippet thingies, but here's a link:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/feat/vite/src/backend/legendary/games.ts#L589-L592

src/frontend/screens/Login/index.tsx
```diff
@@ -4,13 +4,13 @@ import Runner from './components/Runner'
...
 import { useNavigate } from 'react-router'
 
-import { ReactComponent as EpicLogo } from 'src/assets/epic-logo.svg'
-import { ReactComponent as GOGLogo } from 'src/assets/gog-logo.svg'
+import { ReactComponent as EpicLogo } from 'frontend/assets/epic-logo.svg'
+import { ReactComponent as GOGLogo } from 'frontend/assets/gog-logo.svg'
 
-import { LanguageSelector, UpdateComponent } from 'src/components/UI'
-import { FlagPosition } from 'src/components/UI/LanguageSelector'
+import { LanguageSelector, UpdateComponent } from '../../components/UI'
```
arielj, line 10: just wondering, can't these be `frontend/...` like the others?

CommandMC: It absolutely can (and I have no doubt you'll find more of these elsewhere)

Note that the same thing doesn't apply for the backend, for some reason trying to do `import ... from 'backend/someFile'` gives an error that it can't resolve `backend`. *I think* it's caused by some module we have exporting a `Backend` object, although I'm really not sure there.

arielj:
> Since there are _a lot_ of changes here, feel free to only review code you've written/you know yourself & leave the parts you don't know that well up to someone else. Last thing I want is this not getting reviewed because of its size

good thing is that most file changes are just renames or changes in the imports to point to the right place, so it looks bigger than what it really is

I check this, I didn't look at code I've never touched before, I think it looks good at least the parts I understand

I'll try the branch later and use the app for a while to see if there's anything I can find broken

One question, I think we are removing the unreal market stuff, right? I see a lot of `is_game` checks begin removed

CommandMC:
> One question, I think we are removing the unreal market stuff, right? I see a lot of is_game checks begin removed

Yes, I got rid of that since I don't think it was working anymore. Could add it back if it's really necessary though, although I really don't think anyone used it

> EDIT: I can't imagine the amount of conflicts you get here xD

It's fine actually, Git itself is much better at auto-resolving things than GitHub

arielj:
> > One question, I think we are removing the unreal market stuff, right? I see a lot of is_game checks begin removed
> 
> Yes, I got rid of that since I don't think it was working anymore. Could add it back if it's really necessary though, although I really don't think anyone used it

I've never used that either (I think it was actually broken, it never showed anything for me but maybe I just don't know how to use that feature). Just wonder if anybody actually uses that, maybe there was a discussion about this already in discord.

Nocccer (approved):
Let's merge this after resolving conflicts. We need this Pr to progress further with refactoring. Bugs can be fixed later here. This pr is the first step of beta 2.5.0

flavioislima:
I will start to review and test this one. Might take a few days though.

CommandMC:
One issue I've now encountered is that updates don't seem to want to work. They run through fine, but the frontend still displays an update even after it's done. Backend does *not* report this game as update-able, so it's definitely some caching weirdness

I'm not too sure if this is caused by this PR, but it's definitely what I'm gonna look into next

CommandMC:
Another thing I'd like to move away from is watching Legendary's `installed.json`. If the file's updated, it's written in chunks, so when the first chunk gets written & we try to read it, it'll be malformed.  
Instead, I'd like to really only read it out if we need to (so on startup & after installing/importing/updating/uninstalling games)

src/backend/main.ts
```diff
@@ -638,18 +638,11 @@ ipcMain.handle(
...
-      return GOGLibrary.get().listUpdateableGames()
-    default:
-      return [
-        ...(await LegendaryLibrary.get().listUpdateableGames()),
-        ...(await GOGLibrary.get().listUpdateableGames())
-      ]
-  }
+ipcMain.handle('checkGameUpdates', async () => {
+  return [
+    ...(await LegendaryLibrary.get().listUpdateableGames()),
```
Nocccer, line 646: Is here catched that a user only uses one store?

Nocccer (changes requested):

src/backend/main.ts
CommandMC: For Legendary, if you're not logged in, `listUpdateableGames()` will just return `[]`. GOG does a similar thing, looping through the installed games and checking for an update one-by-one

Having GOG games installed, logging out of GOG & then refreshing the library might lead to issues, although that should be easy to fix.  
I'm not quite sure though, maybe @imLinguin can confirm this

flavioislima (approved):
What colossal work, thanks for that.
From what I saw and tested I could not find any bug, even the one you said about the update status not updating. For me it works fine.

Vite is also a lot better to develop and does everything faster so it was a pretty good improvement.

In my opinion it is ready to merge on `beta` so if some bug appears we can fix it afterwards :)

flavioislima:
@Nocccer I believe you need to dismiss the requested changes, otherwise, it can't be merged.

Nocccer (approved):
Pretty good work. I alread developed a little bit against it with the logger refactor.

CommandMC:
> From what I saw and tested I could not find any bug, even the one you said about the update status not updating. For me it works fine.

Well, that would be because I fixed it in 8754058cba8a0b6dbb1379bf28debd655b22cb11.  
I just discarded the old update list in the state of the frontend if `checkUpdates` was true, instead of appending what the backend sent to it. I assume the original change was done to make it possible to refresh the updates of just one library, although as you saw (well, you didn't, but I did), it introduces this "stuck updates" issue if the frontend (for whatever reason) doesn't remove the game from the updateable list

Issue #1625, opened 2022-07-24
make kde inhibit sleep and screen locking when a game is launched


Etaash-mathamsetty:
### Problem description

I'm always annoyed that when I run a game with a controller plugged in, kde always locks the screen after some time. Steam doesn't have this problem because when steam runs a game, steam tells kde to not screen lock/sleep. 

### Feature description

make kde not sleep when playing a game. 

### Alternatives

manually clicking the block screen locking and sleep toggle (I always forget) or gamemode works too

flavioislima:
@Etaash-mathamsetty can you test the PR to see if its working fine now: #1635

Issue #1624, opened 2022-07-24
"current_dxvk" and "current_vkd3d" files not created inside prefix


italoghost:
### Describe the bug

It is happening with The Witcher 2, from GOG - the other games, from Epic, are working just fine: after installing DXVK, a file named "The" would be created on the parent directory:

https://cdn.discordapp.com/attachments/818939754245652490/1000781046419292210/unknown.png

https://cdn.discordapp.com/attachments/818939754245652490/1000781099607261285/unknown.png

Then, every time that I booted the game, Heroic would try to install/update DXVK and VKD3D, even though it was already installed.

Placing manually the "current_" files inside the prefix prevented this from happening.

On the discord server, CommandMC mentioned the that the problem is related to the fact that you are using the command `echo` to write these files.

### Add logs

There is no log

### Steps to reproduce

1. Install "The Witcher 2" from GOG
2. Mark the options to automatic install/update DXVK and VKD3D.
3. Try to boot the game more than once.
4. See that Heroic is trying to install the extensions again.
5. Go to the game prefix and see that the "current_" files are not there.

### Expected behavior

The "current_" files should be created inside the prefix. 

### System Information

- OS [e. g. "Ubuntu"]: Arch Linux GNOME 42 Wayland
- Heroic Version [e. g. 2.1.1]: 2.4.0-beta-2

flavioislima:
Can you try the AppImage from this PR once the build is done? #1635

italoghost:
@flavioislima I tested the AppImage and it successfully installed DXVK and VKD3D and created the "current_" files!

flavioislima:
Nice, thanks for testing. 👍

PR #1635, opened 2022-07-26
[UX] Prevent Display from sleep when playing a game.

flavioislima:
Just add a new handler to prevent the display to go to sleep as asked on #1625 
Added a small fix for #1624  following what @CommandMC said on discord.
I could not reproduce the issue but changing the line to use `writeFileSync` didn't break the functionality.

---


Etaash-mathamsetty:
It doesn't deactivate when I stop playing the game, but steam doesn't even care if the game is running or not so I think it's fine for now

flavioislima:
> It doesn't deactivate when I stop playing the game, but steam doesn't even care if the game is running or not so I think it's fine for now

That's weird because Heroic removes the lock and stop the powersaver blocker so it should get back to normal behavior.
Not a major issue but should work.

flavioislima:
@Etaash-mathamsetty 
pushed a fix, the method to stop the blocker was not always being called, now its working as it should.

Etaash-mathamsetty:
> @Etaash-mathamsetty pushed a fix, the method to stop the blocker was not always being called, now its working as it should.

It's still not working for me...

flavioislima:
> > @Etaash-mathamsetty pushed a fix, the method to stop the blocker was not always being called, now its working as it should.
> 
> It's still not working for me on kde...

Can you run it from terminal and check if it shows a message like: _stopping display power saver blocker_ or something?

Etaash-mathamsetty:
all I get is this
`(16:41:00) INFO:    [Backend]:         Preventing display from sleep`
I do get an error from discord rich presence since I don't have discord installed anymore (I am using webcord), so let me try disabling the discord integration thing

flavioislima:
@Etaash-mathamsetty I totally forgot that KDE showed this information.
I was actually trapped inside a Javascript issue.
The `powerDisplayId` was coming as `0` and since I was checking `if(powerDisplayId)` JS was returning false and not stopping the blocker, changing it to check if it was null fixed it. Now I can see that it disappears from the blocker list, but try again on your machine.

Etaash-mathamsetty:
> @Etaash-mathamsetty I totally forgot that KDE showed this information. I was actually trapped inside a Javascript issue. The `powerDisplayId` was coming as `0` and since I was checking `if(powerDisplayId)` JS was returning false and not stopping the blocker, changing it to check if it was null fixed it. Now I can see that it disappears from the blocker list, but try again on your machine.

works fine now

src/components/UI/SearchBar/index.tsx
```diff
@@ -11,6 +11,11 @@ import ContextProvider from 'src/state/ContextProvider'
 import FormControl from '../FormControl'
 import './index.css'
 
+function fixFilter(text: string) {
+  const regex = new RegExp(/([\\|*|+|(|)|[|]|])+/, 'g')
```
arielj, line 15: we should consider `?` too in this regexp

flavioislima: ok will add it

arielj (approved):
looks good, I haven't had issues with powersaver / screensaver while playing so I don't know how to test this

I was a bit confused about the translations and the extra changes like the searchbar fixes, I think we should try to keep separated fixes in separated PRs so things are easier to follow (I don't think we need to split this PR, but would be nice to not have a mix of unrelated changes xD)

PR #1641, opened 2022-07-28
Show controller hints depending on focused element (WIP)

arielj:
WIP

This PR adds hints with controller icons so users know what each button does depending on the currently focused element.

image
image
image

Adding this to get some builds to help check different controllers.

---


src/types.ts
```diff
@@ -440,3 +440,9 @@ interface AntiCheatReference {
   date: string
   reference: string
 }
+
+declare global {
+  interface WindowEventMap {
+    'controller-changed': CustomEvent<{ controllerId: string }>
```
arielj, line 446: adding this event so I can send info from the controller helper to the controller component in react

Etaash-mathamsetty:
looks good so far, but the spacing seems off here:
image

arielj:
> looks good so far, but the spacing seems off here: image it seems to happen when the text is too long

yes, I still need to improve the css, locally I made each button space constant and if text is too long it wraps to 2 lines for example:
image

Etaash-mathamsetty:
> > looks good so far, but the spacing seems off here: image it seems to happen when the text is too long
> 
> yes, I still need to improve the css, locally I made each button space constant and if text is too long it wraps to 2 lines for example: image
> 
> there will be weird looking spacing in many cases, it's better to have inconsistent space between elements than having the buttons move when text changes, so we'll have to make that concession, I'll try some things but since that will be translatable too spacing will be problematic
> 
> it's nice to see that it's detecting the PS5 controller for the icons!

what if we just put equal spacing between each item no matter how big the text is

arielj:
> what if we just put equal spacing between each item no matter how big the text is

Currently what I'm doing locally (didn't push that to this PR yet) is to divide the available space in equal parts that can't resize horizontally and if the text don't fit it will wrap to 2 lines, I'm not sure if that's what you mean by equal spacing between items

> another bug: the controller hints don't disappear when I disconnect the controller

I'll check that, don't know if it's a common use case but I see if it's easy enough to detect that

> or click on something with the mouse

the thing here is that some of the gamepad interactions actually trigger a simulated mouse click (like when activating a select tag) so it might be hard (maybe impossible in some cases) to detect with certainty that it was the real mouse that executed some action and not the simulated mouse

Etaash-mathamsetty:
> > what if we just put equal spacing between each item no matter how big the text is
> 
> Currently what I'm doing locally (didn't push that to this PR yet) is to divide the available space in equal parts that can't resize horizontally and if the text don't fit it will wrap to 2 lines, I'm not sure if that's what you mean by equal spacing between items
> 
> > another bug: the controller hints don't disappear when I disconnect the controller
> 
> I'll check that, don't know if it's a common use case but I see if it's easy enough to detect that
> 
> > or click on something with the mouse
> 
> the thing here is that some of the gamepad interactions actually trigger a simulated mouse click (like when activating a select tag) so it might be hard (maybe impossible in some cases) to detect with certainty that it was the real mouse that executed some action and not the simulated mouse

what I was trying to say is that instead of them being locked into the same position and overflowing to the next line, we could just put an equal amount of spacing between each element no matter what. This way, longer text can take up more space and shorter text will take up less space

arielj:
> what I was trying to say is that instead of them being locked into the same position and overflowing to the next line, we could just put an equal amount of spacing between each element no matter what. This way, longer text can take up more space and shorter text will take up less space

If I understand your suggestion correctly, the problems with this are:
- the text for each action changes with the context, so that would make the buttons move around every time you change focus
- we can't handle all possible lengths when this gets translated (it already happend that translating `ALL` in some languages is like twice the width)
- if we don't allow wrapping into 2 lines we can't always ensure all text fill always fit in all screen sizes, you can resize heroic (or zoom it with the accessibility features) so text might be too long

that's why I think setting a fixed % for each element is the more stable for now, it has less layout shift and allows wrapping into multiple lines if needed

public/index.html
```diff
@@ -29,7 +29,6 @@
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
-    <div class="simple-keyboard"></div>
```
arielj, line 32: moved this inside the app's grid so I can position this without hacks

flavioislima:
@arielj can we remove WIP from the title as well?

src/components/UI/ControllerHints/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+      setLayout('steam-deck')
+    } else if (controller.match(/microsoft|xbox/i)) {
+      setLayout('xbox')
+    } else {
+      setLayout('steam-deck')
+    }
+  }, [controller])
+
+  // empty if no controller activated
+  if (controller === '') {
```
flavioislima, line 128:
```suggestion
  if (!controller) {
```

Small stuff but that would do it as well and it's a bit cleaner imo.

flavioislima:
So I tested here and I found a few issues:
- On uninstalled games, the Triangle button on the dual sense doesn't do anything even though it says it will open the game details. It works fine on an installed game.
- With the card selected on an uninstalled game, it says that X is for Play Game but doesn't do anything. Should show install game and execute the install function.
- On an Installed game, when selecting the card, it says X for play game but it doesn't work either.
- Besides the library, when using the collapsed sidebar there is no way to access the submenus on Settings, and also there is no way to uncollapse the sidebar so you need a mouse to do that. WE could maybe use the buttons Start/Menu or Select/Share to change the Sidebar behavior if that is possible.

Other than that these new hints look pretty cool, good stuff! ⚔️

arielj:
> * Besides the library, when using the collapsed sidebar there is no way to access the submenus on Settings

I'll leave this issue for a different PR, the problem was not introduced here (the spatial navigation gets confused with the submenu and spacing between elements). I think this should be a separated issue since the solution is not simple (I think there's something that should be re thinked about the collapsed sidebar so we don't have to use `fixed` elements for submenus and the collapse icon, I need to try some things but it's outside of the scope of the controller hints.

> and also there is no way to uncollapse the sidebar so you need a mouse to do that. WE could maybe use the buttons Start/Menu or Select/Share to change the Sidebar behavior if that is possible.

I'm moving the `<` icon to collapse the sidebar to the bottom of the sidebar and also changing the tag to be a `button` tag. The problem with the `<` in the middle of the sidebar is that it overlaps with other elements and the spatial navigation picks the next element in the sidebar instead of the button. When moving the button to the bottom of the sidebar it can be accessed with the spatial navigation and I think it looks cleaner there since it doesn't overlap with elements

image

Etaash-mathamsetty:
Could you swap X and Y in the OSK so Y is space and X is backspace
(like what was discussed in #1651)

arielj:
things I added:
- A in the card opens install dialog
- B closes the install dialog
- magnifying glass next to search input not focusable anymore (it's just decorative)
- `x` icon to clear search was not working, fixed that converting it into a button tag
- fixed the `play` action in game cards
- fixed the `open game details` action on uninstalled cards
- changed the collapse/expand icon to be a button so it can be focused better and moved it down so it doesn't overlap with other elements
- X now works as pressing `space` in the virtual keyboard
- Y now works as pressing `backspace` in the virtual keyboard

src/components/UI/SearchBar/index.css
```diff
@@ -27,3 +27,7 @@
   background-color: var(--accent);
   color: var(--background);
 }
+
+.SearchBar .FormControl__leftButton {
+  pointer-events: none;
```
arielj, line 32: so it can't be focused

Etaash-mathamsetty:
> things I added:
> 
> * A in the card opens install dialog
> * B closes the install dialog
> * magnifying glass next to search input not focusable anymore (it's just decorative)
> * `x` icon to clear search was not working, fixed that converting it into a button tag
> * fixed the `play` action in game cards
> * fixed the `open game details` action on uninstalled cards
> * changed the collapse/expand icon to be a button so it can be focused better and moved it down so it doesn't overlap with other elements
> * X now works as pressing `space` in the virtual keyboard
> * Y now works as pressing `backspace` in the virtual keyboard

I compared it with my PR, Y is space and X is backspace in this PR

arielj:
> > things I added:
> > 
> > * A in the card opens install dialog
> > * B closes the install dialog
> > * magnifying glass next to search input not focusable anymore (it's just decorative)
> > * `x` icon to clear search was not working, fixed that converting it into a button tag
> > * fixed the `play` action in game cards
> > * fixed the `open game details` action on uninstalled cards
> > * changed the collapse/expand icon to be a button so it can be focused better and moved it down so it doesn't overlap with other elements
> > * X now works as pressing `space` in the virtual keyboard
> > * Y now works as pressing `backspace` in the virtual keyboard
> 
> I compared it with my PR, Y is space and X is backspace in this PR

oops, typo in that message, the code is Y = space and X = backspace (this is the expected behavoir, right?)

flavioislima (approved):
Nice, Looks good 👍🏽

CommandMC (approved):
Well, can't say I fully understand the code (React + CSS might still be a little magic to me), but what I can understand is looking good and it's working great

PR #1646, opened 2022-07-28
[Tech] Add Runner param to all routes

### LINKED ISSUE #1644: [Linux] "Select an alternative EXE to run" file chooser icon does nothing when clicked
### Describe the bug

When trying to set an alternative EXE for a game, clicking the "+" button in the text box does nothing, i.e. no file chooser is launched.

### Add logs

No error is printed, it just doesn't do anything.

### Steps to reproduce

1. Go to Game Settings -> Other
2. Click on the "+" button inside the "Select an alternative EXE to run" text box.
3. Nothing happens.

### Expected behavior

A file chooser should be launched in order for the user to find and select the EXE of their choice.

### Screenshots

_No response_

### System Information

- OS: Arch Linux
- Heroic Version: 2.4.0-beta.2

### Additional information

_No response_

flavioislima:
This is a small but breaking change so appreciate if everyone could test this one.

I basically changed how we use the routes for `gamepage` and `settings` so we can have the `runner` in it, making it easier to make the requests and use it everywhere now that we support more than one store.

So before the routes were like this:
`/gamepage/appName` and `/settings/appname/wine`
now they will be:
`/gamepage/runner(gog or legendary)/appName` and `/settings/runner(app, gog or legendary)/appname/wine`

This change also fixes #1644 

---


arielj (commented):
I can confirm that the steps described in the related issue showed the error without this fix and is fixed with this fix!

One comment about something I noticed (but I don't know if it's caused by this PR of just uncovered by it) is that when I open the game settings from the gear icon in the game card, the Settings button or the Report a problem... link in the game card, I can see the `Save-Sync` option under Game Settings in the sidebar; but if I click in `Game Settings` in the sidebar, the `Save-Sync` option doesn't exist.

I don't know if this error existed before though, because the Game Settings link in the sidebar doesn't work for without this fix (I noticed this with Cyberpunk 2077 from GOG). Not sure if it also happens/happend before.

It's a really minor issue though, doesn't seem to be a blocker if this was present before these changes, I just don't have an installed Epic game with save-sync support.

flavioislima:
I will check if I can reproduce this issue. Might be a routing thing.

flavioislima:
Ok, found a fix for it. Although I think its not ideal to have the game information on the sidebar, since we decided to change that section to game settings, I dont think there is any other way to know if there is cloud save support or not.

Nocccer (changes requested):
Can confirm that nothing breaks. The problem that game settings are shown on not installed games is still there. I think we should not allow manipulate game settings of not yet installed games.

flavioislima:
> Can confirm that nothing breaks. The problem that game settings are shown on not installed games is still there. I think we should not allow manipulate game settings of not yet installed games.

ah, good catch, will fix that.

electron/legendary/eos_overlay/eos_overlay.ts
```diff
@@ -198,7 +203,7 @@ async function isEnabled(appName?: string, runner?: Runner) {
   let enabled = false
 
   let prefix = ''
-  if (isLinux || !(appName && runner)) {
+  if (!isWindows && isLinux && appName && runner) {
```
flavioislima, line 206: The previous check was not working on Windows, it was checking for a winePrefix so it was receiving an empty object and passing to legendary.

Now if it is not Windows it also needs to have an appName and Runner.
Let me know if this makes sense or not @CommandMC

electron/utils.ts
```diff
@@ -549,11 +549,12 @@ const formatEpicStoreUrl = (title: string) => {
 }
 
 function quoteIfNecessary(stringToQuote: string) {
-  if (
-    stringToQuote &&
+  const shouldQuote =
+    typeof stringToQuote === 'string' &&
```
flavioislima, line 553: added one more check here because the function was breaking because sometimes it was receiving an object or array as a variable.

flavioislima:
@Nocccer can you check again since you requested changes?

electron/legendary/eos_overlay/eos_overlay.ts
CommandMC: Right, I see the issue, not sure how that `!` got there.

The `isWindows` check is redundant, `isLinux` and `isWindows` will never both be true at the same time.
```suggestion
  if (isLinux && appName && runner) {
```

flavioislima: That's totally true 👍🏽

Nocccer (changes requested):
I get the bug again, where heroic does not render probably. See video:
Uploading simplescreenrecorder-2022-08-04_15.49.18.mp4…

I wonder how you can pass object or array to a function that expects only string?
Make the type check also in `removeQuoteInNessecary` and adapt the test for both fucntions please. There should be atleast one testcase where you pass a array or object.

flavioislima:
@Nocccer I forgot to fix this bug because it is a pretty edge case. Now I sent a fix to it.

Nocccer (approved):
Could not trigger the bug again. LGTM now.

PR #1648, opened 2022-07-28
[GOG Saves] Support multiple save locations

imLinguin:
- Splitted SaveSync settings component to two `gog.tsx` and `legendary.tsx`
- Added `gogSaves` entry into game settings - it's array which will store multiple save locations
- Added `gog_save_location`  for storing locations from api
- Created separate ipc handler for gog saves only


src/screens/Game/GamePage/index.tsx
```diff
@@ -60,6 +66,7 @@ export default function GamePage(): JSX.Element | null {
   const [updateRequested, setUpdateRequested] = useState(false)
   const [autoSyncSaves, setAutoSyncSaves] = useState(false)
   const [savesPath, setSavesPath] = useState('')
+  const [gogSaves, setGOGSaves] = useState([] as Array<GOGCloudSavesLocation>)
```
CommandMC, line 69: Would
```suggestion
  const [gogSaves, setGOGSaves] = useState<GOGCloudSavesLocation[]>([])
```
work?

src/screens/Settings/components/SyncSaves/gog.tsx
```diff
@@ -0,0 +1,251 @@
...
+}: Props) {
+  const [isLoading, setIsLoading] = useState(true)
+  const [isSyncing, setIsSyncing] = useState(false)
+  const [syncType, setSyncType] = useState('--skip-upload')
+
+  const { t } = useTranslation()
+
+  const { platform } = useContext(ContextProvider)
+  const isWin = platform === 'win32'
+  const isMac = ['osx', 'Mac']
```
CommandMC, line 40: This name could be better, with `isMac` you'd think it's a boolean
Maybe `macPlatforms`?

src/screens/Settings/components/SyncSaves/gog.tsx
```diff
@@ -0,0 +1,251 @@
...
+          String(installed_platform)
+        )
+        console.log(saveLocation)
+        let actualPath: string
+        if (!isNative) {
+          const { stdout } = await ipcRenderer
+            .invoke('runWineCommandForGame', {
+              appName,
+              runner: 'gog',
+              command: `cmd /c winepath "${saveLocation}"`
```
CommandMC, line 81: Why not just run `winepath "${saveLocation}"` directly here?

src/screens/Settings/components/SyncSaves/gog.tsx
```diff
@@ -0,0 +1,251 @@
...
+      const isLinuxNative = installed_platform === 'linux'
+      const isNative = isWin || isMacNative || isLinuxNative
+
+      for (const loc of gog_save_location) {
+        const { name, location } = loc
+        const saveLocation = await fixGogSaveFolder(
+          location.replace('<?INSTALL?>', String(install_path)),
+          String(installed_platform)
+        )
+        console.log(saveLocation)
```
CommandMC, line 74:
```suggestion
```

src/screens/Settings/components/SyncSaves/gog.tsx
```diff
@@ -0,0 +1,251 @@
...
+            })
+          actualPath = stdout.trim()
+        } else {
+          actualPath = await ipcRenderer.invoke('getShellPath', saveLocation)
+        }
+        const locationIndex = locations.findIndex(
+          (value) => value.name === name
+        )
+
+        console.log(actualPath)
```
CommandMC, line 95:
```suggestion
```

src/screens/Settings/components/SyncSaves/gog.tsx
```diff
@@ -0,0 +1,251 @@
...
+        if (locationIndex >= 0 && !locations[locationIndex]?.location.length) {
```
CommandMC, line 96: It might be good to have `!locations[locationIndex]?.location.length` in an extra variable, just for code clarity

src/screens/Settings/components/SyncSaves/gog.tsx
```diff
@@ -0,0 +1,251 @@
...
+        let actualPath: string
+        if (!isNative) {
+          const { stdout } = await ipcRenderer
+            .invoke('runWineCommandForGame', {
+              appName,
+              runner: 'gog',
+              command: `cmd /c winepath "${saveLocation}"`
+            })
+            .catch((error) => {
+              console.error('There was an error getting the path', error)
```
CommandMC, line 84: A `logError` IPC call would be better here I'd say

src/screens/Settings/components/SyncSaves/gog.tsx
```diff
@@ -0,0 +1,251 @@
...
+              setIsLoading(false)
+            })
+          actualPath = stdout.trim()
```
CommandMC, line 87: `stdout` will be `undefined` if the IPC call fails, probably easiest if you just add a `return { stdout: '' }` to the .catch

src/screens/Settings/index.tsx
```diff
@@ -603,18 +610,27 @@ function Settings() {
...
+            (runner === 'legendary' ? (
+              <LegendarySyncSaves
+                savesPath={savesPath}
+                setSavesPath={setSavesPath}
+                appName={appName}
+                autoSyncSaves={autoSyncSaves}
+                setAutoSyncSaves={setAutoSyncSaves}
+                isProton={!isWin && wineVersion.type === 'proton'}
+                winePrefix={winePrefix}
+                runner={runner}
```
CommandMC, line 623: Passing `runner` here is a little unnecessary, I'd assume you're still doing it because you want to avoid re-writing a lot of the old IPC calls?

CommandMC (commented):
Judging just by the code: Looks good! Left some suggestions and some questions

Sadly can't test on Win/Mac, will give it a test on Linux later though

src/screens/Settings/components/SyncSaves/gog.tsx
imLinguin: Env vars won't get parsed by the wine shell

CommandMC:
Might be a good idea to follow Symlinks for the save resolution. Save game location for Witcher 3 is giving me  
`/home/commandmc/Games/Heroic/Prefixes/The Witcher 3 Wild Hunt GOTY/dosdevices/c:/users/commandmc/Documents/The Witcher 3`,  
everything up until the `Documents` is a symlink. Node's `fs` module, and with that `graceful-fs` as well probably, can resolve symlinked paths with `realpath`:
```node
> const fs = require('fs')
> fs.realpathSync('/home/commandmc/Games/Heroic/Prefixes/The Witcher 3 Wild Hunt GOTY/dosdevices/c:/users/commandmc/Documents/The Witcher 3')
'/home/commandmc/Documents/The Witcher 3'
```

src/screens/Settings/components/SyncSaves/legendary.tsx
```diff
@@ -0,0 +1,219 @@
...
+        : path.replaceAll(/\\/g, '/') // invert slashes and remove latest on windows
+      setSavesPath(fixedPath)
+      setLoading(false)
+    }
+    getSyncFolder()
+  }, [winePrefix, isProton])
+
+  const isLinked = Boolean(savesPath.length)
+
+  const syncCommands = [
```
flavioislima, line 99: Since this is repeating on all runners, maybe we can store it on a constants file and import it here.

flavioislima (approved):
Looks good. Will do some testing as well on macOS to see if it is working there.

flavioislima:
Tested on macOS and it has a few issues.
- For native game, it starts loading and never loads the inputs.
- For Windows games, it takes a lot of time as well because it needs to start crossover to check the path.
- Still could not properly sync.
- The response is not cached so if I get out of the sync page after it loads, it will take all the time again to check for the path one more time.

flavioislima:
The errors:
GOG Game:
```
winepath "%USERPROFILE%/Documents/OpenTTD/save"
Error occurred in handler for 'getRealPath': Error: ENOENT: no such file or directory, lstat '/Users/flavio/Library/Application Support/CrossOver/Bottles/Heroic/drive_c/users/crossover/Documents'
    at realpathSync (node:fs:2490:7)
    at node:electron/js2c/asar_bundle:5:4831
    at /Users/flavio/Develop/electron/HeroicGamesLauncher/public/main.js:1143:90
    at node:electron/js2c/browser_init:189:579
    at EventEmitter.<anonymous> (node:electron/js2c/browser_init:161:11327)
    at EventEmitter.emit (node:events:526:28) {
  errno: -2,
  syscall: 'lstat',
  code: 'ENOENT',
  path: '/Users/flavio/Library/Application Support/CrossOver/Bottles/Heroic/drive_c/users/crossover/Documents'
```

For Epic games as well:
```
(17:20:25) DEBUG:   [Legendary]:       Ran Wine command: "/Applications/CrossOver.app/Contents/SharedSupport/CrossOver/bin/wine" cmd /c winepath "%USERPROFILE%/AppData/LocalLow/Alt Shift/Crying Suns/saves/save0/"
Error occurred in handler for 'getRealPath': Error: ENOENT: no such file or directory, lstat '/Users/flavio/Library/Application Support/CrossOver/Bottles/Heroic/drive_c/users/crossover/AppData/LocalLow/Alt Shift/Crying Suns/saves'
    at realpathSync (node:fs:2490:7)
    at node:electron/js2c/asar_bundle:5:4831
    at /Users/flavio/Develop/electron/HeroicGamesLauncher/public/main.js:1143:90
    at node:electron/js2c/browser_init:189:579
    at EventEmitter.<anonymous> (node:electron/js2c/browser_init:161:11327)
    at EventEmitter.emit (node:events:526:28) {
  errno: -2,
  syscall: 'lstat',
  code: 'ENOENT',
  path: '/Users/flavio/Library/Application Support/CrossOver/Bottles/Heroic/drive_c/users/crossover/AppData/LocalLow/Alt Shift/Crying Suns/saves'
```

flavioislima:
I am still having the same issues on latest commit, changing the `realPath` handler to something like this fixes it:
```ipcMain.handle('getRealPath', (event, path) => {
  try {
    const realPath = realpathSync(path)
    return realPath
  } catch (error) {
    return path
  }
})```

Because the `realPathSync` emits an error if the path doesn't exist, but the thing is that should be fine for the folder to not exists since the game wasn't played yet. So I believe we can just ignore it and fill the path. 
Another alternative is to create the folder if it doesn't exists but I think Heroic should not handle that since the game will, especially if the game uses specific permissions when creating the folders.

**Comment by flavioislima** [comment]
The save-sync screen for GOG native games are also not opening, just keeps loading. :(

**Comment by imLinguin** [comment]
What is the error for native games, still the same as before about not existing file? Or it's the same error like with realPath?

**Comment by flavioislima** [comment]
> What is the error for native games, still the same as before about not existing file? Or it's the same error like with realPath?

No, it seems to be the same as for windows games:
```
'/Library/Application Support/SteamWorld Dig 2'
    at realpathSync (node:fs:2490:7)
```

But I believe now it is looking into the wrong place. The correct one is:
`/Application Support/SteamWorld Dig 2` 
so it should not start under `Library`

**Comment by flavioislima** [comment]
correction: just played the game a bit and it is under Library but not /Library only Library/Applications Support

```
❯ pwd
/Users/flavio/Library/Application Support/SteamWorld Dig 2
~/Library/Application Support/SteamWorld Dig 2 ❯
```

**Comment by flavioislima** [comment]
Another question, why it takes so long to get the info? Does it need to get from a different endpoint than the one we use for `getInstallInfo` ? I think would be nice if we could cancel when clicking on the circle and then add the path manually. Makes sense?

**Comment by flavioislima** [comment]
Found other issues.
- Some GOG games return the `locationIndex` as -1 because the name and the `value.name` are different. That causes the whole component to break since the value for the `value.location.name` will be undefined on lines 176 and 189
- Legendary logic is also breaking in some cases, I will update with the error messages soon.

File: electron/main.ts [hunk 11]
```diff
@@ -1469,7 +1471,19 @@ ipcMain.handle(
 ipcMain.handle('getShellPath', async (event, path) => {
   return (await execAsync(`echo "${path}"`)).stdout.trim()
 })
-ipcMain.handle('getRealPath', (event, path) => realpathSync(path))
+ipcMain.handle('getRealPath', (event, path) => {
+  let resolvedPath = path
+  try {
+    resolvedPath = realpathSync(path)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  } catch (err: any) {
```
flavioislima, line 1479: instead of `any` try using `Record<string, string>`

src/screens/Settings/components/SyncSaves/legendary.tsx
```diff
@@ -46,6 +46,7 @@ export default function LegendarySyncSaves({
   const { t } = useTranslation()
   const { platform } = useContext(ContextProvider)
   const isWin = platform === 'win32'
+  const macPlatforms = ['osx', 'Mac']
```
flavioislima, line 49: If this is for legendary only using `Mac` would be enough I believe.

flavioislima:
Everything seems to be working fine now on macOS:
- Legendary games are getting the correct path now
- GOG games are getting the correct path as well for both native and windows games
- Sync is working for both as well, tested uploading and downloading

Only one thing that I believe can be improved: after getting the path and updating the settings, when entering the save-sync settings again, it will always start refreshing the path again. I think this is just problematic because it takes some time since we need to call crossover/wine for that. 
I see two ways of solving that:
- Refresh the path only if the Wine version or Wine prefix was changed
- Add a refresh button there so we can do it manually.

I believe that the first solution is better and we could have a key in the storage or somewhere else to store the current and previous wine versions/prefixes and if they are different and the user goes to the save-sync page (or maybe also during auto-sync), heroic should refresh the setting with the new path.

Screenshot 2022-08-03 at 12 14 12

imLinguin:
Now we are calling wine only for paths that are empty. I agree automatic path refreshing should be a thing when we change wine type and wine prefix.

Today I'll improve reading Documents path from the registry for windows games.

electron/main.ts
imLinguin: Getting this if type is other than any or unknown

`error TS1196: Catch clause variable type annotation must be 'any' or 'unknown' if specified.`

flavioislima:
Tested on Windows and I the path discovered is not working, for both legendary and gog all I get is the path from where heroic is running + C: like:
`C:\Users\flavi\Develop\HeroicGamesLauncher\"C:`

flavioislima:
The games on windows are getting the correct path with the fix I sent now.
So I believe all platforms are ok but let's test a bit more on Linux again to check if nothing broke after my commit.

flavioislima:
I think this one is ready to merge now 👍🏽

PR #1650, opened 2022-07-29
[Tech] AppVeyor Configuration

flavioislima:
This will add the Appveyor configuration to sign the windows artifacts with the certificate that SignPath will provide to us. 
The tests already passed and they confirmed that everything is ok. 
We just got the prod certificate and I updated the flow to run only on tags.

---


package.json
```diff
@@ -42,6 +42,7 @@
       }
     ],
     "win": {
+      "artifactName": "Heroic_Setup_${version}.${ext}",
```
flavioislima, line 45: the artifact cannot contain spaces on the name so I needed to change it.

Nocccer (approved):

PR #1652, opened 2022-07-30
[Refactor] Make gamepad code more efficient

Etaash-mathamsetty:
there were a lot of redundant gamepad mappings laying around, so I removed them
Tested with my dualsense and xbox one controller


flavioislima (commented):
I don't have all controllers to test, I can see that PS3 and PS5 use the same configuration but the PS3 clone doesn't.
Can you confirm that they are the same?

Etaash-mathamsetty:
> I don't have all controllers to test, I can see that PS3 and PS5 use the same configuration but the PS3 clone doesn't. Can you confirm that they are the same?

seems like I made an oopsie, latest commit should fix it

arielj:
I'd say to wait until after the realease of 2.4.0 to merge this one. We don't have all the controllers to test this so I think it's safer to do a change like this in the next beta so people can safely revert to stable if this breaks something.

Etaash-mathamsetty:
The other controllers should continue work correctly (I didn't modify anything related to them), we just need someone to test with the PS3 and clone PS3 controllers, since I tested with the dual sense and xbox one controllers

arielj:
Yes sorry, I didn't mean it would break something with the controllers in general, but maybe it breaks something for users of that controller that we can't test. Since this is not critical for the release I'd prefer to merge this after so there's more time for people to test that in a beta (I don't remember who reported the ps3 clone controller when we were developing this, maybe it's a matter of trying to find that person in discord and ask them)

But I think, as a general rule, so close to a release we can wait a bit to merge a refactor.

PR #1655, opened 2022-08-01
[Fix] Skip fetch/check of tools if game is not installed

Nocccer:
Fixed also a bug where we skipped checking of added to steam and shortcuts on windows. This two functions are available on any os.

---

### DIFF
--- a/public/locales/pl/translation.json
+++ b/public/locales/pl/translation.json
@@ -268,7 +268,7 @@
             },
             "remove": {
                 "steam": {
-                    "corrupt": "Gra {{game}} nie mogła zostać usunięta dla wszystkich użytkowników Steam. Sprawdź logi po więcej informacji. Restart Steam'a jest wymagany, by zastosować\u00a0zmiany.",
+                    "corrupt": "Gra {{game}} nie mogła zostać usunięta dla wszystkich użytkowników Steam. Sprawdź logi po więcej informacji. Restart Steam'a jest wymagany, by zastosować zmiany.",
                     "success": "Gra {{game}} została usunięta ze Steam. Restart Steam'a jest wymagany, by zastosować zmiany.",
                     "title": "Usunięto ze Steam"
                 }
--- a/public/locales/pl/translation.json
+++ b/public/locales/pl/translation.json
@@ -268,7 +268,7 @@
             },
             "remove": {
                 "steam": {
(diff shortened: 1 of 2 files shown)


src/screens/Game/GameSubMenu/index.tsx
```diff
@@ -248,51 +248,53 @@ export default function GamesSubmenu({
...
-      try {
-        const { wineVersion, winePrefix }: AppSettings =
-          await ipcRenderer.invoke('requestSettings', appName)
-        let wine = wineVersion.name
-          .replace('Wine - ', '')
-          .replace('Proton - ', '')
-        if (wine.includes('Default')) {
-          wine = wine.split('-')[0]
+    if (isInstalled) {
+      if (isWin) {
```
CommandMC, line 252: Could've done something like
```suggestion
    if (!isInstalled || isWin) {
```
here to avoid the extra indentation if that's desired

Nocccer: Yes good catch.

Nocccer:
Should be cleaned up now.

public/locales/pl/translation.json
```diff
@@ -268,7 +268,7 @@
             },
             "remove": {
                 "steam": {
-                    "corrupt": "Gra {{game}} nie mogła zostać usunięta dla wszystkich użytkowników Steam. Sprawdź logi po więcej informacji. Restart Steam'a jest wymagany, by zastosować zmiany.",
+                    "corrupt": "Gra {{game}} nie mogła zostać usunięta dla wszystkich użytkowników Steam. Sprawdź logi po więcej informacji. Restart Steam'a jest wymagany, by zastosować\u00a0zmiany.",
```
flavioislima, line 271: I think this is not correct, remove this file from the change.

flavioislima (approved):
Looks good, just check that weird issue with that translation. 👍🏽

public/locales/pl/translation.json
Nocccer: If i do i will still get the yarn18n update after every push

Nocccer:
```suggestion
                    "corrupt": "Gra {{game}} nie mogła zostać usunięta dla wszystkich użytkowników Steam. Sprawdź logi po więcej informacji. Restart Steam'a jest wymagany, by zastosować zmiany.",
```

Issue #1118, opened 2022-03-14
Log files can take up to 100gb of disk space


CapsAdmin:
### Describe the bug

This is caused by some bug I guess, but this issue is more that log files are allowed to grow to insane sizes.

This seems to happen if I try to close the launcher. If I close the launcher it hangs and then kde plasma asks if I want to force close it. I guess that leaves some processes hanging.

### Add logs

(01:00:03) INFO:    [Legendary]:       Location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(01:00:03) INFO:    [Gog]:             Location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(01:00:04) INFO:    [Backend]:         
  Heroic Version: 2.2.6 Oden
  Legendary Version:  0.20.25 Our Benefactors
  OS: neon KERNEL: 5.13.0-30-generic ARCH: x64
  CPU: AMD Ryzen 9 5950X 16-Core Processor @3.4 GOVERNOR: ondemand
  RAM: Total: 67.3 GB Available: 45.4 GB
  GRAPHICS: GPU0: Navi 21 [Radeon RX 6900 XT] VRAM: 1MB DRIVER:  
  PROTOCOL: x11
  
(01:00:04) WARNING: [Backend]:         Protocol already registered.
(01:00:04) INFO:    [Frontend]:        Refreshing Library
(01:00:04) INFO:    [Legendary]:       Refreshing library...
(01:00:04) INFO:    [Legendary]:       Refreshing Epic Games...
(01:00:04) INFO:    [Legendary]:       Refreshing Epic Games...
(01:00:05) INFO:    [Legendary]:       [cli] INFO: Logging in...
(01:00:09) INFO:    [Backend]:         checking for heroic updates
(01:00:10) INFO:    [Legendary]:       [Core] INFO: Logging in...
(01:00:10) INFO:    [Legendary]:       [cli] INFO: Getting game list... (this may take a while)
(01:00:11) INFO:    [Legendary]:       Updating game list
(01:00:11) INFO:    [Legendary]:       Game List Updated
(01:00:12) INFO:    [Legendary]:       Checking for game updates
(01:00:12) INFO:    [Legendary]:       Found 0 game(s) to update
(01:00:12) INFO:    [Gog]:             Found 0 game(s) to update

### Steps to reproduce

[...]

PR #1662, opened 2022-08-03
[Fix] Show uncaught exception in frontend and exit heroic.

Nocccer:
- According to #1118 this can fill the logs.
- Also uncaught exception should not be ignored and instead shown as message box in the frontend.
- Close the app to avoid unexpected behavior. We should catch all errors we know can be thrown.

Can be tested by throwing a error somewhere in the backend.

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -471,8 +471,22 @@ ipcMain.on('frontendReady', () => {
 })
 
 // Maybe this can help with white screens
-process.on('uncaughtException', (err) => {
+process.on('uncaughtException', async (err) => {
   logError(`${err.name}: ${err.message}`, LogPrefix.Backend)
+  await showErrorBoxModal(
+    mainWindow,
+    i18next.t(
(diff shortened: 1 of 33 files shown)


flavioislima (approved):
NIce! This is pretty good to have 👍🏽

electron/main.ts
```diff
@@ -471,8 +471,22 @@ ipcMain.on('frontendReady', () => {
 })
 
 // Maybe this can help with white screens
```
CapsAdmin, line 473: What white screens are being mentioned here?

flavioislima: I think was an old Electron bug that sometimes the app became blank in some setups, this was a general issue not only in Heroic so I found this solution when searching on the web.
Not sure if they fixed it already.

Nocccer:
I want to have feedback from the issue i linked but he don't repsonse. We can merge it or not anyway. Opinions?

PR #1667, opened 2022-08-05
[Refactor] Logger

Nocccer:
Refactored logger
- Able to pass objects, arrrays, numbers without making them a string to log functions ( logError, logInfo, ... )
- Make paramters a option interface
- Split logger and logfile functionality
- Wrote tests for logger and logfile functions

---


src/backend/logger/logger.ts
```diff
@@ -32,17 +23,42 @@ export enum LogPrefix {
...
+// helper to convert LogInputType to string
+function convertInputToString(param: LogInputType): string {
+  const getString = (value: LogInputType): string => {
+    switch (typeof value) {
+      case 'string':
+        return value
+      case 'object':
+        // Object.prototype.toString.call(value).includes('Error') will catch all
+        // Error types (Error, EvalError, SyntaxError, ...)
+        if (Object.prototype.toString.call(value).includes('Error')) {
```
CommandMC, line 37: Why the `toString.call`? Wouldn't just
```suggestion
        if (Object.prototype(value).includes('Error')) {
```
work?

src/backend/logger/logger.ts
```diff
@@ -32,17 +23,42 @@ export enum LogPrefix {
...
+          return value!.toString()
```
CommandMC, line 38: Checking for `value.stack` first and returning that (if it exists) might be more helpful

src/backend/logger/logger.ts
```diff
@@ -32,17 +23,42 @@ export enum LogPrefix {
...
-    return param
-  } else if (Array.isArray(param)) {
-    return param.join(' ')
+  if (!Array.isArray(param)) {
+    return getString(param)
   }
-  // if somehow the param is an object and not array or string
-  return JSON.stringify(param)
+
+  const strings = [] as string[]
```
CommandMC, line 57: This is probably nitpicky, but I'd recommend not getting used to using `as` too much
```suggestion
  const strings: string[] = []
```
does the same thing, but doesn't get you used to disabling the type checker :^)

src/backend/logger/logger.ts
```diff
@@ -65,221 +81,134 @@ const getTimeStamp = () => {
...
-  try {
-    if (currentLogFile) {
-      appendFileSync(currentLogFile, `${message}\n`)
-    }
-  } catch (error) {
-    logError(`Writing log file failed with ${error}`, LogPrefix.Backend, true)
+  if (!options?.skipLogToFile) {
+    appendMessageToLogFile(`${messagePrefix} ${text}`)
   }
 }
```
CommandMC, line 214: I feel like I've said this before, but having one common "logBase" function that all of these functions then call with their respective log level and arguments would be easier to read, maintain, and test

src/backend/main.ts
```diff
@@ -980,17 +991,19 @@ ipcMain.handle(
     )
     return game
       .launch(launchArguments)
-      .catch((error) => {
-        logError(error.stack ?? `${error}`, LogPrefix.Backend)
+      .catch((exception: Error) => {
```
CommandMC, line 994: You can't know for sure that what you got here is an `Error`, throwing *anything* is fine
```
$ node
Welcome to Node.js v16.16.0.
Type ".help" for more information.
> throw 'Hey'
Uncaught 'Hey'
```

src/backend/utils.ts
```diff
@@ -712,11 +713,24 @@ function detectVCRedist(mainWindow: BrowserWindow) {
...
+    const notify = new Notification({
+      body,
+      title
+    })
+
+    notify.on('click', () => mainWindow.show())
+    notify.show()
+  }
+}
+
```
CommandMC, line 733: Why was this moved?

CommandMC (commented):
This is very close to being perfect, just got some questions/suggestions/ideas

src/backend/gog/setup.ts
```diff
@@ -9,7 +9,7 @@ import {
 import { copySync } from 'fs-extra'
 import path from 'node:path'
 import { GOGLibrary } from './library'
-import { GameInfo, InstalledInfo } from 'common/types'
+import { GameInfo, InstalledInfo } from '../../common/types'
```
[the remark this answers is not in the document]
Nocccer: I remember i had problems once with `common/types`. I don't know why but i got alot of errors that some functions could not be found, even typescript and lint didn't complain about it. I will revert and we can see it is still working.

src/backend/logger/__tests__/logfile.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  test('createNewLogFileAndClearOldOnces removing old logs fails', () => {
+    jest.spyOn(app, 'getPath').mockReturnValue(tmpDir.name)
+    const spyUnlinkSync = jest
+      .spyOn(graceful_fs, 'unlinkSync')
+      .mockImplementation(() => {
+        throw Error('unlink failed')
+      })
+    const date = new Date()
+    date.setMonth(date.getMonth() > 0 ? date.getMonth() - 1 : 11)
+    const monthOutdated = join(
```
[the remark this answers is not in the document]
Nocccer: `oldLogFilename` does not explicit tells you that this log file is over a month old and should be removed by the function.
Something like `monthOutdatedLogFile` is maybe better.

src/backend/logger/__tests__/logfile.test.ts
```diff
@@ -0,0 +1,186 @@
...
+      tmpDir.name,
+      // @ts-ignore replaceAll error
+      `heroic-${date.toISOString().replaceAll(':', '_')}.log`
```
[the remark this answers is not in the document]
Nocccer: I have no clue. I already tried to setup a new tsconfig for the tests. I feel like vscode does something wrong and does not apply tsconfig to the test files, because everything works fine.

src/backend/logger/__tests__/logfile.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    expect(logError).not.toBeCalled()
+    expect(graceful_fs.existsSync(monthOutdated)).toBeFalsy()
+    expect(graceful_fs.existsSync(yearOutdated)).toBeFalsy()
+  })
+
+  test('getLogFile all possible values', () => {
+    //@ts-ignore Needed override the currentLogFile
+    constants.currentLogFile = 'current.log'
+    //@ts-ignore Needed override the lastLogFile
+    constants.lastLogFile = 'last.log'
```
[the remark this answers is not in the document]
Nocccer: I can mock the values globally, but somehow i can't override them in the test. This was the easiest way. I can try again.

src/backend/logger/logfile.ts
```diff
@@ -0,0 +1,135 @@
...
+ * @param appName     needed to get appName log
+ * @param defaultLast if set getting heroic default last log
+ * @returns path to log file
+ */
+export function getLogFile(
+  isDefault: boolean,
+  appName: string,
+  defaultLast = false
+): string {
+  return isDefault
```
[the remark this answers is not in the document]
Nocccer: Really good idea :).

src/backend/logger/logger.ts
Nocccer: Bildschirmfoto vom 2022-09-11 17-34-38

I was searching the internet and they suggest the `toString.call`

src/backend/logger/logger.ts
Nocccer: I think `value.toString()`returns everything of Error. I will still check Error documentation.

src/backend/logger/logger.ts
CommandMC: Ah sorry, meant `getPrototypeOf`
```
$ node
Welcome to Node.js v16.16.0.
Type ".help" for more information.
> Object.getPrototypeOf(EvalError()).name
'EvalError'
> Object.getPrototypeOf(EvalError()).name.includes('Error')
true
```

src/backend/logger/logger.ts
CommandMC: [`Error.stack`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack) *is* non-standard, but it gives more information than just calling `toString()` on it
Comparison:
```
$ node
Welcome to Node.js v16.16.0.
Type ".help" for more information.
> Error("Some Error").toString()
'Error: Some Error'
> Error("Some Error").stack
'Error: Some Error\n' +
  '    at REPL24:1:1\n' +
  '    at Script.runInThisContext (node:vm:129:12)\n' +
  '    at REPLServer.defaultEval (node:repl:566:29)\n' +
  '    at bound (node:domain:421:15)\n' +
  '    at REPLServer.runBound [as eval] (node:domain:432:12)\n' +
  '    at REPLServer.onLine (node:repl:893:10)\n' +
  '    at REPLServer.emit (node:events:539:35)\n' +
  '    at REPLServer.emit (node:domain:475:12)\n' +
  '    at REPLServer.Interface._onLine (node:readline:487:10)\n' +
  '    at REPLServer.Interface._line (node:readline:864:8)'
```
(in an actual program the stack would of course make more sense)

src/backend/logger/__tests__/logfile.test.ts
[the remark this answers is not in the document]
CommandMC: > There is no other easy way i found. Feel free to find one.

Not sure what you mean, the code snippet in my original comment works (without giving me warnings/having to use `@ts-ignore`)

src/backend/utils.ts
Nocccer: I think somehow the function was implemented twice. I removed one. I can move the function to the end again if you want. But actually does not hurt to let it there. The implementation didn't change.

src/backend/logger/__tests__/logfile.test.ts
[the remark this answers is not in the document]
Nocccer: It removes the type error but has the same problem. You still need to override the constant back to the default one.
I want to find a better way.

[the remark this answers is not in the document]
CommandMC: I have some ideas on how to solve this with Vitest, so it's fine for now

src/backend/utils.ts
CommandMC: Yeah that's fine, was just wondering why it moved

CommandMC (approved):

src/backend/logger/__tests__/logfile.test.ts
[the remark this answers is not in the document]
Nocccer: I changed it to your solution, the problem is the restoring of the default value. 
Please keep the global manual mocks in vitest, because they will pretty helpful if we introduce more tests.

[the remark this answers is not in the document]
CommandMC: Wasn't planning on removing 'em

src/backend/logger/logger.ts
Nocccer: Yes it might work, but my way works better on undefiend scenarios. Your let the test fail:
image

PR #1668, opened 2022-08-05
[GOG] Fix errors when offline

imLinguin:
Added checks for `isOnline` in places where heroic requests gog servers. I think I found all possible cases.

### DIFF
--- a/electron/gog/library.ts
+++ b/electron/gog/library.ts
@@ -13,7 +13,7 @@ import { join } from 'node:path'
 import { existsSync, readFileSync } from 'graceful-fs'
 
 import { logError, logInfo, LogPrefix, logWarning } from '../logger/logger'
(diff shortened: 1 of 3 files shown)


arielj (commented):
I shared this comment on discord buy I think it applies here too. I think the `isOnline` method only checks the system is connected to a network but it doesn't really check if it's actually connected to the internet.

If `isOnline` is true it can still fail to connect to the internet.

I think the `isOnline` helper needs an additional check pinging somewhere like github if it detects a network connection.

flavioislima:
i think this one needs to be merged first before @Nocccer PR since it fixes the GOG stuff.

flavioislima:
This one looks good I think we can merge it and get back to the online/offline handler later.
At least now it is possible to open the game page when offline.

flavioislima (approved):

PR #1675, opened 2022-08-08
[Fix] Check if steam flatpak is really installed

Nocccer:
instead of checking if com.valvesoftware.Steam folder exist, we also check now if `flatpak info com.valvesoftware.Steam` returns succesfully.

Tested with steam flatpak installed and also if the folder exist but steam flatpak is not installed anymore.

---


flavioislima (approved):
LGTM.

Issue #1494, opened 2022-06-17
Heroic must have an active connection when started.


zany130:
### Describe the bug

as the title says opening heroic before you have an active working connection (this happens on my garuda kde system) prevents game downloads and other connection-related things from working

### Add logs

```
(14:39:29) DEBUG:   [Legendary]:       Running Legendary command: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary --version
(14:39:30) INFO:    [Legendary]:       Legendary location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(14:39:30) INFO:    [Gog]:             GOGDL location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(14:39:30) INFO:    [Backend]:         

Heroic Version: 2.3.9 Brook
Legendary Version:  0.20.26 Dark Energy
OS: Garuda KERNEL: 5.18.3-zen1-1-zen ARCH: x64
CPU: AMD Ryzen 5 2600X Six-Core Processor @3.6 GOVERNOR: performance
RAM: Total: 31.27 GiB Available: 26.58 GiB
GRAPHICS: GPU0: Navi 22 [Radeon RX 6700/6700 XT/6750 XT / 6800M] VRAM: 1MB DRIVER:  
PROTOCOL: wayland

(14:39:30) INFO:    [Backend]:         User Not Found, removing it from Store
(14:39:30) INFO:    [Gog]:             Getting data about the user
(14:39:33) WARNING: [Backend]:         Protocol already registered.
(14:39:33) ERROR:   [Gog]:             Error getting user Data Error: getaddrinfo ENOTFOUND embed.gog.com
(14:39:33) WARNING: [Backend]:         Error Downloading Winetricks
(14:39:34) INFO:    [Frontend]:        Refreshing Library
(14:39:34) INFO:    [Legendary]:       Refreshing library...
(14:39:34) INFO:    [Gog]:             Getting GOG library
(14:39:34) ERROR:   [Gog]:             There was an error getting games library data getaddrinfo ENOTFOUND embed.gog.com
(14:39:34) ERROR:   [Gog]:             There was an error Loading games library
(14:39:34) ERROR:   [Frontend]:        Error: Error invoking remote method 'checkGameUpdates': Error: getaddrinfo ENOTFOUND content-system.gog.com
[...]

PR #1677, opened 2022-08-08
[Fix] Added check if heroic can communicate with internet (ping some common websites)

Nocccer:
Heroic checks if internet connection is available.
Using ping command to ping `google.com`, `store.epicgames.com` and `gog.com`.
If one succeed it will return true.

Tested on linux with a invalid url. Was succesfully marked as offline.
TODO: Windows needs to be tested.

---


flavioislima:
Will do some testing on windows and macOS.

flavioislima:
Still not working as it should.
If you start Heroic without internet connection and then connect again it won't update the status and wont update the game info.
And GOG game pages are not working when offline as well.

Nocccer:
> Still not working as it should. If you start Heroic without internet connection and then connect again it won't update the status and wont update the game info. And GOG game pages are not working when offline as well.

Where is the status? I mean it never worked before. Every function which needs a internet connection should listen to an event then?

Btw, what do you think about make the `isOnline` a boolean and create instead a interval function that checks connection each 10s. So all function which relies on the `isOnline` value, does not call ping.

flavioislima (approved):
Now it is working pretty well and completely fixes the bug in #1494

Nocccer:
Actually i think this is was not needed since ariel pr

flavioislima:
Which PR? I still think that pinging the websites is a good thing to have.

Nocccer:
Ah sorry thought this got also merged, because @arielj implemented a ping there aswell.
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1727

flavioislima:
> Ah sorry thought this got also merged, because @arielj implemented a ping there aswell.
> https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1727

A-ha. That's true I'm blind.
We can revert this one then or let's wait until Ariel put it in testing :)

arielj:
hey I'm traveling until the 28th so I won't have much time to continue that PR, I'll do a few more things there after I get back

PR #1678, opened 2022-08-08
[Fix] App breaking after GOG login

flavioislima:
This small fix was introduced on #1646. It breaks heroic on redirect from the login to library.

---


electron/gog/library.ts
```diff
@@ -628,6 +628,9 @@ export class GOGLibrary {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   public readInfoFile(appName: string, installPath?: string): any {
     const gameInfo = this.getGameInfo(appName)
+    if (!gameInfo) {
```
flavioislima, line 631: this is necessary because if there was no cache, heroic was not loading gog games, this forces it to create the cache.

Nocccer (approved):
LGTM

Issue #1681, opened 2022-08-08
"Game removed from Steam" notification for games that were not added to Steam


arielj:
### Describe the bug

When uninstalling a game, a notification saying the game was removed from Steam and that Steam needs to be restarted shows up even if you didn't add the game to Steam.

### Add logs

```
(20:14:04) INFO:    [Legendary]:       Uninstalling 19927295d6e3467887d4e830d8c85963: /home/ariel/dev/oss/HeroicGamesLauncher/public/bin/linux/legendary uninstall 19927295d6e3467887d4e830d8c85963 -y
(20:14:04) INFO:    [Legendary]:       Installed game list updated
(20:14:04) INFO:    [Legendary]:       Installed game list updated
(20:14:04) INFO:    [Backend]:         Desktop shortcut removed
(20:14:04) INFO:    [Backend]:         Applications shortcut removed
(20:14:04) INFO:    [Shortcuts]:       Absolute Drift was successfully removed from Steam.
(20:14:04) INFO:    [Backend]:         Removing prefix /home/ariel/Games/Heroic/Prefixes/AbsoluteDrift
(20:14:04) INFO:    [Backend]:         finished uninstalling
```

### Steps to reproduce

1. Install a game
2. Without adding the game as a non-Steam game on Steam, uninstall it
3. The notifications shows up

### Expected behavior

It should only notify that the game was removed if the game was actually added as a non-Steam game in Steam.

### Screenshots

![image](https://user-images.githubusercontent.com/188464/183530619-b3106b80-803d-4f13-ac1e-d6ab7aed8ad3.png)

After that it shows the `Uninstalled` notification.

### System Information

- OS [e. g. "Ubuntu"]: not relevant
- Heroic Version [e. g. 2.1.1]: latest beta branch

flavioislima:
That should be simple to fix I will add to the 2.4.0 milestone.

PR #1684, opened 2022-08-09
[Fix] Multiple frontend fixes for add to steam feature

Nocccer:
- don't show notification of removing from steam, if the game was never added to steam
- update the frontend state correctly.
    - was marked as added, still if a error occured. Added a boolean value as return for `addNoneSteamGame`
    - If the game was added to steam and uninstalled, the frontend was not updated, if you installed the game again and did not leave the gamepage, it still showed as added. Added `isInstalled` to the useEffect, so it will update the state if game is installed again.
- don't report any error on removing from steam to frontend. We just assume on error that the game is removed.
    - The error will catched and reported if he trys to add it again. (incorrect path, write permissions, etc.)
- removed i18n translations, whish are not needed anymore   

---


electron/shortcuts/nonesteamgame/nonesteamgame.ts
```diff
@@ -342,10 +326,11 @@ async function addNonSteamGame(props: {
 
     const message = i18next.t('notify.finished.add.steam.corrupt', {
       defaultValue:
-        '{{game}} could not be added to all found Steam users. See logs for more info.  A restart of Steam is required, changes to take effect.',
+        '{{game}} could not be added to all found Steam users. See logs for more info. A restart of Steam is required, changes to take effect.',
```
arielj, line 329: I noticed this string `A restart of Steam is required, changes to take effect` when adding translations for spanish the other day. I think `A restart of Steam is required for changes to take effect` makes more sense, it sounds like there's a missing word for me there (this is in many of the strings).

arielj (approved):
I just left a comment about the strings but I can confirm this fix the problem of the removed notification #1681

PR #1693, opened 2022-08-10
[Fix] Disable notifications if they are not supported. (E.g. Steamdeck Gamemode)

Nocccer:
Current heroic flatpak freezes in gamemode, because electron can't show a notification.
- Disable notifications on gamemode
- Disable notifications if they are not supported by the system. (https://www.electronjs.org/de/docs/latest/api/notification#notificationissupported)

---


flavioislima (approved):
That was a good catch. I believe we had this issue in the past with other linux distros without libnotify installed.

Issue #1712, opened 2022-08-12
Environment variables table improvements


arielj:
### Problem description

The environment variables table is a great improvement, but there are some things that can be improved to make it more user friendly.

### Feature description

Some ideas:
- add an `(i) Info` block so people can see how it's used with a simple example like "for a variable like MY_VAR=1, put `MY_VAR` as name and `1` as value, then click the `+` button to add it"
- add a `=` character between the fields so users understand better what goes where
- detect if the value of the name includes an `=` sign and automatically split it and put the correct value in each field
- inside a game's setting, show (as readonly) the globally configured env variables so you can see what's happening
- detect when a user puts a variable of a feature we already handle with a toggle and suggest removing it and using the toggle
- detect duplicated variable names
- add a warning next to the fields saying that the variable is NOT saved automatically and the `+` click is required (since we have a comment in the settings saying that settings are saved automatically!)

Nocccer:
- already detect duplicated variable names
- i would prefer the user envs over our toggles. So the user can override the behavior we add, if he wants too.

arielj:
> * already detect duplicated variable names
oops didn't know, does it detect duplications comparing with the global env variables too?

> * i would prefer the user envs over our toggles. So the user can override the behavior we add, if he wants too.
hmmm does this mean our toggles are not correct? not sure I understand the use case

wild idea: make our toggles add env variables to the table directly instead of handling things in the background? or maybe show in the `(i)` tooltip for the toggles what env variable is being added?

Nocccer:
> > * already detect duplicated variable names
> 
> oops didn't know, does it detect duplications comparing with the global env variables too?
> 
> > * i would prefer the user envs over our toggles. So the user can override the behavior we add, if he wants too.
> 
> hmmm does this mean our toggles are not correct? not sure I understand the use case
> 
> wild idea: make our toggles add env variables to the table directly instead of handling things in the background? or maybe show in the `(i)` tooltip for the toggles what env variable is being added?

Between global and game i don't know. I don't even know how global envs are working. From my understanding global settings are the default values passed to a fresh installed game.

Today a user had problems to run Mangohud on a opengl game. We can't detect if the game runs with vulkan or opengl. He solved it by adding a env.

arielj:
> > > * already detect duplicated variable names
> > 
> > 
> > oops didn't know, does it detect duplications comparing with the global env variables too?
[...]

PR #1720, opened 2022-08-13
Some UX improvements on env variables

arielj:
This PR adds many of the improvements listed in #1712 

- Added an `=` sign between the input fields so it gives users a hint that the `=` sign splits the two inputs

- If an `=` sign is added in the key field and the value is empty, the key is split in 2, the left side is set as key and the right side is set as value (useful if somebody copy paste a `MY_FLAG=something` as the key)

- If an `=` sign is added in the key field and the value is NOT empty, don't do the split but show an error to the user that `=` is not allowed in env variable names
image

- If there's a space in the variable name, show an error and prevent adding it
image

- If the variable name is empty, show an error and prevent adding it
image

- If the fields have `dirty` values (the user edited anything), show a message letting the user know the changes are not saved automatically
image

- Added an info box to explain how this works
image

I also extracted the env variables table in it's own component.

I also renamed the label from `Advanced (Environment Variables)` to `Environment Variables`, I can revert this though, but I don't think it should be called `Advanced`, that's too generic.

---

### DIFF
--- a/public/locales/bg/translation.json
+++ b/public/locales/bg/translation.json
@@ -322,8 +322,17 @@
             "key": "Variable Name",
             "placeHolderKey": "NAME",
             "placeHolderValue": "Пример: път/до/допълните/файлове",
-            "title": "Разширени настройки (променливи на средата):",
+            "title": "Environment Variables",
             "value": "Стойност"
+        },
+        "env_variables": {
+            "error": {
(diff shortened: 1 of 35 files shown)


flavioislima (approved):
Looks and works pretty good 👍🏽

PR #1725, opened 2022-08-14
Remove extra game status change when killing a game

### LINKED ISSUE #1698: `Play` buttons stuck in the `Play` state
### Describe the bug

The Play button takes some time to change its state to the "playing" state (the `x` that allows the user to force-stop the game). During that time, it's still showing the Play action and you can click it multiple times and trigger more launches of the game (which show an error of already running), you may click it more than once if it takes too long to change and you think you didn't click it.

This doesn't happen with the Play button in the game details, that one changes to the `playing` state immediately.

After that happens, both the Play button in the library and the Play button in the game details are always stuck in the `Play` state (they never change to `playing (stop)`) until Heroic is restarted so you can't force-stop a game anymore.

Another way to produce the same issue (not sure if it's the same cause but it generates the same problem) is actually using the `force stop` feature (click the `X` in the library when a game is running) to kill the game. After that, the same problem with the `Play` buttons stuck in the `Play` state is present.

Similar to the other case, killing a game using the `Playing (STOP)` button in the game details page does NOT produce this issue.

### Add logs

There's nothing relevant on the logs, just the normal launch game (and sometimes multiple launch when multiclicking) and the kill commands (when killing the game). But no errors or anything.

Also nothing on the frontend.

### Steps to reproduce

First way:
- double (triple, ...) click the Play button in the Library for a game

Second way:
- click the Play button in the Library for a game only once and when the game started kill it with the `(X)` button in the Library

After doing any of those things, the play buttons are broken.

### Expected behavior

The `Play` button in the Library should behave the same way as the one in the game details page. Clicking it should change it's state immediately. It shouldn't allow multiple click. Killing a game by clicking it in the `(X)` state should not break the play buttons.

### Screenshots

_No response_

### System Information

- Heroic Version [e. g. 2.1.1]: beta branch (I think this error has been here for a while though, I remember experiencing this issue a few months ago too but I didn't know how to reproduce it)

### Additional information

I suspect it's some mix between the process that keeps track of the game that's running and something on the React app not getting updated properly thinking the game never started. We can certainly fix this by comparing the play button in the game cards and the play button in the game details and making sure they work the same way.

arielj:
This PR fixes #1698 . The problem is better explained in the linked issue but basically, when killing a game using the `(x)` button in the library, the libraryStatus global state got corrupted because of the line I'm now removing.

This line is not used when killing a game in the Game Page component so I understand it's safe to remove and it fixes the problem.

---

### DIFF
--- a/src/screens/Library/components/GameCard/index.tsx
+++ b/src/screens/Library/components/GameCard/index.tsx
@@ -427,7 +427,6 @@ const GameCard = ({
       })
     }
     if (status === 'playing' || status === 'updating') {
-      await handleGameStatus({ appName, runner, status: 'done' })
       return sendKill(appName, runner)
     }
     if (isInstalled) {


flavioislima (approved):
I think we probably forgot to remove that line on a previous refactoring.

PR #1726, opened 2022-08-14
[Refactor] Settings

arielj:
This PR is a refactor on how we handle settings to make them more modular and less coupled.

## Before testing this, backup your settings!

Instead of having all config values as different states in the main Settings component, a SettingsContext is added wrapping everything and each setting (or group or related settings) can be a standalone component.

This gives us many benefits:
- moving settings around is really simple, since a setting is just one component with no props
- improves readability a lot in my opinion
- adding new settings involves way less code and touching less files

---


src/components/UI/TwoColTableInput/index.tsx
```diff
@@ -210,24 +210,24 @@ export function TableInput({
...
+                    {t(
+                      'two_col_table.save_hint',
+                      'Changes in this table are not saved automatically. Click the + button'
+                    )}
+                  </span>
+                  <FontAwesomeIcon icon={faArrowUp} />
+                </>
+              )}
+            </td>
+          </tr>
```
arielj, line 230: these changes are only to remove an error in the devtools console, the conditionals were rendering an empty text node when false and react doesn't like that

Nocccer (approved):
Create stuff. Actually i wonder if it would not be better to merge it against beta, so we can test this more with a beta release.

If so i refactored the logger, so you might get some conflicts. You can now directly pass arguments without wrap them into a string or parse them via `JSON.parse`. Also the prefix is now behind a options object

example:
```typescript
const obj = {first: 'Hello', last: 'World'}
const err = new Error('Something failed!')

logError(['Object', obj, 'failed with:', err], {prefix: LogPrefix.Backend})
```

arielj:
> Create stuff. Actually i wonder if it would not be better to merge it against beta, so we can test this more with a beta release.
> 
> If so i refactored the logger, so you might get some conflicts. You can now directly pass arguments without wrap them into a string or parse them via `JSON.parse`. Also the prefix is now behind a options object
> 
> example:
> 
> ```ts
> const obj = {first: 'Hello', last: 'World'}
> const err = new Error('Something failed!')
> 
> logError(['Object', obj, 'failed with:', err], {prefix: LogPrefix.Backend})
> ```

sure! I'll change it to beta and fix conflicts and stuff

arielj:
This is ready to check against beta now.

flavioislima (commented):
I am still not convinced that this is the best approach. I think having one component for each setting is a bit overkill in the long run and also leads to code repetition.

I was thinking in having settings inside `electron-store` for easy sharing between BE and FE and also easier to set since we could set directly into the component instead of having a `toggle`, `useState`, etc on `Settings` that is shared with the others like it is today.

So we could just import the setting store inside WineSettings, for instance, the method to change each setting would use the methods inside `electron-store` and whenever we need to read this settings, we just need to use it's methods as well. It is easier to share I believe.

arielj:
One comment on the repetition, what's repeated is usually the imports and some useContext stuff, but most of the code of each component is just the code that was already there moved as a single component with a single responsibility (or maybe 2 responsibilities if 2 settings are coupled).

What I found complicated with a big component full of settings was:
- it was really bad for readability (we currently have the setting html and the setting read/write scattered all over the place mixed with other functions for other settings)
- props passing all over the place (this was killing me xD)
- bad for maintainability
- confusing to add new settings (it took me a long time one some of my first contributions to understand all the things I have to do just to add 1 boolean setting)

With a setting wrapped in a component, reordering settings is just moving one tag somewhere else and adding settings requires adding 1 file with some predefined pattern and then putting the tag somewhere.

We can still use electron store, the `useSetting` hook could be changed to read/write to the electron store instead, even if we don't use different components I think we should still wrap the interaction with the electron store with a hook.

I do think there's probably some way to refactor the settings to have less repetition, I just didn't want to go too wild in a first iteration of this.

Happy to discuss this more and compare to other ideas, what I want to avoid is big components full of html, getters and setters all mixed in hundreds of lines of code.

flavioislima:
I see your point.
What would be the step-by-step to add a new setting?

Also, currently, we have one useEffect to get the whole bunch of settings and with this change, we need a `useEffect` on each of the components to get that specific setting. How is this affecting performance and re-rendering of the Settings screen?

arielj:
The steps would be:
- create a new component for the setting
- use the `useSetting` to get the value and the setter of the setting (you define the default too)
- get anything else you need (maybe other info from the context like the appName, the `t` function, etc)
- render whatever you need for a particular setting
- put the component tag where you need it

Regarding the useEffect, we still only do one that gets all settings at once in the Settings component, then all the settings are shared in the SettingsContext. The useEffect inside the useSetting hook is used only to detect when the setting changes but not to read the setting.

flavioislima:
Cool. Then I will do some tests and we can proceed 👍🏽

src/backend/api/helpers.ts
```diff
@@ -9,6 +9,7 @@ export const openSupportPage = () => ipcRenderer.send('openSupportPage')
 export const quit = () => ipcRenderer.send('quit')
 export const showAboutWindow = () => ipcRenderer.send('showAboutWindow')
 export const openDiscordLink = () => ipcRenderer.send('openDiscordLink')
+export const openWinePrefixFAQ = () => ipcRenderer.send('openWinePrefixFAQ')
```
arielj, line 12: This got lost at some point https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1774

flavioislima:
Did some testing here and most of the things are running fine except:
- Cannot toggle GameMode
- Add games to start menu is using wrong config name. The correct is `addStartMenuShortcuts` but it is using `addGamesToStartMenu` (the confusion is probably because of the i18n key.
- In the config file `language` is not updating, although the language change is working fine 🤷🏽

arielj:
@flavioislima I fixed the issue with the GameMode option and the key for the `addStartMenuShortcuts` config.

I checked the language thing and it's an `issue` currently in beta (and probably before that too). The language setting is stored at `~/.config/heroic/config.json` (along with screen size, recent games, etc), but the settings from the "settings" screen are stored at `~/.config/heroic/stores/config.json`.

I think I could change this to update the `language` value in the `~/.config/heroic/stores/config.json` file but I don't know what's going on there. Maybe it's better to handle that in a different PR?

I think we need to plan how to reorganize these files because I also see the file that store the actual language value has this for example:

```
	"settings": {
		"altLeg": "",
		"altGogdl": ""
	},
```

So there seem to be some mix of things going on there before these changes.

flavioislima:
> @flavioislima I fixed the issue with the GameMode option and the key for the `addStartMenuShortcuts` config.
> 
> I checked the language thing and it's an `issue` currently in beta (and probably before that too). The language setting is stored at `~/.config/heroic/config.json` (along with screen size, recent games, etc), but the settings from the "settings" screen are stored at `~/.config/heroic/stores/config.json`.
> 
> I think I could change this to update the `language` value in the `~/.config/heroic/stores/config.json` file but I don't know what's going on there. Maybe it's better to handle that in a different PR?
> 
> I think we need to plan how to reorganize these files because I also see the file that store the actual language value has this for example:
> 
> ```
> 	"settings": {
> 		"altLeg": "",
> 		"altGogdl": ""
> 	},
> ```
> 
> So there seem to be some mix of things going on there before these changes.

I think we were testing changing the settings to be in electron-store so at the time I added these settings there but they are not used I believe.

I think having the language setting with the other settings would be better for consistency imo.

But the PR is good now, pretty good refactor 👍🏽

flavioislima (approved):

PR #1732, opened 2022-08-15
Images offline cache

arielj:
This PR adds a mechanism to cache game images in the file system to be overcome the issue with the chrome's cache having a limit in size.

How it works:
- added a custom protocol `imagecache://` intercepted by electron to respond with a local file
- added a CachedImage component that first tries to use the cache, and, if the request fails, it fallback to the original url
- added an images_cache module with a function that will convert a `imagecache://` url into a local file response and, if the file is not already cached, cache it in the background

this was suprisingly way easier than I expected xD

---

### DIFF
--- a/electron/images_cache.ts
+++ b/electron/images_cache.ts
@@ -1,0 +1,34 @@
+import { existsSync, createWriteStream, mkdirSync } from 'graceful-fs'
+import { createHash } from 'crypto'
+import { imagesCachePath } from './constants'
+import { join } from 'path'
+import axios from 'axios'
+import { protocol } from 'electron'
+
+export const initImagesCache = () => {
+  // make sure we have a folder to store the cache
+  if (!existsSync(imagesCachePath)) {
+    mkdirSync(imagesCachePath)
+  }
+
+  // use a fake protocol for images we want to cache
+  protocol.registerFileProtocol('imagecache', (request, callback) => {
+    callback({ path: getImageFromCache(request.url) })
+  })
+}
+
+const getImageFromCache = (url: string) => {
+  const realUrl = url.replace('imagecache://', '')
+  // digest of the image url for the file name
+  const digest = createHash('sha256').update(realUrl).digest('hex')
+  const cachePath = join(imagesCachePath, digest)
+
+  if (!existsSync(cachePath)) {
+    // if not found, download in the background
+    axios
+      .get(realUrl, { responseType: 'stream' })
+      .then((response) => response.data.pipe(createWriteStream(cachePath)))
+  }
+
+  return join(cachePath)
+}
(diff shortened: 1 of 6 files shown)


electron/images_cache.ts
```diff
@@ -0,0 +1,34 @@
+import { existsSync, createWriteStream, mkdirSync } from 'graceful-fs'
```
Nocccer, line 1: Would be good if we already do a strucutre for this.
To many files in `electron` makes things confusing.

What do you think about `electron/common/imagecache` or `electron/imagecache` or `electron/image/cache` ?

Nocccer (changes requested):
LGTM. I would love to get rid of placing files directly into electron folder.

electron/images_cache.ts
flavioislima: electron already creates a Cache directory, maybe use that one.
`heroic/Cache/images`

flavioislima: and I believe you can do something like:
`const imagesCachePath = app.getPath('cache')/images`

flavioislima:
Testing here and it works fine for the library but not for the game page.
Also, would not be better to use the `appNames` for the images instead of a hash? Not sure if that is important, just a thought.

image

electron/images_cache.ts
arielj: I'll use the current heroic/Cache :+1:

arielj:
> Testing here and it works fine for the library but not for the game page. Also, would not be better to use the `appNames` for the images instead of a hash? Not sure if that is important, just a thought.
> 
> image

I'll check the game image!

I thought about using the app name, but then we have 2 different sizes for the main image plus the logo, so I would have to come up with a pattern to generate file names and also I would have to pass the app name and image type around in the url (I guess I could get the requested size from the url), but it sounds like too complicated for something that is just a cache, the digest ensures it's unique and it's really simple

arielj:
@flavioislima I fixed the issue with the game image (added a fallback src to use the smaller one if it fails)

I couldn't use `heroic/Cache/images`, electron clears that folder at boot

electron/images_cache.ts
arielj: oh I see what you mean @Nocccer , you don't mean the cache, you directory, you mean where I puth the images_cache.ts file

arielj: I agree that we need some convention, I put the file there because there are other files there and adding one new folder for just one small file felt weird to me, but I think we need to find some pattern to use to put this accessory files

flavioislima:
NIce, Looks good now! Ready to merge! 👍🏽

flavioislima (approved):

PR #1736, opened 2022-08-18
[Wine] Bottles integration

imLinguin:
Ability to run games within Bottles. Supporting both Flatpak and bottles binary found in `PATH`

After this would get merged the new flatpak manifest flag is needed `--talk-name=org.freedesktop.Flatpak` for ability to launch stuff outside the flatpak container - in our case flatpak version of Bottles. 

Bottles flatpak will need to have permissions to read game library in order to work.

This PR adds new script in `bin/linux` called `bottles-wine` it basically takes provided arguments in wine like factor and runs the them properly in factor `bottles-cli` accepts. This allows us to skip the need to implement Bottles handler in gogdl and legendary

Features: 
- detect bottles version
- list available bottles
- add button to launch bottles ui for selected bottle (for configuration)
- added button manually run setup for gog games


JukoiDev (approved):
lgtm

public/bin/linux/bottles-wine
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
```
Nocccer, line 1: Why we need a shell script? Can we not run it via typescript?
What is the problem not running it from heroic itself?

src/screens/Game/GameSubMenu/index.tsx
```diff
@@ -244,6 +244,10 @@ export default function GamesSubmenu({
     setSteamRefresh(false)
   }
 
+  async function handleGogSetup() {
+    await ipcRenderer.invoke('runGOGSetupForGame', appName)
```
Nocccer, line 248: We need a state here like we have for add to steam and add shortcuts. Was it already setup? Show a circle during setup...

Nocccer (changes requested):
Overall looks pretty good. I still find some stuff we need to talk about.

electron/launcher.ts
```diff
@@ -440,12 +441,13 @@ async function runWineCommand(
...
 
-  const { wineVersion } = gameSettings
+  const { wineVersion, bottlesBottle } = gameSettings
 
   const env_vars = {
     ...process.env,
     ...setupEnvVars(gameSettings),
-    ...setupWineEnvVars(gameSettings, installFolderName)
+    ...setupWineEnvVars(gameSettings, installFolderName),
+    HGL_BOTTLE_NAME: gameSettings.bottlesBottle
```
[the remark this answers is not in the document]
CommandMC: Yup, should be in `setupWineEnvVars` (specifically in the switch statement at the start there)

src/screens/Game/GameSubMenu/index.tsx
imLinguin: I added this button in this PR too, since we can't really tell if it was already ran in Bottles or Crossover

JakobDev:
Maybe you should ask the Bottles Developers to provide a D-Bus Interface, so giving the Flatpak the Permission to run all commands outside a Sandboy is not needed

imLinguin:
I talked with Mirko, he'll adapt xdg-open protocol to handle argument like bottles-cli would https://docs.usebottles.com/advanced/xdg-open

src/backend/bottles/ipc_handler.ts
```diff
@@ -0,0 +1,6 @@
+import { ipcMain } from 'electron'
+import { getBottlesNames } from './utils'
+
+ipcMain.handle('bottles.getBottlesNames', async (event, bottlesBin) => {
+  return getBottlesNames(bottlesBin)
+})
```
CommandMC, line 6:
```suggestion
ipcMain.handle('bottles.getBottlesNames', async (event, bottlesBin) => getBottlesNames(bottlesBin)
```

src/backend/api/wine.ts
```diff
@@ -49,3 +49,7 @@ export const handleProgressOfWineManager = (
     ipcRenderer.removeListener('progressOfWineManager' + version, callback)
   }
 }
+
+// Bottles
+export const getBottlesNames = async (bottlesType: string) =>
```
CommandMC, line 54: Adding a return type here would be great

src/backend/api/menu.ts
```diff
@@ -35,3 +35,5 @@ export const isAddedToSteam = async (appName: string, runner: Runner) =>
   ipcRenderer.invoke('isAddedToSteam', appName, runner)
 export const isEosOverlayEnabled = async (appName?: string) =>
   ipcRenderer.invoke('isEosOverlayEnabled', appName)
+export const runGogSetupForGame = async (appName: string) =>
```
CommandMC, line 38: Same here, now that we *can* add return types to IPC functions, I'd say we should

src/backend/bottles/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+import { logDebug, LogPrefix } from '../logger/logger'
+import { isFlatpak } from '../constants'
+import shlex from 'shlex'
+
+function prepareBottlesCommand(
+  args: string[],
+  bottlesBin: string,
+  noCli?: boolean,
+  join?: boolean
+): string[] | string {
```
CommandMC, line 11: I wouldn't recommend adding options like this that alter the return type of a function. Type checking gets annoying with that rather fast, since TS can't know when you're returning a `string` and when it's a `string[]`
Just a suggestion of course, but maybe do the `shlex.join` in the calling function (or in a small wrapper) & always return a `string[]` here

src/backend/bottles/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  command.push(...args)
+  if (join) {
+    return shlex.join(command)
+  }
+  return command
+}
+
+async function getBottlesNames(
+  bottlesBin: 'flatpak' | 'os'
```
CommandMC, line 36: You sometimes use `'flatpak' | 'os'` as your `bottlesBin`/`bottlesType`, and other times you just use `string`
Consider adding this type as an actual new `type` (either here or in `src/common/types.ts`) and referencing it everywhere, otherwise this is confusing for people who didn't write the code :^)

src/backend/bottles/utils.ts
```diff
@@ -0,0 +1,114 @@
+import { spawn } from 'child_process'
+import { logDebug, LogPrefix } from '../logger/logger'
+import { isFlatpak } from '../constants'
+import shlex from 'shlex'
+
+function prepareBottlesCommand(
+  args: string[],
+  bottlesBin: string,
+  noCli?: boolean,
+  join?: boolean
```
CommandMC, line 10:
```suggestion
  noCli = false,
  join = false
```
This is mostly preference, but it also makes it clear to whoever is calling this function what the default values are (yes, "missing -> false/disabled" might be obvious, but it's one more thing to think about)

src/backend/gog/games.ts
```diff
@@ -383,11 +384,17 @@ class GOGGame extends Game {
...
-          ? ['--no-wine', '--wrapper', `'${wineBin}' run`]
-          : ['--wine', wineBin])
-      )
+      if (wineType === 'proton') {
+        wineFlag.push(...['--no-wine', '--wrapper', `'${wineBin}' run`])
+      } else if (wineType === 'bottles') {
+        commandEnv = {
+          ...commandEnv,
+          HGL_BOTTLE_NAME: gameSettings.bottlesBottle
+        }
```
CommandMC, line 393: Can't test right now, but wouldn't
```suggestion
        commandEnv.HGL_BOTTLE_NAME = gameSettings.bottlesBottle
```
work?

src/backend/gog/setup.ts
```diff
@@ -123,7 +130,31 @@ async function setup(
             keyCommand = `/d "${valueData}" /v "${valueName}" /t ${regType}`
           }
           // Now create a key
-          const command = `${commandPrefix} reg add "${registryPath}" ${keyCommand} /f /reg:32`
+          let command = `${commandPrefix} reg add "${registryPath}" ${keyCommand} /f /reg:32`
+          if (gameSettings.wineVersion.type === 'bottles') {
+            if (!valueName && !valueData) {
+              break // Skip creating just registry path, bottles CLI needs some key to add
```
CommandMC, line 136: Every registry folder has the `(Default)` key (type `REG_SZ` with no data set), can't you add that?

src/backend/gog/setup.ts
```diff
@@ -123,7 +130,31 @@ async function setup(
             keyCommand = `/d "${valueData}" /v "${valueName}" /t ${regType}`
           }
           // Now create a key
-          const command = `${commandPrefix} reg add "${registryPath}" ${keyCommand} /f /reg:32`
+          let command = `${commandPrefix} reg add "${registryPath}" ${keyCommand} /f /reg:32`
```
CommandMC, line 133: *Someone* should port this to using `runWineCommand` someday

src/backend/main.ts
```diff
@@ -645,14 +646,21 @@ ipcMain.handle(
...
       case 'winecfg':
+        if (wineVersion.type === 'bottles') {
+          await runBottlesCommand(
+            ['tools', 'winecfg', '-b', bottlesBottle],
+            wineVersion.bin
+          )
+          break
+        }
         game.runWineCommand('winecfg')
         break
```
CommandMC, line 665:
```suggestion
        } else {
          game.runWineCommand('winecfg')
        }
        break
```

CommandMC (commented):
This is a great addition. At first, I was also team "Why do we need the extra script", but now looking at the full PR, it makes it a lot more convenient to add this without changing too much code
Left a few comments & questions

src/backend/gog/setup.ts
imLinguin: Empty registry keys are not needed either way. There are always other instructions that set some value in the key

src/frontend/screens/Settings/components/WineExtensions/index.tsx
```diff
@@ -1,12 +1,51 @@
...
+      <>
+        <h3 className="settingsSubheader">
+          {t('settings.navbar.wineExt', 'Wine Extensions')}
+        </h3>
+
+        <div className="infoBox saves-warning">
+          <FontAwesomeIcon icon={faExclamationTriangle} color={'yellow'} />
+          {t(
+            'settings.wineExt.bottlesInformation',
+            'You can configure things like DXVK, VKD3D and more, in Bottles UI'
```
arielj, line 34:
```suggestion
            'You can configure DXVK, VKD3D and more, in Bottles UI'
```

Just a suggestion, we are already saying `and more` so we don't need the `things` and it may make translations easier.

src/backend/legendary/games.ts
```diff
@@ -634,11 +635,17 @@ class LegendaryGame extends Game {
           ? wineExec.replaceAll("'", '')
           : wineExec
 
-      wineFlag.push(
-        ...(wineType === 'proton'
-          ? ['--no-wine', '--wrapper', `'${wineBin}' run`]
-          : ['--wine', wineBin])
-      )
+      if (wineType === 'proton') {
+        wineFlag.push(...['--no-wine', '--wrapper', `'${wineBin}' run`])
```
arielj, line 639: We can remove these `...[___]` syntax now that we don't have the ternary operator:

```suggestion
        wineFlag.push('--no-wine', '--wrapper', `'${wineBin}' run`)
```

src/backend/gog/games.ts
```diff
@@ -383,11 +384,14 @@ class GOGGame extends Game {
           ? wineExec.replaceAll("'", '')
           : wineExec
 
-      wineFlag.push(
-        ...(wineType === 'proton'
-          ? ['--no-wine', '--wrapper', `'${wineBin}' run`]
-          : ['--wine', wineBin])
-      )
+      if (wineType === 'proton') {
+        wineFlag.push(...['--no-wine', '--wrapper', `'${wineBin}' run`])
```
arielj, line 388: Same suggestion to remove these `...[____]` pushes

```suggestion
        wineFlag.push('--no-wine', '--wrapper', `'${wineBin}' run`)
```

src/backend/constants.ts
```diff
@@ -82,7 +82,9 @@ const weblateUrl = 'https://hosted.weblate.org/projects/heroic-games-launcher'
 const kofiPage = 'https://ko-fi.com/heroicgames'
 const patreonPage = 'https://www.patreon.com/heroicgameslauncher'
 const wineprefixFAQ = 'https://wiki.winehq.org/FAQ#Wineprefixes'
-
+const bottlesWineBin = fixAsarPath(
+  join(publicDir, 'bin', 'linux', 'bottles-wine')
+)
```
arielj, line 87: would it be useful to add this path in the logs when Heroic starts? we are already printing Gogdl and Legendary

I've never used bottles so I don't know if it might be useful to print the path to bottles for when we need to help people with issues or if the `wineVersion` setting will include all we need

arielj (commented):
looks good! I added some suggestions

I don't really use bottles so I didn't test this but it looks good code-wise

src/backend/constants.ts
CommandMC: Well, the script path isn't all too interesting

imLinguin:
Closing this for now, will rebase it properly soon

Issue #1737, opened 2022-08-19
Empty blue window since latest update


Jpxe:
### Describe the bug

I use the Flatpak version on Fedora (Nobara Linux 36) and since the latest update it just shows a blue screen. Everything worked fine in the last version. Screenshot is attached below.

### Add logs
```
[jpx@fedora ~]$ flatpak run com.heroicgameslauncher.hgl
[3:0819/194543.221123:ERROR:bus.cc(398)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: Filen eller katalogen finns inte
[3:0819/194543.221264:ERROR:bus.cc(398)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: Filen eller katalogen finns inte
(19:45:43) INFO:    [Legendary]:       Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary --version
hostname: ogiltig flagga -- "f"
Försök med ”hostname --help” för mer information.
(19:45:43) INFO:    [Legendary]:       Legendary location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(19:45:43) INFO:    [Gog]:             GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(19:45:43) INFO:    [Backend]:         

Heroic Version: 2.4.0 Chopper
Legendary Version:  0.20.27 Dark Energy (hotfix)
OS: Freedesktop.org KERNEL: 5.18.17-201.fsync.fc36.x86_64 ARCH: x64
CPU: Intel Core™ i5-4670K @3.4 GOVERNOR: schedutil
RAM: Total: 15.57 GiB Available: 9.58 GiB
GRAPHICS: GPU0: GM204 [GeForce GTX 980]  VRAM: 32MB DRIVER:  
PROTOCOL: x11

(19:45:43) INFO:    [Gog]:             Getting data about the user
APPIMAGE env is not defined, current application is not an AppImage
LaunchProcess: failed to execvp:
xdg-settings
(19:45:43) WARNING: [Backend]:         Failed to register protocol with OS.
[3:0819/194543.927343:ERROR:browser_main_loop.cc(270)] Gtk: gtk_widget_add_accelerator: assertion 'GTK_IS_ACCEL_GROUP (accel_group)' failed
[...]

Issue #1707, opened 2022-08-11
[UI] Add information for Nvidia Prime


Nocccer:
### Problem description

Missing info for nvidia prime switch. we need to add information for that and also call it NVIDIA Prime again!
![image](https://user-images.githubusercontent.com/61798668/184217618-f33a564e-4447-45ba-9bf5-95c630664113.png)

### Feature description

none

flavioislima:
We can maybe add both but this setting works for both Nvidia and AMD.
We changed it exactly because people from AMD didn't know they could use that.
Unless we find a way of checking the Graphic card and then show the correct option, what do you think?

Nocccer:
Atm the option is unclear. We need the info hover, like on other options!

Raycert07:
Here's my idea, something along the lines of

> option uses dedicated graphics card to render game on multi gpu systems. Only needed on gaming laptops, or desktops that use a headless gpu for graphics rendering.

Issue #1486, opened 2022-06-16
Remove Prefix needs to be explained better


LiamDawe:
### Problem description

When you remove a game, Heroic asks you if you wish to remove the Prefix.

This is technical jargon, that most people won't have any idea what it actually means.

This needs to be explained better for users.

### Feature description

Less tech jargon, more simple text.

### Additional information

In the age of the Steam Deck and tons of people using it, keep it simple.

LiamDawe:
Example: https://www.reddit.com/r/SteamDeck/comments/vd36hy/heroic_launcher_would_you_like_to_remove_the/

redromnon:
Just as a suggestion, apart from the improved Remove Prefix dialog, a **help button** can be implemented right under the Wine Prefix in the game's Wine settings.
This help button will expand on a click and briefly explain what a Prefix is just like how it does for Wine Version, Advanced Options and Game Arguments.

PR #1740, opened 2022-08-20
[Fix] Multiple small fixes

Nocccer:
- Fixed some TypeErrors according to: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1737
    - we filter out empty game names after the map and not before on recentGames
    - for searchBar we didn't not filter out empty objects, leads to no render of heroic.
    - tested by two users and it fixes there problem.
- Added help information to prime option: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1707
- Added wine prefix information button: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1486

---


src/helpers/library.ts
```diff
@@ -302,14 +298,9 @@ type RecentGame = {
...
-    recentGames
-      .map((game) => {
-        return libraries.find((info) => info.app_name === game.appName)
-      })
-      // With this `.filter`, no empty entries can be returned. TS doesn't understand this, so we have to add the `as GameInfo[]`
-      .filter(Boolean) as GameInfo[]
-  )
+  return recentGames.filter(Boolean).map((game) => {
+    return libraries.find((info) => info.app_name === game.appName)
+  }) as GameInfo[] // With this `.filter`, no empty entries can be returned. TS doesn't understand this, so we have to add the `as GameInfo[]`
```
CommandMC, line 303: Maybe move/modify this comment now as well

src/screens/Settings/components/OtherSettings/index.tsx
```diff
@@ -277,12 +277,23 @@ export default function OtherSettings({
...
+              handleChange={togglePrimeRun}
+              title={t('setting.primerun', 'Use Dedicated Graphics Card')}
+            />
+
+            <FontAwesomeIcon
+              className="helpIcon"
+              icon={faCircleInfo}
+              title={t(
+                'help.primerun',
+                'Use dedicated graphics card to render game on multi gpu systems. Only needed on gaming laptops, or desktops that use a headless gpu for graphics rendering, e.g.: laptops with NVIDIA Optimus'
```
CommandMC, line 293:
```suggestion
                'Use dedicated graphics card to render game on multi-GPU systems. Only needed on gaming laptops or desktops that use a headless GPU for rendering (NVIDIA Optimus, AMD CrossFire)'
```

src/helpers/library.ts
```diff
@@ -185,16 +185,12 @@ async function uninstall({
...
-        `${t(
-          'gamepage:box.uninstall.checkbox_prefix',
-          'Prefix'
-        )}: ${wineprefix}`
-      ].join('\n'),
+      checkboxLabel: t('gamepage:box.uninstall.checkbox', {
+        defaultValue:
+          "Remove prefix: {{prefix}}{{newLine}}Note: This can't be undone and will also remove not backed up save files.",
+        prefix: wineprefix,
+        newLine: '\n'
```
CommandMC, line 192: This feels a little weird, why not just put the newline into the `defaultValue` directly?

CommandMC (commented):
Looks great in general, I especially like the Wineprefix clarifications

src/helpers/library.ts
Nocccer: I don't know how webplate or i18n handles this. Maybe it shows the \n in the translation and break something?
With the variable i know webplate and i18n can handle this. Didn't find a useful entry if `\n` is supported.

CommandMC: *I think* `\n` is handled properly, at least some other parts of the Weblate docs mention it
But doing it like this with the variable also works of course

src/helpers/library.ts
```diff
@@ -300,7 +300,7 @@ function getRecentGames(libraries: GameInfo[]) {
     (configStore.get('games.recent', []) as Array<RecentGame>) || []
   return recentGames.filter(Boolean).map((game) => {
     return libraries.find((info) => info.app_name === game.appName)
-  }) as GameInfo[] // With this `.filter`, no empty entries can be returned. TS doesn't understand this, so we have to add the `as GameInfo[]`
+  }) as GameInfo[] // needed to return empty array instead undefined
```
CommandMC, line 303: Hm, now that I see it, is this `as GameInfo[]` even right anymore? Don't we need *another* `.filter()` to make sure no `undefined` can get into the list?

Nocccer: See filter above (`.filter(Boolean)`). This filter was already there before but wrong placed. It should be before `libraries.find`, not after.

src/helpers/library.ts
Nocccer: I don't mind. Can add the \n. Woould be good to know how the translator prefer it.

src/helpers/library.ts
CommandMC: I'm aware the `.filter` was there already, I wrote the code. What I'm saying is, if `libraries.find(...)` returns `undefined` (as far as I know that's what it does if no element is found), wouldn't we then return an array of `GameInfo | undefined`, not just `GameInfo`?

As far as I understand, this code maps recent games (which are stored locally) to their respective GameInfo. What if you had a game that is for some reason no longer in your account (for example, you refunded it)? Wouldn't that then result in the config store still having the recent game stored, but the `libraries` not including it anymore? So wouldn't that then mean the second `.filter(...)` *is* necessary?

Nocccer: Maybe, but this is not the topic of this PR. The PR is about to fix the empty screen and not to take care of refunded games. Also from my understanding `libraries.find` does actually compare recent against the libraries collection. A removed game should not be in the library on restart of heroic. So the fix should be on another place, if it is a problem.

The GameInfo[] | undefined was a problem also before. Will see how i can solve that.

Nocccer: Ok i checked back. The approach before was correct.
A undefined[] array is still a object but with length 0. Casting it to an empty GameInfo[] array is allowed.

CommandMC: > Maybe, but this is not the topic of this PR

That was the point of the `.filter(Boolean)` before. By moving it elsewhere, you're removing that fix

> The GameInfo[] | undefined was a problem also before. Will see how i can solve that.

No, with the `.filter(Boolean)` it wasn't, since `undefined` is filtered out (this is also not quite the right terminology, so just so we understand each other, it's `(GameInfo | undefined)[]` (so an array whose members are either GameInfo or undefined)). Just getting `undefined` itself isn't possible as far as I know, since `.map(...)` always returns an array

> A removed game should not be in the library on restart of heroic

Yes, that's... what I was saying. It's not in `libraries` (since those are fetched dynamically from the games in the users account(s)), but it *is* going to be in `recentGames` (since that's just a config file we write to when starting a game). And this function goes over each game in `recentGames` and tries to find it in `libraries`, while `libraries` isn't guaranteed to have this game anymore. That's how you might get `undefined`, and that's why you need the `.filter(Boolean)` after

Nocccer: I don't see the problem actually. What we do is filtering out recent games which are `undefined`, but we do that before we searching the library. The old approach was incorrect because we called `game.appName` which is not possible on `undefined`. Fitlering out all `undefined` recent games and then find all matching games will return a empty `GameInfo[]` array or a `GameInfo[]` array with the found games.

CommandMC: Hm, there's something I don't quite understand there. So accessing `game.appName` won't work when `game` is `undefined`, alright, but how is it then fine to return `undefined` as a `GameInfo` object at the end here?  
Can I then do `gameInfoObj.app_name` for example, even if it's `undefined`? And if I can, why couldn't I do `game.appName` earlier if `game` is `undefined`? Isn't that the same situation there?

Nocccer: Ok i finally understand it and found the solution. We should not use map, we should use filter. So it will always be a `GameInfo[]`and not a `(GameInfo[] | undefined)[]`

CommandMC: If you'd just use `.filter(...)`, wouldn't you return part of `recentGames`, so a `RecentGame[]`, and not a `GameInfo[]`? I think the `.map(...)` is still necessary, you'd just do it after `.filter()`ing

Nocccer: Check back if it makes sense now. The mapping was totally wrong. We need to filter over library and not map over recentGames.

CommandMC (approved):

src/helpers/library.ts
Nocccer: Should be faster or same speed. Before we iterate over recent and for each recent game over libraries. Now we just iterate once over libraries.

CommandMC: Right, I guess before the speed was dependent on where exactly the recent games were in the library array
Anyways, if it helps with readability (and it certainly does), it's fine by me even if it would be a little slower

PR #1761, opened 2022-08-31
[Tech] Let's be even more strict when writing TS code

CommandMC:
I've changed some of our TS settings so we're forced to account for more situations that were easy to miss before. I think this will ultimately lead to a more stable experience for our users, at the cost of having to do *a bit* more work when writing new code

Changes in tsconfig.json:
- Ordered options based on where they appear in TS' docs
- Added the following options:
   - `allowUnreachableCode`: `false`
   - `allowUnusedLabels`: `false`
   - `noUncheckedIndexedAccess`: `true`
- Turned off the following options:
   - `suppressImplicitAnyIndexErrors`
- Options that were removed but whose values were set to their default values:
  (Just to clarify, these changes only cleaned up our config file, no functional changes were made)
   - `noImplicitThis`: Already `true` since `strict` is enabled
   - `noImplicitAny`: Same as above
   - `strictNullChecks`: Same as above

---


src/backend/utils.ts
```diff
@@ -365,7 +365,8 @@ async function errorHandler(
   const otherErrorMessages = ['No saved credentials', 'No credentials']
 
   if (!window) {
-    window = BrowserWindow.getFocusedWindow() ?? BrowserWindow.getAllWindows[0]
+    window =
+      BrowserWindow.getFocusedWindow() ?? BrowserWindow.getAllWindows()[0]
```
CommandMC, line 369: As an example on why these changes are necessary, take a look at this diff. Before, we were trying to access the first element of `BrowserWindow.getAllWindows`, which would of course never have worked since it's a function and not an array, but we weren't warned about that since `suppressImplicitAnyIndexErrors` was on

src/backend/gog/games.ts
```diff
@@ -159,7 +159,14 @@ class GOGGame extends Game {
     const etaMatch = data.match(/ETA: (\d\d:\d\d:\d\d)/m)
     const bytesMatch = data.match(/Downloaded: (\S+) MiB/m)
     const progressMatch = data.match(/Progress: (\d+\.\d+) /m)
-    if (etaMatch && bytesMatch && progressMatch) {
+    if (
+      etaMatch &&
```
Nocccer, line 163: If the array gets longer you could do:
```typescript
!etaMatch.some((entry) => entry === null))
```

src/backend/utils.ts
```diff
@@ -95,33 +95,33 @@ function semverGt(target: string, base: string) {
...
   const [tmajor, tminor, tpatch] = target.split('.').map(Number)
 
   let isGE = false
   // A pretty nice piece of logic if you ask me. :P
-  isGE ||= tmajor > bmajor
-  isGE ||= tmajor === bmajor && tminor > bminor
-  isGE ||= tmajor === bmajor && tminor === bminor && tpatch > bpatch
+  isGE ||= tmajor! > bmajor!
+  isGE ||= tmajor === bmajor && tminor! > bminor!
+  isGE ||= tmajor === bmajor && tminor === bminor && tpatch! > bpatch!
```
Nocccer, line 124: I am not sure that this can not be undefined. We actually wrote a test for it where both passed values are undefined. I somehow don't like the `!` operator. With this you trust your code to much. Better are checks and tests.

Nocccer (changes requested):
Looks good overall. 
Still have some questions/suggestions.

src/backend/gog/games.ts
CommandMC: Not sure if that would help here, since I have to check that
1. `etaMatch` itself isn't `null`
2. the group provided in the RegExp (`etaMatch[1]`) matched something
I believe both of those cases can happen independently, at least TS warned me about it

src/backend/utils.ts
CommandMC: I'd argue that it would be the responsibility of the one calling this function to check the inputs first (in this case, making sure you're actually passing SemVer strings). But I agree, a simple check isn't too much to ask

CommandMC:
Arg, got caught by polymorphism here... `0` is of course also `false`. Now what's the most elegant way to check just for `undefined`...

src/backend/gog/games.ts
Nocccer: Should you then not test `etaMatch && etaMatch[0] && etaMatch[1]`. I don't know if typescript arrays work like C arrays where `etaMatch` would be the pointer to `etaMatch[0]`

src/backend/utils.ts
Nocccer: I would not take responsibility to the user. We working in a open source project where everyone have his design choices. I try to catch all possible errors, to avoid any traces and actually that should be our goal, else we never get stable. We got now 3 renderer traces causing empty screen in frontend.

Btw what happens if you still use the operator and write a test with both versions undefined. Does it return false?

Nocccer:
> Arg, got caught by polymorphism here... `0` is of course also `false`. Now what's the most elegant way to check just for `undefined`...

Someone in the internet said checking for null also catches undefined.

CommandMC:
I ended up just using the `semver` library, I think some dependency we have already requires it anyways and it's far easier to just use it rather than implementing everything ourselves

src/backend/gog/games.ts
CommandMC: As long as `null` isn't returned, `etaMatch[0]` will always be the matched text. See [the spec on `RegExp.exec`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#return_value) (called by `String.match` internally)

Nocccer: Ok yes makes sense. This is fine then, aslong as we don't do something like this:
```typescript
if(etaMatch && etaMatch[1] && etaMatch[2] && etaMatch[3] && ...)
```

src/backend/gog/games.ts
```diff
@@ -159,15 +159,8 @@ class GOGGame extends Game {
...
-      etaMatch &&
-      etaMatch[1] &&
-      bytesMatch &&
-      bytesMatch[1] &&
-      progressMatch &&
-      progressMatch[1]
-    ) {
-      const eta = etaMatch[1]
+    if (bytesMatch && bytesMatch[1] && progressMatch && progressMatch[1]) {
+      const eta = etaMatch ? etaMatch[1] : null
```
Nocccer, line 163: If we using only the index one, is it not enough to check for those?

CommandMC: Just checking `bytesMatch[1]` for example?  
No, that would not be enough. `bytesMatch` itself can be null, and trying to access properties of null will throw a TypeError

arielj: We can do `if (bytesMatch?.[1] && progressMatch?.[1])` which looks kinda ugly but it actually works xD

src/frontend/state/GlobalState.tsx
```diff
@@ -92,9 +92,9 @@ export class GlobalState extends PureComponent<Props> {
       []
     for (const igame in games) {
       for (const installedGame of installedGames) {
-        if (installedGame.appName === games[igame].app_name) {
-          games[igame].install = installedGame
-          games[igame].is_installed = true
+        if (installedGame.appName === games[igame]!.app_name) {
```
flavioislima, line 95: should not be an interrogation mark instead of a exclamation one to see if the item is undefined?

src/frontend/screens/WebView/index.tsx
```diff
@@ -32,27 +32,17 @@ export default function WebView() {
...
-    '/loginweb/legendary': epicLoginUrl,
-    '/loginweb/gog': gogLoginUrl
+  const urls: { [index: string]: string } = {
+    '/epicstore': `https://www.epicgames.com/store/${lang}/`,
+    '/gogstore': `https://gog.com`,
+    '/wiki':
+      'https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki',
+    '/loginweb/legendary': 'https://legendary.gl/epiclogin',
+    '/loginweb/gog':
+      'https://auth.gog.com/auth?client_id=46899977096215655&redirect_uri=https%3A%2F%2Fembed.gog.com%2Fon_login_success%3Forigin%3Dclient&response_type=code&layout=galaxy'
```
flavioislima, line 45: Don't you think this made it more unreadable?
Another idea would be to move the list of URLs to a constants file and import them, making the file more cleaner.

src/frontend/screens/Settings/components/Tools/index.tsx
```diff
@@ -70,14 +70,15 @@ export default function Tools() {
...
-      if (ev.dataTransfer.items[0].kind === 'file') {
-        const exe = ev.dataTransfer.items[0].getAsFile()?.path
-        if (exe) {
-          return callTools('runExe', exe)
-        }
+    if (!ev.dataTransfer.items[0]) {
+      return
+    }
+    // Use DataTransferItemList interface to access the file(s)
+    // If dropped items aren't files, reject them
```
flavioislima, line 77: just a reminder here to remove this Drag functionality. I don't think it is pretty useful. Maybe let's just add a new button like the others and select the EXE on click.

src/frontend/screens/Library/components/InstallModal/index.tsx
```diff
@@ -81,7 +81,7 @@ function getInstallLanguage(
       return foundAvailable
     }
   }
-  return availableLanguages[0]
+  return availableLanguages[0]!
```
flavioislima, line 84: What is the name of this feature of adding an exclamation mark at the end? I don't remember using this one before.

src/frontend/screens/Game/GameSubMenu/index.tsx
```diff
@@ -153,7 +153,7 @@ export default function GamesSubmenu({
         defaultPath: defaultInstallPath
       })
       if (path) {
-        await window.api.changeInstallPath([appName, path, runner])
+        await window.api.changeInstallPath(appName, runner, path)
```
flavioislima, line 156: If changing it from an array maybe using an object would be better since we won't need to care about the order of the arguments.

src/backend/utils.ts
```diff
@@ -54,46 +55,10 @@ const { showMessageBox } = dialog
...
-  const [bmajor, bminor, bpatch] = base.split('.').map(Number)
-  const [tmajor, tminor, tpatch] = target.split('.').map(Number)
-
-  let isGE = false
-  // A pretty nice piece of logic if you ask me. :P
-  isGE ||= tmajor > bmajor
-  isGE ||= tmajor === bmajor && tminor > bminor
-  isGE ||= tmajor === bmajor && tminor === bminor && tpatch > bpatch
-  return isGE
+  return semver.gt(target, base)
```
flavioislima, line 61: we need to test if the package covers all use cases. I am worried about when we release a stable from a previous beta.
like:
- 2.2.0-beta.1, then
- 2.2.0

flavioislima (commented):
It is a lot of changes, will test it during the week.
Left a few comments for now.

redromnon (commented):
LGTM :+1:

src/backend/utils.ts
CommandMC: There are tests for this function, testing exactly that. They're all passing

src/frontend/screens/Library/components/InstallModal/index.tsx
CommandMC: It's the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator). It's basically a way of telling Typescript "this thing can't be undefined/null right now", in the rare case that it's not able to detect this on its own

src/frontend/state/GlobalState.tsx
CommandMC: We're looping over all indexes in `games` (see the `for (const igame in games)` above). Thus, it's not possible for `games[igame]` to be undefined here

src/frontend/screens/WebView/index.tsx
CommandMC: I mean, having the URLs as their own variables up top isn't very readable either IMO  
Having a separate file would probably be a good idea, I agree

src/frontend/screens/Library/components/InstallModal/index.tsx
flavioislima: Yes, true. I don't really use it that much because it has some issues imo:
- Adding this operator moves the responsibility to know if the prop can be null/undefined from TS to the developer.
- With that, we might actually cause something to fail because some argument or prop come null
- I like to deal with this like this is using `if` or optional chaining (not the most beautiful way but the safest imo:
`if(prop){
return prop
}`

or 

`if(something && something.value)` or `if(something?.value){`

The thing is that on this refactor there are dozens of cases where we needed this operator so might be a sign that we should maybe review some of our types or add more checks.

flavioislima: but I am not arguing against using it, it is actually pretty useful, we just need to be extra careful :P

CommandMC: I definitely agree that it should be the last thing you use, but there are simply cases where you *have* to use it since an undefined/null simply *can't* happen, and adding checks everywhere for situations that can't happen is pointless IMO

As a simple example, we will always have a BrowserWindow open, but `BrowserWindow.getAllWindows()` returns an array that can technically be empty. In cases like these, where it's *absolutely certain* that the given value isn't undefined/null, it's fine to use. There are just simply cases in a big project like this where the developer *can* know better than TypeScript, and the operator is meant for those situations

src/frontend/screens/Game/GameSubMenu/index.tsx
CommandMC: I think we have a mix of both styles (regular arguments, and a single object containing everything) in the project right now, if we want to unify everything that's fine, but I'd say that belongs into a different PR

src/frontend/screens/WebView/index.tsx
CommandMC: I've ended up not putting them in a separate file, see notes on the commit for details

arielj (approved):
looks good to me!

CommandMC:
Currently waiting on #1887 and #1901 to merge before updating this

Poopooracoocoo:
Wvewvevvwv wWvewvevvwv w

CommandMC:
I doubt I'm going to work on it anytime soon, might as well open up a new PR then

PR #1762, opened 2022-09-01
Finish Login fix based on Nocccer's PR

arielj:
I made a new branch/PR with my changes since I don't know if somebody else was working on the other branch.

This PR starts from the Login fix of this PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1760 and adds the missing pieces:

- get the `preloadEpic.js` path from the backend if we are doing an epic login
- send the exchange token to the backend using ipcRenderer.send
- process login using the `--code` option in Legendary
- notify the frontend of the successful login and update frontend
- removed `external login` method since it's not working anymore (I only commented things out for now)

---

### DIFF
--- a/electron/legendary/user.ts
+++ b/electron/legendary/user.ts
@@ -29,7 +29,7 @@ export class LegendaryUser {
   }
 
   public static async loginWithCode(code: string) {
-    const commandParts = ['auth', '--code', code]
+    const commandParts = ['auth', '--token', code]
 
     logInfo('Logging in with Legendary.', LogPrefix.Legendary)
 
--- a/electron/legendary/user.ts
+++ b/electron/legendary/user.ts
@@ -28,8 +28,8 @@ export class LegendaryUser {
     }
   }
 
-  public static async loginWithToken(token: string) {
-    const commandParts = ['auth', '--code', token]
+  public static async loginWithCode(code: string) {
+    const commandParts = ['auth', '--code', code]
 
     logInfo('Logging in with Legendary.', LogPrefix.Legendary)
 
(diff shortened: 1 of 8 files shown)


derrod:
Note that with https://github.com/derrod/legendary/releases/tag/0.20.28 the command line arguments were changed. For exchange tokens you now want to use `--token` instead of `--code` (blame Epic for inconsistent toke/code usage).

CommandMC:
This can be closed now, right?

arielj:
> This can be closed now, right?

I actually like this approach of the preload script that waits for the page to load better than the current implementation. The current is a bit hacky with the search event and the copy to clipboard part (it also override whatever's in the clipboard without telling the user, I think that's a really bad thing that I didn't know how to solve until this!!)

I would like to have this solution be the one we use, I'll try to revive this and see what happened that it didn't work for some people.

derrod:
I would second that, this is arguably a safer way to handle login as it does not rely on behaviours that aren't used by the launcher itself. This hopefully makes it more stable since Epic might be less likely to introduce breaking changes here.

That said it appears to be still incomplete, `launchexternalurl` should be implemented so OAuth through third party platforms (Google, Steam, etc.) works.

flavioislima:
Yes I also agree. The current solution is too hacky. Let's finish this PR, it's better than the current workaround.

Nocccer:
> @arielj @Nocccer what should we do about this PR?

I don't work on it. Actually we should implement it soon, but i don't know if we should start from scratch.

arielj:
I did try to continue this a few weeks ago but then I had some issue and then some time at work that I didn't have time to do anything for heroic, and then I forgot. I do want to get this done, sounds like the correct way to implement the login.

arielj:
Closing this in favor of https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2435

PR #1774, opened 2022-09-03
[Fix] WinePrefix FAQ link

redromnon:
- Create coloured hyperlink
- Open link in browser

This PR fixes two things -
1. The WinePrefix FAQ under (_Settings->Wine->Wine Prefix InfoBox_) is displayed like a simple text. Thus, users don't know that it is actually a link. 
2. Moreover, the link opens up the web page in the Heroic window itself and there's no way to go back to Heroic.

---


flavioislima (approved):
LGTM

I think that we should have our own wiki page for wine/proton information this one from wine is a bit convoluted. But it is better than nothing :P

redromnon:
> I think that we should have our own wiki page for wine/proton information

That sounds like a nice idea.

PR #1778, opened 2022-09-03
Log frontend errors in log files

arielj:
This PR adds an `error` event listener in the frontend so we can send the error message to the backend to log frontend error messages in the log files.

This should make it easier to debug problems and we won't have to ask users to open the dev tools > go to console steps to find frontend errors.

At least with my tests, things like the images cache miss don't show up so that's a good thing (those are expected for example) but I'm not sure if there are other errors that might not show up that are important, but I tested breaking the app on purpose and the error is handled by the backend.

---

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -468,6 +468,10 @@ ipcMain.on('frontendReady', () => {
   handleProtocol(mainWindow, [openUrlArgument, ...process.argv])
 })
 
+ipcMain.on('frontendError', async (event, error: string) => {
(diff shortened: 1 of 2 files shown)


flavioislima (approved):
Cool. I implemented something similar some time ago but it was specific to the library. Having something like this is better.
Just try to use the same method we already have here:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/electron/main.ts#L890

Instead of creating a new one.
Other than that looks good 👍🏽

PR #1783, opened 2022-09-07
[Tech] Enable sandboxing for ipcRenderer Processes

### LINKED ISSUE #1772: Enable Sandboxing in BrowserWindow
### Problem description

_No response_

### Feature description

We should consider enabling sandboxing in the renderer process in Electron (https://www.electronjs.org/docs/latest/tutorial/sandbox). It's default enabled in electron v20+ (https://www.electronjs.org/blog/electron-20-0#default-changed-renderers-without-nodeintegration-true-are-sandboxed-by-default) and helps prevent "cross-site scripting to content injection to man-in-the-middle attacks on remotely loaded websites, just to name a few". 

Since we're navigating to external websites in the renderer thread (epic and gog store), those sites might be able to utilize those exploits and potentially get file system/nodejs library access. Html5/webgl games running in the electron renderer process might be able to do the same in the future.

To do this, we'll have to use a preload script to expose the main process's interface to the renderer process and refactor any renderer process code that uses node into the main process.

I'll fork and create a branch from beta to implement for review.

### Alternatives

_No response_

### Additional information

_No response_

BrettCleary:
This PR refactors all ipcRenderer calls into preload.ts as mentioned and begins to resolve #1772. 

This allows contextIsolation to be set as true, which limits some of the api's that the renderer process can call and is an important security measure (https://www.electronjs.org/docs/latest/tutorial/context-isolation). 

This refactor should also improve frontend developer productivity by clearly documenting the types and arguments when calling ipcRenderer.send, ipcRenderer.invoke, and ipcRenderer.on (now called via window.api.METHOD and defined in preload.ts). It will also clearly define the ipcRenderer calls/handlers in one place (preload.ts) whereas before these were duplicated across several frontend files in some instances and were pretty difficult to determine the types that could/should be passed.

There is also now the option to share types between frontend and backend ipc calls/handlers. Here is how it could work (documented in src/common/types.ts):

```
// here is a way to type the callback function in ipcMain.on or ipcMain.handle
// does not prevent callbacks with fewer parameters from being passed though
// the microsoft team is very opposed to enabling the above constraint https://github.com/microsoft/TypeScript/issues/17868
// ipcMain.handle('updateGame', async (e, appName, runner) => {}) 
// could be converted to:
// ipcMain.handle('updateGame', typedCallback<WrapApiFunction<typeof updateGame>>() => {})
// this has the benefit of type checking for the arguments typed in the preload api
// but may be overly complex for a small benefit
```

Here are the functions that type the callback from the preload.ts method

```
export function typedCallback<T>(arg: T) {
  return arg
}

export type WrapApiFunction<
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  TFunction extends (...args: any) => any
> = (
  e: Electron.IpcMainInvokeEvent,
  ...args: [...Parameters<TFunction>]
) => ReturnType<TFunction>
```

**I decided to not include this in the current PR as it is quite complex and may only have marginal benefits** since the types are already documented in preload.ts.

Also, the preload.ts file is built separately with es-build now (same build tool used by vite) and not bundled. This is necessary as the preload option for BrowserWindow requires an absolute path to the preload.js file. 

This is an important first step towards enabling sandboxing in the renderer processes. Since it's a pretty big refactor already, allows contextIsolation to be enabled, and has other productivity benefits, I figured it would be better to go ahead and PR now. 

**FURTHER WORK**
The next step towards enabling sandboxing will be to set nodeIntegration: false. The first issue encountered with this was that electron-store methods inside of preload.ts are synchronous but ipcRenderer calls are all async. As it is necessary to refactor these methods into the main process after setting nodeIntegration: false, this will propagate changes to a lot of frontend react components that are expecting these store method calls to be synchronous. This will be another large change, so I think it's best to PR now before beginning experimentation with that.

---


CommandMC (commented):
Thanks for this, was looking into doing something similar for a while now.

Looking into this a bit, it seems like it's also quite easy to add return type definitions to IPC functions now, which is (IMO) also something we need. But that can come gradually as people work on the system, no need to include everything here.

One very small thing, I'm not sure if just `api` is the best name for the folder & property on the `window` object. But then again, I can't really think of a better name right now either :^)

You've described a method of converting the `electron-store` calls so `nodeIntegration` can be turned off. I assume you've refrained from actually implementing that here since the PR's only about `contextIsolation`?

src/frontend/screens/Settings/components/Tools/index.tsx
```diff
@@ -28,20 +27,22 @@ export default function Tools({ appName, runner }: Props) {
...
-    await ipcRenderer.invoke('callTool', {
-      tool,
-      exe,
-      appName,
-      runner
-    })
+    exe
+      ? await window.api.callTool({
+          tool,
+          exe,
```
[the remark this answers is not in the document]
flavioislima: Actually, this change is breaking the WineCFG and Winetricks calls on Linux.
Change the types on `callTool` to have the `exe` as an optional parameter and remove this check here.
The missing `exe` will be treated in the backend and it is only used for the feature to run exe inside the prefix.

src/frontend/helpers/index.ts
```diff
@@ -11,49 +11,45 @@ import { GogInstallInfo } from 'common/types/gog'
...
+const readFile = window.api.readConfig
 
-const writeConfig = async (
-  data: [appName: string, x: unknown]
-): Promise<void> => ipcRenderer.invoke('writeConfig', data)
+const writeConfig = window.api.writeConfig
 
 const notify = ([title, message]: [title: string, message: string]): void =>
-  ipcRenderer.send('Notify', [title, message])
+  window.api.notify([title, message])
```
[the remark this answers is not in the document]
CommandMC: Hm, I think that would hurt the clarity a bit, you'd then think it's just a regular function while it's actually calling the backend. `window.api` isn't too much to write

flavioislima:
Testing here on Linux with my current install and what is not working so far:
- [Linux] Game Settings > Wine > WineCFG and Winetricks won't open
- `process` is coming `undefined` on the Library. I wonder if this was introduced on this PR or on another one, but it is working fine in `main`
- On Install/Uninstall the game page is not updating correctly for GOG games (Epic works fine). Might be that the refresh method is not being called thereafter the action is finished.
- [Linux] On Wine Manager the progress is not working. The wine version installs correctly but no progress while downloading.

flavioislima:
On a clean install the app doesn't open. Somehow it is trying to find Legendary folder but since it doesn't exist it fails:
```
(node:487388) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, scandir '/home/flavio/.config/legendary/metadata'
    at Object.readdirSync (node:fs:1405:3)
    at e.readdirSync (node:electron/js2c/asar_bundle:5:10830)
    at LegendaryLibrary.loadGamesInAccount (/home/flavio/Develop/Electron/HeroicGamesLauncher/build/electron/main.js:1850:16)
    at new LegendaryLibrary (/home/flavio/Develop/Electron/HeroicGamesLauncher/build/electron/main.js:1841:10)
    at LegendaryLibrary.get (/home/flavio/Develop/Electron/HeroicGamesLauncher/build/electron/main.js:1845:41)
    at createWindow (/home/flavio/Develop/Electron/HeroicGamesLauncher/build/electron/main.js:10647:20)
    at /home/flavio/Develop/Electron/HeroicGamesLauncher/build/electron/main.js:10819:11
```

CommandMC:
> On a clean install the app doesn't open. Somehow it is trying to find Legendary folder but since it doesn't exist it fails:

This is most likely my fault and not related to this PR:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/2ba6604f705f67563ac4b49a08c9d4c913bcd0d0/src/backend/legendary/library.ts#L76

This assumes `legendaryMetadata` exists. A quick `if (existsSync(legendaryMetadata))` in front there would solve that issue

package.json
```diff
@@ -162,7 +162,7 @@
     "tslib": "^2.4.0"
   },
   "scripts": {
-    "dev": "vite",
+    "dev": "esbuild src/backend/preload.ts --platform=node --external:electron --bundle --outfile=build/electron/preload.js && vite",
```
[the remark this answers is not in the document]
BrettCleary: > Thanks for this, was looking into doing something similar for a while now.
> 
> Looking into this a bit, it seems like it's also quite easy to add return type definitions to IPC functions now, which is (IMO) also something we need. But that can come gradually as people work on the system, no need to include everything here.
> 
> One very small thing, I'm not sure if just `api` is the best name for the folder & property on the `window` object. But then again, I can't really think of a better name right now either :^)
> 
> ~You've described a method of converting the `electron-store` calls so `nodeIntegration` can be turned off. I assume you've refrained from actually implementing that here since the PR's only about `contextIsolation`?~ Nevermind, should've taken care to read everything first before asking questions

Yes I agree. It'd be nice to have the return types too. Agreed that they can be added gradually moving forward.

I'm open to naming the folder something different. I named it that because I saw some other projects using a similar name, I'm pretty sure it doesn't conflict with window's existing properties, and it's quite simple to remember.


> > Also, the preload.ts file is built separately with es-build now (same build tool used by vite) and not bundled. This is necessary as the preload option for BrowserWindow requires an absolute path to the preload.js file.
> 
> Hm, I've passed the preload script path [like described in the docs](https://github.com/electron-vite/vite-plugin-electron/tree/main/packages/electron#api), and that worked fine (both testing with `vite` and building an AppImage)
> 
> This is what I've added to `vite.config.ts`:
> 
> ```
> preload: {
>   input: {
>     preload: path.resolve(__dirname + '/src/backend/preload.ts')
>   }
> }
> ```

Ah I didn't see this. This is a much cleaner solution. I will update the code. Thank you for this!

src/backend/main.ts
```diff
@@ -223,13 +229,13 @@ async function createWindow(): Promise<BrowserWindow> {
   if (!app.isPackaged) {
     /* eslint-disable @typescript-eslint/ban-ts-comment */
     //@ts-ignore
-    import('electron-devtools-installer').then((devtools) => {
-      const { default: installExtension, REACT_DEVELOPER_TOOLS } = devtools
+    // import('electron-devtools-installer').then((devtools) => {
```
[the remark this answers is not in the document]
BrettCleary: Ah great catch. I was getting the following errors in the console

 ```
"Electron sandbox_bundle.js script failed to run", source: node:electron/js2c/sandbox_bundle (160)
"TypeError: object null is not iterable (cannot read property Symbol(Symbol.iterator))", source: node:electron/js2c/sandbox_bundle (160)
```

which are caused by React DevTools (https://github.com/electron/electron/issues/32133#issuecomment-1107635348). Seems they have a PR in review that fixes the issue, so I guess it will be resolved shortly.

There were several issues preventing the app from starting at that point in time, so in my troubleshooting, I disabled it. It turns out it was not the cause, and the app now runs fine with it enabled. I will uncomment and push another commit. Thank you!

**Comment by BrettCleary** [comment]
> Those are pretty interesting changes and codewise it looks pretty good. Although we need to change the whole app now on all platforms since these changes touch all functionalities.
> 
> can you point out the features and platforms you have already tested? It is also important to test on both a current Heroic installation and a new one, by removing/renaming the config folders for both legendary and heroic.
> 
> I can help test but as more people test the better. :)

Awesome! And noted about the legendary/heroic config folders.

I've just been testing on Windows. I've only run the unit tests and the app in dev mode. I am planning to test the builds and features more comprehensively today on Windows, mac, linux. Appreciate the help testing too!

**Comment by BrettCleary** [comment]
> Thanks for this, was looking into doing something similar for a while now.
> 
> Looking into this a bit, it seems like it's also quite easy to add return type definitions to IPC functions now, which is (IMO) also something we need. But that can come gradually as people work on the system, no need to include everything here.
> 
> One very small thing, I'm not sure if just `api` is the best name for the folder & property on the `window` object. But then again, I can't really think of a better name right now either :^)
> 
> ~You've described a method of converting the `electron-store` calls so `nodeIntegration` can be turned off. I assume you've refrained from actually implementing that here since the PR's only about `contextIsolation`?~ Nevermind, should've taken care to read everything first before asking questions

Yes I agree. It'd be nice to have the return types too. Agreed that they can be added gradually moving forward.

I'm open to naming the folder something different. I named it that because I saw some other projects using a similar name, I'm pretty sure it doesn't conflict with window's existing properties, and it's quite simple to remember.

File: src/frontend/helpers/index.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (BrettCleary) [reply src/frontend/helpers/index.ts:19]: Yeah I agree with @CommandMC. I think developers coming from other electron projects might find it a bit unclear as to whether it's a function or backend call. It also might add a bunch of boilerplate with
`const {loginPage, handleKofi...} = window.api`
and even after that, there's nothing stopping the dev from writing 
`window.api.loginPage(...args)`
which might be confusing as there are now two equivalent ways of calling the api function.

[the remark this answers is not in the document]
Comment (flavioislima) [reply src/frontend/helpers/index.ts:19]: I agree that `window.api` is ok to write. Although `window` at least in my head would never be associated with backend :P
But ok, let's keep as it is.
It is good to decide this kind of stuff on refactors since that will be the pattern we will use from now on.

**Comment by flavioislima** [comment]
> Yes I agree. It'd be nice to have the return types too. Agreed that they can be added gradually moving forward.
> 
> I'm open to naming the folder something different. I named it that because I saw some other projects using a similar name, I'm pretty sure it doesn't conflict with window's existing properties, and it's quite simple to remember.

Imo `api` sounds good as the name for the folder. It is pretty clear since it will be used to call the methods from frontend to backend.

**Comment by flavioislima** [comment]
Login with Epic is not working but that is because the beta branch is not updated with the main branch so we need to updated it to  get the fixes from it.

File: src/frontend/screens/Settings/components/Tools/index.tsx (hunk 1 above)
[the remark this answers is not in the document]
Comment (BrettCleary) [reply src/frontend/screens/Settings/components/Tools/index.tsx:33]: Makes sense. Okay let me know if this is what you had in mind

Inside ipcMain.handle('callTool'
```
case 'runExe':
        if (exe) {
          exe = quoteIfNecessary(exe)
          game.runWineCommand(exe)
        }
        break
```
And remove the `exe ?` and the `quoteIfNecessary` from [src/frontend/screens/Settings/components/Tools/index.tsx].

And then make the exe property options in the Tools type.

**Comment by flavioislima** [comment]
ok, just updated the beta branch, now just update this branch against beta to get the fixes.

**Comment by BrettCleary** [comment]
> > On a clean install the app doesn't open. Somehow it is trying to find Legendary folder but since it doesn't exist it fails:
> 
> This is my fault and not related to this PR:
> 
> https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/2ba6604f705f67563ac4b49a08c9d4c913bcd0d0/src/backend/legendary/library.ts#L76
> 
> This assumes `legendaryMetadata` exists. A quick `if (existsSync(legendaryMetadata))` in front there would solve that issue

Shouldn't it be `mkdirSync(legendaryMetadata, {recursive: true})` or is the directory being created somewhere else?

**Comment by flavioislima** [comment]
> 
> Shouldn't it be `mkdirSync(legendaryMetadata, {recursive: true})` or is the directory being created somewhere else?

This folder is created after login with epic, so legendary creates it.

File: src/common/types.ts [hunk 5]
```diff
@@ -5,6 +5,31 @@ import { VersionInfo } from 'heroic-wine-downloader'
 
 export type Runner = 'legendary' | 'gog'
 
+// here is a way to type the callback function in ipcMain.on or ipcMain.handle
+// does not prevent callbacks with fewer parameters from being passed though
+// the microsoft team is very opposed to enabling the above constraint https://github.com/microsoft/TypeScript/issues/17868
+// for ipcMain.handle('updateGame', async (e, appName, runner) => { for instance could be converted to:
+// ipcMain.handle('updateGame', typedCallback<WrapApiFunction<typeof updateGame>>() => {
+// this has the benefit of type checking for the arguments typed in the preload api
+// but may be overly complex for a small benefit
```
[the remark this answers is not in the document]
BrettCleary: noted. we could add this in gradually like the ipc return types

BrettCleary:
Looks like the windows build check is throwing the same error as the last two commits to the beta branch. 

`FetchError: request to https://artifacts.electronjs.org/headers/dist/v20.0.2/node-v20.0.2-headers.tar.gz failed, reason: connect ETIMEDOUT 13.107.246.69:443`

Maybe the remote server was down during these tests. Doesn't seem to be related to the changes we've made unless I'm missing something. Can we re run the checks? Or should we just ignore for now

flavioislima:
Just ignore the builds. Some days github are a more unstable than others.

BrettCleary:
> Those are pretty interesting changes and codewise it looks pretty good. Although we need to change the whole app now on all platforms since these changes touch all functionalities.
> 
> can you point out the features and platforms you have already tested? It is also important to test on both a current Heroic installation and a new one, by removing/renaming the config folders for both legendary and heroic.
> 
> I can help test but as more people test the better. :)

I've been testing today on Windows and Mac. I'm having a hard time in some cases trying to determine what the expected behavior should be though. I'm also still familiarizing myself with the frontend/repo in general so it's taking me a lot of time to go through every error message as well.

On Windows I received the following error messages:

Can't download from epic cdn (I think this is being discussed in the discord and is unrelated to the current PR)
```
forbidden access to epic store for fortnite download
[Core] INFO: Trying to re-use existing login session...
[cli] INFO: Saving install tags for "Fortnite" to config: ,chunk0,chunk10
[cli] INFO: Preparing download for "Fortnite" (Fortnite)...
Traceback (most recent call last):
  File "legendary\cli.py", line 3054, in <module>
  File "legendary\cli.py", line 2971, in main
  File "legendary\cli.py", line 931, in install_game
  File "legendary\core.py", line 1296, in prepare_download
  File "legendary\core.py", line 1221, in get_cdn_manifest
  File "requests\models.py", line 1021, in raise_for_status
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://download.epicgames.com/Builds/Fortnite/CloudDir/
```

HGL showed a "Can't get game info" error when trying to expand the game in the library while downloading it through Epic Games Launcher.

React devtools is also throwing the following error when checking/unchecking some of the settings components for the first time.
```
Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

The following error is sometimes thrown as well
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at NewLogin (http://localhost:5173/src/frontend/screens/Login/index.tsx:33:7)
```

But I'm able to navigate the Epic games store, install through EGL, and then refresh the library to see my game. The game launches fine through HGL. When I download a game from the GOG store. I don't see it show up in HGL. I'm not sure how to troubleshoot this as it seems there is only a sync option for the EGL.

I've tested on Mac too and I've gotten the following error
```
Error: Error invoking remote method 'checkGameUpdates': Error: ENOENT: no such file or directory, open '/Users/brett/.config/legendary/installed.json'.
```
I'm not sure when legendary is supposed to be installed on MacOS, so I'm having trouble troubleshooting this too.

On Mac I'm able to see games from Epic Games Launcher, but the game that I've downloaded through EGL still shows up as "Not Installed". Clicking into the game gives
```
... legendary info a74 ...
SyntaxError: Unexpected end of JSON input
Cannot get game info
SyntaxError: Unexpected end of JSON input
```
and I get a "Cannot get game info" message on the screen

Nocccer (changes requested):
I tested our beta channel before. Everything worked fine. On your branch i got some errors:
image
image

This happens as soon as i press game settings

https://user-images.githubusercontent.com/61798668/189245446-08e80f29-5955-4aeb-8362-97c6b32134ed.mp4

ProgressDialog should be in frontend/components
image

BrettCleary:
> I tested our beta channel before. Everything worked fine. On your branch i got some errors: 
> 
> This happens as soon as i press game settings
> 
> ProgressDialog should be in frontend/components
@Nocccer 

Thank you for finding this. Looks like two ipcRenderer calls were still left in the frontend. They have been refactored now. Can you try again?

Nocccer:
> > I tested our beta channel before. Everything worked fine. On your branch i got some errors:
> > This happens as soon as i press game settings
> > ProgressDialog should be in frontend/components
> > @Nocccer
> 
> Thank you for finding this. Looks like two ipcRenderer calls were still left in the frontend. They have been refactored now. Can you try again?

Ok will test later. Btw you should move `src/components/UI/ProgressDialog/` to `src/frontend/components/UI/`. Somehow it is missplaced.

arielj:
one question, is this preload thing necessary?

```
import { contextBridge } from 'electron'
import api from './api'

contextBridge.exposeInMainWorld('api', api)
```

Since we have full control over the frontend we should be able to already import the functions without this preload (like we do in the helpers/library.ts), but maybe this has some benefit I don't know.

Related to that, adding the functions to the window object and calling `window.api` feels weird to me. If we have the functions as a .ts file imported in the frontend we can import the functions we need in each file and use them without the `window.ap`.

Nocccer (changes requested):
Gets pretty close to perfect.
I found something i don't know if you introduced it, but i could not reproduce it on beta.

If i uninstall a game it is still shown as installed on the library page till i refresh via the refresh button.

src/frontend/screens/Settings/components/Tools/index.tsx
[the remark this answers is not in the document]
flavioislima: It is working fine now.

src/backend/api/library.ts
```diff
@@ -38,8 +30,16 @@ export const importGame = async (args: ImportGameArgs) =>
...
+    ipcRenderer.removeListener('setGameStatus', callback)
+  }
+}
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// export const setGameStatusRemoveListener = (onGameStatusUpdate: any) => {
+//   ipcRenderer.removeListener('setGameStatus', onGameStatusUpdate)
+// }
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const handleProgressOf = (version: string, callback: any) =>
   ipcRenderer.on('progressOf' + version, callback)
```
Nocccer, line 45: This belongs to the wine-manager, not to the library. Feel free to rename the event to `'progressOfWineManager' + version`

Make sure also to adapt the sender. The callback is actually no callback i think. It is data. See the sender.

https://github.com/BrettCleary/HeroicGamesLauncher/blob/2d2e15b4dfa6c473a1e67e070e6a9e8822f14dbd/src/backend/wine/manager/ipc_handler.ts#L22-L36

src/frontend/index.tsx
```diff
@@ -14,7 +14,7 @@ import { initShortcuts } from './helpers/shortcuts'
 import { configStore } from './helpers/electronStores'
 
 window.addEventListener('error', (ev: ErrorEvent) => {
-  window.api.logError(ev.error.stack)
+  window.api.logError(ev.error)
```
Nocccer, line 17: Prefix is missing in the ipcRenderer call (`LogPrefix.Frontend`)

src/frontend/screens/Settings/components/LogSettings/index.tsx
```diff
@@ -71,7 +71,11 @@ export default function LogSettings({ isDefault, appName }: LogSettingsProps) {
 
   const getLogContent = () => {
     window.api
-      .getLogContent({ isDefault, appName, defaultLast })
+      .getLogContent({
+        isDefault,
+        appName: isDefault ? '' : appName,
+        defaultLast
+      })
```
Nocccer, line 78: The backend changed. There are now two arguments instead of three:
https://github.com/BrettCleary/HeroicGamesLauncher/blob/2d2e15b4dfa6c473a1e67e070e6a9e8822f14dbd/src/backend/logger/logfile.ts#L107-L115

I think the `isDefault` is two much.

src/frontend/screens/Settings/components/LogSettings/index.tsx
```diff
@@ -98,7 +102,11 @@ export default function LogSettings({ isDefault, appName }: LogSettingsProps) {
   }, [isDefault, defaultLast])
 
   function showLogFileInFolder() {
-    window.api.showLogFileInFolder({ isDefault, appName })
+    window.api.showLogFileInFolder({
+      isDefault,
+      appName: isDefault ? '' : appName,
+      defaultLast
+    })
```
Nocccer, line 109: Same here. Two arguments. `isDefault` is to much:
https://github.com/BrettCleary/HeroicGamesLauncher/blob/2d2e15b4dfa6c473a1e67e070e6a9e8822f14dbd/src/backend/logger/ipc_handler.ts#L12-L14

src/backend/gog/library.ts
```diff
@@ -138,16 +138,16 @@ export class GOGLibrary {
...
     }
     this.refreshInstalled()
+    for (const game of libraryStore.get('games', []) as GameInfo[]) {
+      const copyObject = { ...game }
+      if (this.installedGames.has(game.app_name)) {
+        copyObject.install = this.installedGames.get(game.app_name)!
+        copyObject.is_installed = true
+      }
+      this.library.set(game.app_name, copyObject)
+    }
```
Nocccer, line 148: @imLinguin is this useful or is there a better fix?

src/backend/api/library.ts
```diff
@@ -38,8 +30,16 @@ export const importGame = async (args: ImportGameArgs) =>
...
+export const handleSetGameStatus = (callback: any) => {
   ipcRenderer.on('setGameStatus', callback)
+  return () => {
+    ipcRenderer.removeListener('setGameStatus', callback)
+  }
+}
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// export const setGameStatusRemoveListener = (onGameStatusUpdate: any) => {
+//   ipcRenderer.removeListener('setGameStatus', onGameStatusUpdate)
+// }
```
Nocccer, line 42: If this is not needed anymore maybe remove it?

Nocccer (changes requested):
We getting close :)
Left some comments

src/backend/gog/library.ts
flavioislima: This looks like could be written as a filter, no?
The loop with the filter might be slower but I think we would not need to copy the object, modify it and then put it in the library. 🤔

imLinguin: Whole idea of copying is to avoid writing changes into `gog_store/library.json` I wanted it to only store game metadata while `installed.json` contains install informations

src/backend/api/library.ts
BrettCleary: Noted on the naming and folder. On the last point, the sender sends `{state, progress}` to the callback, so the callback arguments for callback passed by the frontend are `(e: IpcRendererEvent, {state, progress})`

src/frontend/screens/Settings/components/LogSettings/index.tsx
BrettCleary: Ah. makes sense. Just changed it

src/backend/api/library.ts
BrettCleary: done

flavioislima:
@BrettCleary the wine downloader still have issues but I found a solution. 
On `WineItem.tsx` you need to add a `useEffect` like this:

```
  useEffect(() => {
    if (version) {
      window.api.handleProgressOfWineManager(
        version,
        (
          e: Electron.IpcRendererEvent,
          progress: {
            state: State
            progress: ProgressInfo
          }
        ) => {
          setProgress(progress)
        }
      )
    }
  }, [])
```

Also change the api call to remove the event:
```
export const handleProgressOfWineManager = (version: string, callback: any) => {
  ipcRenderer.on('progressOfWineManager' + version, callback)
  return () => {
    ipcRenderer.removeListener('progressOfWineManager' + version, callback)
  }
}
```

And finally on wine manager ipc_handlers use the window to send the command:
```
  const window = BrowserWindow.getAllWindows()[0]
  const onProgress = (state: State, progress?: ProgressInfo) => {
    window.webContents.send('progressOfWineManager' + release.version, {
      state,
      progress
    })
  }
```

I think the only mandatory change is the first one but the others are for consistency and best practices.
This way we will be following the same pattern that we are using for the game install progress. :)

Nocccer:
> window.webContents.send('progressOfWineManager' + release.version, {
>       state,
>       progress
>     })

Why `window.webContents.send` and not taking the `event.sender.send` ? Is there any benefit ?

src/backend/api/wine.ts
```diff
@@ -21,6 +21,8 @@ export const installWineVersion = async (release: WineVersionInfo) =>
   ipcRenderer.invoke('installWineVersion', release)
 export const removeWineVersion = async (release: WineVersionInfo) =>
   ipcRenderer.invoke('removeWineVersion', release)
+export const refreshWineVersionInfo = async (fetch?: boolean) =>
+  ipcRenderer.invoke('refreshWineVersionInfo', fetch)
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const handleProgressOfWinetricks = (onProgress: any) =>
```
Nocccer, line 28:
```suggestion
export const handleProgressOfWinetricks = (onProgress: string[]) =>
```

src/backend/api/wine.ts
```diff
@@ -29,3 +31,7 @@ export const handleProgressOfWinetricks = (onProgress: any) =>
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const progressOfWinetricksRemoveListener = (onProgress: any) =>
```
Nocccer, line 32:
```suggestion
export const progressOfWinetricksRemoveListener = (onProgress: string[]) =>
```

src/backend/api/wine.ts
```diff
@@ -29,3 +31,7 @@ export const handleProgressOfWinetricks = (onProgress: any) =>
...
   ipcRenderer.removeListener('progressOfWinetricks', onProgress)
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const handleProgressOfWineManager = (version: string, callback: any) =>
+  ipcRenderer.on('progressOfWineManager' + version, callback)
```
Nocccer, line 37:
```suggestion
export const handleProgressOfWineManager = (version: string, onProgress: { state: State, progress: ProgressInfo}) =>
  ipcRenderer.on('progressOfWineManager' + version, onProgress)
```

Nocccer (changes requested):
Beside the one thing that flavio suggested, i would love to get rid of all 
`// eslint-disable-next-line @typescript-eslint/no-explicit-any. I added some lines where i know the type.

flavioislima:
> > window.webContents.send('progressOfWineManager' + release.version, {
> > state,
> > progress
> > })
> 
> Why `window.webContents.send` and not taking the `event.sender.send` ? Is there any benefit ?

In practice, it doesn't change anything. Was just about consistency because for the game progress we are using one way and for wine progress another way. Thats confusing because they are doing the same thing.

BrettCleary:
> > window.webContents.send('progressOfWineManager' + release.version, {
> > state,
> > progress
> > })
> 
> Why `window.webContents.send` and not taking the `event.sender.send` ? Is there any benefit ?

The difference would be that `event.sender.send` would send the wine download progress back to the window that invoked `installWineVersion` whereas using the `window.webContents.send` approach would send to the first BrowserWindow. Since we're only using one browser window, it's interchangeable.

BrettCleary:
> @BrettCleary the wine downloader still have issues but I found a solution. On `WineItem/index..tsx` you need to add a `useEffect` like this:
> 
> ```
>   useEffect(() => {
>     if (version) {
>       window.api.handleProgressOfWineManager(
>         version,
>         (
>           e: Electron.IpcRendererEvent,
>           progress: {
>             state: State
>             progress: ProgressInfo
>           }
>         ) => {
>           setProgress(progress)
>         }
>       )
>     }
>   }, [])
> ```
> 
> Also change the api call to remove the event:
> 
> ```
> export const handleProgressOfWineManager = (version: string, callback: any) => {
>   ipcRenderer.on('progressOfWineManager' + version, callback)
>   return () => {
>     ipcRenderer.removeListener('progressOfWineManager' + version, callback)
>   }
> }
> ```
> 
> And finally on wine manager ipc_handlers use the window to send the command:
> 
> ```
>   const window = BrowserWindow.getAllWindows()[0]
>   const onProgress = (state: State, progress?: ProgressInfo) => {
>     window.webContents.send('progressOfWineManager' + release.version, {
>       state,
>       progress
>     })
>   }
> ```
> 
> I think the only mandatory change is the first one but the others are for consistency and best practices. This way we will be following the same pattern that we are using for the game install progress. :)

Noted. latest commit addresses this. Also the useEffect will have to return the callback to remove the listener on each render and since TS throws error because not all code paths return a value, I added the ()=>void callback at the end.
```
useEffect(() => {
    if (version) {
      const removeWineManagerDownloadListener =
        window.api.handleProgressOfWineManager(
          version,
          (
            e: Electron.IpcRendererEvent,
            progress: {
              state: State
              progress: ProgressInfo
            }
          ) => {
            setProgress(progress)
          }
        )
      return removeWineManagerDownloadListener
    }
    /* eslint-disable @typescript-eslint/no-empty-function */
    return () => {}
  }, [])
```

src/backend/api/wine.ts
BrettCleary: onProgress is a callback and the api method should return the removeListener call too. I tried using separate api methods for .on and .removeListener, and it was not unsubscribing in the tests that I ran.
```
export const handleProgressOfWinetricks = (
  onProgress: (e: Electron.IpcRendererEvent, messages: string[]) => void
) => {
  ipcRenderer.on('progressOfWinetricks', onProgress)
  return () => {
    ipcRenderer.removeListener('progressOfWinetricks', onProgress)
  }
}
```

src/backend/api/wine.ts
BrettCleary: this should be 
```
export const handleProgressOfWineManager = (
  version: string,
  callback: (
    e: Electron.IpcRendererEvent,
    progress: {
      state: State
      progress: ProgressInfo
    }
  ) => void
) => {
  ipcRenderer.on('progressOfWineManager' + version, callback)
  return () => {
    ipcRenderer.removeListener('progressOfWineManager' + version, callback)
  }
}
```
This is reflected in the latest commit.

BrettCleary:
> Beside the one thing that flavio suggested, i would love to get rid of all `// eslint-disable-next-line @typescript-eslint/no-explicit-any`. I added some lines where i know the type.

Agreed. I would prefer to do this incrementally though especially since before this PR, there were no types in ipcRenderer/main calls

src/frontend/screens/WineManager/components/WineItem/index.tsx
```diff
@@ -33,20 +33,26 @@ const WineItem = ({
...
+              progress: ProgressInfo
+            }
+          ) => {
+            setProgress(progress)
+          }
+        )
+      return removeWineManagerDownloadListener
+    }
+    /* eslint-disable @typescript-eslint/no-empty-function */
+    return () => {}
```
flavioislima, line 54: I am not sure we need that since we already remove the listener on backend. Might be unnecessary although might not cause any harm.

flavioislima (approved):
Looks good now, all bugs I could found were fixed, left a small comment but you should be able to merge whenever you like now :D

Thanks for that by the way, was a pretty good refactor ⚔️

flavioislima:
@Nocccer approve it when you find some time since you asked for changes. ;)

src/frontend/screens/WineManager/components/WineItem/index.tsx
BrettCleary: I'm not seeing where it gets removed on the backend. Also every time it renders or re-renders WineItem it would resubscribe to wine manager progess without the above changes. This would cause excess calls to setProgress in WineItem which might result in excess render calls to WineItem

flavioislima: I mean that we were removing the listener.
OK then 👍

src/frontend/screens/WineManager/components/WineItem/index.tsx
```diff
@@ -33,20 +33,26 @@ const WineItem = ({
...
+  }, [])
```
Nocccer, line 55: Is something like this not possible?
The api need to have a remove function:
```typescript
export const removeProgressOfWineManager = (
  version: string,
  callback: (
    e: Electron.IpcRendererEvent,
    progress: {
      state: State
      progress: ProgressInfo
    }
  ) => void
) => {
    ipcRenderer.removeListener('progressOfWineManager' + version, callback)
}
```

And we just call the remover funcion inside the unmount of useEffect.
```suggestion
  useEffect(() => {
    const onProgress = (e: Electron.IpcRendererEvent, progress: {state: State, progress: ProgressInfo) 
      => {setProgress(progress)}
    if (version) {
        window.api.handleProgressOfWineManager(
          version,
          onProgress
        )
    }
    
    // unmount
    return () => {
        window.api.removeProgressOfWineManager(version, onProgress)
    }
  }, [])
```

Nocccer (approved):
LGTM. I left some comment we need to discuss and maybe fix after the merge.

src/frontend/screens/WineManager/components/WineItem/index.tsx
BrettCleary: I tried this recently and the remove listener wasn't working so on every render it would just subscribe again and then I would get n messages from the back where n is the number of component renders. Here's an issue that describes the same problem and returning the ()=>removeListener callback to the frontend works as a solution https://github.com/reZach/secure-electron-template/issues/43#issuecomment-772303787. This also generally fits the pattern of listening/removing listener in the same context and is easier than a separate call imo

PR #1788, opened 2022-09-08
[Windows - Fix] Replace wmic with powershell command, because it is deprecated.

Nocccer:
- wmic is deprecated in win11 and since win10 21H1. See https://www.bleepingcomputer.com/news/microsoft/microsoft-starts-killing-off-wmic-in-windows-will-thwart-attacks/#:~:text=%22The%20WMIC%20tool%20is%20deprecated,the%20command%2Dline%20management%20tool.
- According to this article it is not good to use wmic and Win32_product:
  https://devblogs.microsoft.com/scripting/use-powershell-to-find-installed-software/

- Replaced wmic with the solution suggest in the article above.
- Added `child.on('error')`, else heroic will close all the time, because of uncaught exception.
- Didn't tested this, because i don't have a windows machine. Will test with people reported the problem.

---

### DIFF
--- a/electron/utils.ts
+++ b/electron/utils.ts
@@ -637,12 +637,16 @@ function detectVCRedist(mainWindow: BrowserWindow) {
   let stderr = ''
 
   // get applications
-  const child = spawn('powershell', [
+  const child = spawn('powershell.exe', [
     'Get-ItemProperty',
     'HKLM:\\Software\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*,',
     'HKLM:\\Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*',
-    '-Name',
-    'DisplayName'
+    '|',
+    'Select-Object',
+    'DisplayName',
+    '|',
+    'Format-Table',
+    '–AutoSize'
   ])
 
   child.stdout.setEncoding('utf-8')
--- a/electron/utils.ts
+++ b/electron/utils.ts
@@ -630,34 +630,48 @@ function detectVCRedist(mainWindow: BrowserWindow) {
     return
   }
 
+  // According to this article avoid using wmic and Win32_Product
+  // https://xkln.net/blog/please-stop-using-win32product-to-find-installed-software-alternatives-inside/
+  // wmic is also deprecated
   const detectedVCRInstallations: string[] = []
   let stderr = ''
-  const child = spawn('wmic', [
-    'path',
-    'Win32_Product',
-    'WHERE',
-    'Name LIKE "Microsoft Visual C++ 2022%"',
-    'GET',
-    'Name'
+
+  // get applications
+  const child = spawn('powershell', [
+    'Get-ItemProperty',
+    'HKLM:\\Software\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*,',
+    'HKLM:\\Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*',
+    '-Name',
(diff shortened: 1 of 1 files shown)


CommandMC (approved):
Tested on a fresh W11 install, works fine
I don't particularly *like* that we're using PowerShell to query the registry here, ideally we'd use something like [`regedit`](https://www.npmjs.com/package/regedit), but it's fine for now

Nocccer:
> Tested on a fresh W11 install, works fine I don't particularly _like_ that we're using PowerShell to query the registry here, ideally we'd use something like [`regedit`](https://www.npmjs.com/package/regedit), but it's fine for now

Can you provide me a example output?

Btw there is this package you might wanna use for registry. Last time i tried i got node-gyp error. Maybe that is fixed now.

https://www.npmjs.com/package/registry-js

CommandMC:
[Example output](https://pastebin.com/Hw5cNgTa); note that the empty lines at the start, in between the entries, as well as at the end are all deliberate

I'll take a look at that package then, thanks. For now though, still feel free to merge

Nocccer:
> [Example output](https://pastebin.com/raw/Hw5cNgTa); note that the empty lines at the start, in between the entries, as well as at the end are all deliberate
> 
> I'll take a look at that package then, thanks. For now though, still feel free to merge

I will do the test in another pr, because on my logger refactor i did alot of mocks which can help here to write easy tests.
Should i still set shell true?

Is this possible:
```typescript
  const child = spawn(
    'Get-ItemProperty',
    [
      'HKLM:\\Software\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*,',
      'HKLM:\\Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\*',
      '|',
      'Select-Object',
      'DisplayName',
      '|',
      'Format-Table',
      '–AutoSize'
    ],
    { shell: true }
  )
```

CommandMC:
Hm, it seems the default shell is CMD, not PowerShell. Could've sworn I ran into PS errors when using `shell: true` before though...
`shell: 'powershell'` works though

CommandMC:
On another note, I now finally know why I was never able to copy-paste that command properly:
image

While you're at it, could you replace that character?

Nocccer:
> Hm, it seems the default shell is CMD, not PowerShell. Could've sworn I ran into PS errors when using `shell: true` before though... `shell: 'powershell'` works though

Ok it seems you can't use powershell on windows. In `spawn` you can just enable shell, which then will use cmd. For `exec` i did not find any infomration if setting shell to `'powershell'` is supported.

CommandMC:
You can set `shell` to either `true`/`false` (to (not) use the default shell for the platform you're on), or you can set it to any program on `%PATH%`/`$PATH`, which will then be treated as a shell
`powershell` is on `%PATH%`, so you can pass that to `shell`

Nocccer:
> You can set `shell` to either `true`/`false` (to (not) use the default shell for the platform you're on), or you can set it to any program on `%PATH%`/`$PATH`, which will then be treated as a shell `powershell` is on `%PATH%`, so you can pass that to `shell`

Yes but only exec supports setting the shell executable. So should i switch to exec ? I mean `spawn('powershell.exe')` should be the same as `exec('', {shell: 'powershell'})`. I don't think they have besides the deault shell better shell detection

CommandMC:
> Yes but only exec supports setting the shell executable

exec, execFile, and spawn all support it. I've literally tried it myself
But you are right, it's functionally equivalent to just doing `spawn('powershell.exe', [...])`

flavioislima (approved):

Issue #1152, opened 2022-03-23
[General UI/UX] Replace Native Dialogs with HTML Dialogs


flavioislima:
### Problem description

Right now Heroic uses the native dialogs to show error messages or to prompt the user for uninstalling, repairing the game, warning in general.
The problem is that those dialogs don't follow Heroic design and they feel like misplaced elements.

### Feature description

Ideally, Heroic should have their own dialogs that work from inside it, just like the Install Dialog. So then it could follow the same design specifications.

Like on this example on Figma:
![Screenshot 2022-03-23 at 11 23 54](https://user-images.githubusercontent.com/26871415/159678917-a2e3d58c-54ce-4727-8d58-91d42f014d92.png)

Link:
https://www.figma.com/file/Zv89aBtLJ8yN4BwFS0AwNb/Heroic-Games?node-id=0%3A1

arielj:
I think this requires a fair amount of changes, I think it shouldn't be labeled as `good first issue`.

I was checking and the current `dialog` we use comes from electron, not react. So, apart from a design change, this will also require to change how the frontend communicates with the backend when triggering those actions in many places.

flavioislima:
Yes, you're right, this is a big complex.
One idea would be to have a PR only to implement the generic dialog and then we could replace the other dialogs in different PRs.

Nocccer:
So this should be not complex anymore. We have now the Dialog component and we can create as first step a Error Dialog and Message Dialog.

flavioislima:
@biliesilva would need some examples to replace the one's bellow:

- `showMessageBox` (Update info)
![image](https://user-images.githubusercontent.com/26871415/192107988-acfb42fb-f821-46e9-8110-39bb5bbcc176.png)
- `showMessageBox` but with a checkbox (uninstall windows game on Linux) 
![image](https://user-images.githubusercontent.com/26871415/192108131-8f3a4543-e7f3-42f3-a3b4-3054a746fa14.png)
- Other example with a bigger message:
![image](https://user-images.githubusercontent.com/26871415/192108157-6d9f7f5d-eec8-4550-a26b-f91f9025b9f7.png)
- One Button
![image](https://user-images.githubusercontent.com/26871415/192108173-50f624f2-ef05-4286-bb50-4670a64a7cb1.png)
- Three button (Stop installation)
[...]

PR #1815, opened 2022-09-14
[UI] Custom Error Dialog

Nocccer:
- Replaced electron `dialog.showErrorBox` with custom error dialog in frontend.
- Adapt `showErrorBoxModalAuto` to send error to custom error dialog

image

---


flavioislima:
@Nocccer 
Does this also cover the dialogs that we talked about on #1152 ?
I mean the other dialogs like the one for uninstall, to skip update, etc?

Nocccer:
> @Nocccer Does this also cover the dialogs that we talked about on #1152 ? I mean the other dialogs like the one for uninstall, to skip update, etc?

For now it only covers the error dialogs. So all other dialogs are still the old once.

flavioislima:
OK. So then lets keep the issue opened so we know that we need to fix the other dialogs as well 👍🏽

flavioislima (approved):
LGTM 👍🏽

PR #1832, opened 2022-09-22
[Feature] Download Manager

Nocccer:
- Added queue system. Pressing install now adds the game to the download manager queue.
- Added new electron-store for download manager queue
- Added new Download Manager screen
  - added ProgressChart for average downlaod speed
- Replaced kill signal with abort controllers
- Sidebar shows which game gets installed and on click navigate to download manager

---


Etaash-mathamsetty:
1. Instead of calling it "Download Manager", I could call it "Downloads" (the way steam does it)
2. I would also put some units behind the numbers of the graph (like MB/s and whatnot)

Nocccer:
> 1. Instead of calling it "Download Manager", I could call it "Downloads" (the way steam does it)
> 2. I would also put some units behind the numbers of the graph (like MB/s and whatnot)

The pr doc is not up to date. I already added more frontend with MiB/s

Etaash-mathamsetty:
Also for the sidebar, the best thing to do (In my opinion) would be to show the game that is currently installing, since showing the ones that are queued would be redundant and unnecessary to show in the sidebar.

flavioislima:
> Also for the sidebar, the best thing to do (In my opinion) would be to only show the game that is currently installing, since showing the ones that are queued would be redundant and unnecessary to show in the sidebar.

we are thinking on removing this and show the graphic instead or maybe dont show anything there and add some other way of seeing the current download. Perhaps using a floating snackbar that when clicked show the list or something like this.
For now, yes, lets show only one game.

Etaash-mathamsetty:
image
is this intentional or where the actual game names supposed to be here?

BrettCleary:
I don't think this was introduced on your branch, but I got this error message in the browser when clicking download initially

```
index.tsx:301 Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'languages' in undefined
    at getInfo (index.tsx:301:11)
```

and on the backend:

```
(21:06:12) ERROR:   [Gog]:              Error with refreshing token, reauth required
(21:06:19) INFO:    [Gog]:              Refreshing access_token
(21:06:19) DEBUG:   [Backend]:          {"free":242982338560,"diskSize":998984978432,"message":"226.29 GiB / 930.38 GiB","validPath":true}
(21:06:19) ERROR:   [Gog]:              Error with refreshing token, reauth required
(21:06:19) ERROR:   [Backend]:          No credentials, cannot get install info
(21:06:19) ERROR:   [Gog]:              Could not get install info for 1692939089, returning empty object. Something is probably gonna go wrong soon
```

So I have to log out and then back into my gog account. Then it works. 

Would be good to send a message to the frontend and prompt the user to re-login automatically for this case.

BrettCleary:
I ran into an issue when following these steps:
1. starting download
2. losing internet connection during download
3. closing out of app and reconnecting internet
4. relaunching app

The download didn't resume and I got the following error message from the backend
```
(13:37:17) INFO:    [Gog]:              Saved user data to config
(13:37:21) INFO:    [Backend]:          Checking for new Heroic Updates
(13:37:21) INFO:    [Frontend]:         Refreshing Library
(13:37:21) INFO:    [Legendary]:        Refreshing library...
(13:37:21) INFO:    [Legendary]:        Refreshing Epic Games...
(13:37:21) INFO:    [Legendary]:        Game list updated, got 4 games & DLCs
(13:37:21) INFO:    [Gog]:              Getting GOG library
(13:37:21) INFO:    [Legendary]:        Installed game list updated
(node:20704) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of undefined (reading 'webContents')
    at installQueueElement (C:\HeroicGamesLauncher\build\electron\main.30919b2d.js:10319:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async initQueue (C:\HeroicGamesLauncher\build\electron\main.30919b2d.js:10391:5)
(node:20704) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
```

I was also able to click download again on the same the game and it queued it up again in the Download Manager. So there were two instances of the same undownloaded game in the download manager. I also couldn't cancel the current download. The native popup would show up but the app was unresponsive to any of the action buttons on that window.

flavioislima:
@Heroic-Games-Launcher/development this one is ready for review

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -0,0 +1,160 @@
...
+#### Public ####
+*/
+
+async function initQueue() {
+  const window = getMainWindow()
+  let element = getFirstQueueElement()
+  queueState = element ? 'running' : 'idle'
+
+  while (element) {
+    // need to type cohert here since this method can return string as well
```
arielj, line 67: should we handle the case when this is a string? if this is a string for some reason the next line will throw an error

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -0,0 +1,160 @@
...
+    const { status } = (await installQueueElement(window, element.params)) as {
```
arielj, line 68: do we need to pass the windows here? I don't know how this will work with the `close heroic to the tray` feature (I think it destroys the main window and creates a new one when re-opened, so this will be pointing to the wrong place)

I think we should get the main window inside the installQueueElement method when it's needed instead

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -0,0 +1,160 @@
...
+      status: 'done' | 'error'
+    }
+
+    if (status === 'done') {
+      addToFinished(element)
+    }
+    removeFromQueue(element.params.appName)
```
arielj, line 75: Haven't tried this case, but what happens if an installation fails? it's removed from the queue with no indication that it failed? maybe we need a `failed` list like the `finished` list to put failed installations so the user can retry them?

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -0,0 +1,160 @@
...
+    })
+    return
+  }
+
+  let elements: DMQueueElement[] = []
+  if (downloadManager.has('finished')) {
+    elements = downloadManager.get('finished') as DMQueueElement[]
+  }
+
+  const elementIndex = elements.findIndex(
```
arielj, line 41: I wonder if this could ever happen, or if this happened, why? I can imagine installing a game, uninstalling it, and installing it again, in that case I think it's ok to show the game twice maybe?

I asking just to simplify this, if this shouldn't happen maybe we don't need to add this complexity and we can simply push elements when done

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -0,0 +1,160 @@
...
+  } else {
+    elements.push(element)
+  }
+
+  downloadManager.set('queue', elements)
+  logInfo([element.params.appName, 'added to download manager queue.'], {
+    prefix: LogPrefix.DownloadManager
+  })
+
+  getMainWindow().webContents.send('changedDMQueueInformation', elements)
```
arielj, line 109: we should check if there's a main window, I think mainWindow is null when heroic is closed to the tray icon

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -0,0 +1,160 @@
...
+      elements.splice(index, 1)
+      downloadManager.delete('queue')
+      downloadManager.set('queue', elements)
+    }
+
+    logInfo([appName, 'removed from download manager.'], {
+      prefix: LogPrefix.DownloadManager
+    })
+
+    getMainWindow().webContents.send('changedDMQueueInformation', elements)
```
arielj, line 133: same concern here about the mainWindow being null if heroic is closed to the tray

src/backend/downloadmanager/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+      title: i18next.t('box.warning.title', 'Warning'),
+      error: i18next.t(
+        'box.warning.epic.install',
+        'Epic Servers are having major outage right now, the game cannot be installed!'
+      )
+    })
+    return { status: 'error' }
+  }
+
+  mainWindow.webContents.send('setGameStatus', {
```
arielj, line 44: I think we should get the mainWindow right here and check if present, check my previous comment on the function that calls this, I'm worried that this will not always reference the right thing after heroic is closed to the tray

src/backend/downloadmanager/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+          prefix: LogPrefix.DownloadManager
+        })
+        notify({ title, body: i18next.t('notify.install.canceled') })
+      } else if (res.status === 'done') {
+        notify({
+          title,
+          body: i18next.t('notify.install.finished')
+        })
+
+        logInfo(['Finished installing of', params.appName], {
```
arielj, line 89:
```suggestion
        logInfo(['Finished installing ', params.appName], {
```

or

```suggestion
        logInfo(['Finished installation of ', params.appName], {
```

src/frontend/screens/DownloadManager/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  const handleClearList = () => {
+    setFinishedElem([])
+    downloadManagerStore.set('finished', [])
+  }
+
+  return (
+    <>
+      <h4 style={{ paddingTop: 'var(--space-md)' }}>
+        {t('download.manager.title', 'Download Manager (Beta)')}
```
arielj, line 70: Not sure if it makes sense translation-wise, but maybe we can split this in 2 strings so when we move this out of beta we don't need to update all the translations?

public/locales/az/gamepage.json
```diff
@@ -94,13 +97,22 @@
             "start": "İndi oyna",
             "stop": "Oynanır (Dayandır)"
         },
+        "queue": {
+            "remove": "Remove from Queue"
```
arielj, line 101: We have this same string in 2 places: `button.queue.remove` and `label.queue.remove`, maybe we can unify them?

public/locales/az/gamepage.json
```diff
@@ -94,13 +97,22 @@
...
             "syncing": "Sinxronizasiya Saxlamalar"
         }
     },
     "launch": {
         "options": "Başlama Seçimləri..."
     },
+    "Open": "",
+    "queue": {
+        "label": {
+            "remove": "Remove from download manager"
```
arielj, line 113: We have this `queue.label.remove` and also `label.queue.remove` with the value `Remove from Queue`, maybe we can unify them?

src/backend/downloadmanager/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  if (!isOnline()) {
+    logWarning(`App offline, skipping install for game '${title}'.`, {
+      prefix: LogPrefix.Backend
+    })
+    return { status: 'error' }
+  }
+
+  const epicOffline = await isEpicServiceOffline()
+  if (epicOffline && runner === 'legendary') {
```
arielj, line 33: maybe move this like this:

```ts
if (runner === 'legendary') {
  const epicOffline = await isEpic....
  if (epicOffile) {
    ...
  }
}
```
so we don't check if the services are online for gog games

src/backend/legendary/user.ts
```diff
@@ -12,32 +16,57 @@ export class LegendaryUser {
...
   public static async logout() {
     const commandParts = ['auth', '--delete']
 
-    const res = await runLegendaryCommand(commandParts, {
-      logMessagePrefix: 'Logging out'
-    })
+    const abortID = 'legendary-logout'
+    const res = await runLegendaryCommand(
+      commandParts,
+      createAbortController(abortID),
```
arielj, line 57: I wonder if we can have the abort controller as an optional argument in the options hash instead of a fixed second argument, I see many places where we call `runLegendaryCommand` and `runGogCommand` that I understand is just adding the abort controller and deleting this after but we won't really abort a login for example, should make things cleaner, like... only pass an abort controller if we want to but not have to do it always

BrettCleary:
I get unknown error when clicking the action icon after downloading league of legends on windows
image

my backend console is 
```
(17:21:34) INFO:    [Legendary]:        Installed game list updated
(17:21:34) INFO:    [DownloadManager]:  Finished installing of 64b0c77d07f644e6a2326a1fd7ab9926
(17:21:34) INFO:    [DownloadManager]:  64b0c77d07f644e6a2326a1fd7ab9926 added to download manager finished.
(17:21:34) INFO:    [DownloadManager]:  64b0c77d07f644e6a2326a1fd7ab9926 removed from download manager.
(17:21:34) INFO:    [Frontend]:         Refreshing Library
(17:21:34) INFO:    [Legendary]:        Refreshing library...
(17:21:34) INFO:    [Legendary]:        Game list updated, got 4 games & DLCs
(17:21:34) INFO:    [Backend]:          operation interrupted
(17:21:43) DEBUG:   [Legendary]:        Using cached install info
(17:21:43) ERROR:   [Frontend]:         TypeError: Cannot read properties of undefined (reading 'name')
[41240:1011/172203.603:ERROR:gpu_init.cc(486)] Passthrough is not supported, GL is disabled, ANGLE is
```

BrettCleary:
> 1. Instead of calling it "Download Manager", I could call it "Downloads" (the way steam does it)
> 2. I would also put some units behind the numbers of the graph (like MB/s and whatnot)

I agree that "Downloads" is better

BrettCleary:
Also I think we should just use MB instead of MiB

I agree with this [Microsoft article](https://devblogs.microsoft.com/oldnewthing/20090611-00/?p=17933)

Etaash-mathamsetty:
this PR, is functioning quite nicely from what I have tested so far, it's just that I would like to see updates showing up in the download manager as well, since they are not currently in there as far as I can tell.

src/backend/downloadmanager/downloadqueue.ts
flavioislima: This makes sense, would be good to have this failed list.

src/backend/downloadmanager/downloadqueue.ts
arielj: I read a bit more about this and looks like `close to try` only hides the window so the reference is not broken, so this shouldn't be an issue.

src/backend/gog/library.ts
```diff
@@ -343,56 +347,76 @@ export class GOGLibrary {
...
 
-    if (
-      !libraryArray[gameObjectIndex]?.gog_save_location &&
-      this.installedGames.get(appName) &&
-      this.installedGames.get(appName)?.platform !== 'linux'
-    ) {
-      gameData.gog_save_location = await this.getSaveSyncLocation(
-        appName,
-        this.installedGames.get(appName)!
+    try {
```
CommandMC, line 374: I know I'm the only one with this opinion, but the solution to incomplete/incorrect type information that causes errors at runtime isn't "Just wrap everything in a `try {} catch`". I know it's not fun, but do research and actually fix the types. If we keep doing this, we might as well switch back to using JS

src/backend/launcher.ts
```diff
@@ -567,34 +568,37 @@ async function callRunner(
...
-      stdout.push(data.toString().trim())
+      stdout.push(data.trim())
     })
 
-    child.stderr.on('data', (data: Buffer) => {
+    child.stderr.setEncoding('utf-8')
+    child.stderr.on('data', (data: string) => {
       if (options?.logFile) {
-        appendFileSync(options.logFile, data.toString())
+        appendFileSync(options.logFile, data)
```
CommandMC, line 594: Being able to call `kill` on the `ChildProcess` isn't the only reason we're passing it here. It's also useful if you have to communicate with it, like I'm doing in #1887 ([exact lines if you're curious](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/cbee1e4c2e58f771315270c8a57e8426403d06db/src/backend/save_sync.ts#L57-L59)). Please don't remove it.

src/backend/legendary/games.ts
```diff
@@ -457,23 +498,29 @@ class LegendaryGame extends Game {
...
+    )
+
+    deleteAbortController(this.appName)
 
     if (res.error) {
       logError(['Failed to uninstall', `${this.appName}:`, res.error], {
         prefix: LogPrefix.Legendary
       })
-    } else {
+    } else if (!res.abort) {
```
CommandMC, line 523: Can `res.abort` ever be true here?

src/backend/legendary/user.ts
CommandMC: I second this. Having to always pass an AbortController is adding a lot of unnecessary code.

I get that for *some actions*, you want to have it so it can be aborted from somewhere else, but it's a little excessive to always provide it.
In this instance (logging out), the process will only live for at most half second, I doubt anything will try to cancel it (and besides, how would that even happen? A "Cancel logging out" button seems very strange to me)

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -90,23 +90,20 @@ export default function GamePage(): JSX.Element | null {
...
+  const isQueued = status === 'queued'
   const isReparing = status === 'repairing'
   const isMoving = status === 'moving'
-  const hasDownloads = Boolean(
-    libraryStatus.filter(
-      (game) => game.status === 'installing' || game.status === 'updating'
-    ).length
-  )
+
+  const backRoute = location?.state?.fromDM ? '/download-manager' : '/'
```
CommandMC, line 97: `location` can't be undefined/null. If `state` can be undefined/null, please mark it as such in the `as { state: ... }` above

CommandMC (dismissed):

src/backend/downloadmanager/utils.ts
flavioislima: Makes sense 👍🏽

src/backend/legendary/user.ts
flavioislima: I agree but I think it is out of the scope of this PR.

src/backend/gog/library.ts
flavioislima: One thing is not related to another.
Types will get most of the things but if a promise fails for unknown reasons, TS won't do anything.
TS won't run at runtime and won't get random issues.
For instance: An expected OS signal, a hang in the system, this type of thing. So having `try/catch `blocks will at least deal with some of these problems.
TS is for the dev not for the client.

src/backend/launcher.ts
flavioislima: I don't get it, what are we removing? The `.toString()`?

src/backend/legendary/games.ts
flavioislima: If aborting the download throws an error this will be true.

src/backend/gog/games.ts
```diff
@@ -421,14 +466,24 @@ class GOGGame extends Game {
...
+          appendFileSync(this.logFileLocation, output)
+        }
       }
-    })
+    )
+
+    deleteAbortController(this.appName)
+
+    if (abort) {
+      return true
```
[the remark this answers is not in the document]
flavioislima: Depends, as I said above about the downloads. Heroic launched the game just fine, if the user clicked on STOP for some reason, it was aborted due to user interaction/choice, Heroic launched the game just fine so I think yes.

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -0,0 +1,160 @@
...
+  return null
+}
+
+function addToFinished(element: DMQueueElement) {
+  if (!element) {
+    logError('Can not add undefined element to finished!', {
+      prefix: LogPrefix.DownloadManager
+    })
+    return
+  }
```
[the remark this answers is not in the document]
flavioislima: yes makes sense

src/backend/gog/library.ts
CommandMC: Let's keep this in the realm of "reasonably realistic" here. Having a random bit flip that erases a property from an object (I believe that's what you're implying by "hang in the system"), while also changing the extra bits of your ECC memory in the exact way so this doesn't get detected & corrected, is not realistic. And if it would be, who's to say it didn't just erase the try-catch block in the process? If everything's susceptible to this issue, why add protections at all then?
An "OS signal" (not sure what you mean there) that modifies your local JS variables does not exist. The OS can kill you, can pause you, and that's about it

What's far more realistic is:
- we're not taking care to check if a value satisfies all aspects of a type before using `as T`
- some of our types have properties that can be optional but aren't marked as such

Some parts of the code in this try-catch block can fail, yes. The system could run out of memory and GOGDL could fail to spawn, which would leave us with no `res.stdout`, and in turn `JSON.parse` throwing an error (on an unrelated note, if you *do* want to account for that here, there are a lot of cases elsewhere in the code where that's not being done)
But *at least* wrap just *that* in a try-catch then. When leaving it like this, it's impossible for someone who didn't run into the issue themselves to even guess what the problem was

Yes, TS doesn't "run" at runtime, because it doesn't "run" at all. It's a pre-processor, it prevents you from shooting yourself into the foot, but it still has ways (as described above) to disable this protection

src/backend/gog/games.ts
```diff
@@ -681,6 +746,12 @@ class GOGGame extends Game {
       status: 'done'
     })
 
+    deleteAbortController(this.appName)
+
+    if (res.abort) {
+      return { status: 'done' }
```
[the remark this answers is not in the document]
CommandMC: Why not add a possible 3rd `abort` return type?

src/backend/legendary/user.ts
Nocccer: I like to have everywhere a abortcontroller, because if someone closes heroic during login and legendary or gogdl stuck, they will not be closed. Because of the abortcontroller on closing all running child_processes will be aborted.

src/backend/gog/games.ts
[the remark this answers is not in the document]
Nocccer: We want to log a abort as abort and not as error. The problem was that we never handled abort correctly. We always showed alot of error messages, like game failed to install and alot more. We don't want to log that because aborts via abortcontroller are intenional.

src/backend/legendary/games.ts
Nocccer: Aborting via abortController always calls the catch block. We really need to handle aborts. We never did before, if someone aborted the logs where full of errors we should not show.

src/backend/legendary/user.ts
arielj: Maybe what we can do is make this argument optional and, if the option is not passed, the `runXYZCommand`  methods can create an abort controller internally.

That way we can have an explicit abort controller when needed if we wan to be able to abort on demand, and fallback to an internally-managed abortController to handle that case of something that could get stuck without having to always pass it.

Also, would it make sense to automatically remove the abort controller right before returning from those methods? so we don't have to do the deleteAbortController everywhere, if the methods get to the last line to return a value it means the abort controller can be deleted so we could hide that code inside the `runXYZCommand` methods.

src/backend/gog/library.ts
arielj: I agree that we need to be clear about what we are trying to catch with a try/catch block, wrapping too many lines in a try catch is confusing and if we know what can throw an exception we should only wrap that and add a comment about what was the error (then with that information we can try to prevent that in the future).

In languages with Exception and sub-classes for Exception this would mean the common best practice of don't catch `Exception`, catch the specific things that can be thrown. In the js world we don't have that but we could add comments about what we are trying to catch.

src/backend/legendary/games.ts
CommandMC: Well, my question was more "Is it possible to abort this specific command". Uninstalling a game should take < 1 second and we're not exposing any buttons/similar in the frontend to cancel the process

flavioislima (approved):
Tested as much as I could.
Some thing that will need to be improved for next iteration:
Add support for updates
Add support for wine manager
Better UI

src/backend/downloadmanager/downloadqueue.ts
flavioislima: just added the list as well

src/backend/downloadmanager/downloadqueue.ts
flavioislima: Ok, changed the method to always return the status now, so this wont be the case anymre.

PR #1862, opened 2022-10-03
[Fix] Game status not updating on uninstall

flavioislima:
The issue was due to the dependencies inside `useEffect` on the Game Page.

---

### DIFF
--- a/src/frontend/components/UI/CachedImage/index.tsx
+++ b/src/frontend/components/UI/CachedImage/index.tsx
@@ -27,7 +27,7 @@ const CachedImage = (props: Props) => {
   let src = useFallback ? props.fallback : props.src
   src = useCache ? `imagecache://${src}` : src
 
-  return <img {...props} src={src} onError={onError} />
+  return <img loading="lazy" {...props} src={src} onError={onError} />
 }
 
 export default CachedImage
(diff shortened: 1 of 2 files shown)


src/frontend/components/UI/CachedImage/index.tsx
```diff
@@ -27,7 +27,7 @@ const CachedImage = (props: Props) => {
   let src = useFallback ? props.fallback : props.src
   src = useCache ? `imagecache://${src}` : src
 
-  return <img {...props} src={src} onError={onError} />
+  return <img loading="lazy" {...props} src={src} onError={onError} />
```
flavioislima, line 30: Not related with this PR but adding here since it is just one line of code.

Nocccer (approved):

arielj:
hey @flavioislima can you explain the issue this is fixing or steps to reproduce it? I can't reproduce the problem in the main branch if I understand it correctly, when I uninstall a game it's updating the GamePage with no issues

flavioislima:
> hey @flavioislima can you explain the issue this is fixing or steps to reproduce it? I can't reproduce the problem in the main branch if I understand it correctly, when I uninstall a game it's updating the GamePage with no issues

Sometimes it updates and sometimes not.
I added some examples in the development channel in discord.
If we use only Gog and epic in the array of dependencies it trigger up to 5 re-renders and this makes the value not update sometimes.
Using epic.library and gog.library makes it re-render only once and I could not reproduce this error anymore.

arielj (approved):
Looks good, I can't reproduce the original issue but the code changes look fine to me.

Issue #1506, opened 2022-06-21
Ability to add external games not related to EGS or GOG


Laar3:
### Problem description

This day and age its annoying to have to have 5 different launchers for everything.

### Feature description

It would be great if we could add a game, give a title, provide a run command, and supply an image and it would be added to the library.

Nocccer:
This is a idea for long now but takes time to implement.
If you use steam you can try out this feature soon:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1331

In steam you can add external executables.

CommandMC:
Resolved with #1864

PR #1864, opened 2022-10-04
[Feat - General] Sideload Apps 

flavioislima:
This adds the possibility of adding a game outside Epic/GOG. That can use the same configuration system we have and support the same features as the previous.
It can be a Game or any type of software.

OBS.: this also serves as POC for a functional-based approach instead of OOP like we have for Epic and GOG. This decision leads to some code duplication and could be improved if we decided to use functional approach on the other runners as well in the future.

Implements #1506 

### Checkliist

- [x] Add backend methods to add a game to a sideload library similar to current ones
- [x] Add an Install dialog that can be used to add a new App/Game
- [x] Handle GamePage and settings for sideloaded app
- [x] Add a way to edit the game
- [x] Handle Uninstall / Add shortcut / Add to Steam for sideloaded apps
- [x] Run Windows games on Linux using Wine/Proton
- [x] Run Native Linux binaries
- [x] Run Native windows Games
- [x] Run native macOS games
- [x] Add new filter for sideloaded games
- [x] Run windows games on macOS using crossover
- [x] App should have a log
- [x] App should show with correct name on Discord Rich Presense
- [x] Changed dev command from `yarn dev` to `yarn start`
- [x] Changed defaults for Install DXVK and VKD3D to be true.

---

### Screenshots:
image
image
image


BrettCleary:
I'm getting `sideloadedLibrary is not iterable` error on first launch on Windows. `sideloadedLibrary` is the empty array [] until refresh after which it is undefined.

I changed the following line in `src/frontend/state/GlobalState.tsx`
`const updatedSideload = sideloadLibrary.get('games')`

to
`const updatedSideload = sideloadLibrary.get('games', []) as SideloadCard[]`

to fix.

BrettCleary:
image

When clicking the game card, it tries to go to a game page and fails with
```
{error: TypeError: Cannot read properties of undefined (reading 'platform')
    at updateConfig
```

Looks like it's a Link to `href="#/gamepage/sideload/sideload"` Does this exist?

I'm not sure if it's intuitive for the user to have to click the download button to add a game too.

Also after reading through the code a bit more, I found that the sideload card is from this empty card in the Library screen:
`return [emptyCard, ...library]`

which is a little bit confusing and I think it's because you would expect that gamesList would contain a list of games and not a list of games + a way to add new games.

I also saw that there were a good number of if checks to get the desired sideload behavior.

What if we created a separate React component that exposed the sideload game functionality? We could put it in the same spot before the gamelist (or somewhere else) but then we could keep the sideload and gamecard/gamelist code separate and there would be a clearly defined element for sideloading a game.

flavioislima:
@BrettCleary I added this card just to have a way of testing the functionality, it is not finished yet. Will be a card like but when clicked it will open the dialog to add the game.
Something like this:


So for now we have that card as a placeholder.

About the approach you mentioned, you mean the logic we have today on backend move it to frontend? 

Some things would be fine like Add, remove the app, can work doing only in FE. But I'm not sure if launching Windows games on Linux and mac could work since we need to get all other configs like environmental variables, wrappers, etc. 

We can investigate what we can do on each side to avoid code duplication and a lot of if checks.

BrettCleary:
> About the approach you mentioned, you mean the logic we have today on backend move it to frontend?
> 
> Some things would be fine like Add, remove the app, can work doing only in FE. But I'm not sure if launching Windows games on Linux and mac could work since we need to get all other configs like environmental variables, wrappers, etc.
> 
> We can investigate what we can do on each side to avoid code duplication and a lot of if checks.

Oh. no. I was just referring to the frontend elements. Moving the sideload code out of the game card/game list components and into its own component. That way we don't add complexity to the existing game card React component.

I might have spoken too soon about it reducing the number of if checks too.

flavioislima:
> > About the approach you mentioned, you mean the logic we have today on backend move it to frontend?
> > 
> > Some things would be fine like Add, remove the app, can work doing only in FE. But I'm not sure if launching Windows games on Linux and mac could work since we need to get all other configs like environmental variables, wrappers, etc.
> > 
> > We can investigate what we can do on each side to avoid code duplication and a lot of if checks.
> 
> Oh. no. I was just referring to the frontend elements. Moving the sideload code out of the game card/game list components and into its own component. That way we don't add complexity to the existing game card React component.
> 
> I might have spoken too soon about it reducing the number of if checks too.

Ah OK, that makes sense.

BrettCleary:
Can you merge the latest changes from the beta branch @flavioislima? I need the new design system to style the "Add Game" button.

BrettCleary:
Just replaced the empty game card with an "Add Game" button. This is the updated look: 

image

I can update the styling to use the new frontend design system when this branch gets the changes from beta or I can do it in a separate branch after merging.

flavioislima:
> FYI: The button doesn't seem to follow the heroic theme atm, otherwise this looks like a great addition image

fixed.

src/backend/sideload/games.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+export async function getAppSettings(appName: string): Promise<GameSettings> {
+  return (
+    GameConfig.get(appName).config ||
+    (await GameConfig.get(appName).getSettings())
+  )
+}
+
+export function addNewApp({
```
flavioislima, line 41: I wonder if this one along with the `removeApp` could be done on the frontend instead of backend. Would be less code to achieve the same.

JakobDev:
Just a Question: Each Game is in it's own prefix and you can set only one executable for each game/prefix. How do deal with games that has more than one executable? I have some (older) games that have a Executable for the Base Game and one for each AddOn. Changing the Executable pa5th each time I want to switching between the Base game and o9ne of the AddOns would be annoying.

flavioislima:
> Just a Question: Each Game is in it's own prefix and you can set only one executable for each game/prefix. How do deal with games that has more than one executable? I have some (older) games that have a Executable for the Base Game and one for each AddOn. Changing the Executable pa5th each time I want to switching between the Base game and o9ne of the AddOns would be annoying.

You can have several games using the same prefix just like for other games right now.

But about having several exe for the same games, how can we solve that?
Adding more selects for other executables and then showing a list on the game page?
I think this is problematic and also would fix just some edge cases.
Most of the games has only one executable.

BrettCleary:
> > Just a Question: Each Game is in it's own prefix and you can set only one executable for each game/prefix. How do deal with games that has more than one executable? I have some (older) games that have a Executable for the Base Game and one for each AddOn. Changing the Executable pa5th each time I want to switching between the Base game and o9ne of the AddOns would be annoying.
> 
> You can have several games using the same prefix just like for other games right now.
> 
> But about having several exe for the same games, how can we solve that? Adding more selects for other executables and then showing a list on the game page? I think this is problematic and also would fix just some edge cases. Most of the games has only one executable.

Couldn't the user just add a separate sideloaded game for each executable? Then you just click on whichever game card corresponds to the base game or addon that you want to launch.

Seems like a cleaner solution to me than an executable drop down and doesn't add any additional code.

Nocccer (changes requested):
Tested with Harry Potter Chamber of secrets.
- Winetricks works
- Game starts
- Winecfg fails with: 
```
(15:59:25) ERROR:   [Gog]:              Could not get game info for nMm5V3XJbktbJoeqQ12vgb, returning empty object. Something is probably gonna go wrong soon
(node:16221) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of undefined (reading 'platform')
    at _GOGGame.isNative (/tmp/.mount_Heroic6uqxjB/resources/app.asar/build/electron/main.3c1baef5.js:27:4306)
    at _GOGGame.runWineCommand (/tmp/.mount_Heroic6uqxjB/resources/app.asar/build/electron/main.3c1baef5.js:31:2893)
    at /tmp/.mount_Heroic6uqxjB/resources/app.asar/build/electron/main.3c1baef5.js:100:3492
    at async node:electron/js2c/browser_init:189:563
(node:16221) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 5)
```
- The Prefix contains spaces if the title contains them aswell, maybe remove all spaces in the prefix by default, to avoid space problems.

redromnon:
Isn't this feature also supposed to support scripts? I'm not able to select a script I created via `Select Executable`

redromnon (commented):
Isn't this feature also supposed to support scripts? I'm not able to select a script I created via `Select Executable`

flavioislima:
> Isn't this feature also supposed to support scripts? I'm not able to select a script I created via `Select Executable`

It is not the main goal and it is not a blocker but should work in theory, yes. Because we just call the binary/script/executable from the shell so I don't know why it didn't.
I tested only appimages on Linux.
Can you check if something happens on the logs?

flavioislima:
> Tested with Harry Potter Chamber of secrets.
> - Winetricks works
> - Game starts
> - Winecfg fails with: 
> ```
> (15:59:25) ERROR:   [Gog]:              Could not get game info for nMm5V3XJbktbJoeqQ12vgb, returning empty object. Something is probably gonna go wrong soon
> (node:16221) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of undefined (reading 'platform')
>     at _GOGGame.isNative (/tmp/.mount_Heroic6uqxjB/resources/app.asar/build/electron/main.3c1baef5.js:27:4306)
>     at _GOGGame.runWineCommand (/tmp/.mount_Heroic6uqxjB/resources/app.asar/build/electron/main.3c1baef5.js:31:2893)
>     at /tmp/.mount_Heroic6uqxjB/resources/app.asar/build/electron/main.3c1baef5.js:100:3492
>     at async node:electron/js2c/browser_init:189:563
> (node:16221) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 5)
> ```
> - The Prefix contains spaces if the title contains them aswell, maybe remove all spaces in the prefix by default, to avoid space problems.

Hmm, OK will check that 👍

public/locales/az/translation.json
```diff
@@ -526,6 +530,7 @@
         "collapse": "Yan paneli yığcamlaşdırın",
         "uncollapse": "Kenar çubuğunu yığışdırın"
     },
+    "Sideload": "Other",
```
arielj, line 533: This as a key might be weird for translators, like... why is this different than the other `Others` added in this file?

src/backend/config.ts
```diff
@@ -475,8 +475,8 @@ class GlobalConfigV0 extends GlobalConfig {
       enableUpdates: false,
       addDesktopShortcuts: false,
       addStartMenuShortcuts: false,
-      autoInstallDxvk: false,
-      autoInstallVkd3d: false,
+      autoInstallDxvk: true,
+      autoInstallVkd3d: true,
```
arielj, line 479: Just to confirm, this is going to change the default value for these options, I'm ok with this default but I'm checking if this is intended in this PR

src/backend/launcher.ts
```diff
@@ -447,23 +455,40 @@ function launchCleanup(rpcClient?: RpcClient) {
...
+  const settings = gameSettings
+    ? gameSettings
+    : await GlobalConfig.get().getSettings()
+  const { wineVersion, winePrefix } = settings
+
+  if (!existsSync(winePrefix)) {
+    mkdirSync(winePrefix, { recursive: true })
+  }
+
+  if (['exe', 'msi', 'bat'].includes(command.split('.')[1])) {
```
arielj, line 475: should we do this always just in case? no idea which file extensions we could have here but I imagine it's safe to quote any command?

src/backend/shortcuts/ipc_handler.ts
```diff
@@ -63,8 +87,16 @@ ipcMain.handle(
...
-    const gameInfo = await game.getGameInfo()
+    let gameInfo: GameInfo
+
+    const isSideload = runner === 'sideload'
+    if (isSideload) {
+      gameInfo = getAppInfo(appName)
+    } else {
+      const game = getGame(appName, runner)
+      gameInfo = await game.getGameInfo()
+    }
```
arielj, line 98: We have this pattern of checking if the runner is sideload or not and then get the gameInfo or the AppInfo in a lot of places (we also do that to get settings for example too). Maybe it's cleaner have a `getInfo(appName, runner)` and inside that method handle the ifSideload check? to avoid repetition

src/backend/sideload/games.ts
arielj: Maybe it's better to leave that here in the backend? for consistency with the other stores, I think it can be confusing to handle similar things some in the backend and some in the frontend

src/backend/utils.ts
```diff
@@ -278,6 +279,7 @@ export const getSystemInfo = async () => {
 
   return `Heroic Version: ${heroicVersion}
 Legendary Version: ${legendaryVersion}
+GogDL Version: ${gogdlVersion}
```
arielj, line 282: Should this be `GOGdl`? cc @imLinguin to confirm, I think I remember some comment about this on discord

public/locales/az/gamepage.json
```diff
@@ -39,11 +48,19 @@
         "add_to_favourites": "Sevimlilərə əlavə edin",
         "cancel": "Pauza/Ləğv et",
         "continue": "Yükləməyə davam edin",
+        "finish": "Finish",
         "force_update": "Mövcuddursa Yeniləməni məcbur edin",
+        "force-innstall": "Force Install",
         "hide_game": "Oyunu Gizlət",
         "import": "İdxal Oyunu",
         "install": "Yüklemek",
+        "install-error": "No path selected",
```
arielj, line 57: maybe use `no-path-selected` as a key here?

public/locales/az/gamepage.json
```diff
@@ -17,6 +17,15 @@
             "message": "Bu oyunu təmir etməyə cəhd etmək istəyirsiniz? Bu uzun müddət çəkə bilər.",
             "title": "Doğrulayın və Təmir edin"
         },
+        "runexe": {
+            "title": ""
```
arielj, line 21: this empty string might create issues, I imagine it's missing a default value somewhere

BrettCleary:
> > FYI: The button doesn't seem to follow the heroic theme atm, otherwise this looks like a great addition image
> 
> fixed.

The button background color still seems out of place when I just tested. I think we should use --accent.
`background-color: var(--accent);`

--primary:
image

--accent:
image

BrettCleary:
I had a couple issues sideloading and launching valorant on windows

1. It's necessary to add the Valorant shortcut as the exe. not the Valorant.exe file. it will throw an access denied error trying to launch this directly. Riot did this to ensure it's launched through their launcher
2. I get this error when launching
`'C:\Riot' is not recognized as an internal or external command,`
It's fixed by putting the entire filepath in double quotes
3. On the cmd line, this will only open the launcher. To open the game directly some flags need to be added as well (specified by the `Target` property on the shortcut). Thus the executable that works is 
`"C:\Riot Games\Riot Client\RiotClientServices.exe" --launch-product=valorant --launch-patchline=live`

Your code gets the RiotClientServices.exe from the shortcut correctly. Is there a way to get and pass the flags too?

Nocccer:
> I had a couple issues sideloading and launching valorant on windows
> 
> 1. It's necessary to add the Valorant shortcut as the exe. not the Valorant.exe file. it will throw an access denied error trying to launch this directly. Riot did this to ensure it's launched through their launcher
> 2. I get this error when launching
> `'C:\Riot' is not recognized as an internal or external command,`
> It's fixed by putting the entire filepath in double quotes
> 3. On the cmd line, this will only open the launcher. To open the game directly some flags need to be added as well (specified by the `Target` property on the shortcut). Thus the executable that works is 
> `"C:\Riot Games\Riot Client\RiotClientServices.exe" --launch-product=valorant --launch-patchline=live`
> 
> Your code gets the RiotClientServices.exe from the shortcut correctly. Is there a way to get and pass the flags too?

You can try launch args under game settings

arielj:
> I had a couple issues sideloading and launching valorant on windows
> 
> 1. It's necessary to add the Valorant shortcut as the exe. not the Valorant.exe file. it will throw an access denied error trying to launch this directly. Riot did this to ensure it's launched through their launcher
> 2. I get this error when launching
>    `'C:\Riot' is not recognized as an internal or external command,`
>    It's fixed by putting the entire filepath in double quotes
> 3. On the cmd line, this will only open the launcher. To open the game directly some flags need to be added as well (specified by the `Target` property on the shortcut). Thus the executable that works is
>    `"C:\Riot Games\Riot Client\RiotClientServices.exe" --launch-product=valorant --launch-patchline=live`
> 
> Your code gets the RiotClientServices.exe from the shortcut correctly. Is there a way to get and pass the flags too?

I don't think that's an issue of this feature though, it's going to happen for any game that requires any special launch method, we can't do anything about that, it will be up to the user to set the correct file and command. Or you mean something is missing in the feature to set the things you need to set?

redromnon:
> It is not the main goal and it is not a blocker but should work in theory, yes. Because we just call the binary/script/executable from the shell so I don't know why it didn't.
> I tested only appimages on Linux.
> Can you check if something happens on the logs?
@flavioislima I didn't find anything related in the log. 

However, I was able to insert the path to the script by manually typing and it launches fine. As @BrettCleary noted, the filepath should include double quotes to avoid the related error.

redromnon:
There's also a **bug** in the Settings which shows Wine and related settings despite the game/app platform version being selected as "Linux". When I click on `Edit App/Game`, the platform is also displayed as Windows instead of Linux. On the game page however, the platform shows Linux correctly. 

Please see if you can replicate this issue.

src/backend/config.ts
flavioislima: This should have been changed a while back, it is an old discussion. since it is 2 lines of code I will change in this PR instead of a separate one.
I will add a note on the PR description.

src/backend/launcher.ts
flavioislima: Don't need this anymore since I refactored it to use `spawn` now. If we need to deal with spaces so it is better to avoid using `exec` and `execAsync` from now on.

public/locales/az/translation.json
flavioislima: fixed

src/backend/utils.ts
flavioislima: fixed

public/locales/az/gamepage.json
flavioislima: fixed

public/locales/az/gamepage.json
flavioislima: fixed

flavioislima:
@Heroic-Games-Launcher/development Fixed all issues that you guys pointed out here and did some refactors as well.

I have changed a few things so you know:
- `runWineCommand` now uses `spawn` instead of `execAsync` This is good since we don't need to worry about spaces on the executable or arguments.
- It is now also more close to `callRunner` and now accepts wrappers and other arguments just like a normal game, but ofc it is optional and the other commands are working just fine (`winecfg`, `runexe`, `wineboot`, etc). Tested it with default wine, Proton-GE, Proton experimental, and wine-ge.
- due to some inconsistencies I needed to change the `installPlatform` parameter for windows to be lowercase, `windows` instead of `Windows`. I saw that GOG already uses the lowercase and legendary also works just fine like this as well. Same for Linux. Didn't change it for Mac since it is not applicable here. 
- the reason was that in some places we were checking as capital-case and in some other places we were checking as by using `toLowerCase()`.
- In the future when I have time to develop on a Mac I might change it as well so we might only use lowercase values.

src/backend/shortcuts/ipc_handler.ts
flavioislima: That's true, will check if I can refactor this.

flavioislima: Ok, added this handled and it is better now.
Not everywhere is possible and since we are using different approaches, Sideload does not depends on `Game` so it is hard to avoid repetition in some places, like the Shortcuts handlers, for instance.
Let's see if we can refactor the others runners in the future so we can cleanup this code.

BrettCleary:
I was getting 
image
when clicking the game card to navigate to the game page for a sideloaded game. 

I added null checks for the wineVersion and winePrefix to fix this on Windows.

I also added logging for errors on the game page for easier troubleshooting in the future.

BrettCleary:
There was also an issue with linting the webview tag. It looked to be an unnecessary eslint error, so I added 
`{/* eslint-disable react/no-unknown-property */}`
to fix

src/backend/sideload/games.ts
flavioislima: Yes, will keep it here 👍🏽

flavioislima:
@Heroic-Games-Launcher/development 
Need a new review, II believe I have addressed most of the comments and fixed most of the bugs.
Also now I added a api call to check if a game is native or not so we can use it to refactor some checks we are doing in the future.

BrettCleary (approved):
Moved some ipc handlers in main that could be encapsulated in the sideload/games module into that module.

Read through the frontend and backend changes and tested on Windows. Everything is working. Looks great!

src/backend/launcher.ts
```diff
@@ -430,23 +438,40 @@ function launchCleanup(rpcClient?: RpcClient) {
...
+    command = quoteIfNecessary(command)
+  }
 
   const env_vars = {
     ...process.env,
-    ...setupEnvVars(gameSettings),
-    ...setupWineEnvVars(gameSettings, installFolderName)
+    ...setupEnvVars(settings),
+    ...setupWineEnvVars(settings, installFolderName)
```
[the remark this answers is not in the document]
BrettCleary: this was fixed

src/backend/sideload/games.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+export async function removeAppShortcuts(appName: string): Promise<void> {
+  return removeShortcuts(getAppInfo(appName))
+}
+
+export async function launchApp(appName: string): Promise<boolean> {
+  const {
+    install: { executable }
+  } = getAppInfo(appName)
```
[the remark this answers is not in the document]
BrettCleary: this was fixed

flavioislima:
> Tested with Harry Potter Chamber of secrets.
> 
> * Winetricks works
> * Game starts
> * Winecfg fails with:
> 
> ```
> (15:59:25) ERROR:   [Gog]:              Could not get game info for nMm5V3XJbktbJoeqQ12vgb, returning empty object. Something is probably gonna go wrong soon
> (node:16221) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of undefined (reading 'platform')
>     at _GOGGame.isNative (/tmp/.mount_Heroic6uqxjB/resources/app.asar/build/electron/main.3c1baef5.js:27:4306)
>     at _GOGGame.runWineCommand (/tmp/.mount_Heroic6uqxjB/resources/app.asar/build/electron/main.3c1baef5.js:31:2893)
>     at /tmp/.mount_Heroic6uqxjB/resources/app.asar/build/electron/main.3c1baef5.js:100:3492
>     at async node:electron/js2c/browser_init:189:563
> (node:16221) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 5)
> ```
> 
> * The Prefix contains spaces if the title contains them aswell, maybe remove all spaces in the prefix by default, to avoid space problems.

Can you check that again?
For me it is working fine now

flavioislima:
> Moved some ipc handlers in main that could be encapsulated in the sideload/games module into that module.
> 
> Read through the frontend and backend changes and tested on Windows. Everything is working. Looks great!

This change made the tests fail, for some reason, it cannot see the methods inside `ipcMain`. We can investigate that later.

arielj:
I found some visual issues:

In some languages, the Add Game button doesn't fit in the space (it's probably using a fixed with, we should make it fluid):
image

Something happened on the merge of beta and `system requirements` is outside the container that gets shown/hidden:
image

flavioislima:
> I found some visual issues:
> 
> In some languages, the Add Game button doesn't fit in the space (it's probably using a fixed with, we should make it fluid):
> image
> 
> Something happened on the merge of beta and `system requirements` is outside the container that gets shown/hidden:
> image
> 

OK, will send a fix later.
Thanks for testing.

arielj:
> > I found some visual issues:
> > In some languages, the Add Game button doesn't fit in the space (it's probably using a fixed with, we should make it fluid):
> > image
> > Something happened on the merge of beta and `system requirements` is outside the container that gets shown/hidden:
> > image
> 
> OK, will send a fix later. Thanks for testing.

sure! I'll test more later (like actually installing things)

arielj:
I found another issue that I can only reproduce on this branch, when I try to install a game something fails after it finishes loading the download information:

https://user-images.githubusercontent.com/188464/197102812-fa701a62-1a19-4509-b07e-7d017be4d0fc.mp4

It only happens when opening the install dialog of a game I've never tried to install before.

And in the console it shows:
```
(00:11:41) INFO:    [Legendary]:        Running command: /home/ariel/dev/oss/HeroicGamesLauncher/public/bin/linux/legendary info 33411619626243f1859d0b7fb5cc5405 --platform windows --json
[cli] WARNING: Platform "windows" may be invalid. Valid ones are: Windows, Win32, Mac.
```

I think the change from `Windows` to `windows` needs some tweak to work with legendary.

arielj:
Another issue I see (but I'm not sure what's causing it), the game card nor the sidebar show any progress update during installation. It's always 0% until it finishes installing the game. There's nothing wrong in the logs anywhere, also nothing wrong in the devtools console.

arielj:
When I open the `Add Game` dialog, I see this error multiple times in the devtools console:
image

flavioislima:
Ok, fixed the UI bugs:
image
image

And also the issue with the install platform. I swear I have tested installing games using 'Windows' on legendary and it worked but now I can reproduce the issue. 

The error you showed related to the icon I cannot reproduce it. Can you maybe try to remove node_modules and run yarn again?

The progress issue I am investigating.

flavioislima:
Ok, fixed the missing progress issue.
Somehow I removed the `onOutput` option from `callRunner` so the progress was not being sent to the frontend.

arielj:
I can't reproduce the console issue when opening the add game modal, that can be ignored :+1:

One thing that will be a problem for sure, we should remove invalid folder name characters, you can put / * ? ! anything there
image

arielj:
Hmmm I can get the devtools console error when I change from `Windows` to `Linux` in the platform selector

arielj:
If you delete the app's title, the prefix has no folder (which makes sense) but you can still execute an installer with the Run Installer First button.

If the title is empty and you change to Linux and then back to Windows, the prefix now uses a `sideload` folder.

arielj:
In the list view, the Store column shows `sideload` but the filter at the top is `Other`, it should be `Other` too.

arielj:
Maybe this is something for a future iteration, but we can run an installer with a given prefix folder and then change the prefix. I can imagine users changing the prefix by mistake like maybe changing the title after running the installer to fix a typo or something, for people not knowing what a prefix is it can be not clear that it's a problem.

I can maybe we can default to some autogenerated folder name for the prefix (hidden to the user) and allow an `advanced` option to set a prefix folder for the people that knows what it is, that way we show that only for users that know what it is.

flavioislima:
@ariel those were great catches, will work on then.
About the prefix, yes, that can be complicated. We can either disable the input after running a installer there. Or I don't know. It is a really complicated issue to solve because we have users of all levels.
Maybe adding some help texts. 
We can maybe also do those in steps. Like enabling/disabling fields one by one. That is a more complicated approach but could make it more stable.
I believe it is better to test first with real users on the beta and then we can improve it.

flavioislima:
@liamdawe @arielj 
Did these changes and fixes that you guys have commented on here and on #1936 
image

I think in the future we need to show a Dialog on Startup for new Installs with some instructions, that would make things clear.
And we should on that for next release imo.

src/backend/launcher.ts
```diff
@@ -526,28 +544,84 @@ async function runWineCommand(
...
+    prefix: LogPrefix.Backend
   })
-  return execAsync(finalCommand, { env: env_vars })
-    .then((response) => {
-      logDebug(['Ran Wine command:', finalCommand], {
-        prefix: LogPrefix.Legendary
-      })
-      return response
+
+  return new Promise((res) => {
```
CommandMC, line 556: Not that big of a fan of all the repeated code here, *someone* should write a general "runCommand" wrapper soon:tm:

src/backend/launcher.ts
```diff
@@ -283,7 +283,7 @@ function setupWineEnvVars(gameSettings: GameSettings, gameId = '0') {
   }
   if (gameSettings.enableFSR) {
     ret.WINE_FULLSCREEN_FSR = '1'
-    ret.WINE_FULLSCREEN_FSR_STRENGTH = gameSettings.maxSharpness.toString()
+    ret.WINE_FULLSCREEN_FSR_STRENGTH = gameSettings?.maxSharpness?.toString()
```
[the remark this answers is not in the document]
flavioislima: Yes, I added it here because it was causing an error when opening heroic. This fixed it.

src/backend/launcher.ts
flavioislima: I agree. Let's work on that in some refactor in the future.
Maybe a PR to address only this 👍

src/backend/launcher.ts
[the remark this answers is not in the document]
flavioislima: ok updated the types

src/backend/sideload/games.ts
```diff
@@ -122,7 +122,7 @@ export async function launchApp(appName: string): Promise<boolean> {
       mangoHudCommand,
       gameModeBin,
       steamRuntime?.length
-        ? [...steamRuntime, `--filesystem=${gameInfo.install.install_path}`]
+        ? [...steamRuntime, `--filesystem=${gameInfo.folder_name}`]
```
CommandMC, line 125: This isn't right. `gameInfo.folder_name` contains only the game folder name (either `metadata.customAttributes.FolderName.value` or the game's AppName if the former is not set). E.g. let's say I install Fortnite, `gameInfo.folder_name` only contains `Fortnite`

flavioislima: Yes, this is only for Sideload Games, I a not changing for the others.
On this one, we don't have any metadata so I added the folder name on the main key object.

CommandMC: Hm, I see. I'd still say that we should keep this in line with other metadata (so setting `install.install_path` to the path & maybe leaving `folder_name` empty), but I guess this works for now

PR #1865, opened 2022-10-04
Refactor/sidebar

BrettCleary:
sidebar elements are now centered correctly.

before
image

after
image

Also applied some of the .button style to the button element. Applying other fields breaks some of the existing design.

Renamed some of the styles to match the `style-xx` naming convention.

---

### DIFF
--- a/src/frontend/styles/_buttons.scss
+++ b/src/frontend/styles/_buttons.scss
@@ -1,22 +1,30 @@
+@mixin buttonDefaultStyle {
+  font-family: var(--secondary-font-family);
+  border-radius: var(--space-3xs);
+  color: #161616;
+}
+
+button {
+  @include buttonDefaultStyle();
+}
+
 .button {
+  @include buttonDefaultStyle();
   --line-height: 17px;
   cursor: pointer;
   padding: var(--space-sm);
   background: var(--primary);
   color: #1a1b1c;
   text-transform: uppercase;
-  font-family: var(--secondary-font-family);
   font-style: normal;
   font-weight: var(--medium);
   font-size: var(--text-md);
   line-height: var(--line-height);
-  color: #161616;
   border: none;
   max-height: 40px;
   text-align: center;
   transition: 500ms;
   box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
-  border-radius: var(--space-3xs);
 }
 
 .button:disabled {
--- a/src/frontend/styles/_typography.scss
+++ b/src/frontend/styles/_typography.scss
@@ -80,8 +80,13 @@ h2,
 h3,
 h4,
 h5,
-h6 {
+h6,
+.subtitle {
   font-family: var(--primary-font-family);
+}
+
+.content-sm {
+  font-family: var(--secondary-font-family);
 }
 
 .text--6xl {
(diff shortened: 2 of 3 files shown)


src/frontend/styles/_typography.scss
```diff
@@ -119,22 +124,44 @@ h5,
...
 }
+
+.subtitle,
+.content-sm,
+h5,
+h4 {
+  font-weight: var(--medium);
+}
+
+.subtitle-sm,
```
arielj, line 155: do we need these changes? we are not using these utility classes anywhere, and it's not clear when we'll use utility classes

src/frontend/styles/_buttons.scss
```diff
@@ -1,22 +1,30 @@
+@mixin buttonDefaultStyle {
```
arielj, line 1: not sure this changes are related to this PR, the sidebar uses link tags, wondering if these changes were committed by mistake?

src/frontend/styles/_typography.scss
BrettCleary: yeah we can remove all the utility classes if you want. Billie and I came up with these and are going to start trying to use them in new designs. I went ahead and added them so that other frontend devs could build with them. This way it prevents rework/refactoring later on

BrettCleary: what do you think? leave them in and start integrating them now or wait until later and do a more formal refactor?

src/frontend/styles/_buttons.scss
BrettCleary: the sidebar links are <button> and <a> elements. iirc the <button> elements were using the primary font instead of the secondary font (hard to tell unless you change the fonts to look very different). I went ahead and made a mixin so that we could share properties between `<button>` and `.button` css classes because sharing all the properties breaks other <button> elements.

arielj: oh weird, I checked a few of them and thought they were all `a` tags. I wonder why Manage Accounts is a button when other items that change the page are `a` tags :S.

I imagine the Discord/Patreo/Ko-fi are buttons because they open external urls.

Would be nice to standardize what we use there, but I guess that's a different thing to refactor haha.

src/frontend/styles/_typography.scss
arielj: hmmm we don't have a definition on how to write the css so we currently have a mess of naming conventions and methodologies; we have camelCase, kebab-case, BEM, utility classes, in some places I even saw inline styles

unless it's clear what should be used in new code people might miss these new utility classes since the code doesn't instruct people to use them

I don't have a strong opinion on what we should aim for in heroic's css conventions (I personally don't like utility classes, but it's just a personal opinion) but I would say to keep these utility classes only if it's clear that we want to go in that direction of writing styles

BrettCleary: Gotcha. Yeah I think it would be good to try and standardize and restrict them to some predefined styles defined in a design system here and in the figma file. It'll make it easier to change later to have it defined in one place and help prevent the mess that you're describing.

It'll be a pretty big refactor to take the current text css styles and redefine them here. It should probably start in a figma file too, so I haven't started working on that (and why these are unused currently)

src/frontend/styles/_buttons.scss
BrettCleary: Yeah they should probably be standardized but I agree that should be a separate pr haha

src/frontend/styles/_typography.scss
BrettCleary: I think we should move this discussion into an issue @arielj since it's not really related to this PR

CommandMC (approved):
I can't really review this change as I have little CSS knowledge, but from the provided images & small bit of testing myself, seems to be alright :+1:

src/frontend/styles/_typography.scss
arielj: sure!

arielj (approved):

Issue #1867, opened 2022-10-05
Unicode doesn't display correctly in error when using Add To Steam with invalid steam path


idbrii:
### Describe the bug

If your steam path is invalid, when Heroic notifies you of an error, it has unicode characters which make the incorrect path hard to read.

### Add logs

```
(20:07:34) INFO:    [Legendary]:       Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary --version
(20:07:35) INFO:    [Legendary]:       Legendary location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(20:07:35) INFO:    [Gog]:             GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(20:07:35) INFO:    [Backend]:         

Heroic Version: 2.4.3 Chopper
Legendary Version:  0.20.28 Dark Energy (hotfix #2)
OS: Freedesktop SDK KERNEL: 5.13.0-valve21.1-1-neptune-02211-gc54cda5a36f3 ARCH: x64
CPU: AMD Custom APU 0405 @1.82 GOVERNOR: schedutil
RAM: Total: 14.49 GiB Available: 12.71 GiB
GRAPHICS: GPU0: VanGogh [AMD Custom GPU 0405] VRAM: 512MB DRIVER:  
PROTOCOL: x11

(20:07:35) INFO:    [Gog]:             Getting data about the user
(20:07:35) WARNING: [Backend]:         Failed to register protocol with OS.
(20:07:35) INFO:    [Backend]:         AreWeAntiCheatYet data downloaded
(20:07:36) INFO:    [Gog]:             Saved user data to config
(20:07:36) INFO:    [Frontend]:        Refreshing Library
(20:07:36) INFO:    [Legendary]:       Refreshing library...
(20:07:36) INFO:    [Legendary]:       Refreshing Epic Games...
(20:07:36) INFO:    [Gog]:             Getting GOG library
(20:07:36) INFO:    [Legendary]:       Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary list
(20:07:36) INFO:    [Gog]:             Number of library pages: 1
(20:07:37) INFO:    [Gog]:             Saved games data
(20:07:37) INFO:    [Backend]:         Downloaded Winetricks
(20:07:38) INFO:    [Legendary]:       Updating game list
[...]

PR #1873, opened 2022-10-06
[Fix] Unicode chars in non-steam-game error dialog

redromnon:
The unicode appears in the dialog box because i18next escapes the value to avoid XSS injection attacks by default. 
Solved by turning off the escaping flag.

Related Issue:
#1867 
---


flavioislima (approved):
LGTM

PR #1876, opened 2022-10-06
Fix setting not updating after another setting changes

arielj:
This PR fixes an issue after the settings refactor that the useSetting was not triggering a re-render other components after setting the new value.

It was using some useEffects/useStates but they were actually not needed.

---


flavioislima (approved):
LGTM

PR #1887, opened 2022-10-09
[TECH] Move default save computation into the backend

CommandMC:
This PR does two main things:
- Default save path computation is now done in the backend
- Instead of parsing the path ourselves (and filling in any variables), we now let Legendary do it for us

Since I got a bit carried away, it also features:
- Types for a few of the GOG-related data structures

TODO:
- [X] Move GOG's path resolution to the backend as well
- [X] Pass `WINEPREFIX`/`STEAM_COMPAT_DATA_PATH`/`CX_BOTTLE` to Legendary

---


CommandMC:
Some of the changes in the last 3 commits are a bit off-topic, but I just couldn't live with everything just being the `any` type, so I annoyed @imLinguin for the last 2 days about all of the structure info he had, and put that into 29422b90b6acb8980181964280545d7b302820f2 and by extension 14c3c8dac085c2b218305d125c4dc4a6f714e80e and 1b35d93b921df0ec8e7ed03e6a003498fd758831

flavioislima:
Testing this one now.
It is nice to have the option to manually select the path again now, but at the same time I believe we are solving the path but not showing it in the frontend. For instance, for CP2077, on the backend I can see that we get the proper path, but on the frontend the input is not updated:
```
  "gogSaves": [
    {
      "name": "saves",
      "location": "/home/flavio/Games/Prefixes/CP2077/pfx/drive_c/users/steamuser/Saved Games/CD Projekt Red/Cyberpunk 2077"
    }
  ],
  "gogSavesLocations": [
    {
      "name": "saves",
      "location": ""
    }
  ]
```

Probably we need to update the second item with the value from the first one,

flavioislima:
Some other GOG games don't even show the input anymore:
image

CommandMC:
I can't find a reference to `gogSaves` anywhere in the code, everything is using `gogSavesLocations`. I also can't find `savesPath` getting set for GOG games (it's only used by Epic games, since those only have 1 possible save path)
I suspect you have some older settings in your config files

After much debugging, I've now found the issue: `winepath` indeed just... doesn't run when ran inside of Proton. Will see if I can find why that happens
I'm not sure how that 2nd issue happened though, what's in `gogSavesLocations` for that one now?

flavioislima:
Hmm, I am 90% sure that we use `savesPath` for GOG as well but I am not sure anymore after the latest save-sync refactor. 🤔

CommandMC:
Well GOG games don't just have one `savesPath`, so I can't really imagine what we'd put there

Anyways, found the issue with Proton: There's a specific verb (`getnativepath`) just for running `winepath` on a path (and another, `getcompatpath` for converting from Unix to Windows paths)
Looks like I'll have to add a new `runWinePath` function that uses that then

CommandMC:
Proton path computation should be fixed now
As for your second issue, I don't quite know how it happened, and I can't reproduce it. I've added a lot of debug logging now (intended to be removed before merging this of course), hopefully that helps in finding the issue

CommandMC:
I've now managed to reproduce this 2nd issue. It's indeed caused by *something* set in the cache when using older development versions of Heroic
Clearing your cache (and optionally restarting Heroic) fixed it for me though

flavioislima (approved):
Looks good and it is working fine now with both Proton and Wine

PR #1891, opened 2022-10-11
[UI/UX] Replace native dialogs

BrettCleary:
This refactors native error and message dialog boxes into modals.

The `GlobalState`  ContextProvider now has `showDialogModal` and `dialogModalOptions`. Any frontend component can call `showDialogModal` to show a dialog. This is achieved with the `DialogHandler` component which uses `dialogModalOptions` to determine what dialog to show. The `DialogHanlder` also listens for dialogs that the backend requests to show.

Before:
Native Uninstall Modal
Native Reset Heroic
Native Cache Cleared

After:
Uninstall Modal
EAC runtime enabled modal
Cache Cleared Modal

---


flavioislima:
so far is looking pretty good, just a few comments on the current state on Linux:
You can see on the video below that when uninstalling a windows game on linux it shows the uninstall dialog without the checkbox to delete the prefix and then the checkbox appears. So maybe we could delay the checkbox first or check if it is not a native game and is linux, then show the checkbox by default.
https://user-images.githubusercontent.com/26871415/195089384-40ec4255-9017-46ec-9688-90cb0b0ca794.mp4

And after clicking uninstall the dialog doesn't close until the action is complete, which makes Heroic hang for a few seconds. So perhaps adding or an indicator that the operation is ongoing or closing the dialog without waiting for the promise to be fulfilled. Other than that looks cool 👍🏽

src/backend/main.ts
```diff
@@ -700,31 +700,19 @@ ipcMain.handle('getLatestReleases', async () => {
...
-    ),
-    buttons: [i18next.t('box.ok', 'Ok')]
-  })
+  event.sender.send(
+    'showDialog',
+    i18next.t('box.cache-cleared.title', 'Cache Cleared'),
+    i18next.t('box.cache-cleared.message', 'Heroic Cache Was Cleared!'),
+    false,
+    [i18next.t('box.ok', 'Ok')]
+  )
```
Nocccer, line 711: I would love to have a generic function, else we need to adapt code later if we wanna change stuff.
So we need a `showDialogModalAuto` function, which replaces the `showErrorBoxModalAuto`

src/backend/api/misc.ts
```diff
@@ -83,12 +83,18 @@ export const clipboardReadText = async () =>
...
 
-export const handleShowErrorDialog = (
-  onError: (e: Electron.IpcRendererEvent, title: string, error: string) => void
+export const handleShowDialog = (
+  onMessage: (
+    e: Electron.IpcRendererEvent,
+    title: string,
+    message: string,
+    isError: boolean,
+    buttons: Array<string>
```
Nocccer, line 92: Maybe we add here a map instead just the button names. Maybe we need to map a action later to each button. So each entry is a pair of the button name and the callback/action to trigger on click.

src/frontend/components/UI/Dialog/MessageBoxModal/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  DialogFooter,
+  DialogHeader
+} from 'frontend/components/UI/Dialog'
+
+interface MessageBoxModalProps {
+  title: string
+  message: string
+  onClose: () => void
+  buttons: Array<string>
+  buttonsOnClick: Array<() => void>
```
Nocccer, line 15: Combine this to a map or new interface ?
`Map<string, Function>` or 
```typescript
interface Button {
  name: string
  action: () => void
}
```

Nocccer (dismissed):
Looks really promising. I left some suggestions i find useful.
Even if we even not have a single error dialog with actions at the moment, maybe we need one in the future where we provide actions on a error dialog, which mostly comes from the backend.

src/backend/api/misc.ts
BrettCleary: We need to be able to specify the order the buttons will appear though. If we have a map like `{button_a: ()=>{}, etc.}` and we iterate through the keys, we won't be able to specify the order.

src/backend/api/misc.ts
```diff
@@ -83,12 +83,18 @@ export const clipboardReadText = async () =>
...
   ipcRenderer.send('clipboardWriteText', text)
 
-export const handleShowErrorDialog = (
-  onError: (e: Electron.IpcRendererEvent, title: string, error: string) => void
+export const handleShowDialog = (
+  onMessage: (
+    e: Electron.IpcRendererEvent,
+    title: string,
+    message: string,
+    isError: boolean,
```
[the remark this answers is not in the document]
BrettCleary: sure. we can add an enum

src/frontend/components/UI/Dialog/MessageBoxModal/index.tsx
BrettCleary: this was a great suggestion. simplified the code a bit and will allow us to add other button options in the future like a styling enum by only changing the code in one place

src/backend/main.ts
BrettCleary: sure thing. will change

BrettCleary:
> so far is looking pretty good, just a few comments on the current state on Linux: You can see on the video below that when uninstalling a windows game on linux it shows the uninstall dialog without the checkbox to delete the prefix and then the checkbox appears. So maybe we could delay the checkbox first or check if it is not a native game and is linux, then show the checkbox by default.
> 
> And after clicking uninstall the dialog doesn't close until the action is complete, which makes Heroic hang for a few seconds. So perhaps adding or an indicator that the operation is ongoing or closing the dialog without waiting for the promise to be fulfilled. Other than that looks cool 👍🏽

This is fixed now. Closes immediately. Also uses the context provider to get the platform instead of awaiting for it. Doesn't show the dialog until it checks this so no re-render now

BrettCleary:
I am having an issue with absolute imports now though

`[vite]: Rollup failed to resolve import "common/types" from "src/backend/main.ts".`

Relative imports are working fine and it seems to only be common/types after I added a couple types to it.

BrettCleary:
Okay so adding an enum in the common/types module breaks all the imports. 

In troubleshooting this, I went deep into the rabbit hole regarding absolute imports and managed to get it working for the frontend and backend without the vite-tsconfig-paths plugin. Before, we could only use imports in the frontend.

The prefix alias is `@@/` which maps to the `src` directory.

BrettCleary:
This is working for me and tests are updated so I'll take it out of draft now

flavioislima:
Testing here and now the linux dialog for uninstall looks good. Maybe just add a limiit to the size of the dialogs in general, this one is huge, for instance:
image

src/backend/utils.ts
```diff
@@ -483,7 +485,7 @@ async function getSteamRuntime(
   const steamLibraries = await getSteamLibraries()
   const runtimeTypes: SteamRuntime[] = [
     {
-      path: 'steamapps/common/SteamLinuxRuntime_soldier/run',
+      path: 'steamapps//src/common/SteamLinuxRuntime_soldier/run',
```
arielj, line 488: this change seems not intended (maybe a search/replace thing?)

src/frontend/screens/Settings/components/OtherSettings/GameMode.tsx
```diff
@@ -48,7 +53,7 @@ const GameMode = () => {
     <div className="toggleRow">
       <ToggleSwitch
         htmlId="gamemode"
-        value={useGameMode}
+        value={false}
```
arielj, line 56: is this intentional?

src/backend/logger/__tests__/logfile.test.ts
```diff
@@ -85,10 +85,7 @@ describe('logger/logfile.ts', () => {
     const data = logfile.createNewLogFileAndClearOldOnces()
 
     expect(logError).toBeCalledWith(
-      [
-        expect.stringContaining('Removing old logs in /tmp/'),
-        Error('unlink failed')
-      ],
+      [expect.stringContaining('Removing old logs in'), Error('unlink failed')],
```
arielj, line 88: is this `/tmp/` removal intentional?

BrettCleary: it fails on windows if I keep it

BrettCleary: I can add it back. we should probably use path module so tests aren't expecting linux formatted file paths

BrettCleary:
> Testing here and now the Linux dialog for uninstalling looks good. Maybe just add a limit to the size of the dialogs in general, this one is huge, for instance: image

done.

src/backend/utils.ts
BrettCleary: yup my mistake. fixed now

src/frontend/screens/Settings/components/OtherSettings/GameMode.tsx
BrettCleary: nope. dunno why that was changed. my mistake. fixed now. nice catch

flavioislima (approved):
Looks great. Thanks for doing this ⚔️

CommandMC (approved):

src/frontend/screens/Library/components/LibraryHeader/index.tsx
```diff
@@ -62,6 +62,8 @@ export default function LibraryHeader({
     return 'gog'
   }
 
+  console.log('LIBRARY CATEGRORY = ', category)
+
```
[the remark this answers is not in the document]
BrettCleary: it was not. removed now. I also read through all the changes again to make sure nothing else like this was left in

biliesilva:
> > Testing here and now the Linux dialog for uninstalling looks good. Maybe just add a limit to the size of the dialogs in general, this one is huge, for instance: image
> 
> done.

It's important to differentiate the buttons **No** and **YES**, because this can confuse the user. The button that has more importance has to be all filled with color.  Also is better to reduce the size of the box as you said.

flavioislima:
@biliesilva the size was fixed. But yes, that is a good point about the colors.
I think we will need to open a new Issue here on GH so we can fix that.

@BrettCleary 
I think we can add a new prop to the dialog to choose which button is important or not.
Or maybe just highlight the first (or last) button by default with a different color.

BrettCleary:
> @biliesilva the size was fixed. But yes, that is a good point about the colors. I think we will need to open a new Issue here on GH so we can fix that.
> 
> @BrettCleary I think we can add a new prop to the dialog to choose which button is important or not. Or maybe just highlight the first (or last) button by default with a different color.

We can add a style enum or string type to the `ButtonOptions` type in `src/common/types` for `buttonStyle` and then set have predefined css styles for each applied in the `DialogHandler` component

PR #1895, opened 2022-10-12
[Tech] Make it possible to disable the React developer tools

CommandMC:
This PR allows for disabling the installation of the React dev tools by setting the `HEROIC_NO_REACT_DEVTOOLS` env var before running `yarn dev`.

I'm currently developing on a somewhat older machine, and I've noticed that this leads to just a *little* bit faster load times. And that little time goes a long way when you're restarting Heroic ~50 times a day :^)

---


arielj (approved):
Looks good, would be good to update the README to show this next to the `yarn dev` instructions.

PR #1897, opened 2022-10-12
[Tech] System info cache

BrettCleary:
This improves game launch speed by caching system info that won't change while the app is running.

Time to launch was reduced from 20 seconds to 9 seconds with League of Legends on Windows. 

It also speeds up Heroic launch time by ~5 seconds. Before, we were unnecessarily awaiting getSystemInfo before launching the main window. Now it runs asynchronously.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1002,6 +1002,7 @@ ipcMain.handle(
       game.logFileLocation,
       'System Info:\n' +
         `${systemInfo}\n` +
+        '\n' +
         `Game Settings: ${gameSettingsString}\n` +
         '\n' +
         `Game launched at: ${startPlayingDate}\n` +
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -344,7 +344,9 @@ if (!gotTheLock) {
   app.whenReady().then(async () => {
     initOnlineMonitor()
 
-    const systemInfo = await getSystemInfo()
+    getSystemInfo().then((systemInfo) =>
+      logInfo(`\n\n${systemInfo}\n`, { prefix: LogPrefix.Backend })
+    )
 
     initImagesCache()
 
@@ -355,7 +357,6 @@ if (!gotTheLock) {
     logInfo(['GOGDL location:', join(...Object.values(getGOGdlBin()))], {
       prefix: LogPrefix.Gog
     })
-    logInfo(`\n\n${systemInfo}\n`, { prefix: LogPrefix.Backend })
     // We can't use .config since apparently its not loaded fast enough.
     const { language, darkTrayIcon } = await GlobalConfig.get().getSettings()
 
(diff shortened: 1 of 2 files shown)


src/backend/main.ts
```diff
@@ -1001,7 +1002,6 @@ ipcMain.handle(
       game.logFileLocation,
       'System Info:\n' +
         `${systemInfo}\n` +
-        '\n' +
```
CommandMC, line 1004: Won't this remove the empty line between the System Info & the settings? Right now, the log looks like this:
```
System Info:
...
GRAPHICS: GPU0: 3rd Gen Core processor Graphics Controller  VRAM: 256MB DRIVER:  
PROTOCOL: wayland

Game Settings: {
...
```
I'd say the empty line separates these two items quite well. Why remove it?

BrettCleary: In my GamesConfig folder, it showed two new lines. I see now this was from this line and the fact that I'm not on Linux
`${isLinux ? `PROTOCOL: ${xEnv}` : ''}``

I'll add it back

CommandMC:
That 11 second metric seemed a bit high for me, so I've had a look into the `systeminformation` module we're using
Seems like they're executing PowerShell commands on Windows, which is why it's so slow.

Maybe we should *someday* implement this on our own, using Win32 API calls and/or WMI queries. Those should be significantly faster

BrettCleary:
> That 11 second metric seemed a bit high for me, so I've had a look into the `systeminformation` module we're using Seems like they're executing PowerShell commands on Windows, which is why it's so slow.
> 
> Maybe we should _someday_ implement this on our own, using Win32 API calls and/or WMI queries (there are npm modules for making both of those things easy, as far as I know). Those should be significantly faster

I was getting 5-10 seconds improvements consistently. Might be some user error in those numbers. Only tested on windows too

There's still no need for getSystemInfo to be synchronous and not cached. The only improvement we'd get by making the getSystemInfo call faster is if the app opens faster than getSystemInfo runs and the user launches a game before it finishes. In that case, we'd run getSystemInfo twice and the game open would be slow. The app doesn't open that fast and I think most users won't launch a game in the first second the app opens.

It's also only used for logging so if this edge case becomes problem, we could disable system info logging on game launch since it's already logged on app start.

CommandMC:
> I was getting 5-10 seconds improvements consistently. Might be some user error in those numbers. Only tested on windows too

To clarify, I was not at all questioning your measurements, I was just observing & suspecting what the reason for a difference this significant might be

> The app doesn't open that fast and I think most users won't launch a game in the first second the app opens.

This might happen when games are launched via the protocol

> It's also only used for logging so if this edge case becomes problem, we could disable system info logging on game launch since it's already logged on app start.

This might be a bad idea, since users usually share just the game log & not the general one

CommandMC (approved):

flavioislima (approved):
Looks good 👍🏽

Issue #1870, opened 2022-10-06
Intercept store login and redirect to Heroic's login


arielj:
### Problem description

Many users log in in the store inside heroic instead of the Manage Accounts section, then they are not actually logged in from Heroic's point of view.

### Feature description

We can use the preload script functionality to inject some code in the stores and make the users log in in the right place.

I can imagine 2 options:
- detect when the user is trying to log in and redirect them to the Manage Accounts section
- or, detect when a user is not logged in in the store and show a popup telling them to log in using the Manage Accounts section

### Alternatives

Maybe check if the user is logged in when they click the Store in the sidebar and do this at that point?

Or maybe hide the Store options until the users are logged in

PR #1900, opened 2022-10-13
Try to prevent login using the store webview

arielj:
Many users log in using the store webview when something fails in the login, creating more issues because then the user is already logged in the store but not in Heroic.

This PR adds a `Log in` dialog/warning covering the page that would take the user to the login process we want for each store. This fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1870

https://user-images.githubusercontent.com/188464/197361806-68c526c7-62c2-4336-9ffb-db75db1235ed.mp4

---


CommandMC (commented):
First of all, I like it. The "Log in" button is basic, but it works.

That being said, I have some concerns:
While testing this locally, I noticed that the button was "trailing behind" when resizing the window, making it possible to still click the actual button below it if you're fast enough
There are also other ways of getting to the login prompt, for example by trying to check out in GOG. I also fear that this'll break with a site redesign

Maybe disabling the "Stores" tab for the respective store if you're not logged in would be easier

arielj:
> First of all, I like it. The "Log in" button is basic, but it works.
> 
> That being said, I have some concerns: While testing this locally, I noticed that the button was "trailing behind" when resizing the window, making it possible to still click the actual button below it if you're fast enough There are also other ways of getting to the login prompt, for example by trying to check out in GOG. I also fear that this'll break with a site redesign
> 
> Maybe disabling the "Stores" tab for the respective store if you're not logged in would be easier

I agree that a user can find ways to break this like resizing the window or going somewhere else ignoring the `Log In` button, but if you actually try to be fast enough to click the login button while resizing then you are looking to have an issue on purpose.

I think I can detect when the user navigates to any of the checkout pages and redirect them to the login if they are not logged in (my first implementation was detecting the current URL after a user change but had to discard that because the GOG login is not a separated URL, but I can make a combination of both strategies)

I thought about disabling the stores, I just didn't know if we really wanted to do that (I don't know what was the reason to have the stores display when logged out in the first place).

CommandMC:
> but if you actually try to be fast enough to click the login button while resizing then you are looking to have an issue on purpose

Speaking from unpleasant past experiences, I'd never underestimate how determined a user is to breaking your software

> I don't know what was the reason to have the stores display when logged out in the first place

I don't think there is a reason.
This, to me, sounds much easier than adding this element and watching for certain URL/navigation changes, on top of it of course not breaking when sites change

flavioislima:
A more simple approach would be to hide the Store until the user is logged in.
Or show a warning window if the user clicks on the store link but it is not logged in.
Like:
`Are you trying to log in? Click here to Login to your GOG/Epic account.`
And another button saying: `I am just looking at the store, continue.`

I think having this button is not sustainable in the long run because both epic and GOG can change their layout and this will break on our end as well, so we would need another update to fix that.

arielj:
I'll wait until `feat/replace_native_dialogs` is merged so I can use that feature for this to make a dialog on top of the stores to prevent users from logging in there

arielj:
@flavioislima @CommandMC I replaced this with a dialog component covering the screen, I think it looks better now and it's more obvious to the user. They can still close it and navigate the stores.

flavioislima (approved):
LGTM!
Now looks a lot better imo 👍🏽

PR #1901, opened 2022-10-13
[Tech] Add types to all IPC functions

CommandMC:
This PR improves our IPC system by adding return types wherever possible and refining some already existent return & parameter types  
It's one of the necessary steps to finally get full confidence over our variable typings, right in line after the general IPC rework in #1783 

Some non-required, but "would be nice" TODOs:
- [X] Having some sort of linter that can automatically look at the IPC call's implementation, and make sure the paramter/return types there match up with the definition in the preload file
- [X] Some of the GOG types could use more polish. Mostly implemented in #1887, so I'd like to wait for that one to merge first

---


src/frontend/components/UI/Sidebar/components/SidebarLinks/index.tsx
```diff
@@ -70,13 +70,15 @@ export default function SidebarLinks() {
...
-        setGameInfo(info)
-        if (info?.is_installed) {
-          setIsDefaultSetting(false)
-          const wineOrOther = isWin
-            ? `/settings/${runner}/${appName}/other`
-            : `/settings/${runner}/${appName}/wine`
-          setSettingsPath(wineOrOther)
+        if (info) {
+          setGameInfo(info)
+          if (info?.is_installed) {
```
arielj, line 75: I think the `?` is not needed now?

CommandMC: There are probably all kinds of optional chaining and non-null assertion operators in the frontend that can now be removed

There's [an eslint rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md) that makes finding these easy, I'll give that a go

CommandMC: Uh oh, that was a bad idea, >200 problems with that rule

It might be better to make those changes in a separate PR then

CommandMC: Got them all solved now, but it's ~600 lines of changes, so I'll add those in another PR

src/backend/main.ts
```diff
@@ -1007,227 +1064,255 @@ ipcMain.handle(
...
-  const { filePaths, canceled } = await showOpenDialog(mainWindow, {
-    ...args
-  })
-  if (filePaths[0]) {
-    return { path: filePaths[0] }
+ipcMain.handle(
+  'openDialog',
+  async (e, args: Electron.OpenDialogOptions): Promise<string | false> => {
+    const { filePaths, canceled } = await showOpenDialog(mainWindow, args)
+    if (canceled) {
```
arielj, line 1125: this should be

```suggestion
    if (!canceled) {
```

src/backend/main.ts
```diff
@@ -1007,227 +1064,255 @@ ipcMain.handle(
...
-        if (isCLINoGui) {
-          app.exit()
-        } else {
-          mainWindow.show()
-        }
-      })
+    // Show main window if we've minimized it before
+    if (minimizeOnLaunch) {
+      mainWindow.show()
+    }
```
arielj, line 1110: do we need this separation of the if/else? this is checking the `minimizeOnLaunch` setting. I think that if we launch a game with the `no-gui` flag and we have the minimizeOnLaunch flag turned on, this 2 ifs will first show the app and then close it, while the original if/else is either closing the app or showing it.

src/backend/main.ts
```diff
@@ -791,106 +819,132 @@ ipcMain.handle(
...
+ipcMain.handle(
+  'isLoggedIn',
+  async (): Promise<boolean> => LegendaryUser.isLoggedIn()
+)
 
 ipcMain.handle('login', async (event, sid) => LegendaryUser.login(sid))
 ipcMain.handle('authGOG', async (event, code) => GOGUser.login(code))
 ipcMain.handle('logoutLegendary', async () => LegendaryUser.logout())
-ipcMain.handle('logoutGOG', async () => GOGUser.logout())
+ipcMain.on('logoutGOG', async () => GOGUser.logout())
```
arielj, line 836: any reason this one changes from `handle` to `on`? I understand they are slightly different but I don't know why it's changed only for this logout and not the other

arielj: is it because LegendaryUser.logout() is async and GOGUser.logout() is sync? I wonder if we should have both logout methods to be async for consistency

src/backend/main.ts
CommandMC: Hm, I see the issue. The reason I've changed this `if` was that Heroic would take focus after a game quit even if we weren't minimized/hidden  
I'll just return it to how it was before, not sure why I've added this change here anyways

src/backend/main.ts
CommandMC: Yup, that was the reason I've made that change. I was also thinking about leaving it as-is because of consistency, but decided against that since you can easily see which functions are async/sync in the frontend, so consistency isn't really needed

flavioislima (dismissed):
good stuff 👍🏽

flavioislima:
I needed to dismiss my reviews since I am having some bugs like the GamePage is not working, it is throwing an error for both GOG and Legendary:
```
(09:36:28) ERROR:   [Gog]:              ReferenceError: game is not defined
    at /home/flavio/Develop/Electron/HeroicGamesLauncher/build/electron/main.01dcdc58.js:11328:34
    at node:electron/js2c/browser_init:189:579
    at EventEmitter.<anonymous> (node:electron/js2c/browser_init:161:11327)
    at EventEmitter.emit (node:events:527:28)
```

the same happens when clicking install on the game card.

flavioislima (changes requested):
Let me know once the bugs are fixed so I can test it again 👍🏽

flavioislima:
> Looks like one of my next TODOs is "finally cutting down on the `@ts-expect-error` comments". This is one issue caused by them:
> 
> I renamed the `game` variable to `appName` in that function, and since the only time that was used was below a `@ts-expect-error` comment, I didn't get a warning that it was undefined

I see, ok I will check again 👍🏽

BrettCleary:
The ipcMain.on/handle callback function almost always has the same function type as the preload script function. In `src/common/types` I described a way where we would only have to type these in the preload script instead of defining it in two different places. Here's the code/comment describing the approach.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/a03347e329b7a6736099534e0f927c4637d56a9b/src/common/types.ts#L18-L35

This way we only need to change the function signature in the preload script and still get fullstack type checking.

What are your thoughts on this approach @CommandMC?

CommandMC:
Hm, I think I'm mostly getting how this works, but I'm not sure if I *like* how it works :sweat_smile:
IMO that can be in its own PR though, this is already getting fairly large

Thanks for the reminder though, I honestly totally missed that comment

CommandMC:
@BrettCleary I've now implemented full-stack type checking my way. It's more compex and certainly less readable, but also has some benefits
- No more typos, you can only `send`/`invoke` and `on`/`handle` IPC channels that you've defined in the .d.ts file
- Parameter types are checked even in `backend/api` definitions (already caught some issues there)
- Parameter & return types are defined once (single source of truth)

The main drawback is that you now have to jump through one more hoop to add a new IPC function (adding it to `common/typedefs/ipcBridge.d.ts`, either to `SyncIPCFunctions` or `AsyncIPCFunctions`), although since this gives us more features and makes the other two hoops (adding the actual implementation in the backend, adding the callback in `backend/api`) easier, I'd say that's sensible

Let me know what you think

flavioislima (approved):
OK, looks good now! :D

src/frontend/screens/Settings/components/DefaultSteamPath.tsx
```diff
@@ -46,7 +45,7 @@ const DefaultSteamPath = () => {
             title: t('box.default-steam-path', 'Steam path.'),
             defaultPath: defaultSteamPath
           })
-          .then(({ path }: Path) =>
+          .then((path) =>
             setDefaultSteamPath(path ? `${path}` : defaultSteamPath)
```
arielj, line 49: Maybe replace this as `path || defaultSteamPath` as you did in other places?

src/frontend/screens/Settings/components/DefaultInstallPath.tsx
```diff
@@ -21,7 +20,7 @@ const DefaultInstallPath = () => {
         title: t('box.default-install-path'),
         defaultPath: defaultInstallPath
       })
-      .then(({ path }: Path) =>
+      .then((path) =>
         setDefaultInstallPath(path ? `${path}` : defaultInstallPath)
```
arielj, line 24: Same here as in other places

src/frontend/screens/Settings/components/CustomWineProton.tsx
```diff
@@ -33,8 +32,8 @@ export default function CustomWineProton() {
         properties: ['openFile'],
         title: t('box.customWine', 'Select the Wine or Proton Binary')
       })
-      .then(({ path }: Path) => {
-        if (!customWinePaths.includes(path)) {
+      .then((path) => {
+        if (path && !customWinePaths.includes(path)) {
```
arielj, line 36: do we want this `path &&` here? if this is true then the `path ? [...customWinePaths, path] : customWinePaths` at line 38 will always use the true condition

maybe it should be something like:

```ts
if (!path) {
  setCustomWinePaths(customWinePaths)
} else if (!customWinePaths.includes(path)) {
  setCustomWinePaths([...customWinePaths, path])
}
```

or something like that

arielj:
hey, I left 2 comments, the `path || defaultSteamPath` one is not really that important though, I just saw it was changed in other places

src/frontend/screens/Settings/components/CustomWineProton.tsx
CommandMC: Well, `setCustomWinePaths(customWinePaths)` is essentially a no-op (not sure if that would trigger a re-render as well), so I'm just going to remove that 2nd ternary operator there & always use `[...customWinePaths, path]`

PR #1902, opened 2022-10-14
Fix error when clicking "Settings" when the current screen is a login webview

arielj:
This PR fixes an issue I just noticed while working on another PR.

Steps to reproduce the issue before this fix:
- go to Manage Accounts
- click Login for any store
- click Settings
- react app crashes with an error in the console, blank screen

The issue is that the url for the logins is `/loginweb/legendary` or `/loginweb/gog` and the sidebar component does a `split` on that url and was considering `legendary` and `gog` as runners like it happens for game page urls `/gamepage/legendary/someGame` but the was no appName in the url.

This PR fixes that by only linking to a game setting if we have an app name

---


flavioislima (approved):
LGTM

Issue #1868, opened 2022-10-05
Feature: Automatically add to Steam


idbrii:
### Problem description

It's awesome that with #1331 Heroic has an option to add games to Steam, but it would be great if there was an option to automatically add games to steam so I don't need to click that every time I install a new game.

### Feature description

Add a checkbox in "Settings > Other" after "Add games to start menu automatically" add a new option "Add games to Steam automatically". When you install a game, it does the same thing as the Add to Steam button. When you remove a game, it removes it from steam (looks like that's already there?).

### Alternatives

Add a postinstall command to run boilr or steamsync to add games and art to steam. But heroic already has Add to Steam, so might as well use it.

PR #1904, opened 2022-10-14
Add shortcuts to Steam automatically

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1868 adding the setting to automatically add games to Steam after installation.

I'll add some notes in the diff to clarify some of the changes.

This branch is branched from https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1903, if 1903 gets merged first we may need to change the `base` of this PR before merging!

---

### DIFF
--- a/src/backend/api/menu.ts
+++ b/src/backend/api/menu.ts
@@ -22,13 +22,8 @@ export const installEosOverlay = async () =>
   ipcRenderer.invoke('installEosOverlay')
 export const removeFromSteam = async (appName: string, runner: Runner) =>
   ipcRenderer.invoke('removeFromSteam', appName, runner)
-export const addToSteam = async (
-  appName: string,
-  runner: Runner,
-  bkgDataURL: string,
-  bigPicDataURL: string
-) =>
-  ipcRenderer.invoke('addToSteam', appName, runner, bkgDataURL, bigPicDataURL)
+export const addToSteam = async (appName: string, runner: Runner) =>
+  ipcRenderer.invoke('addToSteam', appName, runner)
 export const shortcutsExists = async (appName: string, runner: Runner) =>
   ipcRenderer.invoke('shortcutsExists', appName, runner)
 export const isAddedToSteam = async (appName: string, runner: Runner) =>
--- a/src/backend/gog/games.ts
+++ b/src/backend/gog/games.ts
@@ -630,12 +630,7 @@ class GOGGame extends Game {
     await removeShortcuts(this.appName, 'gog')
     syncStore.delete(this.appName)
     const gameInfo = await this.getGameInfo()
-    const { defaultSteamPath } = await GlobalConfig.get().getSettings()
-    const steamUserdataDir = join(
-      defaultSteamPath.replaceAll("'", ''),
-      'userdata'
-    )
-    await removeNonSteamGame({ steamUserdataDir, gameInfo })
+    await removeNonSteamGame({ gameInfo })
     return res
   }
 
--- a/src/backend/gog/games.ts
+++ b/src/backend/gog/games.ts
@@ -274,6 +274,7 @@ class GOGGame extends Game {
       })
       await setup(this.appName, installedData)
     }
+    this.addShortcuts()
     return { status: 'done' }
   }
 
--- a/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
+++ b/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
@@ -18,11 +18,17 @@ import {
   generateShortAppId,
   removeImagesFromSteam
 } from './steamhelper'
-import { app } from 'electron'
+import { app, BrowserWindow } from 'electron'
 import { isFlatpak, isWindows, tsStore } from '../../constants'
 import { logError, logInfo, LogPrefix, logWarning } from '../../logger/logger'
 import i18next from 'i18next'
 import { showErrorBoxModalAuto } from '../../dialog/dialog'
+import { GlobalConfig } from '../../../backend/config'
+
+const getSteamUserdataDir = async () => {
+  const { defaultSteamPath } = await GlobalConfig.get().getSettings()
+  return join(defaultSteamPath.replaceAll("'", ''), 'userdata')
+}
 
 /**
  * Opens a error dialog in frontend with the error message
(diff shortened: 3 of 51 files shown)


src/backend/api/menu.ts
```diff
@@ -22,13 +22,8 @@ export const installEosOverlay = async () =>
...
   ipcRenderer.invoke('removeFromSteam', appName, runner)
-export const addToSteam = async (
-  appName: string,
-  runner: Runner,
-  bkgDataURL: string,
-  bigPicDataURL: string
-) =>
-  ipcRenderer.invoke('addToSteam', appName, runner, bkgDataURL, bigPicDataURL)
+export const addToSteam = async (appName: string, runner: Runner) =>
+  ipcRenderer.invoke('addToSteam', appName, runner)
```
arielj, line 26: removed the need to pass the images from the frontend, now the backend executes some javascript in the main window instead when needed

src/backend/gog/games.ts
```diff
@@ -630,12 +630,7 @@ class GOGGame extends Game {
     await removeShortcuts(this.appName, 'gog')
     syncStore.delete(this.appName)
     const gameInfo = await this.getGameInfo()
-    const { defaultSteamPath } = await GlobalConfig.get().getSettings()
-    const steamUserdataDir = join(
-      defaultSteamPath.replaceAll("'", ''),
-      'userdata'
-    )
-    await removeNonSteamGame({ steamUserdataDir, gameInfo })
+    await removeNonSteamGame({ gameInfo })
```
arielj, line 633: moved this steamUserdataDir inside the ..NonSteam.. methods to simplify calling them

src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
```diff
@@ -173,17 +179,36 @@ function checkIfAlreadyAdded(object: Partial<ShortcutObject>, title: string) {
...
  * @returns boolean
  */
 async function addNonSteamGame(props: {
-  steamUserdataDir: string
   gameInfo: GameInfo
-  bkgDataUrl: string
-  bigPicDataUrl: string
+  steamUserdataDir?: string
+  bkgDataUrl?: string
+  bigPicDataUrl?: string
```
arielj, line 192: made these 3 props optional so we can pass them during tests but we don't need to pass those props outside of tests

src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
```diff
@@ -173,17 +179,36 @@ function checkIfAlreadyAdded(object: Partial<ShortcutObject>, title: string) {
...
+    props.steamUserdataDir || (await getSteamUserdataDir())
+
+  let bkgDataUrl = props.bkgDataUrl
+  let bigPicDataUrl = props.bigPicDataUrl
+
+  if (bkgDataUrl === undefined || bigPicDataUrl === undefined) {
+    const window = BrowserWindow.getAllWindows()[0]
+
+    bkgDataUrl = await window.webContents.executeJavaScript(
+      `window.imageData("${props.gameInfo.art_cover}", 1920, 620)`
```
arielj, line 204: I made the `imageData` method global in the window object so we can ask the frontend to generate the image on demand instead of passing them with the click event. This is needed since the after install has no context of a user click to receive these images.

src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
```diff
@@ -259,8 +284,8 @@ async function addNonSteamGame(props: {
         otherGridAppID: generateShortAppId(newEntry.Exe, newEntry.AppName)
       },
       gameInfo: props.gameInfo,
-      bkgDataUrl: props.bkgDataUrl,
-      bigPicDataUrl: props.bigPicDataUrl
+      bkgDataUrl: bkgDataUrl!,
+      bigPicDataUrl: bigPicDataUrl!
```
arielj, line 288: because of the `bkgDataUrl === undefined || bigPicDataUrl === undefined`, typescripts can't infer these should be strings at this point so I needed the `!`

CommandMC: Not quite. This is caused by the return type of `window.webContents.executeJavaScript` being `Promise<any>`. So at this point, those two variables are `string | undefined` (not quite sure why it isn't `string | any`, I suspect some Type Narrowing is happening there)

Adding `as string` after the calls to `executeJavaScript` might make it more obvious what's happening.
It might be a good idea to make a helper function for this that has the return type `Promise<string>`, so you don't accidentally miss the `as string` if you ever need to call this function anywhere else

src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
```diff
@@ -18,11 +18,17 @@ import {
   generateShortAppId,
   removeImagesFromSteam
 } from './steamhelper'
-import { app } from 'electron'
+import { app, BrowserWindow } from 'electron'
 import { isFlatpak, isWindows, tsStore } from '../../constants'
 import { logError, logInfo, LogPrefix, logWarning } from '../../logger/logger'
 import i18next from 'i18next'
 import { showErrorBoxModalAuto } from '../../dialog/dialog'
+import { GlobalConfig } from '../../../backend/config'
```
CommandMC, line 26: This relative import doesn't look quite right

src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
arielj: that's right!

src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
arielj: hmmm this is the only way I can get this to work

flavioislima (approved):
Tested here and looks fine 👍🏽

src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
arielj: I'm leaving this like this for now since I can't find any other way to reference this file without an error, it will be easier to fix once the PR changing the import config is merge (then it will just be `backend/config`)

PR #1907, opened 2022-10-15
Remove `draggable` from `Run EXE on Prefix` button

arielj:
This PR removes the `draggable` attribute of the button. The button supports dragging and dropping a file ON the button, but the button itself does not need to be draggable.

I think this is causing issues when clicking that button on the Steam Deck (it might be trying to trigger the `drag` instead of clicking it). I'm not sure how to test this directly on the device though but it's the only difference I can see comparing this button with the other buttons that open a file picker dialog.

Click and dropping files on the button still work fine after this change.

The `return`s were not needed

---

### DIFF
--- a/src/frontend/screens/Settings/components/Tools/index.tsx
+++ b/src/frontend/screens/Settings/components/Tools/index.tsx
@@ -61,9 +61,8 @@ export default function Tools() {
     })
     if (path) {
       exe = path
-      return callTools('runExe', exe)
+      callTools('runExe', exe)
     }
-    return
   }
 
   function dropHandler(ev: React.DragEvent<HTMLSpanElement>) {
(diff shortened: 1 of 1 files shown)


CommandMC (approved):
According to the [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable), just adding `draggable` without setting it to a value will not actually do anything, so I doubt this will change anything at all, but it's still good to remove an unneeded property

arielj:
> According to the [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable), just adding `draggable` without setting it to a value will not actually do anything, so I doubt this will change anything at all, but it's still good to remove an unneeded property

The generated HTML was `draggable="true"` and it was actually draggable. I imagine JSX was doing something fishy there.

flavioislima:
I think we should remove this feature and just add a new button like the others. This dragable feature is not really used and breaks the consistency of the settings page imo.

arielj:
> I think we should remove this feature and just add a new button like the others. This dragable feature is not really used and breaks the consistency of the settings page imo.

It also works as a button though (it's just broken on the steam deck for me, but works fine with a click on desktop).

arielj:
I'm merging this, I think moving this button to be a different thing would be better done in another PR since it will have to change design and functionality.

Issue #1906, opened 2022-10-15
Disable the `Winetricks` button if winetricks is not available


arielj:
### Problem description

Currently, if Winetricks is not available, you can click the button in the game settings, the modal to show the log/progress shows up but nothing happens and there's no way to close the modal (the only way to close it is to close heroic itself)

EDIT: looks like my issue was that the Wine version configured for a given game was deleted, so maybe we need to check that instead? I understand that winetricks should always be available because we download it, but it fails to start if the wine configuration is incorrect

### Feature description

If winetricks is not available, the WINETRICKS button should be disabled with a message stating that we couldn't find winetricks.

EDIT: since the issue in my case was actually the select wine being deleted, maybe the feature would be to disable the button and warn the user that the wine version configured for the game could not be found

arielj:
Closing this, the associated PR addresses this problem, the button is not disabled but now it shows an error if used without a valid wine

PR #1912, opened 2022-10-15
Show an error if the select wine doesn't exist

arielj:
This PR adds an error dialog when a user tries to do an action that requires Wine but the binary doesn't exist.

This is really common when deleting wine versions without updating the wine version of the game for example, and trying to run the game/winetricks/winecfg/exe on prefix were failing silently in the frontend or confusingly (winetricks dialog stuck doing nothing).

This would address https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1906 and also show more info for the user to quickly address the problem without having to check the logs.

image

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -498,15 +498,12 @@ async function runWineCommand(
   forceRunInPrefixVerb = false
 ) {
   const gameSettings = await game.getSettings()
+  const { folder_name: installFolderName } = game.getGameInfo()
   const { wineVersion } = gameSettings
 
   if (!(await validWine(wineVersion))) {
     return { stdout: '', stderr: '' }
   }
-
-  const wineBin = wineVersion.bin.replaceAll("'", '')
-
-  const { folder_name: installFolderName } = game.getGameInfo()
 
   const env_vars = {
     ...process.env,
@@ -537,6 +534,7 @@ async function runWineCommand(
     }
   }
 
+  const wineBin = wineVersion.bin.replaceAll("'", '')
   let finalCommand = `"${wineBin}" ${command}`
   if (additional_command) {
     finalCommand += ` && ${additional_command}`
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -388,7 +388,7 @@ function setupWrappers(
 
 /**
  * Checks if the game's selected Wine version exists
- * @param game The game to check the Wine version of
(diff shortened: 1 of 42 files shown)


src/backend/legendary/games.ts
```diff
@@ -614,10 +614,12 @@ class LegendaryGame extends Game {
           this.logFileLocation,
           `Launch aborted: ${wineLaunchPrepFailReason}`
         )
-        showErrorBoxModalAuto({
-          title: t('box.error.launchAborted', 'Launch aborted'),
-          error: wineLaunchPrepFailReason!
-        })
+        if (wineLaunchPrepFailReason) {
```
arielj, line 617: Added this check instead of `!` since this param can actually be undefined in many cases and it was throwing an error in the frontend and killing the react app

src/backend/launcher.ts
```diff
@@ -394,6 +386,45 @@ function setupWrappers(
   return wrappers.filter((n) => n)
 }
 
+/**
+ * Checks if the game's selected Wine version exists
+ * @param game The game to check the Wine version of
```
CommandMC, line 391: Shouldn't this be `wineVersion` instead of `game`?

src/backend/legendary/games.ts
CommandMC: Looks like *someone* (me) didn't check this properly. Sorry!

Having *some* kind of message, even if no failure reason was given, might still be a good idea

arielj: In this case I'm already displaying an error dialog before this, I think it's fine to make this optional or it would try to display 2 errors messages (one for the missing wine files and then one here).

src/backend/launcher.ts
arielj: yep!

arielj:
```suggestion
 * @param wineVersion an object of type WineInstallation with binary path and name to check
```

arielj:
```suggestion
 * @param wineVersion an object of type WineInstallation with binary path and name to check
```

src/backend/legendary/games.ts
CommandMC: Ah right, now I remember how I originally planned this function... Either you show an error message yourself (that can be more specific/include buttons and so on), or you set `failureReason` and get the generic "Launch aborted" message

CommandMC (approved):

PR #1919, opened 2022-10-17
[Fix] Pass `--offline` correctly again

CommandMC:
*Someone* (me) mistakenly applied the spread operator (...) to `offlineFlag` when passing it to Legendary. I'd say that instead of removing it there again, let's put the variable in line with all the other flags being `string[]` (even if it's just one string)

---


arielj (approved):
Looks good

PR #1921, opened 2022-10-17
[Linux] Remove Resizable BAR toggle

CommandMC:
This was already on by default in VKD3D-Proton for over a year (so toggling the option did nothing at all)

---


arielj (approved):
PRs that only remove code are so satisfying haha

flavioislima (approved):
Nice!
I wonder if we still need FSR options as well. It seems now it is on by default as well for Wine-GE and Proton. 🤔

Etaash-mathamsetty:
> Nice! I wonder if we still need FSR options as well. It seems now it is on by default as well for Wine-GE and Proton. thinking

It can still be disabled using the same env variable (as far as I can remember), so I guess it's only useful to disable it now

PR #1925, opened 2022-10-20
[Linux] Display a warning if "Use dedicated graphics card" is enabled with just one GPU

CommandMC:
This PR adds a warning message to the "Use dedicated graphics card" setting. We've had a lot of users in the past toggling this option with only one GPU in their system, thinking that they had to toggle it. Hopefully this message clears up that confusion

In the process, I've also made `useSetting` a little smarter. Types of settings are now inferred from the type the property has on the `AppSettings` interface, instead of having to specify it in the call. During this, I've found a few inconsistencies, and hopefully corrected them. A 2nd look would be appreciated there

---


src/frontend/screens/Settings/components/OtherSettings/UseDGPU.tsx
```diff
@@ -6,22 +6,41 @@ import ContextProvider from 'frontend/state/ContextProvider'
...
+    if (!useDGPU && numOfGpus === 1) {
+      showDialogModal({
+        title: t('setting.primerun.confirmation.title', 'Only 1 GPU detected'),
+        message: t(
+          'setting.primerun.confirmation.message',
+          'Only one graphics card was detected in this system. ' +
+            'Please note that this option is intended for multi-GPU systems with headless GPUs (like laptops). ' +
+            'On single-GPU systems, the GPU is automatically used & enabling this option can cause issues'
+        ),
+        buttons: [{ text: t('box.ok', 'Ok') }],
```
arielj, line 31: Maybe we can change this `Ok` button to have to buttons: Confirm and Cancel, and only set the state if `Confirm` is clicked?

I think it's more user friendly to warn the user and make them confirm the change instead of applying the change and if they realize it's a bad idea then they have to uncheck it. User's may not realize that it actually changed it it was some mistake or other checkboxes are checked around it.

arielj (commented):
The useSetting typing change is so good, I hated to pass the type for each of them haha. (yet another refactor for the future, remove the `default` argument too and keep defaults somewhere else all in one place)

I left a comment with a question/suggestion.

We have to remember to update the i18n files before merging this (I see changes removing the resizablebar but not adding the new strings)

CommandMC:
Hm, something's not quite right with the `i18next` command. Even after running it, this new key does not appear in the translation files

arielj (approved):
Looks good to me!

arielj:
> Hm, something's not quite right with the `i18next` command. Even after running it, this new key does not appear in the translation files

oh I see what's the issue, we already have the key `settings.primerun` so nothing gets added, maybe change the keys to `settings.primerun_confirmation.title` and `.message` (with a `_`), that should fix it

CommandMC:
I think that (renaming `setting.primerun` to `setting.primerun.description`) should have a similar effect. Thanks for the tip!

flavioislima (approved):
LGTM 👍🏽

PR #1932, opened 2022-10-22
[Steam Deck] Restart xdg-desktop-portal when Heroic is launched

CommandMC:
In SteamOS, the `xdg-desktop-portal` service is misconfigured, preventing URLs from opening unless it's restarted.  
Some Epic games need external account access (which'll open up a web page for the user to grant it) and will fail silently if it's not granted

This PR restarts the service every time Heroic is launched, which I admit is a pretty crude solution, but it *is* a solution
As with #1736, to use `flatpak-spawn --host`, it's necessary to add `--talk-name=org.freedesktop.Flatpak` when building the app

Note that I can't test this myself, since I don't own a deck  

---

### DIFF
--- a/flatpak/local.heroic.yml
+++ b/flatpak/local.heroic.yml
@@ -21,6 +21,7 @@ finish-args:
   - --allow=devel
   - --device=all
   - --talk-name=org.freedesktop.Notifications
+  - --talk-name=org.freedesktop.Flatpak
   - --persist=.wine
   - --filesystem=host
   - --env=LD_LIBRARY_PATH=/app/lib:/app/lib32
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -74,7 +74,8 @@ import {
   getFirstExistingParentPath,
   getLatestReleases,
   notify,
-  quoteIfNecessary
+  quoteIfNecessary,
+  isSteamDeck
 } from './utils'
 import {
   configStore,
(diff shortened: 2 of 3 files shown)


flavioislima:
I know it is just a user service that we are restarting but should heroic do that? Or should we ask steam to fix this issue since it affects all apps?
Because I can reproduce the issue on all other apps that I have on the deck.
I'm not sure if this could cause other side effects on the system now or in the future.

Nocccer:
> I know it is just a user service that we are restarting but should heroic do that? Or should we ask steam to fix this issue since it affects all apps?
> Because I can reproduce the issue on all other apps that I have on the deck.
> I'm not sure if this could cause other side effects on the system now or in the future. 

There is already a ticket open for this issue on the steam side. Don't know if this is resolved yet

CommandMC:
This is absolutely a temporary thing and yes, it *should* be fixed by Valve (see the issue I linked in the comment above this). But before an official fix is rolled out, this fixes the issue for now

I cannot comment on side-effects myself, but there are a lot of people in our Discord with issues that are fixed by running this command (with no known side-effects there)

arielj:
I think it's better to not include this since we don't know if there are side effects of doing this always. Also, we would be doing something that touches the system's services without informing the user about it, I wouldn't want an app restarting things on my system without me knowing it.

I think it's safer to allow people to run this only on demand if they have issues.

CommandMC:
Fair enough.
I've modified our pinned messages in the `steam-deck` channel on Discord now, advising people to run this command if they're experiencing any sort of issue (sadly, as we saw, the issues caused by this can be rather diverse). Hopefully that's enough until Valve manage to fix it

Issue #1930, opened 2022-10-22
Uninstall dialog adds an empty space in grid


arielj:
### Describe the bug

With the new HTML dialogs, when uninstalling a game, a dialog to confirm the uninstall appears and it leaves an empty space in the grid:
![image](https://user-images.githubusercontent.com/188464/197317130-f2a12c9b-3977-4ebc-adb2-f108bdc22612.png)

### Add logs

Nothing relevant

### Steps to reproduce

0. Use beta branch
1. Make sure Grid view is selected
2. Right click > Uninstall on a game card

### Expected behavior

Dialogs should be added somewhere that doesn't affect the layout

### System Information

- Heroic Version [e. g. 2.1.1]: beta branch

Issue #1931, opened 2022-10-22
Dialogs are closed when clicking the `buttons` area


arielj:
### Describe the bug

When clicking on the empty space at the left of the dialog's buttons, the dialog gets closed. The area marked in red here:
![image](https://user-images.githubusercontent.com/188464/197318228-f187d8ee-2da9-402a-8b87-681de02e8d96.png)

This shouldn't happen since the user is clicking inside the dialog.

### Add logs

Nothing relevant

### Steps to reproduce

1. Do any action that triggers a Dialog component
2. Click the empty space at the left of the buttons
3. The dialog is closed

### Expected behavior

The dialog should not be closed clicking that area.

### System Information

- Heroic Version [e. g. 2.1.1]: happens in main and in beta

wbrtm:
Looks like this issue can be closed.

PR #1933, opened 2022-10-22
Some dialog fixes

arielj:
This PR fixes 3 issues with dialogs:
- Uninstall dialog was adding an empty space in grid view in library Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1930
- Clicking at the left empty are of the buttons was closing the dialog https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1931
- Selecting text in some cases was closing the dialog (no issue for this I just noticed it while testing the other fixes)

A comment on the last fix, the bug is present only when selecting text from different HTML elements, for example selecting text from the title and the content. That triggered a `click` event with the parent dialog as the target (since it can trigger an event with the 2 targets). To reproduce this issue before this fix, select text like this for example:
image

---

### DIFF
--- a/src/frontend/components/UI/Dialog/components/Dialog.tsx
+++ b/src/frontend/components/UI/Dialog/components/Dialog.tsx
@@ -40,7 +40,16 @@ export const Dialog: React.FC<DialogProps> = ({
   const onDialogClick = useCallback(
     (e: SyntheticEvent) => {
       if (e.target === dialogRef.current) {
-        onClose()
+        const ev = e.nativeEvent as MouseEvent
+        const tg = e.target as HTMLElement
+        if (
+          ev.offsetX < 0 ||
+          ev.offsetX > tg.offsetWidth ||
+          ev.offsetY < 0 ||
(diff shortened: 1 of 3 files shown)


src/frontend/components/UI/Dialog/components/Dialog.tsx
```diff
@@ -40,7 +40,16 @@ export const Dialog: React.FC<DialogProps> = ({
...
+        const ev = e.nativeEvent as MouseEvent
+        const tg = e.target as HTMLElement
+        if (
+          ev.offsetX < 0 ||
+          ev.offsetX > tg.offsetWidth ||
+          ev.offsetY < 0 ||
+          ev.offsetY > tg.offsetHeight
+        ) {
+          onClose()
+        }
```
arielj, line 52: we want to close this only if the click is outside of the content, this fixes the third issue in the PR description

flavioislima (approved):
LGTM 👍🏽 
that is for sure a pretty annoying thing.

Issue #1319, opened 2022-05-16
Ability to remove games from the list of recently played


arielj:
### Problem description

Recently played games are handled as first in => first out. The problem is that, if you stop playing a game and you don't want it in the list of recently played, the only way to get it out is to start other games until that game is removed from the list.

### Feature description

I imagine an option in the context menu only when right clicking an element in the Recently Played list that allows the user to remove the game from that list.

### Alternatives

Edit the json file that stores this information manually.

mattpnolan:
+1 - Found this ticket via Google when looking for this option in Heroic

PR #1937, opened 2022-10-23
[Feature] Allow removing elements from recently played

arielj:
This PR adds the feature to allow users to remove elements from the Recently Played area manually. Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1319

It adds a few more things:
- now we don't limit the list the store of the recently played list, we limit the list when displaying it, it allows to fill the list with more elements when removing one
- refactored the code a bit moving code to a `recent_games` file
- there was a bug where the order of the recently played games was not correct, it was ordered by the libraries order (epic games before gog, and also ordered within the same library). now the order is the order of the list

There's one issue that this PR doesn't fix (I'll add a report for it): the list of games in the tray icon menu doesn't get updated on the fly, it needs a restart of Heroic. Fixing that with the current code needs a lot of hacky coupled code between main, the context menu handler and the recently played list, if we implement an event system that's going to be a lot easier and cleaner to do.

---


src/frontend/screens/Library/components/RecentlyPlayed/index.tsx
```diff
@@ -1,44 +1,65 @@
...
+
+  const games: GameInfo[] = []
+
+  for (const recent of recentGames) {
+    const found = libraries.find(
+      (game: GameInfo) => game.app_name === recent.appName
+    )
+    if (found) {
+      games.push(found)
+      if (games.length === limit) break
```
arielj, line 25: since the list of recent games can grow a lot, stop the loop once we have enough games to show

flavioislima:
> There's one issue that this PR doesn't fix (I'll add a report for it): the list of games in the tray icon menu doesn't get updated on the fly, it needs a restart of Heroic. Fixing that with the current code needs a lot of hacky coupled code between main, the context menu handler and the recently played list, if we implement an event system that's going to be a lot easier and cleaner to do.

Maybe adding a listener to the `recent_played` store so we could generate the tray menu again could fix that.
We can reload the tray icon just like we are doing when we change the icon color.

arielj:
> > There's one issue that this PR doesn't fix (I'll add a report for it): the list of games in the tray icon menu doesn't get updated on the fly, it needs a restart of Heroic. Fixing that with the current code needs a lot of hacky coupled code between main, the context menu handler and the recently played list, if we implement an event system that's going to be a lot easier and cleaner to do.
> 
> Maybe adding a listener to the `recent_played` store so we could generate the tray menu again could fix that. We can reload the tray icon just like we are doing when we change the icon color.

You mean watching the .json file for changes? I think it would work yeah, I wonder how performant that is (not sure how listening for file changes work).

What I imagined was that idea of having an events system, so then we can emit the event `recentGamesChanged` in the backend and listen to it somewhere else in another module.

flavioislima:
@arielj about performance I think it is fine. 
We can get feedback from the beta release, if it affects it then we try another solution.

arielj:
> @arielj about performance I think it is fine. We can get feedback from the beta release, if it affects it then we try another solution.

I just realized when coding this that we store the recently played games in the same file we store other things like favorites, the window size/position, log paths and more. I'll implement the watch for now, but I'll try to improve this after the beta is released

flavioislima (approved):
LGTM.
Tested here and it is working fine.

PR #1944, opened 2022-10-26
[Fix] Duplicate kill listener

BrettCleary:
Removes the duplicate 'kill' handler added in from download manager PR which causes issues with other ipcMain handlers.

Also removed the pre-push test command and added os checks to certain tests that are only written for linux.

---

### DIFF
--- a/src/backend/logger/__tests__/logfile.test.ts
+++ b/src/backend/logger/__tests__/logfile.test.ts
@@ -14,9 +14,17 @@ jest.mock('../logger')
 jest.unmock('../logfile')
 
 let tmpDir = {} as DirResult
-const skipIfOnWin = platform() === 'win32' ? describe.skip : describe
 
-skipIfOnWin('logger/logfile.ts', () => {
+const shouldSkip = platform() === 'win32'
+const skipMessage = 'on windows so skipping test'
+const emptyTest = it('should do nothing', () => {})
+
+describe('logger/logfile.ts', () => {
+  if (shouldSkip) {
+    console.log(skipMessage)
+    emptyTest
+    return
+  }
   beforeEach(() => {
     tmpDir = dirSync({ unsafeCleanup: true })
   })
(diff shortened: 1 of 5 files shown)


src/backend/logger/__tests__/logfile.test.ts
```diff
@@ -13,8 +14,9 @@ jest.mock('../logger')
 jest.unmock('../logfile')
 
 let tmpDir = {} as DirResult
+const skipIfOnWin = platform() === 'win32' ? describe.skip : describe
 
-describe('logger/logfile.ts', () => {
+skipIfOnWin('logger/logfile.ts', () => {
```
arielj, line 19: I wonder if it's better to do:

```
describe(..... () => {
  if (platform() === 'win32') {
    console.log('ignored on windows')
    return
  }

  ....
}
```

maybe it's just me though

arielj (approved):
Looks good to me, just added a comment but it's all good

CommandMC:
The 2nd `kill` method is missing the extra functionality added by #1864. To resolve this, replace
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/1e01cb25693dc36ebcd6f346aaa05d28aa3dc24c/src/backend/main.ts#L1329
with
```ts
runner === 'sideload' ? stop(appName) : getGame(appName, runner).stop()
```
(import `stop` from `./sideload/games`)

Nocccer (approved):

Nocccer (dismissed):
Aggree with @CommandMC. There is something missing from the sideload app. Easy fix.

src/backend/logger/__tests__/logfile.test.ts
BrettCleary: yeah this will make it easier to find the tests that are being skipped. I'll add this

flavioislima (approved):
Looks good. Tested here and now Login is working again.
Thanks for fixing that.

PR #1950, opened 2022-10-28
Fix some issues when uninstalling games

arielj:
This PR fixes (and partially fixes) some issues I've noticed:

- after uninstalling a game (but not a sideloaded app), the library status was not updated properly (it was a new but, no the one that was fixed haha)
- in slow systems, since you confirm uninstalling a game until it actually gets uninstalled and the UI is updated you can still click the play button which will try to start the game and make everything act weird

This PR disabled the play buttons when uninstalling and shows more visual feedback with an `uninstalling` status label, both in the game card and in the game page.

There are still issues (you can click uninstall multiple times, and you can click any of the game tools/menu options while it's getting uninstalled, but I'll add a separated issue for that since I think it requires more discussions)

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -99,6 +99,7 @@ export default function GamePage(): JSX.Element | null {
   const isQueued = status === 'queued'
   const isReparing = status === 'repairing'
   const isMoving = status === 'moving'
+  const isUninstalling = status === 'uninstalling'
 
   const backRoute = location.state?.fromDM ? '/download-manager' : '/'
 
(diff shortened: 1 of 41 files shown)


src/frontend/screens/Library/components/GameCard/index.tsx
```diff
@@ -155,6 +164,13 @@ const GameCard = ({
   }
 
   const renderIcon = () => {
+    if (isUninstalling) {
+      return (
+        <button className="svg-button iconDisabled">
+          <svg />
+        </button>
```
arielj, line 171: doing this so I don't have to add an arbitrary height to the icons wrapper

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -427,15 +438,15 @@ export default function GamePage(): JSX.Element | null {
...
-                  </>
+                  <button
+                    disabled={
+                      isReparing || isMoving || isUpdating || isUninstalling
+                    }
+                    onClick={handlePlay()}
+                    className={`button ${getPlayBtnClass()}`}
+                  >
+                    {getPlayLabel()}
+                  </button>
```
arielj, line 449: I'm just adding the `|| isUninstalling` and removing the wrapping `<></>` fragment, but the diff makes it look like more things changed

flavioislima (approved):
LGTM

PR #1953, opened 2022-10-28
Optimize eslint settings

arielj:
Using these settings can reduce the time required by eslint to check the project before pushing. On my case it went from 1 minute to less than 4 seconds (if we don't use the --cache option we still have a good benefit -from 1 minute to less than 10 seconds on my system-).

---


flavioislima (approved):
Looks good

Issue #1911, opened 2022-10-15
Remember scroll position when leaving the Library and restore when back


arielj:
### Problem description

Currently, if you are navigating the Library and open a game page to read more about it, if you go back you are back at the top and you have to scroll back again to where you were.

### Feature description

Would be convenient to remember the scroll position of the Library and restore that scroll position when showing the Library again.

Not sure if it's useful in all cases, maybe do this only when "going back" from other screens, but keep the scroll at the top when clicking the `Library` item in the sidebar?

### Alternatives

Keep the current behavior, it's a minor inconvenience only.

flavioislima:
done

PR #1954, opened 2022-10-29
Scrolls to previous location when opening library

Snoodelz:
Based on this issue: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1911


src/frontend/screens/Library/index.tsx
```diff
@@ -82,6 +82,24 @@ export default function Library(): JSX.Element {
...
+        listing.scrollTo(0, scrollPosition)
+      }
+      listing.addEventListener('scroll', () => {
+        scrollPosition = listing.scrollTop
+      })
+    }
+    return () => {
+      storage.setItem('scrollPosition', scrollPosition.toString())
+    }
+  }, [])
```
arielj, line 101: maybe it's better to use a `useRef` hook to avoid the `document.querySelector` (that is less `react-ish`), so if for some reason in the future the class changes this will still work

then we can add the variable holding that hook as a dependency in the dependencies array

src/frontend/screens/Library/index.tsx
```diff
@@ -82,6 +82,24 @@ export default function Library(): JSX.Element {
   const { t } = useTranslation()
   const backToTopElement = useRef(null)
 
+  //Remember scroll position
+  useEffect(() => {
```
arielj, line 86: let's use `useLayoutEffect` here instead to avoid the `jump` from the top to the current scroll

Snoodelz: Ah! Cheers didn't know that hook existed :)

flavioislima (approved):
Looks pretty good, tested it here and it is working nicely.
Thanks for that. ⚔️

PR #1955, opened 2022-10-29
Use yarn start instead of yarn dev

wbrtm:
Just an update to the setup documentation.


flavioislima (approved):
Thanks, I forgot to update that when I changed in the code.

Issue #1949, opened 2022-10-27
Login into a store changes the filter of the Library to only that store


arielj:
### Describe the bug

If I'm installing Heroic for the first time and want to login to both epic and gog accounts, the library will be filtered by the last store I log in and it can be confusing for users that miss the filters as the top.

### Add logs

Nothing relevant

### Steps to reproduce

1. Log in with Epic
2. Log in with GOG
3. Go to Library
4. Library is showing only GOG games

### Expected behavior

After login into a new store, the Library should display what it was showing before and the new store, not only the new store.

Currently, this means changing to `All` but in the future, if we implement this before https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1881, it should keep the same filtering and turn on the store we just logged into.

### System Information

- Heroic Version [e. g. 2.1.1]: beta (should happen in main and 2.4.3 too)

PR #1956, opened 2022-10-29
[UI/UX] Pre-release Fixes

flavioislima:
Just some Small Fixes:

- [x] Show the store filter only for logged-in stores
- [x] Don't show Platform filter on windows
- [x] Invert the Grid/List icon
- [x] Trigger refresh library when clicking on Got to Library and on the Sidebar Library button, since after being logged in the library was not refreshing properly and it was showing an empty list. This also makes it easier to show new games after getting then on the stores, since when clicking the library button the new game will appear right away.
- [x] Fix filters on the header not proper aligned.
- [x] Removed `gawk` from the list of dependencies since it is not needed for a long time now.
- [x] Added a request to the Bottle SteamGrid API to get the image for the sideloaded game. Mirko said I could use it for that.
- [x] Fix current download not showing in the sidebar after restarting heroic.
- [x] Fix BE error message when no log was found for a game.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ You can alternatively download the file ending in .rpm from the [latest release]
 
 #### Other Distributions (AppImage and TAR.XZ)
 
-Since these two distribution formats don't have a form of dependency management, make sure the `curl` commands is available. You might run into weird issues if it's not.
+Since these two distribution formats don't have a form of dependency management, make sure the `curl` command is available. You might run into weird issues if it's not.
 
 ##### AppImage
 
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ You can alternatively download the file ending in .rpm from the [latest release]
 
 #### Other Distributions (AppImage and TAR.XZ)
 
-Since these two distribution formats don't have a form of dependency management, make sure the `curl` and `gawk` commands are available. You might run into weird issues if they aren't.
+Since these two distribution formats don't have a form of dependency management, make sure the `curl` commands is available. You might run into weird issues if not.
 
 ##### AppImage
 
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ You can alternatively download the file ending in .rpm from the [latest release]
 
 #### Other Distributions (AppImage and TAR.XZ)
 
-Since these two distribution formats don't have a form of dependency management, make sure the `curl` commands is available. You might run into weird issues if not.
+Since these two distribution formats don't have a form of dependency management, make sure the `curl` commands is available. You might run into weird issues if it's not.
 
 ##### AppImage
 
--- a/src/frontend/components/UI/PlatformFilter/index.tsx
+++ b/src/frontend/components/UI/PlatformFilter/index.tsx
@@ -14,7 +14,12 @@ export default function PlatformFilter() {
 
   const isMac = platform === 'darwin'
   const isLinux = platform === 'linux'
+  const isWindows = platform === 'win32'
   const disabledIcon = isLinux && category === 'legendary'
+
+  if (isWindows) {
+    return null
+  }
 
   return (
     <div className="platformFilters">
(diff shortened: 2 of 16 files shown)


src/frontend/screens/WebView/index.tsx
```diff
@@ -104,7 +104,7 @@ export default function WebView() {
 
             setTimeout(() => {
               webview.findInPage('authorizationCode')
-            }, 500)
+            }, 50)
```
flavioislima, line 107: Lowered the timeout here on purpose since it was not necessary to wait that much. Tested 4 or 5 times and the login is working fine.

src/frontend/state/GlobalState.tsx
```diff
@@ -295,7 +295,7 @@ export class GlobalState extends PureComponent<Props> {
   }
 
   handleSuccessfulLogin = (runner: Runner) => {
-    this.handleCategory(runner)
+    this.handleCategory('all')
```
flavioislima, line 298: Setting the category on `all` instead of runner works better imo.

README.md
```diff
@@ -171,7 +171,7 @@ You can alternatively download the file ending in .rpm from the [latest release]
 
 #### Other Distributions (AppImage and TAR.XZ)
 
-Since these two distribution formats don't have a form of dependency management, make sure the `curl` and `gawk` commands are available. You might run into weird issues if they aren't.
+Since these two distribution formats don't have a form of dependency management, make sure the `curl` commands is available. You might run into weird issues if not.
```
CommandMC, line 174:
```suggestion
Since these two distribution formats don't have a form of dependency management, make sure the `curl` commands is available. You might run into weird issues if it's not.
```

src/frontend/components/UI/PlatformFilter/index.tsx
```diff
@@ -14,8 +14,13 @@ export default function PlatformFilter() {
 
   const isMac = platform === 'darwin'
   const isLinux = platform === 'linux'
+  const isWindows = platform === 'win32'
   const disabledIcon = isLinux && category === 'legendary'
 
+  if (isWindows) {
+    return null
```
CommandMC, line 21: I'm no frontend guy, but wouldn't
```suggestion
    return <></>
```
be better here, since the function is expected to return JSX elements? Or is `null` the same as `<></>`?

src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
```diff
@@ -102,6 +102,22 @@ export default function SideloadDialog({
     setWine()
   }, [title])
 
+  useEffect(() => {
+    setTimeout(async () => {
+      try {
+        const res = await fetch(
+          `https://steamgrid.usebottles.com/api/search/${title}`
```
CommandMC, line 109: Adding a link to the documentation of this API might be helpful for the future

CommandMC (changes requested):

src/frontend/components/UI/PlatformFilter/index.tsx
flavioislima: `null` is better since it won't draw anything. The React fragments return an empty span element.

src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
flavioislima: Yes, Mirko just sent a message to me on Discord saying that we could use it. Although I think we should host our own in the future.

src/frontend/state/GlobalState.tsx
arielj: agree, this would fix https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1949

README.md
arielj:
```suggestion
Since these two distribution formats don't have a form of dependency management, make sure the `curl` command is available. You might run into weird issues if not.
```

src/frontend/components/UI/PlatformFilter/index.tsx
CommandMC: I see
TODO for the future would then be to do this elsewhere as well, I believe we have a few cases in the settings where we draw `<></>` for Linux-specific settings when on Windows

CommandMC (approved):

PR #1958, opened 2022-10-30
[MacOS/Linux] Fix up runWineCommand

CommandMC:
This brings runWineCommand closer to how callRunner works for example. The same `commandParts` idiom is used to make it possible to exactly specify what the arguments are that should be passed to `wine`  
In practice, this now means that commands like `winepath` or `reg` can be ran successfully again

In the process, I've also made the `wait` variable actually work again (with spawn, we're no longer using a shell, so we have to put in a little more work than adding `&& <extra_command>` to run another command after the first one finishes)

---

### DIFF
--- a/src/backend/api/helpers.ts
+++ b/src/backend/api/helpers.ts
@@ -43,15 +43,14 @@ export const getInstallInfo = async (
   runner: Runner,
   installPlatform?: InstallPlatform | string
 ) => ipcRenderer.invoke('getInstallInfo', appName, runner, installPlatform)
-interface runWineCommand {
-  appName: string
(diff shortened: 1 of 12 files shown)


CommandMC:
Note that this implementation is still somewhat crude; this is very much a band-aid fix. Ideally we'd add (as I've discussed a few times before) a wrapper around `spawn`, but that can come after the beta release since it's quite a large code change

flavioislima:
Codewise looks good, will do some testing on Linux.

flavioislima (approved):
Looks good, tested here and it seems to be working just fine

Issue #1908, opened 2022-10-15
Context menu focuses 2 items when using a controller


arielj:
### Describe the bug

When using a controller to navigate Heroic, the context menu of a game card highlights the first item with a bright color (the mouse hover color) and the actual focused element by the controller with a dim color, making it hard to see and confusing:

![image](https://user-images.githubusercontent.com/188464/196003809-6412c25f-9bf4-4086-a4ee-68e91ba42b99.png)
`Play Now` is not the one focused
`Uninstall` is the one focused

Moving the mouse removes the `hover` color, even though the mouse is not hovering that item anyway.

### Add logs

nothing relevant on logs

### Steps to reproduce

1. Go to the Library using a controller
2. Open the context menu for any game card (X on an Xbox controller)
3. Move the stick/pad down to focus different items
4. The colors issue shows up
5. Move the mouse
6. The colors issue is gone

### Expected behavior

The controller-focused element should be bright like the color used for the mouse hover. The first element should not stay highlighted when changing items.

### Screenshots

Added above

### System Information

- OS [e. g. "Ubuntu"]: happens on any system
- Heroic Version [e. g. 2.1.1]: main branch

PR #1961, opened 2022-10-30
[Fix] context menu selection/hovering

wbrtm:
This is an **incomplete** fix to issue #1908 
It fixes the issue most of the times. It does not always work for games near/at the bottom of the window/screen, where the context menu is rendered on top of the clicked item (GameCard or install button) triggering the hover effect (darker color).

Another idea is to disable the hover effect entirely when using the keyboard/gamepad navigation, not sure how to go about that though.

---


wbrtm:
Latest commit fixes the issue for all game cards. If the offset was only introduced for aethetic reasons we should be fine. It still remains for list items near the bottom of the window/screen.

flavioislima (approved):
Works good now, thanks for fixing that.

arielj (approved):
great!

PR #1968, opened 2022-10-31
[Fix] controller hints for library list view

wbrtm:
Controller hints for the list view were not specified, leading to wrong hints being displayed.
I added the correct hints. Note the different key-bindings between library view and list view which make another case necessary and should probably be unified at some point.

---


wbrtm:
Included in #1971

PR #1969, opened 2022-10-31
[General/Protocol] Fix protocol not launching gog games

imLinguin:
This fixes the issue we discovered on latest 2.5.0.beta.1 release that prevents gog games from launching using protocol.  
Fixed it by checking for underlying value of a object returned by getGameInfo methods, since empty object `{}` is always true in boolean expressions

### DIFF
--- a/src/backend/protocol.ts
+++ b/src/backend/protocol.ts
@@ -1,9 +1,8 @@
 import { BrowserWindow, dialog } from 'electron'
 import { logError, logInfo, LogPrefix } from './logger/logger'
 import i18next from 'i18next'
-import { getGame } from './utils'
+import { getInfo } from './utils'
 import { Runner } from 'common/types'
-import { getAppInfo } from './sideload/games'
 
 export async function handleProtocol(window: BrowserWindow, args: string[]) {
   // Figure out which argv element is our protocol
@@ -36,15 +35,11 @@ export async function handleProtocol(window: BrowserWindow, args: string[]) {
     const runners: Runner[] = ['legendary', 'gog', 'sideload']
 
     const game = runners
-      .map((runner) =>
-        runner === 'sideload'
-          ? getAppInfo(arg)
-          : getGame(arg, runner).getGameInfo()
-      )
+      .map((runner) => getInfo(arg, runner))
       .filter(({ app_name }) => app_name)
       .shift()
 
-    if (!game?.app_name) {
+    if (!game) {
       return logError(`Could not receive game data for ${arg}!`, {
         prefix: LogPrefix.ProtocolHandler
       })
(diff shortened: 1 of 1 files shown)


src/backend/protocol.ts
```diff
@@ -31,11 +31,14 @@ export async function handleProtocol(window: BrowserWindow, args: string[]) {
...
-    const game =
-      getGame(arg, 'legendary').getGameInfo() ||
-      getGame(arg, 'gog').getGameInfo()
+    let game = getGame(arg, 'legendary').getGameInfo()
 
-    if (!game) {
+    // getGameInfo returns {} so we need to check for underlying value
+    if (!game?.app_name) {
+      game = getGame(arg, 'gog').getGameInfo()
+    }
```
CommandMC, line 39: Maybe something like this would be more elegant:
```ts
const allRunners: Runner[] = ['legendary', 'gog', 'sideload']
const game = allRunners
  .map((runner) => getGame(arg, runner).getGameInfo())
  .filter(({ app_name }) => app_name)
  .shift()

Comment (imLinguin) [reply src/backend/protocol.ts:39]: totally forgot sideload is a thing, yeah that would be much more scalable

Comment (imLinguin) [reply src/backend/protocol.ts:39]: However getGame doesn't handle sideload runner

Comment (CommandMC) [reply src/backend/protocol.ts:39]: Well, nothing says that it *can't*

File: src/backend/protocol.ts [hunk 2]
```diff
@@ -31,12 +33,16 @@ export async function handleProtocol(window: BrowserWindow, args: string[]) {
...
+    const game = runners
+      .map((runner) =>
+        runner === 'sideload'
+          ? getAppInfo(arg)
+          : getGame(arg, runner).getGameInfo()
+      )
+      .filter(({ app_name }) => app_name)
+      .shift()
 
     if (!game?.app_name) {
```
CommandMC, line 47: This could now be simplified to just
```suggestion
    if (!game) {
```

flavioislima:
I am not sure why the tests are falling only now but might be related to some changes I did on my Fixes PR before the beta. 
I changed where we check if the logs exist to fix an error on BE when the game has no logs. 
So the current tests for that might not be relevant and maybe could be removed/Adapted. But we would need to investigate on that.

src/backend/protocol.ts
```diff
@@ -31,11 +33,18 @@ export async function handleProtocol(window: BrowserWindow, args: string[]) {
...
   if (command === 'launch') {
-    const game =
-      getGame(arg, 'legendary').getGameInfo() ||
-      getGame(arg, 'gog').getGameInfo()
+    const runners: Runner[] = ['legendary', 'gog', 'sideload']
 
-    if (!game) {
+    const game = runners
+      .map((runner) =>
+        runner === 'sideload'
```
flavioislima, line 40: Instead of doing like this use the method `getInfo`  from `utils`

CommandMC (approved):

imLinguin:
Tested Epic, GOG and sideload all work fine

PR #1971, opened 2022-10-31
Gamepad navigation tweaks

wbrtm:
Several tweaks regarding gamepad controls. The goal is app-wide consistency and better integration with steam deck controls.

- "A"-button on game cards and list items _always_ opens the game info page (list/grid/installed or not), launching/installing directly from library still works with "Y"-button
- Relabeled Controller hints
  - **–** instead of **None**
  - **Select** instead of **Activate**
  - **Options** instead of **Context menu**
- Back button closes the context menu if open
- Added controller hints to library list view
- Autofocus play/install button on game page
- Icons in library grid view (install, play, settings, update) are hidden when gamepad input is active
- More options in game card context menu (play, stop, settings, remove from queue, cancel installation)

---

### DIFF
--- a/public/locales/az/translation.json
+++ b/public/locales/az/translation.json
@@ -152,21 +152,20 @@
     },
     "controller": {
         "hints": {
-            "activate": "Aktivləşdirin",
             "back": "Geri",
             "backspace": "Backspace",
-            "close_context_menu": "Kontekst menyusunu bağlayın",
             "close_dialog": "Dialoqu bağlayın",
             "close_keyboard": "Klaviaturanı bağlayın",
-            "context_menu": "Kontekst menyusu",
+            "close_options": "Close Options",
             "game_details": "Oyun təfərrüatları",
             "game_settings": "Oyun parametrləri",
             "install_game": "Oyun quraşdırın",
             "move_cursor": "Kursoru köçürün",
-            "none": "Heç biri",
             "open_virtual_keyboard": "Virtual klaviaturanı açın",
+            "options": "Options",
             "play_game": "Oyun oynamaq",
             "scroll": "Sürüşdürün",
+            "select": "Select",
             "space": "Kosmos",
             "update_game": "Oyunu yeniləyin"
         }
--- a/public/locales/be/translation.json
+++ b/public/locales/be/translation.json
@@ -152,21 +152,20 @@
     },
     "controller": {
         "hints": {
-            "activate": "Activate",
             "back": "Back",
             "backspace": "Backspace",
-            "close_context_menu": "Close context menu",
             "close_dialog": "Close dialog",
             "close_keyboard": "Close keyboard",
-            "context_menu": "Context menu",
+            "close_options": "Close Options",
             "game_details": "Game details",
             "game_settings": "Game settings",
             "install_game": "Install game",
             "move_cursor": "Move cursor",
-            "none": "None",
             "open_virtual_keyboard": "Open virtual keyboard",
+            "options": "Options",
             "play_game": "Play game",
             "scroll": "Scroll",
+            "select": "Select",
             "space": "Space",
             "update_game": "Update game"
         }
--- a/public/locales/bg/translation.json
+++ b/public/locales/bg/translation.json
@@ -152,21 +152,20 @@
     },
     "controller": {
         "hints": {
-            "activate": "Activate",
             "back": "Back",
             "backspace": "Backspace",
-            "close_context_menu": "Close context menu",
             "close_dialog": "Close dialog",
             "close_keyboard": "Close keyboard",
-            "context_menu": "Context menu",
+            "close_options": "Close Options",
             "game_details": "Game details",
             "game_settings": "Game settings",
             "install_game": "Install game",
             "move_cursor": "Move cursor",
-            "none": "None",
             "open_virtual_keyboard": "Open virtual keyboard",
+            "options": "Options",
             "play_game": "Play game",
             "scroll": "Scroll",
+            "select": "Select",
             "space": "Space",
             "update_game": "Update game"
         }
--- a/public/locales/bs/translation.json
+++ b/public/locales/bs/translation.json
@@ -152,21 +152,20 @@
     },
     "controller": {
         "hints": {
-            "activate": "Aktiviraj",
             "back": "Nazad",
             "backspace": "Backspace",
-            "close_context_menu": "Zatvorite kontekstni meni",
             "close_dialog": "Zatvori dijalog",
             "close_keyboard": "Zatvorite tastaturu",
-            "context_menu": "Kontekstni meni",
+            "close_options": "Close Options",
             "game_details": "Detalji igre",
             "game_settings": "Postavke igre",
             "install_game": "Instaliraj igru",
             "move_cursor": "Pomeri kursor",
-            "none": "Nema",
             "open_virtual_keyboard": "Otvorite virtuelnu tastaturu",
+            "options": "Options",
             "play_game": "Igrati igru",
             "scroll": "Scroll",
+            "select": "Select",
             "space": "Prostor",
             "update_game": "Ažurirajte igru"
         }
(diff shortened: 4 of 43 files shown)


flavioislima:
If you have the time, one good idea is to hide the buttons to play, install, etc on the card when a controller is detected, this will make the navigation smoother since right now it goes through the icons for every card.

You can just hide the element and also would need to enable the settings option on the context menu and the play option as well.

wbrtm:
So you are not a fan of disabling the buttons permanently but only when gamepad is being used?

flavioislima:
> #[e52644c](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1971/commits/e52644c3c74baef0bd9dba79c33951f807224099) does that, give it a try.
> 
> So you are not a fan of disabling the buttons permanently but only when gamepad is being used? Personally, I would remove the icons for good and not just when a gamepad is being used. They make the library view too cluttered and this would result in a consistent and clean look.

Yes but clean look wont make it practical because then we would need to always open or the context menu (that some people dont know about it) or open the game page to access the settings and play the game, etc.

Adding the buttons there actually was a request from several people when Heroic started because we didn't have them at first :)

flavioislima:
I am changing this to ready for review so we can grab the build files to test 👍🏽

flavioislima:
Ok, the navigation without the buttons now looks good.
But we have no way of playing the game without entering the game page first. So adding the option to the context menu (play and settings) would solve that. 👍🏽

wbrtm:
Thanks for the feedback. Context menu options up next then.

arielj:
I would prefer not to lose the ability to open the game's setting or launch the game or trigger the install dialog with a single button.

With this change we don't have the buttons to fake the clicks, but we can still navigate to the settings page with javascript or launch the game sending a message to the backend for example.

wbrtm:
Not sure I understand what you mean. Something like launching/installing with button Y?

flavioislima:
> Not sure I understand what you mean. Something like launching/installing with button Y?

Currently, If you are using a controller you can select the card and use A to launch the game, or Y to open the game page.
On your changes we lost this behavior and we can only play a game by opening the game page.

wbrtm:
Yes, that was my intention. I think using the game page or the context menu to do anything with a game is fine. That's how it is done on the Steam Deck. Switching to heroic you find yourself launching games by pressing A you did not want to.
Do you want me to add launch/install to button Y instead?

flavioislima:
> Yes, that was my intention. I think using the game page or the context menu to do anything with a game is fine. That's how it is done on the Steam Deck. Switching to heroic you find yourself launching games by pressing A you did not want to. Do you want me to add launch/install to button Y instead?

Thats true. on the Deck using A makes the Game Page open.
Cool, let's keep it like the steam does then, so it is more familiar to everyone. :)

arielj:
I would still keep something like Y to launch/install too, not instead. I think it's good to have all the options in the context menu but it's also nice to be able to access those options with less clicks like it is now.

I imagine a combination of:
- making A consistent (always open the game page, like it is in this PR)
- adding missing options in the context menu (like discussed in this PR)
- having shortcuts with the gamepad to trigger actions on the specific game with a single button (to not lose the current behavior)

wbrtm:
Cool, looks like everyone can agree on play/install on Y in addition to the current PR.
Not everything has to be done the same as on Steam Deck but we should also not underestimate the amount of users the Deck brings to Heroic. By welcoming them Heroic might quickly become the standard alternative launcher available on the Deck. Steam put alot of effort and thoughts into their deck interface, it is very polished and well designed.

Going to add that and then the PR can use some more testing.

arielj:
Speaking about steam deck, what are your thoughts on using the menu (the 3 lines) button to show the context menu? currently it's X but the steam deck uses that one to show the options

arielj:
> Did not notice the deck uses X but we should probably use the options button as well then. Just to not confuse users of both too much.

to clarify, the deck does not use the X for the context, we do

wbrtm:
Got it. Something to keep in mind for the future if someone is at it. The reduced button usage in heroic works for now, other quality of life improvements like autofocus in library and other screens is more urgent I think.

arielj:
I'm not sure if navigating the stores with a controller is even possible

flavioislima:
> I'm not sure if navigating the stores with a controller is even possible

perhaps if we find a way of at least accepting the touch input like Steam does with external URLs on the deck.
But yes, don't know if it is possible or not.

arielj:
> > I'm not sure if navigating the stores with a controller is even possible
> 
> perhaps if we find a way of at least accepting the touch input like Steam does with external URLs on the deck. But yes, don't know if it is possible or not.

hmmm not sure I understand the issue, I can use the stores with the touch screen, maybe it's an issue that happens only in gaming mode? I always use the heroic interface in desktop mode

flavioislima:
> > > I'm not sure if navigating the stores with a controller is even possible
> > 
> > 
> > perhaps if we find a way of at least accepting the touch input like Steam does with external URLs on the deck. But yes, don't know if it is possible or not.
> 
> hmmm not sure I understand the issue, I can use the stores with the touch screen, maybe it's an issue that happens only in gaming mode? I always use the heroic interface in desktop mode

I use it in Gaming Mode most of the time and it is not possible to use the touch for that.

wbrtm:
Should open another issue for that.

arielj:
> > > > I'm not sure if navigating the stores with a controller is even possible
> > > 
> > > 
> > > perhaps if we find a way of at least accepting the touch input like Steam does with external URLs on the deck. But yes, don't know if it is possible or not.
> > 
> > 
> > hmmm not sure I understand the issue, I can use the stores with the touch screen, maybe it's an issue that happens only in gaming mode? I always use the heroic interface in desktop mode
> 
> I use it in Gaming Mode most of the time and it is not possible to use the touch for that.

I'll try later, sure. Can you try one thing? when you hold down the STEAM button it should allow the touch controls to move the mouse and act as a click, is that working?

flavioislima:
@arielj tried that already, tried everything to make it work. The only thing that works is clicking the scrollbar using the touchscreen or the Steam button + R2.
On desktop the touch works fine, that's true.

flavioislima:
@wbrtm testing it again here and one thing I noticed, not sure if it was introduced on this PR or on a previous one, is that after you get on a game page and then get back to library, the controller do not work anymore because heroic loses focus so you need to click on the window to activate the controller again.

wbrtm:
@flavioislima Yeah that was quite annoying while testing but I thought that was not introduced by my changes, didn't check though. Does that also occur on the deck?

flavioislima:
> @flavioislima Yeah that was quite annoying while testing but I thought that was not introduced by my changes, didn't check though. Does that also occur on the deck?

found the issue, was an `if` statement that was preventing the joystick to work. It was introduced several weeks ago in a refactor. Now it is working fine. If you have the time take a look at my PR with this and other fixes #1979

public/locales/fr/gamepage.json
```diff
@@ -33,7 +33,7 @@
         },
         "uninstall": {
             "checkbox": "Remove prefix: {{prefix}}{{newLine}}Note: This can't be undone and will also remove not backed up save files.",
-            "message": "Voulez-vous désinstaller ce jeu ?",
+            "message": "Voulez-vous désinstaller ce jeu\u202f?",
```
flavioislima, line 36: remove these characters otherwise they will break the translations.

wbrtm: Will do. I have no idea how this happened.

wbrtm: Seems resolved after rebase, please confirm.

arielj (approved):
Looks good to me!

arielj:
> @arielj tried that already, tried everything to make it work. The only thing that works is clicking the scrollbar using the touchscreen or the Steam button + R2. On desktop the touch works fine, that's true.

Weird, I just tried running heroic in gaming mode and I can navigate the stores just fine. Holding down the Steam button I can move around and click things in the store by either clicking the right touchpad or pressing R2.

flavioislima (approved):
Good stuff!
Thanks for that 👍

PR #1979, opened 2022-11-02
[Fix] 2.5.0 Beta 1 Fixes

flavioislima:

---

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -1,7 +1,7 @@
 import React, { useContext } from 'react'
 
 import './App.css'
-import { HashRouter, Route, Routes } from 'react-router-dom'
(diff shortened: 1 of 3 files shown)


src/backend/launcher.ts
```diff
@@ -537,6 +537,7 @@ async function runWineCommand({
 
   return new Promise<{ stderr: string; stdout: string }>((res) => {
     const wrappers = options?.wrappers || []
+    const { launcherArgs = '' } = settings
```
CommandMC, line 540: This should be done in the launch function for sideloaded games instead. You don't want these options all the time

CommandMC (changes requested):

src/backend/launcher.ts
flavioislima: good point, just fixed it.

PR #1980, opened 2022-11-02
[Fix] offline notification text

wbrtm:
Fixes the current offline notification string resulting in something like
`OfflineRetrying in ... 5 seconds`
to
`Offline. Retrying in 5 seconds.`

Will require adaption by translators before it shows.

---


arielj (commented):
should we also update the translation files? since this is used as reference for more accurate translations

one concern I have is that the leading `. ` could be a bit weird for translations without much context. maybe we have to refactor this a bit to be a complete string `Offline. Retrying in {{seconds}} seconds.` and then use 2 different keys (one for just `Offline` and one for `Offline. Retrying...`)

wbrtm:
Although I saw translation projects where such strings were common and translators seemed to have no problem to adopt, I thought about the same fixing this. I will add another commit, should be the cleaner solution.

flavioislima (approved):

PR #1981, opened 2022-11-02
[Tech] Update React to v18, Vite, TS and Vite Plugins to latest

flavioislima:
Updated React and other packages

---


src/frontend/screens/WineManager/index.tsx
```diff
@@ -49,7 +49,7 @@ export default function WineManager(): JSX.Element | null {
       }
     }
 
-    return refreshWineVersionInfo(true)
+    return () => refreshWineVersionInfo(true)
```
flavioislima, line 52: The only issue I found after the upgrade was this one. React 18 complains that `useEffect `can only return a clean-up function so it was failing. Needed to change this approach to fix the issue.

CommandMC: I doubt that's what was intended here, seems more like
```suggestion
    refreshWineVersionInfo(true)
```
would be correct

CC @Nocccer

flavioislima: Yes, in practice wont make a difference. The thing is that returning a cleanup function like I did makes the method being called at the end of the `useEffect `always.

CommandMC (approved):

Issue #1983, opened 2022-11-02
Settings are missing from general settings until a game setting screen is opened


arielj:
### Describe the bug

When opening Heroic, the `Wine` and `Wine Extensions` subitems of the general settings are not present. They only appear after clicking in a game card or opening the settings for a game.

### Add logs

Nothing relevant.

### Steps to reproduce

1. Open Heroic
2. Click Settings
3. Verify no Wine option is there
4. Click Library
5. Click on the gears icon to open any game's settings or in a game card to open the game page
6. Click Library (to reset the sidebar to show `Settings` instead of `Game Settings`)
7. Click Settings
8. Verify that now `Wine` and `Wine Extensions` are present

### Expected behavior

The options should show up from the beginning.

### Screenshots

https://user-images.githubusercontent.com/188464/199621138-4af364f4-ddcf-4576-999c-49921b567714.mp4

### System Information

- Heroic Version [e. g. 2.1.1]: beta (probably happens in main too)

arielj:
Closing since this options are removed form the sidebar

Issue #1716, opened 2022-08-12
Make it clear when global settings are actually defaults for when installing a game and not really global


arielj:
### Problem description

There are a lot of settings that don't really affect currently installed games even though the name is `Global Settings`, they only work as default values when installing a new game.

### Feature description

I imagine a section called `Default game settings` grouping all these settings. Also the section should have copy at the beginning making it clear that those settings are only the defaults for when you install a game but they have no effect on games that are already installed.

### Alternatives

An alternative that I think is better but more comples is to have a 3 states setting for games settings when we currently display toggles: instead of `on/off` we could have `on / off / global`. That way we could allow a user to enable/disable the Show FPS settings once for all games with the global config instead of going through each game.

Issue #1696, opened 2022-08-10
Settings grouping is not consistent


arielj:
### Describe the bug

A user might be confused on why options are in some places. For example, `Sync with installed Epic Games` is the third option in the General settings but it's not relevant to users that only use GOG, and there's no equivalent option for GOG. Then we have other Epic related settings (like the legendary path and the EOS overlay or the non-https download) in the `Advanced` settings, but there's a gog binary configuration in the middle and the non-https function I think only applies to Epic.

Maybe we need an `Epic` and a `GOG` sections in the settings?

Also, in the `Others` section there's a mix of toggles that affect default game settings when a game is installed (like show FPS, use GameMode, Pulse Audio fix, env variables, and more) and toggles or settings for heroic's config like automatically creating shorcuts, discord rich presence or add as non-steam configuration. Maybe elements should be moved to other screens or group them more consistently.

In the General settings we have the Max number of workers when downloading that probably works better in the Advanced section.

### Add logs

no logs

### Steps to reproduce

Navigate Heroic's settings and check the options in each section and you'll see some things seems to be related but are in unrelated places.

### Expected behavior

It should be easier for a user to see related options close to each other.

### System Information

- Heroic Version [e. g. 2.1.1]: beta branch

PR #1984, opened 2022-11-03
Reorder settings

arielj:
This PR moves a lot of things around reordering the settings.

There are a few issues with the current organization:
- settings order is not consistent (https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1696)
- users don't always know where to look for settings ("Other" is not clear to describe what they'll find inside)
- It's not clear when a setting is a global setting (affects the whole app) or a default for a new game installation (wine setting, dxvk, mongohud, etc) (https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1716)
- there's a bug where sometimes the Wine and Wine Extensions entries don't show up (https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1983)

With this change, I'm aiming to address those with these changes:
- Everything that is a general Heroic configuration is now in the General Settings section or the Advanced section (we might need to move some things around there)
- All the global settings that are actually defaults are grouped in a `Game Defaults` section and the section is divided by subsections (Wine, Wine Extensions, Others). There's also a message at the top of this section indicating that these are defaults and changing them won't affect already installed games
- When changing the settings of a game, most of the options are in a single screen so users don't have to try the different sections
- By removing Wine and Wine Extensions from the sidebar, the bug linked above is removed

I have one more idea to add to this, a search box at the top of each setting section to filter settings (like the way chrome://config works).

Any feedback would be great. I know there might be some concerns about the Game Settings section being too long.

Some extra notes:
- I moved some files around because when moving the settings it was clear the imports would be really messy, now all the individual settings are in Settings/components and each of the sections is in Settings/sections
- I changed the style of the `Run EXE on Prefix` button that was breaking the design a lot in the place it's now.

Video of how it looks and behaves currently:
https://user-images.githubusercontent.com/188464/199638462-58c06790-8c03-4f35-8217-65b45d2a133b.mp4

---


CommandMC (commented):
I quite like these changes.  
Some oddities I've noticed:
- In the Game Defaults, the Wine Version dropdown is too short:
  image
- The spacing of these options looks a bit off:
  image

flavioislima (approved):
Cool, tested here and everything is working fine. Lets get some feedback from the next beta now :D

flavioislima:
> I quite like these changes. Some oddities I've noticed:
> 
> * In the Game Defaults, the Wine Version dropdown is too short:
>   image
> * The spacing of these options looks a bit off:
>   image

sorry man, saw your comments only after merging the PR. You are right about those, I can fix on another PR.

PR #1987, opened 2022-11-03
[Fix] Verify that a valid Wine/Proton prefix exists before running any Wine command

CommandMC:
This will prevent Heroic from running any Wine commands unless all files required in a Wine/Proton prefix are actually present, which will in turn prevent us from creating invalid Proton prefixes (note that this will not fix already broken Proton prefixes)

This is achieved by checking for known-required prefix files in `runWineCommand`. If any are missing, `verifyWinePrefix` is used to create / hopefully fix the prefix.  
Because of that, Proton will no longer get skipped in `verifyWinePrefix` (even though `wineboot` itself doesn't actually get ran, which is fine since Proton creates the prefix itself)

If you really have to disable this check when running commands (`verifyWinePrefix` uses `runWineCommand` to create the prefix for example), pass `skipPrefixCheckIKnowWhatImDoing` to `runWineCommand`

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -437,10 +437,6 @@ export async function verifyWinePrefix(
     return { res: { stdout: '', stderr: '' }, updated: false }
   }
 
-  if (!(await validWine(wineVersion))) {
-    return { res: { stdout: '', stderr: '' }, updated: false }
-  }
-
   if (wineVersion.type === 'crossover') {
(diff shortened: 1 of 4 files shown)


imLinguin (approved):
Looks good to me

flavioislima (approved):
Looks good!
Thanks for fixing it! 👍

PR #1988, opened 2022-11-03
[General] Avoid unnecessary config writes

CommandMC:
This PR checks settings values passed to `setSetting` and skips re-writing the config file if they haven't changed.

In the process of this, I've applied a similar treatment our `useSetting` hook got to `getSetting` and `setSetting`, which in turn involved setting a proper default value for `currentConfig`. Since *that* can be either `AppSettings` or `GameSettings`, depending on whether we're editing the global or game-specific settings, I've also made `AppSettings` extend `GameSettings` (so even if we only have a `GameSettings` object under the hood, we can still interact with it like it's a `Partial<AppSettings>` (since that's what it is))

---


flavioislima (approved):
Looks good.
Tested here and it is working fine.

PR #2004, opened 2022-11-06
Use new `--accept-path` flag in save sync

CommandMC:
With Legendary 0.20.31, a new hidden `--accept-path` flag got added to the sync-saves command. With it, we can just accept the path Legendary computed for us, instead of fiddling around with parsing its output.
This makes our job a lot easier

---

### DIFF
--- a/src/backend/save_sync.ts
+++ b/src/backend/save_sync.ts
@@ -11,12 +11,19 @@ import {
   logWarning
 } from './logger/logger'
 import { getGame, getShellPath } from './utils'
-import { existsSync, realpathSync } from 'graceful-fs'
+import {
+  existsSync,
+  readFileSync,
+  realpathSync,
+  writeFileSync
+} from 'graceful-fs'
 import { app } from 'electron'
 import {
   createAbortController,
   deleteAbortController
 } from './utils/aborthandler/aborthandler'
+import { legendaryConfigPath } from './constants'
+import { join } from 'path'
 
 async function getDefaultSavePath(
   appName: string,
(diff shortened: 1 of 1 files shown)


CommandMC:
That's not really something we can influence at the moment. Legendary stores the save path itself & if it's already stored, it'll not re-compute it.
The only sensible option I can see here would be to add *another* hidden flag to `sync-saves` to clear out the path (I guess we could modify the .json file directly, but eh)
CC @derrod

flavioislima:
> That's not really something we can influence at the moment. Legendary stores the save path itself & if it's already stored, it'll not re-compute it.
> The only sensible option I can see here would be to add *another* hidden flag to `sync-saves` to clear out the path (I guess we could modify the .json file directly, but eh)
> CC @derrod

Maybe add a reload button on Heroic that clean the legendary config and start a new one.
The logic should be similar to the one we have to change the game path I guess.

CommandMC:
@flavioislima Got that implemented now, Heroic will modify Legendary's config file & delete the stored save path there if it's set
With this, the save path will be correctly updated when switching Wine versions (provided you clear out the path of course)

flavioislima:
> @flavioislima Got that implemented now, Heroic will modify Legendary's config file & delete the stored save path there if it's set
> With this, the save path will be correctly updated when switching Wine versions (provided you clear out the path of course)

OK, will test again and let you know 👍

flavioislima (approved):
Ok, tested here and it works fine now. I will try to implement a refresh button for the path in the future though.

PR #2006, opened 2022-11-07
[Fix] crash on install from context menu

wbrtm:
Fixes the bug of heroic crashing upon selecting install from context menu of a GameCard.
More a dirty workaround than a solution to the problem, I will have another look into it later this week. But this works for now.

---


flavioislima:
Closing this one since it is being fixed by #2010

PR #2007, opened 2022-11-07
[UI] Hide Clear button for Library's Search Bar

redromnon:
By default, the clear button (X) is always visible even if there's no input or text. 

Instead, display the clear button whenever the user starts typing into the search bar and disable it when cleared.

https://user-images.githubusercontent.com/74495920/200341520-b262f7ba-ad62-4bee-bc14-884880180ae4.mp4

---


flavioislima (approved):
Good catch!

Issue #1998, opened 2022-11-06
[Bug] Wine-related options appearing for Linux Native games


redromnon:
### Describe the bug

All the wine-related options used for Windows games are appearing for Linux native games as well (OpenTTD in my case). 
I'm not sure if this occurs for Mac native games too.

### Add logs

(12:17:34) INFO:    [Connection]:       Connectivity: check-online
(12:17:34) INFO:    [Connection]:       Pinging external endpoints
(12:17:34) INFO:    [Legendary]:        Running command: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary --version
(12:17:34) INFO:    [Legendary]:        Legendary location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(12:17:34) INFO:    [Gog]:              GOGDL location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(12:17:34) WARNING: [Backend]:          Protocol already registered.
(12:17:35) INFO:    [Gog]:              Running command: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --version
(12:17:35) INFO:    [Connection]:       Connectivity: online
(12:17:35) INFO:    [Gog]:              Getting data about the user
(12:17:36) INFO:    [Backend]:          

Heroic Version: 2.5.0-beta.2 Caesar Clown
Legendary Version:  0.20.30 Dark Energy (hotfix #4)
GOGdl Version: 0.5
OS: Pop KERNEL: 6.0.3-76060003-generic ARCH: x64
CPU: AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx @2.1 GOVERNOR: powersave
RAM: Total: 13.57 GiB Available: 6.72 GiB
GRAPHICS: GPU0: Picasso/Raven 2 [Radeon Vega Series / Radeon Vega Mobile Series] VRAM: 512MB DRIVER:  
PROTOCOL: x11

(12:17:37) INFO:    [Backend]:          Checking for new Heroic Updates
(12:17:37) INFO:    [Frontend]:         Refreshing Library
(12:17:37) INFO:    [Legendary]:        Refreshing library...
(12:17:37) INFO:    [Legendary]:        Refreshing Epic Games...
(12:17:37) INFO:    [Legendary]:        Game list updated, got 33 games & DLCs
(12:17:37) INFO:    [Gog]:              Getting GOG library
[...]

PR #2008, opened 2022-11-07
[UI] Hide wine-related settings for games when Wine is not used

arielj:
This PR fixes a regresion after refactoring the settings that some wine settings where showing up for native games (native linux, window or mac).

With this PR, all the wine-related settings (wine, prefix, extensions, fps, fsr, fsync, esync, system libraries) are only shown in linux for non-native games.

Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1998

---

### DIFF
--- a/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
+++ b/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
@@ -40,11 +40,9 @@ export default function GamesSettings() {
   const { isDefault, gameInfo } = useContext(SettingsContext)
   const isLinux = platform === 'linux'
 
-  let shouldRenderWineSettings = false
+  let usesWine = isLinux
   if (gameInfo) {
-    shouldRenderWineSettings = isLinux && !gameInfo.is_linux_native
-  } else {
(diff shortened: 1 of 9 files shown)


redromnon (approved):
Looks good to me :+1:

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -90,15 +94,15 @@ export default function GamesSettings() {
 
         <AlternativeExe />
 
-        <ShowFPS />
+        {usesWine && <ShowFPS />}
```
flavioislima, line 97: This one needs to appear on macOS as well. Would be good to check how settings are there before merging it.

arielj: oops! I fixed that and a few more, I also moved that logic inside the component so the component is also controls if it's shown (the logic was there but it depended in some location state that is not that reliable, not it uses the current game's game info)

flavioislima (approved):
Looks good!
Maybe one improvement would be to remove the specific native checks and only use one `isNative`

PR #2010, opened 2022-11-08
[Tech] UI performance improvements + small fixes

flavioislima:
This one add some performance improvements mostly on Library and GamePage.
I am doing that by using `memo` from React, since after some inspection I noticed that some components were being re-rendered even when the props were not changing. This avoid that. For me the performance improvement is pretty noticeble especially after walking throught the app for some time.
The memo just works after the second render so the first one does not change too much.

Other improvements that this PR does is regarding the GamePage. 
Instead of querying the GameInfo again, now the object is being passed from the GameCard, since it already has this information.

Also, not the settings are stored using `electron-store` so we can share them easily.
A good example is on the Download Dialog, where we were using querying the settings just to know the default install path, causing a unecessary re-render with that.

I did not dig deeply into this changes on Settings since would be too much for this PR, but we can start using it more from now on on future PRs.

For now we are still storing the settings on both places to avoid breaking current Heroic users, after a few releases will be safe to remove the original approach.

Other than that I added some small fixes introduced on previous PRs.
---


arielj:
I'm always a bit worried when I see memo cause I'm never sure what will trigger the re-renders so can't review it in that sense. But I played around with the PR and it does feel faster and I tried clicking things, install/uninstall, filtering, etc, all seems to be working fine.

flavioislima:
> I'm always a bit worried when I see memo cause I'm never sure what will trigger the re-renders so can't review it in that sense. But I played around with the PR and it does feel faster and I tried clicking things, install/uninstall, filtering, etc, all seems to be working fine.

I agree 100%, memo, useMemo and useCallback should not be used always, but I think here makes sense because it is a static app. But it is good to test if all states that needs update are being updated. Since we have the context the control most things, I think they are working fine.

But my feeling is that Heroic is a bit faster to navigate and also the removal of the getGameInfo on gamepage makes it loads instantly now, that's good as well.

PR #2011, opened 2022-11-09
[Logs] Catch invalid JSON outputs from legendary when launching games

arielj:
When we launch an Epic game, we request a json object with launch information. If legendary fails to return this information for any reason, the output is invalid JSON, causing an exception and the game is not executed.

This PR adds a try/catch there so if there's any issue that prevents legendary to respond with valid JSON then we just ignore this info in the logs and print the actual output from legendary to help debugging.

---


flavioislima (approved):
LGTM

CommandMC (approved):

PR #2016, opened 2022-11-11
[Tech] Type-checking our Electron Stores

CommandMC:
This PR introduces strict type checks to our various Electron Stores. Every store now has exact definitions of which keys it has & which types these keys have. Unlike `ElectronStore`/`Conf`'s built-in type definitions, this also considers accessing object properties with dot notation (`store.get('foo.bar')`) and defines appropriate types there. Due to every Store now having a (required) name, types are also consistent between FE and BE.

The most important change made here is that in the new classes (`TypeCheckedStoreBackend` & `TypeCheckedStoreFrontend`), you can no longer call `store.get` with just a key, you *have to* provide a default value. If you, for whatever reason, don't want to provide a default value, you have to explicitly call `.get_nodefault`, which adds `undefined` to the return type

I've already caught some (minor) inaccuracies with our type system with this, and I hope it'll lead to a more stable experience in the future

---


flavioislima (approved):
LGTM!
Took a bit for me to understand how things are working now so perhaps we should write some docs about how to create a new store and add new keys as well.

But this was a pretty good job, thanks for that! 👍🏽

PR #2022, opened 2022-11-14
[Linux - UI/UX] Show Checkbox to use default Wine Settings on new Installs

flavioislima:
Although having a prefix for each game is ideal, in most cases one prefix will work for most games, especially when you don't have much space like when using the Steam Deck.

So I added this option to use the default wine settings when installing a game. Checked by default, if unchecked, will create a new prefix on the default prefix folders with the game name like before.

image

image

---

### DIFF
--- a/src/frontend/screens/Library/components/InstallModal/WineSelector/index.tsx
+++ b/src/frontend/screens/Library/components/InstallModal/WineSelector/index.tsx
@@ -9,6 +9,7 @@ import React from 'react'
 import { AppSettings, WineInstallation } from 'common/types'
 import { useTranslation } from 'react-i18next'
 import { configStore } from 'frontend/helpers/electronStores'
+import { removeSpecialcharacters } from 'frontend/helpers'
 
 type Props = {
   setWineVersion: React.Dispatch<
@@ -48,7 +49,9 @@ export default function WineSelector({
       setWinePrefix(defaultPrefix)
       setWineVersion(wineVersion)
     } else {
-      const sugestedWinePrefix = `${prefixFolder}/${title}`
+      const sugestedWinePrefix = `${prefixFolder}/${removeSpecialcharacters(
+        title
+      )}`
       setWinePrefix(sugestedWinePrefix)
       setWineVersion(wineVersion || undefined)
     }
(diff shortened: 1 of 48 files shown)


src/frontend/screens/Library/components/InstallModal/WineSelector/index.tsx
```diff
@@ -24,64 +27,86 @@ export default function WineSelector({
   winePrefix,
   wineVersionList,
   wineVersion,
-  appName,
-  runner
+  title = 'sideload'
 }: Props) {
   const { t } = useTranslation('gamepage')
 
+  const [useDefaultSettings, setUseDefaultSettings] = React.useState(true)
```
CommandMC, line 35: A default value of `true` might be ideal for your specific setup right now, but most people have plenty of space for Prefixes & it's *really* the safer option to have one per game

CommandMC (changes requested):

src/frontend/screens/Library/components/InstallModal/WineSelector/index.tsx
flavioislima: Do you think so? Would be good to have some input from the users someday because this is more speculation in my experience.
Also, for a UI point of view it makes the Install screen cleaner.

flavioislima: A clean prefix could be 200MB but if you install some dependencies with winetricks that can go way up. I have prefixes here that are taking more than 2GB, and that is with btrfs compresion enabled.

I still think that we should test with it default first, is not like I am removing the option, the person always can go and click the checkbox if they want to have a separate prefix for each game.

CommandMC: If you are knowledgeable enough to install dependencies via Winetricks, you are also knowledgeable enough to toggle this option yourself when installing games (and sidenote, if you're installing things with Winetricks, you *absolutely should use a separate prefix*). I'm not advocating against the option as a whole, I just have an issue with it being the default.
Most people are just going to go "Oh, why would I change the default settings" when in reality, unless you're in a specific situation (low on space), *not* checking that option is safer

flavioislima: ok, mate. changed to false now.

PR #2024, opened 2022-11-14
[Feature] Tool Manager

Nocccer:
This PR adds the heroic-wine-downloader source code into the heroic repository.
Besides that it also provides now dxvk, dxvk-async, soda-bottles, dxvk-nvapi and vkd3d.
The Wine Manager was renamed to Tool Manager.

Todo in other PR:
- Make dxvk, dxvk-async, dxvk-nvapi and vkd3d versions selectable in the settings option.
- Adapt heroic-wine-downloader code and fix some problems
- Adapt tests to correct jest approach

---


Etaash-mathamsetty:
we should also add the caffe runners and kron4ek's wine builds (just upstream wine released with normal builds, staging builds and staging + tkg builds, it's also updated often)

Etaash-mathamsetty:
it's not following the theme atm, and the new tools that were added (dxvk, dxvk-async, and etc) have no entries. 
image

src/frontend/helpers/electronStores.ts
```diff
@@ -67,6 +67,6 @@ export {
   libraryStore,
   timestampStore,
   sideloadLibrary,
-  wineDownloaderInfoStore,
+  toolDownloaderInfoStore as wineDownloaderInfoStore,
```
Nocccer, line 70: Fix this

Nocccer:
> it's not following the theme atm, and the new tools that were added (dxvk, dxvk-async, and etc) have no entries. I also don't see any option to enable dxvk-nvapi right now, but maybe that's cuz I don't have it installed in heroic (it's not showing up in the tool manager) image
> 
> otherwise seems like a cool addition

To use the new tools we need to make them selectable. This is not implemented yet.

src/backend/legendary/games.ts
```diff
@@ -751,19 +751,11 @@ class LegendaryGame extends Game {
       "Legendary's config from config.ini (before Heroic's settings):\n"
     )
 
-    try {
-      const json = JSON.parse(stdout)
-      // remove egl auth info
-      delete json['egl_parameters']
+    const json = JSON.parse(stdout)
```
arielj, line 754: we should keep the try/catch here, if legendary fails to run the previous command for any reason it won't return a json output and `JSON.parse` fails, stopping the game launch

src/backend/toolmanager/__test__/toolmanager/getter.test.ts
```diff
@@ -0,0 +1,88 @@
+import { logWarning } from '../../../logger/logger'
+import { Repositorys, VersionInfo } from '../../../../common/types/toolmanager'
+import { getAvailableVersions } from '../../toolmanager'
+import { test_data_release_list } from '../test_data/github-api-test-data.json'
+import * as axios from 'axios'
+
+jest.mock('../../../logger/logger')
+jest.mock('../../../logger/logfile')
+
+describe('Main - GetAvailableVersions', () => {
```
arielj, line 10:
```suggestion
describe('ToolManager - GetAvailableVersions', () => {
```

So we it's more clear in the test output what is this about

src/backend/toolmanager/__test__/toolmanager/install.test.ts
```diff
@@ -0,0 +1,594 @@
...
+import * as axios from 'axios'
+import * as crypto from 'crypto'
+import { clearInterval } from 'timers'
+
+jest.mock('../../../logger/logger')
+jest.mock('../../../logger/logfile')
+
+const workDir = process.cwd()
+
+describe('Main - InstallVersion', () => {
```
arielj, line 20:
```suggestion
describe('ToolManager - InstallVersion', () => {
```

src/backend/toolmanager/__test__/utils/download.test.ts
```diff
@@ -0,0 +1,125 @@
+import { existsSync, mkdirSync, rmSync } from 'graceful-fs'
+import { downloadFile } from '../../utils'
+
+jest.mock('../../../logger/logfile')
+
+const workDir = process.cwd()
+
+describe('Utilities - Downlaod', () => {
```
arielj, line 8:
```suggestion
describe('Utilities - Download', () => {
```

src/backend/toolmanager/toolmanager.ts
```diff
@@ -0,0 +1,346 @@
...
+    Repositorys.DXVK_NVAPI,
+    Repositorys.VKD3D
+  ],
+  count = 100
+}: getVersionsProps): Promise<VersionInfo[]> {
+  const releases: Array<VersionInfo> = []
+  for await (const repo of repositorys) {
+    switch (repo) {
+      case Repositorys.WINEGE: {
+        await fetchReleases({
```
arielj, line 67: If we use the syntax:

```ts
const result = await fetchReleses({...})
releases.push(result)
```

I think we can move all the duplicated `.catch(error...` outside of the switch

src/backend/toolmanager/toolmanager.ts
```diff
@@ -0,0 +1,346 @@
...
+ * If no repository is given, all {@link Repositorys} are checked.
+ * @param repositorys array of {@link Repositorys}.
+ * @defaultValue all {@link Repositorys}
+ * @param count max versions to fetch for each {@link Repository}
+ * @defaultValue 100
+ * @returns * resolves with an array of {@link VersionInfo}
+ *          * rejects with an {@link Error}
+ */
+async function getAvailableVersions({
+  repositorys = [
```
arielj, line 52:
```suggestion
  repositories = [
```

src/backend/toolmanager/toolmanager.ts
arielj: Also, in my experience, mixing `await` and `then` in the same chain can lead to weird race conditions in some cases.

src/backend/toolmanager/toolmanager.ts
```diff
@@ -0,0 +1,346 @@
...
+        await fetchReleases({
+          url: WINEGE_URL,
+          type: 'Wine-GE',
+          count: count
+        })
+          .then((fetchedReleases: VersionInfo[]) => {
+            releases.push(...fetchedReleases)
+          })
+          .catch((error: Error) => {
+            throw error
```
arielj, line 76: do we want to throw an error here? like... what if a user only cares about updating the wine versions? if the dxvk versions fails I understand that would prevent any other tool list to be updated

maybe we should just log that the request to fetch the new versions failed and continue

src/frontend/screens/ToolManager/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+import { Tab, Tabs } from '@mui/material'
+import {
+  StoreIpc,
+  toolDownloaderInfoStore
+} from 'frontend/helpers/electronStores'
+import { ToolVersionInfo, Type } from 'common/types/toolmanager'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faSyncAlt } from '@fortawesome/free-solid-svg-icons'
+
+const ToolItem = lazy(
```
arielj, line 17: lazy loading of react components triggers the `Suspense` component, generating a weird flicker of the screen showing a `loading` screen for a moment

we shouldn't use lazy load of components unless there's a really good performance benefit

src/common/types/toolmanager.ts
```diff
@@ -0,0 +1,63 @@
...
+  download: string
+  downsize: number
+  disksize: number
+  checksum: string
+}
+
+/**
+ * Enum for the supported repositorys
+ */
+export enum Repositorys {
```
arielj, line 35:
```suggestion
export enum Repositories {
```

src/backend/toolmanager/toolmanager.ts
```diff
@@ -0,0 +1,346 @@
...
+  },
+  abortSignal
+}: installProps): Promise<{ versionInfo: VersionInfo; installDir: string }> {
+  /*
+   * VARIABLE DECLARATION
+   */
+
+  const tarFile =
+    installDir + '/' + versionInfo.download.split('/').slice(-1)[0]
+  const installSubDir = installDir + '/' + versionInfo.version
```
arielj, line 218: is `/` cross-platform compatible? I know I had issues in the past and I always fallback to something like the `path.join` from the path module that ensures compatibility

I guess it's fine though since I don't think we show the tool manager on windows anyway

src/backend/toolmanager/utils.ts
```diff
@@ -0,0 +1,576 @@
...
+  logError,
+  logInfo,
+  LogPrefix,
+  logWarning
+} from '../../backend/logger/logger'
+import { getAvailableVersions, installVersion } from './toolmanager'
+import { heroicToolsPath } from '../../backend/constants'
+import { sendFrontendMessage } from '../main_window'
+
+const wineDownloaderInfoStore = new Store({
```
arielj, line 29: we have a `src/frontend/helpers/electroStores.ts` file, we should keep them all there

src/common/types.ts
```diff
@@ -438,19 +437,12 @@ export type RefreshOptions = {
...
 }
 
-export interface WineVersionInfo extends VersionInfo {
-  isInstalled: boolean
-  hasUpdate: boolean
-  installDir: string
-}
-
-export type GamepadActionStatus = Record<
-  ValidGamepadAction,
```
arielj, line 448: we shouldn't remove this ValidGamepadAction here

arielj:
I have one question, I know we are adding tools like DXVK, VKD3D, etc, how does that work? I mean, how are those tools used after downloaded? Maybe I'm missing something

Nocccer:
> I have one question, I know we are adding tools like DXVK, VKD3D, etc, how does that work? I mean, how are those tools used after downloaded? Maybe I'm missing something

This is a Todo. We should add something like the wine selection, but i don't want to add this in the big PR. Should be done in another PR.

src/backend/legendary/games.ts
Nocccer: I did not touched that code. Probably a merge bug.

src/backend/toolmanager/toolmanager.ts
Nocccer: Good point. Actually this is the old code from heroic-wine-downloader and i don't want to change things in this PR.
We should do it in small PR's after this is merged.

src/backend/toolmanager/toolmanager.ts
Nocccer: Good point. Actually this is the old code from heroic-wine-downloader and i don't want to change things in this PR. We should do it in small PR's after this is merged.

src/backend/toolmanager/toolmanager.ts
Nocccer: Good point. Actually this is the old code from heroic-wine-downloader and i don't want to change things in this PR. We should do it in small PR's after this is merged.

src/backend/toolmanager/utils.ts
Nocccer: This is not the fronted, so should not work. We need something similiar in the backend. Maybe in another PR?

arielj: yeah sure, not for this PR, I missed it was in the backend

public/locales/ar/translation.json
```diff
@@ -638,14 +649,6 @@
...
     "wine": {
-        "actions": "الإجراء",
-        "manager": {
-            "link": "مدير واين",
-            "not-found": "لم يُعثر على إصدارات واين. انقر على رمز التحديث للمحاولة مرةً أخرى.",
-            "title": "مدير واين",
-            "unzipping": "جارٍ فك الضغط"
-        },
-        "release": "تاريخ الإصدار",
-        "size": "الحجم"
```
arielj, line 649: I'm not sure how to handle this, but we are losing all the translations because of the main key changing from `wine` to `tool`

I wonder if there's a way to preserve them, I'm not really sure

arielj: maybe they just have to change and that's it

src/backend/legendary/games.ts
arielj: We still have to revert this change, even more so if this is a merge bug. This change should not happen.

src/frontend/screens/ToolManager/index.tsx
arielj: this was marked as resolved, it's like this intentionally? we've had issues with lazy loaded components

src/common/types.ts
arielj: is this change intentional? this is marked as resolved with no comment

src/frontend/screens/ToolManager/index.tsx
arielj: nevermind, it was showing me an outdated/cached diff, this was changed

src/backend/legendary/games.ts
arielj: nevermind, github was showing me an outdated/cached diff

arielj:
should we do something with this PR? can we close it for now?

it's been here for almost a year, I'm trying to cleanup the list

Nocccer:
I think we can close it

PR #2029, opened 2022-11-18
[UI/UX] Download Manager Improvements

flavioislima:
Main changes:

- [x] Add support for games updates on DM
- [x] Improve UI with Game Art, download size, show play and download buttons, show type of the download (Update / Install)
- [x] Show Enqueued, started and finished times.
- [x] Remove unused code (old download and update methods)
- [x] Move status handling to the backend (setStatus was being used sometimes on BE and others on FE)
- [x] Improved empty queue screen

image
image

---


public/locales/az/translation.json
```diff
@@ -170,16 +170,17 @@
             "update_game": "Oyunu yeniləyin"
         }
     },
+    "disk": "Disk",
     "docs": "Sənədlər",
+    "down": "Down",
```
arielj, line 175: this might be confusing for translators without context (like `down` in spanish would be `abajo` but in the context of downloads would be more like `download speed` / `velocidad de descarga` so `descarga` would be a better translation)

maybe `speed` is better than `down`?

src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.tsx
```diff
@@ -119,11 +191,18 @@ const DownloadManagerItem = ({ element, current }: Props) => {
...
+        <span className="titleSize">
+          {title}
+          <span title={path}>
+            {downloadSize ?? ''}
+            {canceled ? ` (${t('queue.label.canceled', 'Canceled')})` : ''}
+          </span>
+        </span>
       </span>
+      <span>{getTime()}</span>
+      <span style={{ textTransform: 'capitalize' }}>{type}</span>
```
arielj, line 204: this should be a translatable string

public/locales/az/translation.json
flavioislima: fixed

CommandMC (approved):

PR #2031, opened 2022-11-19
[UI/UX] Detect if the game folder is available and show if the game is installed or not

flavioislima:
This is a small PR that might be useful on the Steam Deck or other portable devices or just using external SSD/HDD/SDCards.

It adds a new method to check if the game is available, it is a simple check on the folder to see if it exists.
If not, it sets the game as not installed.
It is implemented on the `GameCard` and also on `getGameInfo` call.

As a future improvement (maybe once we have Libraries implemented like Steam), we can check on how detect the change of devices and add a listener to it. 
The tricky part would be to also check the if the device is mounted, since by default some storage devices might not mount automatically depending on the system configuration.

But so far this one works fine. When a new device is connected just reload the library or going to the Game Page changes the status of the game.

---


arielj:
I wonder if we should have a new state for games `unavailable`.

Because I imagine this scenario:
- I have multiple SD cards where I install games
- When I want to play a game, I put the memory card in (like a normal console with cartridges)

If we consider `the folder is not there` as `this is uninstalled`, all the games I have installed in my SD cards will show up as uninstalled, when they are not actually

flavioislima:
I just followed how steam does, though.
If you remove your sdcard from the deck all games that are there will show as uninstalled.
I think that's fine since they really are really not installed on your computer right now.
But we can improve this later, let's get some feedback on next release.

src/backend/main.ts
```diff
@@ -760,6 +760,19 @@ ipcMain.on('createNewWindow', (e, url) => {
   new BrowserWindow({ height: 700, width: 1200 }).loadURL(url)
 })
 
+ipcMain.handle('isGameAvailable', async (e, args) => {
+  const { appName, runner } = args
+  const info = getGame(appName, runner).getGameInfo()
+  if (info && info.is_installed) {
+    if (existsSync(info.install.install_path!)) {
```
CommandMC, line 767:
```suggestion
    if (info.install.install_path && existsSync(info.install.install_path)) {
```

We currently aren't constraining `install_path`'s type based on `is_installed`, so it *could* be undefined even if `is_installed` is true. In the real world this is rather unlikely to happen, but point is that it *can* happen

I plan on revising our core types (`AppSettings`, `GameInfo`, and so on) soon:tm: so this can then no longer happen, but for now this would be safer

src/backend/main.ts
```diff
@@ -771,9 +784,18 @@ ipcMain.handle('getGameInfo', async (event, appName, runner) => {
   try {
     const game = getGame(appName, runner)
     const info = game.getGameInfo()
+
     if (!info.app_name) {
       return null
     }
+    //detects if the game folder is available
+    if (info && info.is_installed) {
+      if (existsSync(info.install.install_path!)) {
```
CommandMC, line 793: See above comment

CommandMC (changes requested):
I'd definitely agree that, in the future, we should *somehow* convey that unavailable games are just... unavailable & not uninstalled entirely, but this would work for now

Got one small concern about that non-null assertion. Other than that, this is fine

src/backend/main.ts
flavioislima: good catch.

CommandMC (approved):

Issue #1742, opened 2022-08-21
Show information from howlongtobeat, areweanticheatyet and protondb


Verantor:
### Problem description

I like to know when I am looking at a game how long I will need to play though. [howlongtobeat](https://howlongtobeat.com/)
(if exist) its status on [protondb](https://www.protondb.com/)
and data from [areweanticheat.net](https://areweanticheatyet.com/)

### Feature description

Pull data from [howlongtobeat](https://howlongtobeat.com/)
(if exist) its status on [protondb](https://www.protondb.com/)
and [areweanticheat.net](https://areweanticheatyet.com/)

JakobDev:
areweanticheat is already included

ScanuNicco:
Would love to see protondb status integrated directly into the library

arielj:
We include a link to open protondb in the search result page in game page > tools.

I don't think it makes sense to actually show the info from protondb inside Heroic because:
- protondb is specific to steam (many epic games are not listed and many will be really hard to match since naming is not the same)
- protondb is specific to proton (reports from protondb won't always be correct since heroic users would probably be using wine-ge or crossover instead of proton)
- a game working in steam may not work in heroic (like epic games may be missing EAC runtime for example)
- there's too much information in protondb to actually show it inside Heroic (that's why I think the link is enough)

Info from areweanticheatyet shows up in the game's page and in the install dialog already.

We could maybe add a link to howlongtobeat if it's a feature requested by more users, but I wouldn't add more than a link to the search result page (again, hard to match if name is not consistent, a lot of information to fit inside the heroic UI)

flavioislima:
I don't think adding how long to beat into the game info is relevant and will also be another api to request information and to maintain. Don't really see much value on this.

flavioislima:
Reopening this one since I found this API here for How Long to Beat, it seems simple to implement it on the Game Page:
https://github.com/ckatzorke/howlongtobeat

flavioislima:
Does this looks good? @arielj @Verantor 
![image](https://user-images.githubusercontent.com/26871415/202871376-18c92716-51de-4c0d-a447-ce92a79ee177.png)

Verantor:
Sick it looks so good

Maltavius:
> We include a link to open protondb in the search result page in game page > tools.
> 
> I don't think it makes sense to actually show the info from protondb inside Heroic because:
> 
[...]

PR #2032, opened 2022-11-19
[UI] Show How Long to Beat Data on Game Page

flavioislima:
Requested on #1742 

image

image

---

### DIFF
--- a/src/backend/howlongtobeat/utils.ts
+++ b/src/backend/howlongtobeat/utils.ts
@@ -1,0 +1,12 @@
+import { HowLongToBeatEntry, HowLongToBeatService } from 'howlongtobeat'
+
+export async function getHowLongToBeat(
+  title: string
+): Promise<HowLongToBeatEntry> {
+  return new Promise((resolve) => {
+    const hltb = new HowLongToBeatService()
+    hltb.search(title).then((results) => {
+      resolve(results[0])
+    })
+  })
+}
--- a/src/backend/howlongtobeat/utils.ts
+++ b/src/backend/howlongtobeat/utils.ts
@@ -1,12 +1,17 @@
+import { logError, LogPrefix } from 'backend/logger/logger'
 import { HowLongToBeatEntry, HowLongToBeatService } from 'howlongtobeat'
 
 export async function getHowLongToBeat(
   title: string
-): Promise<HowLongToBeatEntry> {
-  return new Promise((resolve) => {
+): Promise<HowLongToBeatEntry | null> {
+  try {
     const hltb = new HowLongToBeatService()
-    hltb.search(title).then((results) => {
-      resolve(results[0])
+    const results = await hltb.search(title)
+    return results[0]
+  } catch (error) {
+    logError(['Was not able to get HowLongToBeat data', error], {
+      prefix: LogPrefix.Backend
     })
-  })
+    return null
+  }
 }
--- a/src/frontend/components/UI/HowLongToBeat/index.tsx
+++ b/src/frontend/components/UI/HowLongToBeat/index.tsx
@@ -26,8 +26,6 @@ export default function HowLongToBeat({ title }: Props) {
     return null
   }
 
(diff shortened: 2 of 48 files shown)


src/frontend/components/UI/HowLongToBeat/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+            {t('how-long-to-beat.main-story', 'Main Story')}
+          </div>
+          <div className="howLongToBeat__square__value">
+            {gameplayMain || '?'} {t('hours', 'Hours')}
+          </div>
+        </div>
+        <div className="howLongToBeat__square">
+          <div className="howLongToBeat__square__title">
+            {t('how-long-to-beat.main-plus', 'Main')} +{' '}
+            {t('how-long-to-beat.main-extras', 'Extras')}
```
arielj, line 47: I think it's better to have this string to be translated as a whole `Main + Extras`, I don't know about grammar rules for other languages but maybe these are the other way around in some language or something different is more common than the `+` symbol (like &)

We don't need the 2 string separated anyway

src/backend/howlongtobeat/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { HowLongToBeatEntry, HowLongToBeatService } from 'howlongtobeat'
+
+export async function getHowLongToBeat(
+  title: string
+): Promise<HowLongToBeatEntry> {
+  return new Promise((resolve) => {
+    const hltb = new HowLongToBeatService()
+    hltb.search(title).then((results) => {
+      resolve(results[0])
+    })
```
arielj, line 10: should we add a `catch` function here? in case the API is down or there's no internet connection

src/frontend/components/UI/HowLongToBeat/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+// react component with 3 square with how long to beat info
+
+type Props = {
+  title: string
+}
+
+export default function HowLongToBeat({ title }: Props) {
+  const [howLongToBeatInfo, setHowLongToBeatInfo] =
+    React.useState<HowLongToBeatEntry | null>(null)
```
arielj, line 14: let's use `useState` directly and imported in line 2, for consistency with other components and hooks

src/frontend/components/UI/HowLongToBeat/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+      }
+    })
+  }, [title])
+
+  if (!howLongToBeatInfo) {
+    return null
+  }
+
+  const { gameplayMain, gameplayMainExtra, gameplayCompletionist } =
+    howLongToBeatInfo || {}
```
arielj, line 30: Is this `|| {}` needed? we should already by a non-null/undefined/false values from the previous if

arielj (commented):
Looks good, I added some comments.

One thing I'm concerned is how it's going to look in other languages, but we have to wait and see.

src/backend/howlongtobeat/utils.ts
flavioislima: Done

src/frontend/components/UI/HowLongToBeat/index.tsx
flavioislima: Done

src/frontend/components/UI/HowLongToBeat/index.tsx
flavioislima: fixed

flavioislima:
> Looks good, I added some comments.
> 
> One thing I'm concerned is how it's going to look in other languages, but we have to wait and see.

Yes, since it is a new feature it might take some time until we get proper feedback.
About the caching, we just need to know if this data tend often or not, I guess some games, especially after launch, might change this info a lot, while old games not much.

CommandMC (approved):

PR #2038, opened 2022-11-20
[Fix/Epic] Handle metadata files with dots in their name correctly

CommandMC:
When reading out Legendary's `metadata` folder & adding the names of those files to our `allGames` set, we were assuming that there are no dots (`.`) in the actual app_name. If there were, we would fail once we try to read out the file
In practice this makes little difference, since the affected "games" are Unreal Engine-related files, which are skipped anyways, but it's still neat to get rid of those "Failed to parse ..." messages on startup

---


flavioislima (approved):
Looks good!

PR #2041, opened 2022-11-21
[UI] Small Improvements

flavioislima:
A few UI fixes and improvements:
- [x] Sideload: search image after losing focus instead of typing
- [x] Sideload: turn the image to black/white when searching
- [x] Make AppName selectable on settings
- [x] Hide install button for not available sideloaded games.

---

### DIFF
--- a/src/frontend/components/UI/CachedImage/index.tsx
+++ b/src/frontend/components/UI/CachedImage/index.tsx
@@ -8,9 +8,8 @@ interface CachedImageProps {
 type Props = React.ImgHTMLAttributes<HTMLImageElement> & CachedImageProps
 
 const CachedImage = (props: Props) => {
-  const srcLocation = props.src ?? props.fallback
   const [useCache, setUseCache] = useState(
-    srcLocation.startsWith('http') || false
+    props.src?.startsWith('http') || false
   )
   const [useFallback, setUseFallback] = useState(false)
 
--- a/src/frontend/components/UI/CachedImage/index.tsx
+++ b/src/frontend/components/UI/CachedImage/index.tsx
@@ -8,7 +8,10 @@ interface CachedImageProps {
 type Props = React.ImgHTMLAttributes<HTMLImageElement> & CachedImageProps
 
 const CachedImage = (props: Props) => {
-  const [useCache, setUseCache] = useState(props.src.startsWith('http'))
+  const srcLocation = props.src ?? props.fallback
+  const [useCache, setUseCache] = useState(
+    srcLocation.startsWith('http') || false
+  )
   const [useFallback, setUseFallback] = useState(false)
 
   const onError = () => {
--- a/src/frontend/screens/Library/components/GameCard/index.tsx
+++ b/src/frontend/screens/Library/components/GameCard/index.tsx
@@ -291,8 +291,7 @@ const GameCard = ({
       // install
       label: t('button.install'),
       onclick: () => buttonClick(),
-      show:
-        (!isInstalled && !isQueued) || (runner === 'sideload' && !isInstalled)
+      show: !isInstalled && (!isQueued || runner === 'sideload')
     },
     {
       // cancel installation/update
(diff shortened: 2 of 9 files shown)


CommandMC:
I have a feeling this was meant to target the `beta` branch

flavioislima:
> I have a feeling this was meant to target the `beta` branch

yes, forgot to change, I am still working on this one.

src/frontend/screens/Library/components/GameCard/index.tsx
```diff
@@ -291,7 +291,8 @@ const GameCard = ({
       // install
       label: t('button.install'),
       onclick: () => buttonClick(),
-      show: !isInstalled && !isQueued
+      show:
+        (!isInstalled && !isQueued) || (runner === 'sideload' && !isInstalled)
```
CommandMC, line 295: Not sure what this is doing, `isQueued` is always `false` for sideloaded games, so that first condition also covers them

src/frontend/components/UI/CachedImage/index.tsx
```diff
@@ -8,7 +8,10 @@ interface CachedImageProps {
 type Props = React.ImgHTMLAttributes<HTMLImageElement> & CachedImageProps
 
 const CachedImage = (props: Props) => {
-  const [useCache, setUseCache] = useState(props.src.startsWith('http'))
+  const srcLocation = props.src ?? props.fallback
```
CommandMC, line 11: This also doesn't look right, `props.src` is of type `string`, so (assuming everything's following the type definitions), the `?? props.fallback` should never happen

src/frontend/components/UI/CachedImage/index.tsx
```diff
@@ -8,7 +8,10 @@ interface CachedImageProps {
...
+  const [useCache, setUseCache] = useState(
+    srcLocation.startsWith('http') || false
```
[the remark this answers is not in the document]
flavioislima: When using this component with the Sideloading it was breaking all the time so I needed to add these checks, I will change them to be on the getImage method then.

src/frontend/screens/Library/components/GameCard/index.tsx
flavioislima: actually, those are different things if you look closely.
First part is for non-sideloaded games:
- The game is installed
- The game is queued. 
So we dont show the install button on these cases.

The second part is for sideloading games that are not available
- Runner is sideload
- Is not installed

src/frontend/components/UI/CachedImage/index.tsx
flavioislima: Yes, in some cases the props.src is coming as undefined, I believe this happens when some game does not have a keyImage key on the metadata. 
I think this check here causes no harm and is better to keep it unless we are sure that all keyImages will come correctly.

src/frontend/components/UI/CachedImage/index.tsx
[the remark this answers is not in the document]
flavioislima: fixed

src/frontend/components/UI/CachedImage/index.tsx
CommandMC: > Yes, in some cases the props.src is coming as undefined, I believe this happens when some game does not have a keyImage key on the metadata.
I think this check here causes no harm and is better to keep it unless we are sure that all keyImages will come correctly.

Right, but the only reason TS is not yelling at you there is because it thinks `props.src` will always be a `string` (try changing the type of it to `string | undefined`, you'll see what I mean)
Either `props.src` or `props.fallback` should be *just* `string` for this to be safe. If `props.src` isn't, `props.fallback` should be

src/frontend/screens/Library/components/GameCard/index.tsx
flavioislima: True, just sent a fix 👍🏽

PR #2042, opened 2022-11-21
Add a timeout when watching Legendary's `installed.json`

CommandMC:
After making the recent changes in #2004, I've noticed that an issue became more frequent: We're (trying to) reload our LegendaryLibrary before Legendary/we have finished writing the file.
Apparently, NodeJS' `watch` function is notorious for double-firing events like this (I initially assumed it was caused by the file being written in chunks, but I'm not so sure anymore).

One solution would be to switch to a 3rd party library, but I believe that would be overkill for such a small part of Heroic.
Instead, we now start a timeout of 500ms when a write occurs. If another write happens during this timeframe, our old timeout is stopped & a new one is started

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -915,9 +915,16 @@ ipcMain.handle('writeConfig', (event, { appName, config }) => {
 
 // Watch the installed games file and trigger a refresh on the installed games if something changes
 if (existsSync(installed)) {
+  let watchTimeout: NodeJS.Timeout | undefined
   watch(installed, () => {
-    logInfo('Installed game list updated', { prefix: LogPrefix.Legendary })
-    LegendaryLibrary.get().refreshInstalled()
+    logInfo('installed.json updated, refreshing library', {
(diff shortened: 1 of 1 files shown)


flavioislima:
We need to use a bouncing function to proper do that.
Timeout might not be the best option. Unless the library uses a timeout to bounce it 🙈
We can search for a small library that does that and we actually might need it in other places as well.
On the frontend we can have a React Hook that handles them.

flavioislima:
Maybe this 12kb one:
https://www.npmjs.com/package/debounce

Although in the end it might not make any difference.

CommandMC:
The library you linked also uses a timeout, as far as I can tell: https://github.com/component/debounce/blob/master/index.js#L23
The last release of it, not counting the latest since it just added Contributing/Licensing stuff, was also in 2018

I'm fine with extracting this into a separate function once it's necessary elsewhere. So far, I don't know of any other use cases for it though.

flavioislima (approved):

PR #2044, opened 2022-11-21
[Backend] Add support for auto enabling anti cheat runtimes for legendary on linux

Etaash-mathamsetty:
Previously, users would have to manually enable the EAC runtimes, with this PR if the game uses EAC or Battleye (using areweanticheatyet data), we can automatically enable it

### DIFF
--- a/src/backend/legendary/games.ts
+++ b/src/backend/legendary/games.ts
@@ -515,7 +515,7 @@ class LegendaryGame extends Game {
         setAnticheatInfo(anticheatInfo)
       })
     if (anticheatInfo && isLinux) {
-        let gameSettings = await this.getSettings()
+        const gameSettings = await this.getSettings()
 
         gameSettings.eacRuntime =
           anticheatInfo.anticheats.includes('Easy Anti-Cheat')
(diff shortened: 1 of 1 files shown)


Etaash-mathamsetty:
probably needs more testing tho

src/backend/legendary/games.ts
```diff
@@ -502,6 +505,24 @@ class LegendaryGame extends Game {
       return { status: 'error', error: res.error }
     }
     this.addShortcuts()
+
+    const [anticheatInfo, setAnticheatInfo] = useState<AntiCheatInfo | null>(
+      null
+    )
+    window.api
```
arielj, line 512: `window.api...` is meant to be used only in the frontend, you should import `gameAnticheatInfo` directly from `anticheat/utils` since you are already in the backend side

src/backend/legendary/games.ts
```diff
@@ -502,6 +505,24 @@ class LegendaryGame extends Game {
       return { status: 'error', error: res.error }
     }
     this.addShortcuts()
+
+    const [anticheatInfo, setAnticheatInfo] = useState<AntiCheatInfo | null>(
```
arielj, line 509: this is a react feature, it can't be used like this in the backend that is not a react app

flavioislima (approved):
LGTM

PR #2045, opened 2022-11-21
[FIX] Auto Sync Saves not working on No-GUI mode

### LINKED ISSUE #2043: Add Cloud-Save-Syncing when launching via CLI/Steam-Shortcut
### Problem description

It appears that, when launching via CLI/Steamshort, there is not cloud-save synchronization. 
After looking at the logs, there are no entries like 
```
INFO:    [Gog]:             Syncing saves for <appid>
```
when launching the game on the CLI via
```
flatpak run com.heroicgameslauncher.hgl --no-gui --no-sandbox "heroic://launch/<appid>"
```

I had a brief look at the source-code, and it appears as if `doAutoSyncSaves` in is never called when launching a game via `handleProtocol`, instead it only seems to be called by `handlePlay`, which in turn only seems to be attached to click-handlers. I might be wrong though, as mentioned i only browsed the source-code for a short time. 

### Feature description

Adding Cloud-Save-Syncing for CLI-launches will greatly help Steam-Deck players, that do want to utilize the game-mode features. 

### Alternatives

_No response_

### Additional information

_No response_

Zarklord:
this resolves #2043

Adds cloud syncing when launching from the Game Card page, or when launched via the CLI.

---

### DIFF
--- a/src/frontend/helpers/library.ts
+++ b/src/frontend/helpers/library.ts
@@ -8,7 +8,7 @@ import {
 } from 'common/types'
 
 import { TFunction } from 'react-i18next'
-import { getGameInfo, sendKill } from './index'
+import { getGameInfo, sendKill, syncSaves } from './index'
 import { DialogModalOptions } from 'frontend/types'
 
 const storage: Storage = window.localStorage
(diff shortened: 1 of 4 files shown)


flavioislima:
I think this will work but at the same time I wonder if would be easier to just check the NoGui variable that we have on backend on launch.
So on the launch command, check if the autoSyncSaves option is true, if yes, call the sync saves command.
Then after the launch command ends, sync again if it is true.

What do you think?

flavioislima (approved):
Looks good, thanks for fixing this. ⚔️

PR #2052, opened 2022-11-22
[Fix] errorHandler: Handle `grep` exit code of 1 better

CommandMC:
`grep` exits with a return code of 1 when no lines were found, making Node assume that it failed. However, in this case, it's absolutely fine for it to not find anything

In practice, this gets rid of the "Operation interrupted" message which was showing up in my logs after installing any game

---


flavioislima (approved):
LGTM

PR #2053, opened 2022-11-23
[Tech] Frontend performance improvements

flavioislima:
After doing some benchmarks and performance testing on the frontend I noticed that Heroic was loading on start some components that are not needed right away, for instance, components from Downloads, wine manager, and settings screen and also some others like the DownloadDialog and so on. So I lazy loaded all components that are not needed on startup, basically all but the ones on Library Screen.

Besides that, I noticed that the `Library` and `GameList` was being re-rendered up to 4 times and causing the Gamecards to re-renders up to 6 times so I started on a hunt for what was causing that.
The root problem is the context provider, but also some bad practices when defining methods and using `useState` and `useEffect` as well.

So with this changes I was able to get a bit more performance on frontend, but ideally we need to change to a state management library like Redux so we can avoid all those re-renders that are related with the context.

The branch name implies that I wanted to implement a way of Virtualize the game list, that was the idea but I was not able to find a way of doing that yet. I tested a few libraries but, althought they were working in the sense of making the cards render only on viewPort, the other issues with the context were still making the whole list of games to load at once so were not making any difference in the end. I am still investigating on that though.

---

### DIFF
--- a/src/frontend/hooks/useWhyDidYouUpdate.ts
+++ b/src/frontend/hooks/useWhyDidYouUpdate.ts
@@ -1,0 +1,35 @@
+import { useRef, useEffect } from 'react'
+
+export function useWhyDidYouUpdate(
+  name: string,
+  props: Record<string, string>
+) {
+  // Get a mutable ref object where we can store props ...
+  // ... for comparison next time this hook runs.
+  const previousProps = useRef({})
+  useEffect(() => {
+    if (previousProps.current) {
+      // Get all keys from previous and current props
+      const allKeys = Object.keys({ ...previousProps.current, ...props })
+      // Use this object to keep track of changed props
+      const changesObj = {}
+      // Iterate through keys
+      allKeys.forEach((key) => {
+        // If previous is different from current
+        if (previousProps.current[key] !== props[key]) {
+          // Add to changesObj
+          changesObj[key] = {
+            from: previousProps.current[key],
+            to: props[key]
+          }
+        }
+      })
+      // If changesObj not empty then output to console
+      if (Object.keys(changesObj).length) {
+        console.log('[why-did-you-update]', name, changesObj)
+      }
+    }
+    // Finally update previousProps with current props for next hook call
+    previousProps.current = props
+  })
+}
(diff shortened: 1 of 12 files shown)


src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx
```diff
@@ -292,45 +295,6 @@ export default function DownloadDialog({
     getSpace()
   }, [installPath, gameInstallInfo?.manifest?.disk_size])
 
-  useEffect(() => {
```
flavioislima, line 295: this `useEffect` was duplicated somehow, probably on some merge.

src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx
```diff
@@ -215,7 +215,10 @@ export default function DownloadDialog({
         )
         setGameInstallInfo(gameInstallInfo)
 
-        if (gameInstallInfo && 'languages' in gameInstallInfo.manifest) {
+        if (
```
flavioislima, line 218: needed to add this condition otherwise if a game was shown as installed on legendary installed.json file but was being marked as not available, clicking on install button was causing it to fail.
Now the install screen shows normally.

src/frontend/state/ContextProvider.tsx
```diff
@@ -31,7 +31,7 @@ const initialContext: ContextType = {
   libraryStatus: [],
   libraryTopSection: 'disabled',
   handleLibraryTopSection: () => null,
-  platform: 'unknown',
+  platform: 'linux',
```
flavioislima, line 34: set this as default since it is the majority that uses Heroic. It was causing a unnecessary re-render since it was changing from `unknown`.
Will still cause this re-render on other OSes on the first load, but then the value will be stored on browser cache and loaded from there.

src/frontend/hooks/useWhyDidYouUpdate.ts
```diff
@@ -0,0 +1,35 @@
+import { useRef, useEffect } from 'react'
+
+export function useWhyDidYouUpdate(
```
flavioislima, line 3: for debugging purposes only.
Can be used to check props, state and context changes on a functional component.
It logs all changes on the passed variables.
got from:
https://usehooks.com/useWhyDidYouUpdate/

flavioislima:
After testing more this PR I believe it is too unstable so I will close it and work on fixing the root of the issue on another PR.

PR #2054, opened 2022-11-23
[Fix/Epic] Only delete `userInfo` key from configStore when logging out

### LINKED ISSUE #1965: After logout, settings are reseted
### Describe the bug

When you log out from a store, some settings are deleted (I didn't check exactly which ones, but at least the theme is reseted). This is an issue because you are still logged in with an account.

### Add logs

Nothing relevant.

### Steps to reproduce

1. Login with both epic and gog
2. Change your theme
3. Log out from one of the stores
4. The theme (along with other settings most likely) are reseted as the initial default values

### Expected behavior

It should not reset the settings.

### Screenshots

_No response_

### System Information

- Heroic Version [e. g. 2.1.1]: main and beta

### Additional information

_No response_

CommandMC:
When logging out from Legendary, we were **clearing out our entire configStore**, which in turn reset all settings and, since the `settings` key was no longer present in the store, lead to issues where this new functionality was used
As far as I can tell, the only thing that actually needs to be deleted when logging out is the `userInfo` key. So let's only delete that

Closes #1965

---

### DIFF
--- a/src/backend/legendary/user.ts
+++ b/src/backend/legendary/user.ts
@@ -78,7 +78,7 @@ export class LegendaryUser {
     await ses.clearCache()
     await ses.clearAuthCache()
     await ses.clearHostResolverCache()
-    configStore.clear()
+    configStore.delete('userInfo')
     clearCache()
   }
 


flavioislima:
@CommandMC I believe that Favorites and Recent games should be removed as well since the person might need to change to another account that does not have those games.

CommandMC:
Those are already filtered where we use then:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/8a46220add47f8ded47564da310251b833b76173/src/frontend/screens/Library/components/RecentlyPlayed/index.tsx#L14-L30

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/8a46220add47f8ded47564da310251b833b76173/src/frontend/screens/Library/index.tsx#L201-L214

flavioislima (approved):
Ok, looks good 👍🏽

PR #2055, opened 2022-11-24
[UI] Fix downloads styling

Etaash-mathamsetty:
There were these white borders around all the items in downloads view, I changed a small thing in the CSS so that now it's not an ugly white :D

### DIFF
--- a/src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.css
+++ b/src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.css
@@ -31,7 +31,7 @@
   position: relative;
   cursor: default;
   place-self: center;
-  border-bottom: 1px solid var(--divider);
+  border-top: 1px solid var(--divider);
   padding: var(--space-md) 0;
 }
 
--- a/src/frontend/screens/DownloadManager/components/ProgressHeader/index.css
+++ b/src/frontend/screens/DownloadManager/components/ProgressHeader/index.css
@@ -7,7 +7,7 @@
 }
 
 .progressHeader {
-  border: solid 2px var(--text-default);
+  border: solid 2px var(--navbar-background);
   border-radius: 6px;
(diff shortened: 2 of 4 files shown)


Etaash-mathamsetty:
whoops this is not meant for main

flavioislima:
@Etaash-mathamsetty can you add screenshots with before and after?

flavioislima:
I tested here and I think it is even better if we use this color for the background like this:
image

flavioislima:
maybe also remove the border of the last element

Etaash-mathamsetty:
> maybe also remove the border of the last element

wdym by this?

flavioislima:
> > maybe also remove the border of the last element
> 
> wdym by this?

Can you see that the last game on the list has a line at the bottom? It is ugly, we should have a line to divide the items when having more than one, but the last one should not have imo.
So you can create a css rule to remove the border from last item.

Etaash-mathamsetty:
> > > maybe also remove the border of the last element
> > 
> > 
> > wdym by this?
> 
> Can you see that the last game on the list has a line at the bottom? It is ugly, we should have a line to divide the items when having more than one, but the last one should not have imo. So you can create a css rule to remove the border from last item.

I think I came up with a simpler solution than that, I switched to use border-top, which just removes the bottom border

flavioislima:
> > > > maybe also remove the border of the last element
> > > 
> > > 
> > > wdym by this?
> > 
> > 
> > Can you see that the last game on the list has a line at the bottom? It is ugly, we should have a line to divide the items when having more than one, but the last one should not have imo. So you can create a css rule to remove the border from last item.
> 
> I think I came up with a simpler solution than that, I switched to border-top, which just removes the bottom border, the next element's top border will become the bottom border for the previous element

You are pretty smart, that works as well 🤣

flavioislima (approved):
Looks cool now 👍🏽

PR #2079, opened 2022-11-27
[Fix] Beta 3 Fixes + some small improvements

flavioislima:
### A few QOL improvements

- Heroic will list Origin games but will show a warning and the install button will be disabled:

image
- Not available games will have a different effect on the game card and a warning message on both GameCard and GamePage. Clicking on play will ask if the user wants to uninstall it:

image

### Fixes

- [macOS] Fix: Launching a windows sideloaded game 
- [macOS] Fix: proper get save path for windows games
- [macOS] Fix: native GOG games failing to install
- [Linux] Wine manager should show a Refresh button when the list is empty
- Fix Sideloaded GameCards showing the install button instead of Play
- Fix the game card status not changing after importing the game
- Fix some settings appearing on the wrong platform
- Fix the platform not being passed when importing Epic Games
- Fix the Linear Progress color on the Downloads Manager
- Fix hiding a game not reloading the library

### Other
- Refactor Winetricks download to use Axios instead of CURL, since it is a plain text
- Removed AudioFix setting

---


redromnon:
>Fix some settings appearing on the wrong platform

As already mentioned in Discord, the Sideload games feature still shows Wine options for Linux platform games.

redromnon:
Also once a game is uninstalled, it shows "not available" which it shouldn't. Just uninstalled one now and I see this.

Uninstalled_game_not_available

src/backend/games.ts
```diff
@@ -1,3 +1,4 @@
+import { InstallPlatform } from './../common/types'
```
CommandMC, line 1: Might want to import this after `ProtonVerb` below (l9/10)

src/backend/legendary/library.ts
```diff
@@ -486,6 +486,8 @@ export class LegendaryLibrary {
     const CloudSaveFolder = customAttributes?.CloudSaveFolder
     const FolderName = customAttributes?.FolderName
     const canRunOffline = customAttributes?.CanRunOffline?.value === 'true'
+    const thirdPartyManagedApp =
```
CommandMC, line 489: Thinking about this, maybe we should just store all `customAttributes` in the `GameInfo`

Anyways, for now this is fine

src/backend/tools.ts
```diff
@@ -193,25 +196,29 @@ export const DXVK = {
...
-      .then(() => {
-        exec(`chmod +x ${path}`)
-        logInfo('Downloaded Winetricks', { prefix: LogPrefix.WineTricks })
-      })
-      .catch(() => {
-        logWarning('Error Downloading Winetricks', {
-          prefix: LogPrefix.WineTricks
-        })
+    try {
+      logInfo('Downloaded Winetricks', { prefix: LogPrefix.WineTricks })
```
CommandMC, line 212: This would be
```suggestion
      logInfo('Downloading Winetricks', { prefix: LogPrefix.WineTricks })
```
now, right?

src/backend/tools.ts
```diff
@@ -193,25 +196,29 @@ export const DXVK = {
...
+      const res = Axios.get(url, { timeout: 1000 })
+      const file = (await res).data
```
CommandMC, line 214:
```suggestion
      const res = await Axios.get(url, { timeout: 1000 })
      const file = res.data
```

src/backend/tools.ts
```diff
@@ -1,10 +1,10 @@
 import { WineInstallation } from 'common/types'
-import * as axios from 'axios'
+import Axios from 'axios'
```
CommandMC, line 2:
```suggestion
import axios from 'axios'
```
(we've imported it as "axios" just about everywhere else, but of course this is nitpicking)

flavioislima:
> > Fix some settings appearing on the wrong platform
> 
> As already mentioned in Discord, the Sideload games feature still shows Wine options for Linux platform games.

totally forgot about that, fixed now.

flavioislima:
> Also once a game is uninstalled, it shows "not available" which it shouldn't. Just uninstalled one now and I see this.
> 
> Uninstalled_game_not_available

fixed

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -99,23 +99,25 @@ export default function GamesSettings() {
...
+            <EnableFsync />
 
-        <GameMode />
+            <GameMode />
+          </>
+        )}
 
         <UseDGPU />
 
-        <AudioFix />
```
CommandMC, line 116: Was this intentional?

CommandMC (changes requested):

src/backend/tools.ts
flavioislima: yes, never really understood why so we always need to use:
`axios.default.get` instead of only `Axios.get`

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
flavioislima: Just commented on Discord that most distros now use PipeWire and I dont really know a case where some game need this. If so, the person can always use the Env variables, what do you think?

src/backend/tools.ts
CommandMC: Oh I'm aware, I was just commenting on the name (we've used lowercase "axios" everywhere else, but capitalized "Axios" here)

src/backend/legendary/library.ts
flavioislima: yes, lets think about it in the future because it also will have downsides if the props wont be used anywhere.

src/backend/tools.ts
flavioislima: fixed

src/backend/tools.ts
flavioislima: fixed

src/backend/games.ts
flavioislima: fixed

src/backend/tools.ts
flavioislima: very nitpicking really, fixed now :P

CommandMC (approved):

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
CommandMC: Do we want to remove the actual `AudioFix` component as well then?

flavioislima: yes, will do

PR #2099, opened 2022-11-29
[Refactor] [Branch renamed] Refactoring frontend Library screen and adding mobx to state management 

misaeldossantos:
I add mobx and refactored parts of Library page. The idea is gradually moving the logic from the context api to mobx and separating core logics out of the components

---


arielj:
I have a suggestion for this, since it's a pretty big change.

Maybe we can add support for mobx (or any other state management) without migrating everything at once, then we can make smaller PRs to migrate different parts of the state.

I'm worried that with a change with this much impact the PR will be hard to review and too many things will change at the same time. It will also help with the conflicts, the longer this PR is in progress the harder will be to resolve conflicts.

If we have the setup to use the state we can gradually move things.

Maybe I'm too pessimistic but I'd prefer to do that to be more safe.

misaeldossantos:
> I have a suggestion for this, since it's a pretty big change.
> 
> Maybe we can add support for mobx (or any other state management) without migrating everything at once, then we can make smaller PRs to migrate different parts of the state.
> 
> I'm worried that with a change with this much impact the PR will be hard to review and too many things will change at the same time. It will also help with the conflicts, the longer this PR is in progress the harder will be to resolve conflicts.
> 
> If we have the setup to use the state we can gradually move things.
> 
> Maybe I'm too pessimistic but I'd prefer to do that to be more safe.

I agree, so I started with just the library page (maybe I changed it too much) to have a kind of hybrid state (legacy context api and mobx state). I talked to @flavioislima about this via discord

misaeldossantos:
I'm still working on some bugs that appeared due to my refactor, I hope that this week I can solve them all

arielj:
hey @misaeldossantos , do you have any update on this? would be great to have the setup to use a state manager so we can start moving things into it. I was checking the diff in this PR and I feel it's doing too much (looks like it includes extra stuff like renaming classes, functions, refactoring code, etc)

arielj:
hmmm rewriting the frontend sounds like too much of a change and seems unrelated to the addition of mobx here? should we close this PR for now then?

do you plan to make a PR reimplementing the complete frontend? I don't think that's a good idea, I mean you can do it if you want but I a massive PR replacing the whole frontend will be a pain to review and so many things can break, and you'll have to keep updating it with the changes that keep happening in heroic

if it's for your own fork it's ok, but if your intention is to get those changes in heroic I'd suggest to push smaller PRs instead

misaeldossantos:
Yes, we can close this PR. I agree that merging code today is a bad idea, my idea was to develop a proof of concept, and the refactoring turned out to be bigger than I thought. For now, I'll just keep my fork without merging with the existing frontend and I'll code as I can. If in the future it is possible to implement everything that heroic implements today, I can think of doing a PR and Heroic can keep both versions running in parallel (as an alternative version to the classic version, users can choose which version to install). Hypothetically we would arrive at a version in the future where we could completely replace the current frontend. But these are ideas I thought of, if that's not possible I can keep my fork separate from the "official heroic", keeping the development separate and maybe even changing the name. This is just a hobby for me, not sure if I'll be able to keep it active in the project, anyway my code is available in case anyone wants to evolve

misaeldossantos:
So as not to compromise the existing frontend code, I decided to create another folder (new-frontend) and stop making changes to the frontend folder. So if you need to merge there will be no break (in theory)

PR #2108, opened 2022-11-30
[Feat/macOS] Add Wine and DXVK support for macOS

flavioislima:
This PR adds Wine and Wineskin support to macOS.
Although, packaged applications cannot access system wine so versions like Wine-Crossover should be installed on the user applications folder so we know in case someone asks why it is not being detected. So if it is installed using BREW it needs to be moved to the user Applications folder.

Wineskin doesn't matter since Heroic will use the Wrapper inside ~/Applications/Wineskin.
Not sure if we should support the whole wrapper/prefix/bottle for WineSkin, will need to check the effort for that.

Besides that, DXVK toggle is now enabled for Wine version type on Mac and feedback was added to when it is being installed because, so the settings will be disabled and the text will change. 

Ofc the DXVK methods to get latest version and install were also changed since they differ from the Linux version.

image

image

---


flavioislima:
@Heroic-Games-Launcher/development 
Refactored how we install DXVK/VKD3D to not rely on the install script anymore since it will be removed. So if you guys cannot test it on a mac there is no problem, just check if the method is properly working on linux and the logic I added to get the Wine versions. 👍🏽

src/backend/sideload/games.ts
```diff
@@ -163,7 +177,9 @@ export async function launchApp(appName: string): Promise<boolean> {
         logWarning('File not executable, changing permissions temporarilly', {
           prefix: LogPrefix.Backend
         })
-        await chmod(executable, 0o775)
+        if (!isMac && !executable.includes('app')) {
```
CommandMC, line 180: Shouldn't the `!isMac` be enough here? The `.includes` could lead to false positives, and I don't think you can run MacOS' .app files on other OSs anyways

src/backend/tools.ts
```diff
@@ -146,51 +181,151 @@ export const DXVK = {
...
+
+      // unregister the dlls on the wine prefix
+      await new Promise((resolve) => {
+        dlls.forEach((dll) => {
+          const unregisterDll = `WINEPREFIX='${winePrefix}' '${wineBin}' reg delete 'HKEY_CURRENT_USER\\Software\\Wine\\DllOverrides' /v ${dll} /f`
+          logInfo(`Unregistering ${dll}`, { prefix: LogPrefix.DXVKInstaller })
+          exec(unregisterDll)
         })
+        resolve(true)
+      })
```
CommandMC, line 244: Are all of these `await new Promise`s necessary? Why not just do this directly?
```suggestion
    dlls.forEach((dll) => {
        const unregisterDll = `WINEPREFIX='${winePrefix}' '${wineBin}' reg delete 'HKEY_CURRENT_USER\\Software\\Wine\\DllOverrides' /v ${dll} /f`
        logInfo(`Unregistering ${dll}`, { prefix: LogPrefix.DXVKInstaller })
        exec(unregisterDll)
    })
```
`exec` will not block regardless, if that's what you wanted to do. For that to work, you'd need a callback:
```ts
exec(unregisterDll, undefined, () => resolve(true))
```
(or you just use `execAsync` & `await` it)

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -383,17 +388,25 @@ export default React.memo(function GamePage(): JSX.Element | null {
...
-                          <b>Wine:</b> {wineVersion}
-                        </div>
-                        <div
-                          className="clickable"
-                          onClick={() => window.api.openFolder(winePrefix)}
-                        >
-                          <b>Prefix:</b> {winePrefix}
-                        </div>
+                        <b>Wine:</b> {wineVersion?.name}
+                        {wineVersion && wineVersion?.type === 'crossover' ? (
```
CommandMC, line 394: `wineVersion` can't be `undefined` if you're checking it for truthy-ness first, so
```suggestion
                        {wineVersion && wineVersion.type === 'crossover' ? (
```
should be fine

src/frontend/components/UI/ToggleSwitch/index.tsx
```diff
@@ -21,12 +22,13 @@ export default function ToggleSwitch(props: Props) {
...
     extraClass,
-    description = ''
+    description = '',
+    fading
   } = props
   const { isRTL } = useContext(ContextProvider)
 
   return (
-    <>
+    <span className={classNames()}>
```
CommandMC, line 31: Won't this `classNames()` call just give you an empty string?

src/frontend/screens/Settings/components/AutoDXVK.tsx
```diff
@@ -16,17 +16,28 @@ const AutoDXVK = () => {
...
   const [winePrefix] = useSetting('winePrefix', `${home}/.wine`)
   const [wineVersion] = useSetting('wineVersion', defaultWineVersion)
+  const [installingDxvk, setInstallingDxvk] = React.useState(false)
 
-  const isProton = wineVersion.type === 'proton'
+  const dontRender =
+    wineVersion.type === 'proton' || wineVersion.type === 'crossover'
 
-  if (isProton) {
+  if (dontRender) {
```
CommandMC, line 24: Why not just do
```suggestion
  if (wineVersion.type !== 'wine') {
```

src/backend/main.ts
```diff
@@ -938,18 +938,15 @@ ipcMain.handle('requestSettings', async (event, appName) => {
...
+)
 
 ipcMain.on('toggleVKD3D', (event, { winePrefix, winePath, action }) => {
   DXVK.installRemove(winePrefix, winePath, 'vkd3d', action)
 })
 
 ipcMain.handle('writeConfig', (event, { appName, config }) => {
-  logInfo(`Writing config for ${appName === 'default' ? 'Heroic' : appName}`, {
-    prefix: LogPrefix.Backend
-  })
```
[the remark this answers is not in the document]
flavioislima: no, intentional, It is impossible to read the logs while debugging when changing config. We should log only the changed config, not the whole object.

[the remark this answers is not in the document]
CommandMC: Right, but the object gets logged in line 951. This is only the "Writing config for AppName" text

src/backend/sideload/games.ts
flavioislima: No, it is not enough because you can sideload other things on mac. Not only .app.
So it will be on mac and needs to have .app on the name, if there it not. A bin, a python script or anything else, we would not run that.

src/backend/tools.ts
flavioislima: We need the await to avoid the racing condition. We can start copying before backing up, so we will overwrite it. This happened during my tests.

The registering is not the issue, we can register the DLLs even if they do not exist in the path, or of they are the wrong ones. The problem is only with backing up them. But I wanted to do it in the right order so on the logs we can have the correct execution.

src/frontend/screens/Game/GamePage/index.tsx
flavioislima: Ok. Wont commit your suggestion because Gihub always messes up the lint when I do that, but I will change and push it.

src/frontend/components/UI/ToggleSwitch/index.tsx
flavioislima: Hmm, I think something was there and I removed the class and forgot the call. good catch

src/frontend/screens/Settings/components/AutoDXVK.tsx
flavioislima: Ah, the thing here is that I wanted to support a new type called `wineskin`, but the maintainer said that he will change how the wrappers works and will bring all the features to the main bin, I don't know how. so that wont be necessary. So yes, your solution will work in this case.

src/backend/config.ts
```diff
@@ -150,7 +150,74 @@ abstract class GlobalConfig {
...
+   *
+   * @returns Promise<Set<WineInstallation>>
+   */
+  public async getWineOnMac(): Promise<Set<WineInstallation>> {
+    const wineSet = new Set<WineInstallation>()
+    if (!isMac) {
+      return wineSet
+    }
+    const { stdout } = await execAsync(`ls ${userHome}/Applications`)
+    const apps = stdout.split('\n')
```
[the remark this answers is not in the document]
flavioislima: yes, will change it

src/backend/tools.ts
CommandMC: Right, but this does not change this. The race condition is caused by `exec` not blocking, and just putting it in a Promise won't change that (at least if you're doing it like that)

A simple example:
```ts
import { exec } from 'child_process'

await new Promise<void>((res) => {
  exec('sleep 5', undefined, () => console.log('Process done'))
  res()
})
console.log('Promise resolved')
```
Output:
```
$ ts-node-esm index.ts
Promise resolved
Process done
```

If you want to wait for the process to finish, either resolve the Promise in the callback function of `exec`, or just use `execAsync`

src/backend/sideload/games.ts
CommandMC: I'm confused. This condition is saying "Run `chmod` if we're not on MacOS & the executable path does not contain `app`". That does not sound like what you describe there.

Seems to me like swapping the `&&` to an `||` *would* be what you want though (always run it if we're not on MacOS, or if we are on MacOS, run it if the target's not an `app`)

src/backend/main.ts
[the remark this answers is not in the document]
flavioislima: Yes, got confused. Now I added a way of logging only the changed config. It is a lot better.

src/backend/sideload/games.ts
flavioislima: Yes, you are right. fixed now.

src/backend/tools.ts
flavioislima: Yes, funny thing is that I did use `execAsync` when registering but when registering I used `exec`.
I should stop programming 18h a day 🤦🏽

src/backend/tools.ts
```diff
@@ -218,9 +219,13 @@ export const DXVK = {
           const dllPath64 = `${winePrefix}/drive_c/windows/syswow64/${dll}.bak`
           logInfo(`Restoring ${dll}`, { prefix: LogPrefix.DXVKInstaller })
           if (existsSync(dllPath)) {
-            renameSync(dllPath, `${winePrefix}/drive_c/windows/system32/${dll}`)
+            if (existsSync(dllPath)) {
```
CommandMC, line 222: Aren't we checking this twice now?

src/backend/tools.ts
CommandMC: It's impossible to know/remember everything, especially about a language as vast as JS/TS. That's why reviews exist, after all

CommandMC (approved):

Issue #2098, opened 2022-11-29
Launcher Connectivity issue


Tear4Pixelation:
### Describe the bug

The launcher says I'm offline although I have perfect internet. Also when I launch a game, it says I'm not logged in to epic, but I logged in in the launcher.
This issue only occurs in 2.5 but the old 2.4.3 stopped working with the release of the new version.
Thank you for your time and have a great day!

### Add logs

```text
(15:07:03) INFO:    [Connection]:       Connectivity: check-online
(15:07:03) INFO:    [Connection]:       Pinging external endpoints
(15:07:03) INFO:    [Legendary]:        Running command: /tmp/.mount_HeroicA3zkh6/resources/app.asar.unpacked/build/bin/linux/legendary --version
(15:07:03) INFO:    [Legendary]:        Legendary location: /tmp/.mount_HeroicA3zkh6/resources/app.asar.unpacked/build/bin/linux/legendary
(15:07:03) INFO:    [Gog]:              GOGDL location: /tmp/.mount_HeroicA3zkh6/resources/app.asar.unpacked/build/bin/linux/gogdl
(15:07:04) WARNING: [Backend]:          Failed to register protocol with OS.
(15:07:04) INFO:    [Gog]:              Running command: /tmp/.mount_HeroicA3zkh6/resources/app.asar.unpacked/build/bin/linux/gogdl --version
(15:07:04) INFO:    [Backend]:          

Heroic Version: 2.5.0 Yamato
Legendary Version:  0.20.31 Dark Energy (hotfix #5)
GOGdl Version: 0.5
OS: Ubuntu KERNEL: 5.15.0-52-generic ARCH: x64
CPU: Intel Core™ i7-3770K @3.5 GOVERNOR: schedutil
RAM: Total: 23.43 GiB Available: 20.34 GiB
GRAPHICS: GPU0: NVIDIA GeForce GTX 1050 Ti VRAM: 4036MB DRIVER: 470.141.03 
PROTOCOL: x11

(15:07:05) INFO:    [Backend]:          Checking for new Heroic Updates
(15:07:05) INFO:    [Connection]:       All ping requests failed:
(15:07:05) INFO:    [Connection]:       AggregateError: All promises were rejected
(15:07:05) INFO:    [Backend]:          Checking for current version changelog
(15:07:05) INFO:    [Frontend]:         Refreshing Library
(15:07:05) INFO:    [Legendary]:        Refreshing library...
[...]

PR #2113, opened 2022-12-01
[Connectivity] Connectivity check tweaks: ignore button, more ways to check, longer timeout

arielj:
This PR adds some of the ideas from this comment https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2098#issuecomment-1332970241

- increased timeout to 10 seconds
- when the webview opens a page successfully, tell the backend that we are online
- added a button to ignore the `Offline` message

I didn't add the option to disable the check for now, I think it can lead to issues and we should instead use other ideas places that we can use to detect connectivity (like the webview, maybe we can check if an image is loaded from an external source for without cache for example in the future)

---


flavioislima (approved):
LGTM

Issue #2112, opened 2022-11-30
Cannot install Roller Coaster Tycoon (GOG)


haffmans:
### Describe the bug

When trying to install Roller Coaster Tycoon from GOG, the dialog keeps spinning with the "Geting download size..." and "Geting install size..." 

The log shows it's trying to fetch the info:
```
(23:16:28) INFO:    [Gog]:              Getting game metadata: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl info 1207658945 --token <redacted> --lang=en-US --os windows
(23:16:28) INFO:    [Gog]:              Getting game metadata: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl info 1207658945 --token <redacted> --lang=en-US --os windows
```

The output of this command is a download and disk size of 0:

```
API] INFO: Getting info from products endpoint for id: 1207658945
[DOWNLOAD_MANAGER] INFO: Checking compatibility of Roller Coaster Tycoon® Deluxe with windows
[DOWNLOAD_MANAGER] INFO: Game is compatible
[DOWNLOAD_MANAGER] INFO: Depot version: 2
[API] INFO: Getting Dependencies repository
[API] INFO: Getting repository manifest
[DOWNLOAD_MANAGER] INFO: Calculating download size
{"download_size": 0, "disk_size": 0, "dlcs": [], "buildId": "50860644332681884", "languages": ["de-de", "en-us", "es-es", "fr-fr", "it-it"], "folder_name": "RollerCoaster Tycoon Deluxe", "versionEtag": "\"4159993833\"", "versionName": "1.20.015"}
```

Changing the language argument of gogdl to `en-us` (lowercase), the result works:

```
[API] INFO: Getting info from products endpoint for id: 1207658945
[DOWNLOAD_MANAGER] INFO: Checking compatibility of Roller Coaster Tycoon® Deluxe with windows
[DOWNLOAD_MANAGER] INFO: Game is compatible
[DOWNLOAD_MANAGER] INFO: Depot version: 2
[API] INFO: Getting Dependencies repository
[API] INFO: Getting repository manifest
[DOWNLOAD_MANAGER] INFO: Calculating download size
[...]

PR #2114, opened 2022-12-01
[GOG] Retry GOG install info if invalid language

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2112 (and probably that problem for other games).

We were passing `en-US` for all games when getting the download info, but some games like Roller Coaster Tycoon display `0`. When using `en-us` it works fine.

What this PR is doing is:
- first try with `en-US` (most games are fine with this from my tests)
- if the list of languages in the install info does NOT include `en-US`, check which languages are valid
- if `en-us` is a valid language, use it, else, use the first language in the list

---

### DIFF
--- a/src/backend/gog/library.ts
+++ b/src/backend/gog/library.ts
@@ -1,3 +1,4 @@
+import { install } from 'frontend/helpers/library'
 import axios, { AxiosError, AxiosResponse } from 'axios'
 import { GOGUser } from './user'
 import {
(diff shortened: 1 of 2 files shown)


src/backend/gog/library.ts
```diff
@@ -316,7 +316,8 @@ export class GOGLibrary {
    */
   public async getInstallInfo(
     appName: string,
-    installPlatform = 'windows'
+    installPlatform = 'windows',
```
flavioislima, line 319: I believe we could also use the platform we have on context, this could improve the re-render process I guess.

flavioislima (approved):

PR #2119, opened 2022-12-01
[Logs] Add GOGdl and Legendary verbose log for any command output for better troubleshooting

arielj:
This PR adds new log files to store all the output of Legendary and GOGdl.

The new files are stored next to heroic's general logs but the names are `legendary-....log` and `gogdl-.....log` instead of `heroic-....log` and the same logic to delete old logs still applies to these new ones.

There's no frontend view of these logs but they can be accessed using the same `Show log file in folder` button for the general logs.

I also fixed an issue where logs were being deleted when the month changed instead of when they were a month old (the 1st of each month it was we were deleting all files from last month).

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -674,6 +674,13 @@ async function callRunner(
     })
   }
 
+  if (options?.verboseLogFile) {
+    appendFileSync(
+      options.verboseLogFile,
+      `[${new Date().toLocaleString()}] ${safeCommand}\n`
+    )
+  }
+
   if (options?.logFile && existsSync(options.logFile)) {
     writeFileSync(options.logFile, '')
   }
@@ -705,6 +712,10 @@ async function callRunner(
         appendFileSync(options.logFile, data)
       }
 
+      if (options?.verboseLogFile) {
+        appendFileSync(options.verboseLogFile, data)
+      }
+
       if (options?.onOutput) {
         options.onOutput(data, child)
       }
@@ -716,6 +727,10 @@ async function callRunner(
     child.stderr.on('data', (data: string) => {
       if (options?.logFile) {
         appendFileSync(options.logFile, data)
+      }
+
+      if (options?.verboseLogFile) {
+        appendFileSync(options.verboseLogFile, data)
       }
 
       if (options?.onOutput) {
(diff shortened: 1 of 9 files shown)


src/backend/launcher.ts
```diff
@@ -737,6 +737,13 @@ async function callRunner(
         rej('Process terminated with signal ' + signal)
       }
 
+      if (options?.runnerStdoutLog) {
```
arielj, line 740: stores verbose logs if an option is provided

src/backend/logger/logfile.ts
```diff
@@ -19,50 +19,61 @@ import { logError, LogPrefix } from './logger'
...
+        if (log.match(/(heroic|legendary|gogdl)-/)) {
           const dateString = log
-            .replace('heroic-', '')
+            .replace(/(heroic|legendary|gogdl)-/, '')
             .replace('.log', '')
             .replaceAll('_', ':')
           const logDate = new Date(dateString)
-          if (
-            logDate.getFullYear() < date.getFullYear() ||
-            logDate.getMonth() < date.getMonth()
```
arielj, line 64: this was deleting files if they were from last month instead of `a month old`

src/backend/launcher.ts
```diff
@@ -737,6 +737,13 @@ async function callRunner(
...
+        const f = options.runnerStdoutLog
+        appendFileSync(f, `[${new Date().toLocaleString()}] ${safeCommand}\n`)
```
CommandMC, line 742: It might be a good idea to do this before any command runs...

src/backend/launcher.ts
```diff
@@ -737,6 +737,13 @@ async function callRunner(
...
+        appendFileSync(f, stdout.join('\n') + '\n\n')
+        appendFileSync(f, stderr.join('\n') + '\n\n')
```
CommandMC, line 744: ...and these in the `data` callback above

Two reasons:
- If the Child Process is terminated/fails for some other reason, the `rej` above makes us skip writing the log
- All stdout will always be written after stderr, which might make troubleshooting more difficult

CommandMC (changes requested):

src/backend/launcher.ts
arielj: done

src/backend/launcher.ts
arielj: done

arielj:
thanks @CommandMC this is ready for another review

CommandMC (approved):

Issue #2115, opened 2022-12-01
Version 2.50 game can be launched multiple times and run in background (Steam Deck)


mwhesotech:
### Describe the bug

In Game Mode with Steam Deck:

When I start a game (tested with Witcher 3 from gog.com) from the HeroicGamesLauncher start screen with the Y key, 
nothing seems to happen (for about 30 seconds) until the game starts.

If I press the Y key several times, the game also starts several times. 
And at some point a message appears saying xx is already started.

You cannot see that the game is already started, because is runs behind the HeroicGamesLauncher.
(I have selected the minimize after start option in the HGL settings)

Workaround:
When I click the game, open the details page and start the game from there with a touch click, everything works as expected.

In Version 2.4.X all this worked without problems. And the start was a lot faster ca. 10 sec. to about 30-40 sec. in Version 2.5.

### Add logs

```text
no logs
```

### Steps to reproduce

1. Start HeroicGamesLauncher from Steam (Steam Deck Game Mode)
2. Select the game with the builtin gamepad buttons (Witcher 3 in my case)
3. Press Y to start the game

### Expected behavior

- The game is only started once. 
- HGL shows that the game is launched and doesn't allow a second lauch.
- HGL is minimized / hidden after I click start as in Version 2.4.X

### System Information

- SteamOS on Steam Deck in Game Mode
- Heroic Version 2.50

flavioislima:
Tested here and thats true, although for me the launch of the witcher 3 took only 4 seconds (it is installed on the SSD). But in the meantime, it is possible to launch the game again so we should disable the input after launching.

flavioislima:
@arielj @wbrtm maybe something for you guys to take a look at when you have the time.

wbrtm:
Can someone test if clicking the play button in desktop mode library view (not game page) does have the same bug/results?

mwhesotech:
In desktop mode it is the same.

arielj:
This started with this change https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2045/files#diff-438ee3761ea4339bf45ec2b1f10f5581dd13538338f97f9472ef681c198543fcR484

There are 2 issues:
- that code is only checking if the game supports cloud saves and it's triggered even if the user disables cloud saves
[...]

PR #2120, opened 2022-12-02
[Library] Disable Play button while launching game

arielj:
This is a workaround for https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2115

There are two changes:
- added a new `isLaunching` state in the card component to be able to disable the button while we wait for the game to start (we should detect that it's doing a save sync and show that in the UI)
- only do sync saves if the user enables the option instead

---


flavioislima (approved):
Cool, thanks for fixing that.

PR #2138, opened 2022-12-04
[Tech] Adds useLibrary hook to manage the library

flavioislima:
This PR adds a new hook called `useLibrary` so this way we can get rid of it on the context and improve the loading and avoid re-renders.
Most of the computation is done by the hook as well as changing the category.

the library component is still handling the other filters since I could not think the performance will improve by adding them to the hook, but we can discuss that to see what is the best approach.

---

---

### DIFF
--- a/src/frontend/hooks/useLibrary.ts
+++ b/src/frontend/hooks/useLibrary.ts
@@ -23,7 +23,10 @@ const useLibrary = ({ category }: Props): Array<GameInfo> => {
 
   const getLibrary = (): Array<GameInfo> => {
     const games = gogLibraryStore.get('games', []) as GameInfo[]
-    const installedGames = gogInstalledGamesStore.get('installed', []) as Array<InstalledInfo>
+    const installedGames = gogInstalledGamesStore.get(
+      'installed',
+      []
+    ) as Array<InstalledInfo>
     for (const igame in games) {
       for (const installedGame of installedGames) {
         if (installedGame.appName === games[igame].app_name) {
--- a/src/frontend/hooks/useLibrary.ts
+++ b/src/frontend/hooks/useLibrary.ts
@@ -1,0 +1,72 @@
+import { Category } from 'frontend/types'
+import React from 'react'
+
+import { GameInfo, InstalledInfo } from 'common/types'
+import {
+  gogLibraryStore,
+  gogInstalledGamesStore,
+  sideloadLibrary,
+  libraryStore
+} from 'frontend/helpers/electronStores'
+
+type Props = {
+  category: Category
+}
+
+const useLibrary = ({ category }: Props): Array<GameInfo> => {
+  const [library, setLibrary] = React.useState<Array<GameInfo>>([])
+  const [gogLibrary, setGogLibrary] = React.useState<Array<GameInfo>>([])
+  const [epicLibrary, setEpicLibrary] = React.useState<Array<GameInfo>>([])
+  const [sideloadedLibrary, setSideloadedLibrary] = React.useState<
+    Array<GameInfo>
+  >([])
+
+  const getLibrary = (): Array<GameInfo> => {
+    const games = gogLibraryStore.has('games')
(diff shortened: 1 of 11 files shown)


flavioislima:
Test Checklist:
- [ ]  Should load all libraries
- [ ] Library filter should work
- [ ] Card should update after install finished
- [ ] Card should update on uninstall
- [ ] Search bar should work as before
- [ ] Other filters and sorting should work as before
- [ ]

src/frontend/hooks/useLibrary.ts
```diff
@@ -0,0 +1,72 @@
...
+  const [gogLibrary, setGogLibrary] = React.useState<Array<GameInfo>>([])
+  const [epicLibrary, setEpicLibrary] = React.useState<Array<GameInfo>>([])
+  const [sideloadedLibrary, setSideloadedLibrary] = React.useState<
+    Array<GameInfo>
+  >([])
+
+  const getLibrary = (): Array<GameInfo> => {
+    const games = gogLibraryStore.has('games')
+      ? (gogLibraryStore.get('games', []) as GameInfo[])
+      : []
```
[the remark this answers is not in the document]
flavioislima: good point

arielj:
One concern I have about this, let's say I am in the Library screen which has these components: RecentlyPlayed, Library, SearchBar and SidebarLinks. Each of those components uses the `useLibrary` hook, each has to load the information from the stores when rendering this page, and the library is in memory multiple times.

Related to that, each time we change screens we have to read the stores again when a component needs to be created.

I wonder if there's a way to keep only one copy of the library in memory and use that in the useLibrary hook and only refresh it when needed.

src/frontend/hooks/useLibrary.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const useLibrary = ({ category }: Props): Array<GameInfo> => {
+  const [library, setLibrary] = React.useState<Array<GameInfo>>([])
+  const [gogLibrary, setGogLibrary] = React.useState<Array<GameInfo>>([])
+  const [epicLibrary, setEpicLibrary] = React.useState<Array<GameInfo>>([])
+  const [sideloadedLibrary, setSideloadedLibrary] = React.useState<
+    Array<GameInfo>
+  >([])
+
+  const getLibrary = (): Array<GameInfo> => {
```
arielj, line 24: maybe we can rename this as `getGogLibrary`

src/frontend/hooks/useLibrary.ts
```diff
@@ -0,0 +1,71 @@
...
+        setLibrary(gogLibrary)
+        break
+      case 'legendary':
+        setLibrary(epicLibrary)
+        break
+      case 'sideload':
+        setLibrary(sideloadedLibrary)
+        break
+      default:
+        setLibrary([])
```
arielj, line 63: should we fallback to all the games instead of an empty array? I mean, if something goes wrong we would show no games, and we have had issues with an empty least because of some issue with the category.

I agree that it's more correct to return `[]`, but maybe it's safer to fallback to all on something unexpected

flavioislima:
> One concern I have about this, let's say I am in the Library screen which has these components: RecentlyPlayed, Library, SearchBar and SidebarLinks. Each of those components uses the `useLibrary` hook, each has to load the information from the stores when rendering this page, and the library is in memory multiple times.
> 
> Related to that, each time we change screens we have to read the stores again when a component needs to be created.
> 
> I wonder if there's a way to keep only one copy of the library in memory and use that in the useLibrary hook and only refresh it when needed.

Good point. I will improve this PR soon. I found a few bugs as well, especially if you log out of one store. All of them will disappear.

I think if we use memo on the library it won't reload when changing screens, I mean, even now it doesn't. The only thing that makes the library re-render is changing the props it uses from the context.

So this would be the first step, but I have some ideas to remove other things from context and in the future have only basic stuff there.

Issue #2139, opened 2022-12-04
dependency "vite" not listed in build instructions


lever1209:
### Describe the bug

dependency "vite" not listed in build instructions

### Add logs

```text
yarn dist:linux

yarn run v1.22.19
$ vite build && electron-builder --linux
/bin/sh: line 1: vite: command not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Steps to reproduce

1. open readme.md
2. go to build instructions
3. read steps

### Expected behavior

it should be there but is not

### System Information

- archlinux
- 2.5.0

PR #2140, opened 2022-12-05
include vite in build instructions

lever1209:
fix for #2139

adds dependency vite to build instructions

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ Otherwise, download the file ending in .dmg from the [latest release](https://gi
 
 This part will walk you through setting up a development environment so you can build Heroic binaries yourself or make changes to the code.
 
-1. Make sure Git, NodeJS, and Yarn are installed
+1. Make sure Git, NodeJS, Vite (via npm), and Yarn are installed
 2. Clone the repo and enter the cloned folder, for example with these commands:
 
    ```bash


CommandMC:
Uh, Vite should be auto-installed when running `yarn`/`npm install`, no?

lever1209:
it may have at some point, but it hasnt during my testing

lever1209:
interesting, i think vite should still be listed in some way

i wasnt aware running yarn by itself sorted out dependencies, but doesnt do so when building like how gradle does

CommandMC:
> interesting, i think vite should still be listed in some way

Well, it's listed in the "Web Technologies" section of the README. Other than that, it's not really anything special. You'd get similar errors if you'd miss any other package in the dependencies

> i wasnt aware running yarn by itself sorted out dependencies, but doesnt do so when building like how gradle does

The lines right below the ones you edited mention this:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/aa98cf2455cd2e29a9f7819b3d835f03c4abc5aa/README.md?plain=1#L216

lever1209:
i was under the impression all it would do is the same as running `yarn dist:linux` where instead of building something, it would just verify deps and the project, not install deps as well

lever1209:
for all i knew vite wasnt even needed because it wasnt listed, so if i manually installed all the deps listed, yknow, i wouldnt have to verify the deps are installed and it would work

anyway, ill just leave it up to the maintainers, maybe they can come up with a better solution

CommandMC:
> for all i knew vite wasnt even needed because it wasnt listed

Every dependency (along with its compatible version) is listed in `package.json`. But you don't have to worry about this, since `yarn`/`npm install` will install everything for you

> so if i manually installed all the deps listed, yknow, i wouldnt have to verify the deps are installed and it would work

And if you instead just *ran the command that the README tells you to run*, you wouldn't have to worry about any of this. This is just simply how this system works, it isn't our fault if you don't follow the instructions, be it on purpose or because of a misunderstanding

PR #2152, opened 2022-12-06
[i18n] correct Geting typo

paperbenni:
I usually don't do stupid pull requests like this, but this typo was annoying me enough to do this


flavioislima:
We cannot change the translations Jsons like this, you need to go to our Weblate and make a suggestion there.

https://hosted.weblate.org/projects/heroic-games-launcher/

totoroot:
@paperbenni It was bugging me as well, so I just fixed the source string on Weblate, but I'm not sure how long it will take until the change makes it into a release.

PR #2153, opened 2022-12-07
[Tech] update electron to 22.0.2

Etaash-mathamsetty:
yet another electron update 

### DIFF
--- a/package.json
+++ b/package.json
@@ -212,7 +212,7 @@
     "@typescript-eslint/eslint-plugin": "^5.20.0",
     "@typescript-eslint/parser": "^5.20.0",
     "@vitejs/plugin-react": "^2.2.0",
-    "electron": "^21.2.1",
+    "electron": "^22.0.2",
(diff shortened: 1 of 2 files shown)


flavioislima:
Can you add the change log for it when you open these PRs? So it is easier for us to review. 👍

Etaash-mathamsetty:
https://github.com/electron/electron/releases/tag/v22.0.0

flavioislima:
Ok but I think I will keep the current version on 2.5 and update Electron on the next major release. We had issues on Linux in the past by using the latest Electron version so we need to be more cautious and test in more distros.

Etaash-mathamsetty:
> Ok but I think I will keep the current version on 2.5 and update Electron on the next major release. We had issues on Linux in the past by using the latest Electron version so we need to be more cautious and test in more distros.

yeah this was more targeted towards beta, it's just that I don't think work on the next beta has started

flavioislima (approved):

PR #2165, opened 2022-12-09
[FIX] Crossover not found on macOS

flavioislima:
The promise to get all wine versions for macOS was failing when no WineSkin folder was found due to a missing check to see if the folder exists or not.

---

### DIFF
--- a/src/backend/config.ts
+++ b/src/backend/config.ts
@@ -191,25 +191,30 @@ abstract class GlobalConfig {
     if (!isMac) {
       return wineSet
     }
-    const apps = readdirSync(`${userHome}/Applications/Wineskin`)
-    for (const app of apps) {
-      if (app.includes('.app')) {
-        const wineBin = `${userHome}/Applications/Wineskin/${app}/Contents/SharedSupport/wine/bin/wine64`
-        try {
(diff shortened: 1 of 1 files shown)


src/backend/config.ts
```diff
@@ -191,25 +191,30 @@ abstract class GlobalConfig {
...
-            name: `Wineskin - ${version}`,
-            type: 'wine',
-            bin: wineBin
-          })
-        } catch (error) {
-          logError(`Error getting wine version for ${wineBin}`, {
-            prefix: LogPrefix.GlobalConfig
-          })
+    const wineSkinPath = `${userHome}/Applications/Wineskin`
+    if (existsSync(wineSkinPath)) {
```
arielj, line 195: just a commend on this, maybe nit picky, but if you do something like:

```
if (existsSync(wineSkinPath)) {
  return wineSet
}

const apps = ....
```

then it will be easier to track changes on lines, if you nest the code in the if all lines will show up as if you changed them in this commit when it was actually just moved inside the if

not a big deal though (I do prefer early returns, but it's just my preference)

arielj (approved):
looks good, I just added a minor comment

PR #2166, opened 2022-12-09
[Fix] DXVK + Winetricks on macOS

flavioislima:
- Fix dlls overrides for dxvk
- use `wineboot -u` to restore dlls
- remove unnecessary promises
- Fix winetricks envs

---


src/backend/tools.ts
```diff
@@ -212,123 +217,93 @@ export const DXVK = {
...
     // register dlls on the wine prefix
-    await new Promise((resolve) => {
-      dlls.forEach(async (dll) => {
-        await execAsync(
-          `WINEPREFIX='${winePrefix}' '${wineBin}' reg add 'HKEY_CURRENT_USER\\Software\\Wine\\DllOverrides' /v ${dll} /d native /f `,
-          execOptions
-        )
-          .then(() => {
-            logInfo(`${dll} registered!`, {
+    const overrideType = isMac ? 'native,builtin' : 'native'
```
CommandMC, line 282: `native,builtin` is the "default" DLL override as far as I know, so DLL overrides shouldn't be necessary at all on MacOS

I still wonder if they could also be skipped on Linux, *in theory* they shouldn't do anything (since the default load order is "native,builtin" as mentioned above, and a `native` DLL will now always be found)

flavioislima: Still, it's not causing any harm adding this here.

The point is. It is native only if you have both dlls for 32 and 64 bits. Native built-in it's when yiu have only one of them.

I think it's no harm to keep this until we proper check on both platforms that it's fine

flavioislima: Ok, removed then.

PR #2168, opened 2022-12-10
[WineManager] Update wine item status after install/uninstall

### LINKED ISSUE #2159: Wine versions show up as not installed after installed and as installed after uninstall
### Describe the bug

When installing a new Wine version with the Version Manager, after the installation ends, the wine version shows up as not installed.

Similar issue happens after uninstalling a wine version, it still shows up as installed.

The `refresh` button does not fix the state of the list.

Restarting Heroic shows the correct state.

### Add logs

```text
Nothing relevant, it is installing/uninstalling it correctly.

This started in 2.5.1
```

### Steps to reproduce

1. Go to Wine Manager
2. Install any wine
3. Wait for the process to finish
4. It shows as not installed
5. Restart Heroic
6. It shows as installed

### Expected behavior

Status should be installed after install and not installed after uninstall.

### Screenshots

_No response_

### System Information

- Heroic Version [e. g. 2.1.1]: 2.5.1

### Additional information

_No response_

arielj:
This PR fixes #2159

After installing/uninstalling/refreshing, a message is sent from the backend to the frontend to re-read the store.

---

### DIFF
--- a/src/backend/wine/manager/utils.ts
+++ b/src/backend/wine/manager/utils.ts
@@ -7,6 +7,7 @@ import Store from 'electron-store'
 import { existsSync, mkdirSync, rmSync } from 'graceful-fs'
 import { logError, logInfo, LogPrefix, logWarning } from '../../logger/logger'
 import { WineVersionInfo } from 'common/types'
+import { BrowserWindow } from 'electron'
 
 import {
   getAvailableVersions,
@@ -88,6 +89,7 @@ async function updateWineVersionInfos(
   }
 
   logInfo('wine versions updated', { prefix: LogPrefix.WineDownloader })
+  BrowserWindow.getAllWindows()[0].webContents.send('wineVersionsUpdated')
   return releases
 }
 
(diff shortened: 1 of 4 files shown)


src/frontend/screens/WineManager/index.tsx
```diff
@@ -113,10 +127,7 @@ export default React.memo(function WineManager(): JSX.Element | null {
             {!refreshing &&
               !!wineVersions.length &&
               wineVersions.map((release, key) => {
-                if (release.type === repository) {
```
arielj, line 116: I'm moving the filter when setting the array so we don't have to loop through all the elements in every render

src/backend/wine/manager/utils.ts
```diff
@@ -88,6 +89,7 @@ async function updateWineVersionInfos(
   }
 
   logInfo('wine versions updated', { prefix: LogPrefix.WineDownloader })
+  BrowserWindow.getAllWindows()[0].webContents.send('wineVersionsUpdated')
```
CommandMC, line 92: Shame we can't use `getMainWindow()` here (I assume Jest doesn't like all of our aliased imports in `utils`)

arielj: I'm planning to refactor all these calls creating something like a `frontendEvents` module with `send` function to hide this implementation detail so we can use it always instead, it will also help with testing (to test something like this we have to mock 3 different modules for example)

flavioislima (approved):
LGTM

PR #2121, opened 2022-12-02
[Tech] Don't rely on `setup_dxvk.sh` and similar setup scripts when installing tools

CommandMC:
Starting work on a manual DXVK installer (and DLL installer in general), to no longer depend on the (now removed) `setup_dxvk.sh` script. Still lots to do, but the big parts are done

This is designed to be expanded easily, to eventually become the "Game Workarounds" feature

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #2176, opened 2022-12-11
[Feature] [macOS/Linux] Workarounds

CommandMC:
Picking up where we left off in #2121, this PR will (when it's done) implement the Workarounds system described in https://github.com/WineGameDB/db

TODO:
- [ ] Download workarounds DB
- [ ] Display list of all workarounds in game settings (ordered in Required -> Optional -> Not listed).
Note that some filtering has to be done here, since some workarounds (copying files, setting environment variables) don't make sense here
- [ ] Display list of required & optional workarounds for the given game when creating prefixes

Workarounds themselves should be able to be...
- [X] Installed
- [X] Removed
- [X] Queried (checking if they're installed)
- [X] Updated

I'd also like to have some
- [X] better types

since right now, nothing specifies that a Workaround has to have certain methods & you'll get errors in the wrong places if you miss one.

---


arielj:
@CommandMC can we close this PR? just to clean up the list of PRs

CommandMC:
Oh I'm still planning on continuing this. Currently, this is blocked by us not supporting newer TypeScript versions (#3218)

arielj:
I was reading this and I have some questions:
- what is the idea of the usage of this? because I see that workarounds can be installed and removed and updated, seems like a manual process for the user to actually apply the workaround?
- I see some code doing things like enabling dxvk/vkd3d, we do that by default already, is that some leftover from old heroic when we didn't do that by default?

at least in my experience, many games can be fixed by running one or two winetricks commands, so I was thinking about having something really simple as a first version:
- while a game is being installed, we query some place to get the list of winetricks packages needed and store that info somewhere
- when the prefix is created, we check if we have that information somewhere and install those packages

that should solve a big % of the games and we can then focus in more complex workarounds? because I see the definition in the winegamedb all cases but some seem to be uncommon or optional

CommandMC:
> what is the idea of the usage of this? because I see that workarounds can be installed and removed and updated, seems like a manual process for the user to actually apply the workaround?

The general idea of this is a system alike to Lutris install scripts. However, individual workarounds might also be helpful for the user (as an example, DXVK and VKD3D-Proton).  
They are explicitly *not* Winetricks verbs, as Winetricks doesn't cover everything, and not everything is in the "format" of Winetricks (for example, the EAC runtime just isn't compatible with the "run some commands, and now *thing* is installed" model). We could of course still implement some Workarounds as "run Winetricks with these parameters" behind the scenes.

> I see some code doing things like enabling dxvk/vkd3d, we do that by default already, is that some leftover from old heroic when we didn't do that by default?

For one, *I think* the code is that old, but this system also offers a 3-state system for game "install scripts". Let's take DXVK for example: The system currently implemented in Heroic always respects user choice, while the intent of workarounds is that DXVK can either be required (if WineD3D won't be able to run this game), optional (both DXVK and WineD3D would work with the game) or disabled (DXVK causes issues, only WineD3D works).

> at least in my experience, many games can be fixed by running one or two winetricks commands, so I was thinking about having something really simple as a first version

The point of this PR is to focus on these uncommon cases, and to be as-extensible-as-possible for future game releases.  
Focusing on just the common cases wouldn't be that much easier actually. The system itself is the complicated part; whether we have 3 or 20 workarounds isn't a concern. 

Of course, we could go with a really simple system for "just enough" to work, but I'd prefer not rewriting the whole thing once we need some new functionality.

CommandMC:
With us now having a system for most common workarounds (https://github.com/Heroic-Games-Launcher/known-fixes) and [ULWGL](https://github.com/Open-Wine-Components/ULWGL-protonfixes) eventually covering everything else, I don't think it makes sense to work on a 3rd system (which would then of course be incompatible with other launchers)
A lot of the other technical changes in this PR are now also either already implemented or obsolete

o7 WineGameDB, long live ULWGL

PR #2180, opened 2022-12-11
[UI/UX] Cache HowLongtoBeat Data and move squares down

flavioislima:
Moved the squares with the info a bit down since it was causing a annoying layout shift when opening the game page, even with cached data.

Screenshot 2022-12-11 at 16 49 49

---

### DIFF
--- a/src/backend/howlongtobeat/utils.ts
+++ b/src/backend/howlongtobeat/utils.ts
@@ -18,7 +18,7 @@ export async function getHowLongToBeat(
       return cachedResponse as HowLongToBeatEntry
     }
 
-    logInfo([`Getting HowLongToBeat data for ${title}`, ''], {
+    logInfo(`Getting HowLongToBeat data for ${title}`, {
       prefix: LogPrefix.Backend
     })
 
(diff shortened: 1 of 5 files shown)


flavioislima:
> Tested here, works fine!

Can you approve it?

src/backend/howlongtobeat/utils.ts
```diff
@@ -1,12 +1,35 @@
...
 export async function getHowLongToBeat(
   title: string
 ): Promise<HowLongToBeatEntry | null> {
   try {
+    title = removeSpecialcharacters(title)
+
+    // check if we have a cached response
+    const cachedResponse = howLongToBeatStore.get(title)
+    if (cachedResponse) {
+      logInfo([`Using cached HowLongToBeat data for ${title}`, ''], {
```
arielj, line 15: Just being nit picky here, but I think this is supposed to be used like:

```ts
logInfo(['Using cached HowLongToBeat data for' title], {
```

arielj (approved):
Looks good, I just added one comment.

Should this be merged against `stable` instead? seems like a small enhancement that could go in the next stable release

src/backend/howlongtobeat/utils.ts
flavioislima: I think I started one approach, changed it and left that empty string there. Good catch.

flavioislima:
If I choose `stable` to merge it will bring changes from the main branch so I will do it in `main` and then cherry-pick to the `stable` branch

PR #2181, opened 2022-12-11
[FIX] Not being able to change GOG saves path

flavioislima:
Small fix, since 2.5.1 for some reason GOG saves were stuck, not sure if the best approach since I am assuming the selected path should always be the first element of the array. I mean, since the user defined it, so I believe that makes sense. But the auto discovery is working fine as well.

---


CommandMC:
The problem isn't that we're not saving the saves correctly (that was actually perfectly fine, and these changes are... not how that works). The issue is that we're not re-rendering the component after we change
Try deleting the save path & clicking "Save-Sync" in the sidebar again, you'll see that your changes are now reflected

flavioislima:
@CommandMC it is not possible to delete the path. Have you tried to replicate the bug?

I will see if I can find another approach but what's wrong with that?

In my conception, in the array of saves, the user defined should be always the first item since should have priority.

PR #2182, opened 2022-12-11
Theme selector in general settings

arielj:
This PR adds the theme-related options in the General Settings screen.

It was reported a few times that some people cannot find the theme selector because they expect it to be in the general settings and don't think it would be in the accessibility screen. With this change it shows in both places so it's in a common place for most users and it's also in the Accessibility screen for users that need to change the theme for accessibility reasons.

image

---


flavioislima (approved):
LGTM

PR #2184, opened 2022-12-12
[UI] Update Tray icon when recent games change

arielj:
This PR includes a fix for the tray icon not updating when the list of recent games changes until heroic was restarted.

This PR also includes more changes:
- I moved the tray icon related code into it's own module so it's easier to test
- Added tests and also some mocking to allow testing other things more easily (like being able to set recent games or settings, and a mock of i18next so we can also test that we use the correct translation keys)
- Refactored a lot of the code to be able to add more tests and to remove duplications
- Removed the `--noStackTrace` flag for `yarn test`, it's really difficult to track down where a test fails without the stack trace
- Added the `backend_events` module that we can use as an events system (used in this PR for the fix of the bug mentioned at the beginning)

I'll add some comments in the diff too

---


src/backend/__mocks__/config.ts
```diff
@@ -0,0 +1,20 @@
+const GlobalConfig = (() => {
+  const config = {
+    darkTrayIcon: false
```
arielj, line 3: here we can add default values for the mock

src/backend/backend_events.ts
```diff
@@ -0,0 +1,12 @@
...
+// This can be used to emit/listen to events to decouple components
+// For example:
+//   When the list of recent games changes, a `recentGamesChanged` event is emitted.
+//   The `tray_icon` module listens to this event with `backendEvents.on` to update
+//   the list of recent games in the tray icon's context menu asynchronously
+//
+// Usage:
+//   Emit events with `backendEvents.emit("eventName", arg1, arg2)
+//   Listen to events with `backendEvents.on("eventName", (arg1, arg2) => { ... })
+export const backendEvents = new EventEmitter()
```
arielj, line 12: using this EventEmitter will allow us to decouple code

src/backend/recent_games/ipc_handler.ts
```diff
@@ -0,0 +1,6 @@
+import { ipcMain } from 'electron'
+import { removeRecentGame } from './recent_games'
+
+ipcMain.handle('removeRecent', async (_event, appName) =>
+  removeRecentGame(appName)
+)
```
arielj, line 6: moved this into it's own file so it's not loaded during tests if not needed

src/backend/recent_games/recent_games.ts
```diff
@@ -14,6 +15,16 @@ const getRecentGames = async (options?: { limited: boolean }) => {
...
+const setRecentGames = (recentGames: RecentGame[]) => {
+  // store
+  configStore.set('games.recent', recentGames)
+
+  // emit
+  const window = BrowserWindow.getAllWindows()[0]
+  window.webContents.send('recentGamesChanged', recentGames)
+  backendEvents.emit('recentGamesChanged', recentGames)
+}
+
```
arielj, line 27: this helps remove duplicated code and also makes testing easier

flavioislima (approved):
Teste here and looks good 👍🏽

package.json
```diff
@@ -169,7 +169,7 @@
   "scripts": {
     "start": "vite",
     "codecheck": "tsc --noEmit",
-    "test": "jest --noStackTrace",
+    "test": "jest",
```
CommandMC, line 172: Last time I checked, the stack trace was quite inaccurate. Is this no longer the case now?

src/backend/tray_icon/tray_icon.ts
```diff
@@ -0,0 +1,129 @@
...
+  })
+
+  backendEvents.on('recentGamesChanged', async (recentGames: RecentGame[]) => {
+    loadContextMenu(recentGames)
+  })
+
+  return appIcon
+}
+
+const iconSizesByPlatform = {
```
CommandMC, line 52: Adding a type hint here might be a good idea
```suggestion
const iconSizesByPlatform: Record<'win32' | 'darwin' | 'linux', { width: number, height: number }> = {
```
(don't commit this directly, I doubt our linter would be happy with it)

PR #2185, opened 2022-12-12
[Release] v2.5.2

flavioislima:
PR to add all fixes and improvements for the next release.

---

### DIFF
--- a/.github/workflows/build-prs-linux.yml
+++ b/.github/workflows/build-prs-linux.yml
@@ -2,7 +2,7 @@ name: Build Linux
 
 on:
   pull_request:
-    branches: [main, beta]
+    branches: [main, stable]
   workflow_dispatch:
     inputs:
       version:
--- a/.github/workflows/build-prs-mac.yml
+++ b/.github/workflows/build-prs-mac.yml
@@ -2,7 +2,7 @@ name: Build macOS
 
 on:
   pull_request:
-    branches: [main, beta]
+    branches: [main, stable]
   workflow_dispatch:
     inputs:
       version:
--- a/.github/workflows/build-prs-windows.yml
+++ b/.github/workflows/build-prs-windows.yml
@@ -2,7 +2,7 @@ name: Build Windows
 
 on:
   pull_request:
-    branches: [main, beta]
+    branches: [main, stable]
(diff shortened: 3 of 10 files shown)


arielj:
Should this be the other way around? merging main into stable instead of stable into main? if we want to use main for beta releases and stable for regular releases

flavioislima:
> Should this be the other way around? merging main into stable instead of stable into main? if we want to use main for beta releases and stable for regular releases
> 
> I mean, we do need to have these changes in `main` too so this PR is fine, but `main` will have beta changes eventually (like my PR for the Tray Icon)

It is a bit complicated, I am still not 100% sure we will have any betas soon. I am trying to evaluate if it is worth it or not.

I am just afraid that we might lose some changes with those two branches.

So we should only merge fixes on stable and reactors and new big features in `main`?, but the problem is that we need to start creating all new branches from `stable` otherwise other changes from previous commits will come, it is not intuitive especially for new contributors to understand that.

arielj:
> > Should this be the other way around? merging main into stable instead of stable into main? if we want to use main for beta releases and stable for regular releases
> > I mean, we do need to have these changes in `main` too so this PR is fine, but `main` will have beta changes eventually (like my PR for the Tray Icon)
> 
> It is a bit complicated, I am still not 100% sure we will have any betas soon. I am trying to evaluate if it is worth it or not.
> 
> I am just afraid that we might lose some changes with those two branches.
> 
> So we should only merge fixes on stable and reactors and new big features in `main`?, but the problem is that we need to start creating all new branches from `stable` otherwise other changes from previous commits will come, it is not intuitive especially for new contributors to understand that.

I have an idea, we can add a github workflow to automatically create a new PR after something gets merged into `stable` to merge those changes in `main`, that way we don't need to remember to replicate changes. I have to do some research but I know github actions can interact with the repo, at least something that creates the PR (and we might need to fix conflicts) but that would remove the burden of having to remember the change in the other branch

what I'm worried is that if we merge changes that require some more testing but we need to make a quick stable release we'll have issues with changes we do not want to ship to users

flavioislima:
Yes, but we can try.
Sound alike a good idea.

PR #2189, opened 2022-12-13
[Tech] Make it possible to specify only `prefix` when logging messages

CommandMC:
When logging messages using `logDebug`, `logInfo`, `logWarning`, and `logError`, we usually only specify the `prefix` option & leave all other properties undefined.
This added some unnecessary boilerplate (`{ prefix: <actualValue> }`), and also made the code less readable (our linter would spread out the options object between 3 separate lines, cramming the actual message into just one)

With this PR, you can now just specify a `LogPrefix` directly (note that this is an addition, the old behavior is still there & perfectly functional for anyone who might prefer it)

In case you're worried about losing the clarity that `prefix: LogPrefix.Foo` had, I recommend enabling the "JavaScript" -> "Inlay Hints" -> "Parameter Names" option in VSCode. It'll automatically construct this hint again:  
image

---


arielj (approved):
Looks good to me! :+1:  to the removal of boilerplate

Nocccer (approved):
Really good idea.

PR #2191, opened 2022-12-13
[macOS] Add arm64 build

flavioislima:
This adds an arm64-optimized build for macOS.

Note:
ARM packages will show as damaged since they are not signed by an Apple ID certified Developer, so this command is required to make it work:
`xattr -d com.apple.quarantine | mdfind kMDItemCFBundleIdentifier = "*heroic"`

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
           path: dist/Heroic-*.AppImage
           retention-days: 14
   build_mac:
-    runs-on: macos-11
+    runs-on: macos-latest
     steps:
       - name: Checkout repository.
         uses: actions/checkout@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,9 +59,13 @@ jobs:
         run: yarn
       - name: Build artifacts.
         run: yarn dist:mac --publish=never
-      - name: Upload DMG.
+      - name: Upload x64.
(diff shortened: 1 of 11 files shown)


flavioislima:
Well, reopening this one now since to make it work only a small command is required and the benefits are good because performance on native is a lot better so I think it is ready to review again.

.github/workflows/build.yml
```diff
@@ -31,36 +31,36 @@ jobs:
...
         with:
-          node-version: '16'
+          node-version: '18'
       - name: Install modules.
         run: yarn
       - name: Build artifacts.
         run: yarn dist:mac --publish=never
       - name: Upload DMG.
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
```
arielj, line 63: just one question cause I don't know how this works, I see this `Upload DMG.` step that was not split in two like the one in the `build-prs-mac.yml` file that now has `Upload x64.` and `Upload ARM64.` steps.

Should this one be split too?

arielj:
This is for the M1/2 chips, right? I only have access to an intel mac so I can't really test this. Code looks fine to me, I just left one question.

.github/workflows/build.yml
flavioislima: good point

arielj (approved):
Looks good to me, though I can't test this cause I don't have an M1 chip

PR #2200, opened 2022-12-15
[Refactor] Main window creating and send message to the frontend

arielj:
This PR is a refactor related to how we interact with the main window. It removes some duplication and also while adding tests I found a bug.

Some comments on the changes:
- moved the code that creates the main BrowserWindow object into its own module
- the module takes care of the BrowserWindow object creation with the saved size/position and the sending of the messages to the frontend (abstracting the `window.webContents.send` logic)
- the module takes care of the logic of checking if a window exists before sending a message, so we can remove a lot of duplication
- by doing this I was able to remove the reference to the main window that Game objects had as `this.window`, it makes more sense for games to not depend on that
- I found a bug in the window creation logic, the condition to check if the window size fits the screen was backwards
- I added some more code and patterns to test more things

I'll add more comments in the diff

---

### DIFF
--- a/src/backend/__mocks__/electron.ts
+++ b/src/backend/__mocks__/electron.ts
@@ -1,5 +1,9 @@
 import { EventEmitter } from 'node:events'
-import { MenuItemConstructorOptions } from 'electron'
+import {
+  BrowserWindowConstructorOptions,
+  Display,
+  MenuItemConstructorOptions
+} from 'electron'
 import { tmpdir } from 'os'
 import { join } from 'path'
 
@@ -28,8 +32,23 @@ class Notification {
 }
 
 class BrowserWindow {
-  public constructor() {
-    return {}
+  static windows: BrowserWindow[] = []
+  options: BrowserWindowConstructorOptions = {}
+
+  constructor(options: BrowserWindowConstructorOptions) {
+    this.options = options
+  }
+
+  static getAllWindows() {
+    return this.windows
+  }
+
+  static setAllWindows(windows: BrowserWindow[]) {
+    this.windows = windows
+  }
+
+  public getOptions() {
+    return this.options
   }
 }
 
@@ -44,6 +63,17 @@ const nativeImage = {
     resize: (size: { width: number; height: number }) =>
       `${path} width=${size.width} height=${size.height}`
   })
+}
+
+const screen = {
+  getPrimaryDisplay: () => {
+    return {
+      workAreaSize: {
+        height: 1280,
+        width: 1920
+      }
+    }
+  }
 }
 
 class Tray {
@@ -76,5 +106,6 @@ export {
   Menu,
   nativeImage,
   Tray,
-  ipcMain
+  ipcMain,
+  screen
 }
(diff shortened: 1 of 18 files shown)


src/backend/__mocks__/electron.ts
```diff
@@ -28,8 +32,23 @@ class Notification {
...
+  static getAllWindows() {
+    return this.windows
+  }
+
+  static setAllWindows(windows: BrowserWindow[]) {
+    this.windows = windows
+  }
+
+  public getOptions() {
+    return this.options
```
arielj, line 51: with this we can make assertions about the options passed to the `new BrowserWindow` constructor, even though it's not a real method of a BrowserWindow

src/backend/dialog/dialog.ts
```diff
@@ -37,12 +33,18 @@ function showDialogBoxModalAuto(props: {
...
+
       switch (props.type) {
         case 'ERROR':
           showErrorBox(props.title, props.message)
           break
         default:
-          showMessageBox(BrowserWindow.getAllWindows()[0], {
+          if (!window) {
+            break
+          }
```
arielj, line 46: we were not checking this before and it could fail if for some reason we call this without a window

src/backend/main.ts
```diff
@@ -266,6 +220,13 @@ async function createWindow(): Promise<BrowserWindow> {
       autoUpdater.checkForUpdates()
     }
   }
+
+  ipcMain.on('setZoomFactor', async (event, zoomFactor) => {
+    mainWindow.webContents.setZoomFactor(
```
arielj, line 225: moved this inside since it depends on the mainWindow object that was just created

src/backend/main_window.ts
```diff
@@ -0,0 +1,84 @@
...
+      tmpWindowProps.y !== undefined &&
+      tmpWindowProps.x !== undefined
+    ) {
+      windowProps = tmpWindowProps
+    }
+  } else {
+    // make sure initial screen size is not bigger than the available screen space
+    const screenInfo = screen.getPrimaryDisplay()
+
+    if (screenInfo.workAreaSize.height < windowProps.height) {
```
arielj, line 58: we had this condition backwards

src/backend/dialog/dialog.ts
Nocccer: I thought all electron dialogs were replaced with custom once?

Nocccer (approved):
LGTM.
Get rid of the showMessageBox is another PR i would say.

src/backend/dialog/dialog.ts
flavioislima: We forgot a few, at least two of them are still using the old one.

To be fair, I think because of things like the Steam deck, we should keep at least the one related to the game not found as native, because otherwise, Heroic fails. For now it simply returns to steam, but the Game needs to update one (that is native now), was making Steam completely freeze since it was waiting for the user input.

another idea is to extend these methods and make them accept a parameter like: `nativeDialog: boolean`, so when using NO-GUI we can specify which ones are native or not.

flavioislima (approved):
LGTM codewise but would be good to test all use cases because it touched several components and methods.

PR #2203, opened 2022-12-15
[Fix] GOG games Setup on Windows

flavioislima:
Changed `execAsync ` for `spawn `so the Setup won't fail when there is space in the game name folder, that was the main cause of this issue. Also, `Start-Process` was not being recognized since it was being used on cmd instead of `powershell`. so it was failing because of that as well.

Another issue was that sometimes it was failing because it was trying to access Wine configurations and they were undefined since it is not Linux/mac.

---


imLinguin (approved):
Looks good

PR #2210, opened 2022-12-16
[Tech] Add ts-prune to find dead code

Nocccer:
This PR adds ts-prune to find dead code or useless exports. 

- All dead code reported by ts-prune were removed
- Adapt tests to still test the old now not anymore exported functions
- Moved function notify to dialog. I didn't find a way to mock it from utils while testing utils functions. Utils is a large file and making it smaller makes sense anyway

---


src/backend/legendary/games.ts
```diff
@@ -857,19 +857,11 @@ class LegendaryGame extends Game {
       "Legendary's config from config.ini (before Heroic's settings):\n"
     )
 
-    try {
-      const json = JSON.parse(stdout)
-      // remove egl auth info
-      delete json['egl_parameters']
+    const json = JSON.parse(stdout)
```
arielj, line 860: we shouldn't remove this try/catch

src/backend/dialog/dialog.ts
```diff
@@ -52,4 +53,22 @@ function showDialogBoxModalAuto(props: {
...
 
-export { showDialogBoxModalAuto }
+type NotifyType = {
+  title: string
+  body: string
+}
+
+function notify({ body, title }: NotifyType) {
+  if (Notification.isSupported() && !isSteamDeckGameMode) {
+    const mainWindow = BrowserWindow.getAllWindows()[0]
```
arielj, line 63: we can use `getMainWindow` from the `main_window` module here to hide the implementation

src/backend/sideload/games.ts
```diff
@@ -240,42 +229,6 @@ export async function launchApp(appName: string): Promise<boolean> {
   return false
 }
 
-export async function moveInstall(
```
arielj, line 243: I wonder if this is ok to be removed or if it means that at some point we lost the code to trigger the move of the installation for sideloaded apps?

cc @flavioislima to confirm

Nocccer: Got fixed by merging with newest commits.

Nocccer: Ok vscode github pull request extension tricked me. If i click on your comment it jumps to this function and shows me that it is added?

Nocccer: @flavioislima do we need the moveInstall function for sideload games?

.github/workflows/codecheck.yml
```diff
@@ -17,3 +17,5 @@ jobs:
         run: yarn
       - name: Check Typescript syntax
         run: yarn codecheck
+      - name: Find dead code
+        run: yarn find-deadcode
```
CommandMC, line 21: *Maybe* do this in a separate Workflow?

CommandMC (approved):

.github/workflows/codecheck.yml
Nocccer: We should avoid to make a new workflow for each small checker.
I think the limits are 6 workflows in parallel. We would waste time here to setup a ubuntu server and install dependencies.

Nocccer: Ok 20 is the limit, still i would not cluster the checker with to much workflows.

Nocccer: Added to linter workflow

src/backend/online_monitor.ts
```diff
@@ -119,11 +119,11 @@ export const initOnlineMonitor = () => {
   })
 }
 
-export const makeNetworkRequest = (callback: () => unknown) => {
-  if (isOnline()) {
-    callback()
-  }
-}
+// export const makeNetworkRequest = (callback: () => unknown) => {
```
flavioislima, line 122: Why to keep this commented code?

PR #2214, opened 2022-12-16
Use caching to speed up GitHub workflows

arielj:
Testing

---


arielj:
closing this, it saves a few seconds on linux builds but those are already fast, for the mac build it even made it slower (?)

maybe somebody finds a better way to do this (in theory caching the result of `yarn install` should always be good)

Issue #1881, opened 2022-10-07
Proposal: Change behavior of `All / Epic / GOG` and `All / Window / Linux` filtering


arielj:
### Problem description

Currently we have 3 options to filter by store (all, epic, gog) and 3 options to filter by system (all, windows, linux/mac). If we turn epic/gog into elements that can be toggled on/off we wouldn't need the `all` option, making it cleaner.

### Feature description

I the library, I can pick which store to show/filter. By default we can have both Epic and Gog games enabled, the icons shown with full opacity. If I want to see only Epic games I can toggle off the GOG games (show the icon with lower opacity to indicate is not ON).

That way we don't need the `All` option which would mean having both store toggled ON.

Same for the operating system filters, instead of picking All (windows + linux) / windows / linux, we can have both OS turned ON and then you can filter out by support toggling things off.

We need a decided what we do if a user disables all the options, like show a message saying `hey, you disabled all stores... you sure you want to do that?` or `we are showing all stores since you disabled all of them`.

### Alternatives

Leave the current behavior.

flavioislima:
Would be good to improve that for sure.
Maybe we could also use a list of options like we did before?
A multi-select list in this case because now that we will add Sideloaded games then it will be another filter for the 'stores'.

arielj:
My problem with multi selects is that the UX is pretty bad (you have to hold Ctrl and click the options you want to multiselect) and I don't know how that would translate to the controller for example.

One idea I had was to have a `Filters` area that expands on hover with buttons: Epic / Gog / Heroic / Only Favorites / Show Hidden, that you can toggle on/off. I don't think changing the filters is done that often to have to always display all of them all the time, but hovering is quick and it makes the UI more consistent.

I can imagine also a `Layout` area that expands on hover with: grid vs list / asc vs des order / installed first vs last

That will allow us to add more options without cluttering the general layout without hurting the discoverability or speed to toggle those things.

I'll try to make a POC during this weekend.

flavioislima:
Cool, yes we need to think about the controllers as well.
A filter area makes sense since in the future we want to add game filters as well, based on genre, etc.

biliesilva:
@arielj @flavioislima 

We can use a normal box for the filter and checkboxes to be easy to interact with a controller as in the example below:

![image](https://user-images.githubusercontent.com/61093873/194714886-b93c43b4-cb06-42ff-8f4d-d8cd39d19377.png)


[...]

PR #2216, opened 2022-12-17
[UI] Remove 'active' color from library icons

### LINKED ISSUE #1878: Colors in Library config are inconsistent
### Describe the bug

image

Some buttons have an accentuated color in one state and white in the other, but the accentuated color make it look like an on/off thing when it is not an on/off:
- the layout type is accentuated in grid and white in list
- the alphabetical order is accentuated in descending and white in ascending
- the sort by installed is accentuated if installed first and white if installed last
- show hidden is accentuated if active and white if inactive

All these icons actually change to another icon already to indicate the status. The icon change should be enough indication, since the color seems to suggest a meaning that is not really there.

The color of the favorites icon changes, which seems ok to indicate an on/off status of the filter so it can make sense in this one.

It also mixes sorting with filtering with the refresh icon, but the top left `all / epic / gog` buttons also do filtering and work completely different.

Also icons have light blue hover effect when the text is white but no hover effect when is accentuated.

### Add logs

not relevant

### Steps to reproduce

1. Go to the Library
2. Change the list filter/sorting/layout

### Expected behavior

Icons shouldn't change color if the color is not relevant to the status of the icon.

### Screenshots

_No response_

### System Information

- Heroic Version [e. g. 2.1.1]: main branch

### Additional information

_No response_

arielj:
This PR fixes #1878 

Icons in the library had an `active` color that had no real meaning, since the meaning is the icon itself and not the color.

Now icons are always white except when hovered.

image

Only exception for this change was the Favorites icon that should eventually be moved to a `filters` dropdown to turn filters on/off instead as described here https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1881#issuecomment-1272342402 by billie.

---

### DIFF
--- a/src/frontend/components/UI/ActionIcons/index.tsx
+++ b/src/frontend/components/UI/ActionIcons/index.tsx
@@ -67,9 +67,7 @@ export default React.memo(function ActionIcons({
       <FormControl segmented small>
         {layout === 'grid' ? (
           <button
-            className={classNames('FormControl__button', {
-              active: layout === 'grid'
-            })}
+            className="FormControl__button"
             title={t('library.toggleLayout.list', 'Toggle to a list layout')}
             onClick={() => handleLayout('list')}
           >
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
LGTM

PR #2217, opened 2022-12-17
Indicate latest heroic version in bug report template

arielj:
This PR adds a small change in the PR template to let users know which one is the latest version of Heroic, in case someone tries to report a bug in an older version.

Also moved the version to a dropdown so it's easier to pick between `Latest Stable` `Latest Beta` and `main branch from GitHub`.

This is how it looks:
image

Putting this here as an idea, I understand if there's a concern that this can get outdated and you prefer not to use this.

---

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -57,6 +57,8 @@ body:
         - Latest Stable
         - Latest Beta
         - main branch from GitHub
+    validations:
+      required: true
   - type: textarea
     id: systeminfo
     attributes:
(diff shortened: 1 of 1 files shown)


redromnon (approved):
Looks cool :+1:

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -45,14 +45,26 @@ body:
     attributes:
       label: Screenshots
       description: If applicable, add screenshots to help explain your problem.
+  - type: markdown
+    attributes:
+      value: |-
+        ## Latest stable version of Heroic is 2.5.2. Please, update Heroic before reporting a bug if you have an older version.
```
redromnon, line 51: One will need to update this line every time a new stable and beta version gets released.

flavioislima: That's a good point. Although not a blocker, just one more thing to prepare when doing a new release, ahahaha

flavioislima (approved):
Looks good 👍🏽

PR #2221, opened 2022-12-17
[WineManager] Fix status update

arielj:
This PR fixes a couple of issues in the Wine Manager present in the latest release:

- Proton items were not showing any status update (installing/progress/unzipping)
- After making a change in a Proton item, it was re-rendering the list for Wine-GE instead

---

### DIFF
--- a/src/frontend/screens/WineManager/index.tsx
+++ b/src/frontend/screens/WineManager/index.tsx
@@ -77,7 +77,7 @@ export default React.memo(function WineManager(): JSX.Element | null {
     return () => {
       removeListener()
     }
-  }, [])
+  }, [repository])
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


src/frontend/screens/WineManager/index.tsx
```diff
@@ -69,15 +69,16 @@ export default React.memo(function WineManager(): JSX.Element | null {
...
+  useEffect(() => {
     const removeListener = window.api.handleWineVersionsUpdated(() => {
       setWineVersions(getWineVersions(repository))
     })
-
     return () => {
       removeListener()
     }
-  }, [])
+  }, [repository])
```
arielj, line 81: we have to re add the listener after `repository` changes

src/frontend/screens/WineManager/index.tsx
```diff
@@ -126,8 +127,8 @@ export default React.memo(function WineManager(): JSX.Element | null {
             {refreshing && <UpdateComponent />}
             {!refreshing &&
               !!wineVersions.length &&
-              wineVersions.map((release, key) => {
-                return <WineItem key={key} {...release} />
+              wineVersions.map((release) => {
+                return <WineItem key={release.version} {...release} />
```
arielj, line 131: now that the array is pre-filtered, we can't use the index as the loop key

Nocccer (approved):
LGTM. Thanks for the fast fix

PR #2224, opened 2022-12-18
[UI] Prevent flash when opening an installation dialog for the first time

arielj:
This PR changes the InstallModal import to not be lazy-loaded.

This fixes 2 problems:
- since it was lazy loaded, Suspense was triggered and rendering a Loading screen for a moment, making the UI flash
- because of the Suspense changing the content, the list was re-rendered completely after

---

### DIFF
--- a/src/frontend/screens/Library/index.tsx
+++ b/src/frontend/screens/Library/index.tsx
@@ -1,7 +1,6 @@
 import './index.css'
 
 import React, {
-  lazy,
   useContext,
   useEffect,
   useMemo,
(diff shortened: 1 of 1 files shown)


Nocccer (approved):

flavioislima (approved):
Ok, maybe was not a good idea to lazy load this one. What about lazy loading the Download Dialog and the Sideload dialog inside the Install Modal? it was having the same side effect?

The problem is that these components are being loaded when the app loads so makes the first load slower.

arielj:
> Ok, maybe was not a good idea to lazy load this one. What about lazy loading the Download Dialog and the Sideload dialog inside the Install Modal? it was having the same side effect?
> 
> The problem is that these components are being loaded when the app loads so makes the first load slower.

hmmm I don't know how much is the impact of loading that component with the page (those are really simple components anyway), any lazy loaded component will generate this flicker the first time it's used because it will trigger the Suspense feature and I can't imagine a significant benefit (sounds like the wrong kind of optimization since our main performance problem is not the loading of things but the extra renders)

I would argue to not use any lazy loading of components unless we can justify it with a noticeable performance benefit (with a before/after metric), because it will produce this sideeffect.

PR #2228, opened 2022-12-19
Extra info tweaks and improvements

arielj:
This PR includes a few tweaks and fixes related to the extra info of games:

- The extra info is now only fetched when entering the game page and not when fetching any gameInfo data, that should reduce the first render of the library significantly since we where waiting for 2 extra requests per epic game to get the extra info (one graphql request to get a page slug and one to get the actual info)
- We were using `language` from the app config but the language value is stored in the configStore actually
- If the language was undefined for any reason, the url was invalid, now it has an `en` fallback
- If we couldn't find a page slug we were falling back to the appName but that's probably never valid, now it converts the game title into a slug (it fixes a missing description for rocket league for example)

Note that some games (I only found Alba - A Wildlife Adventure) don't seem to return a page slug and I couldn't find what to put in the api query, so some games will still not show description/requirements, but I'll investigate those more at a different time.

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -305,7 +305,7 @@ export default React.memo(function GamePage(): JSX.Element | null {
                     appName={appName}
                     isInstalled={is_installed}
                     title={title}
-                    storeUrl={gameInfo.store_url}
+                    storeUrl={extraInfo?.storeUrl || gameInfo.store_url}
                     runner={gameInfo.runner}
                     handleUpdate={handleUpdate}
                     disableUpdate={isInstalling || isUpdating}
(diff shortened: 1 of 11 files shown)


flavioislima (approved):
LGTM

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -313,11 +320,11 @@ export default React.memo(function GamePage(): JSX.Element | null {
               <div className="infoWrapper">
                 <div className="developer">{developer}</div>
                 <div className="summary">
-                  {extra && extra.about
-                    ? extra.about.description
-                      ? extra.about.description
-                      : extra.about.longDescription
-                      ? extra.about.longDescription
+                  {extraInfo && extraInfo.about
```
flavioislima, line 323: Maybe put this into a variable to make it cleaner:
`const description = extra?.about?.description || extra?.about?.longDescription || '';`

PR #2229, opened 2022-12-19
[Fix] Get store links for Epic from API response

### LINKED ISSUE #1160: EpicGames Store URL generation is not always correct
### Describe the bug

We are not getting the actual url for the store page of epic games so we are using some code to generate it using the game's title.

This is not always correct though, one example:

https://store.epicgames.com/en-US/p/batman-arkham-asylum <- when navigating using the store
https://www.epicgames.com/store/product/batman-arkham-asylum-game-of-the-year-edition <- url generated with the game's title

The domain, `store`, `product` and language are not relevant (but maybe it's worth to update the generated url anyway, though the language is not really needed), but the name of the game, in this case, shouldn't have the `game-of-the-year-edition` substring

### Add logs

Not relevant

### Steps to reproduce

1. You must own Batman™ Arkham Asylum Game of the Year Edition
2. Click on the game details
3. Go to Tools
4. Click on Store Page

the epic store page will show a 404 error

### Expected behavior

The id in the url for that game should be `batman-arkham-asylum`.

This probably applies to other games too with different problems. For example `Among the Sleep - Enhanced Edition`'s generated url is https://www.epicgames.com/store/product/among-the-sleep--enhanced-edition while the correct url is https://store.epicgames.com/en-US/p/among-the-sleep-enhanced-edition (single dash between sleep and enhanced)

### Screenshots

_No response_

### System Information

Not relevant

### Additional information

_No response_

arielj:
This PR fixes #1160 

Instead of trying to generate a url for an Epic game based on its title, this PR changes that to use the response of the API we use to get the requirements which also includes the `_slug` property.

It's important to clear heroic's cache to test this (or at least clear the `./config/heroic/lib-cache/gameinfo.json` file), since we cache the games extra info in an electron store.

---

### DIFF
--- a/src/backend/gog/games.ts
+++ b/src/backend/gog/games.ts
@@ -84,8 +84,15 @@ class GOGGame extends Game {
     }
 
     const extra: ExtraInfo = {
-      about: gameInfo.extra.about,
-      reqs: await GOGLibrary.get().createReqsArray(this.appName, targetPlatform)
+      about: {
+        description: gameInfo.description || '',
+        longDescription: gameInfo.longDescription || ''
+      },
+      reqs: await GOGLibrary.get().createReqsArray(
+        this.appName,
+        targetPlatform
+      ),
+      storeUrl: gameInfo.store_url
     }
     return extra
   }
--- a/src/backend/gog/games.ts
+++ b/src/backend/gog/games.ts
@@ -84,10 +84,7 @@ class GOGGame extends Game {
     }
 
     const extra: ExtraInfo = {
-      about: {
-        description: gameInfo.description || '',
-        longDescription: gameInfo.longDescription || ''
-      },
+      about: gameInfo.extra.about,
       reqs: await GOGLibrary.get().createReqsArray(
         this.appName,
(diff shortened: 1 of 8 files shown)


flavioislima:
Can we use these calls to get other information about the game to improve the filters? things like categories, ratings, genre, tags, controller support, release date, etc?

arielj:
> Can we use these calls to get other information about the game to improve the filters? things like categories, ratings, genre, tags, controller support, release date, etc?

I did more testing today, the graphql stuff works for some games and fails for others, I left the original method using the API and added the graphql method as a fallback (it fixes the issue I found with Alba - A Wildlife Adventure).

The graphql stuff is undocumented, I was using this https://gist.github.com/woctezuma/8ca464a276b15d7dfad475fd6b6cbee9 and a lot of trial and error. There are some queries that give some info like ratings, but I think it's too unreliable. Maybe if we find a better resource for documentation we could use it more but the lack of docs makes it hard.

flavioislima (approved):
Looks good 👍🏽

PR #2253, opened 2022-12-24
[UI] Add scaling effect on sidebar hover

redromnon:
The sidebar's main items undergo a minimal scaling effect when hovered, just like the game cards.
This effect does not apply to the sub-items. 

_Demo:_

https://user-images.githubusercontent.com/74495920/209443184-1279f2f5-d0d1-4567-b61d-e9a6bbb44674.mp4

---

### DIFF
--- a/src/frontend/components/UI/Sidebar/index.css
+++ b/src/frontend/components/UI/Sidebar/index.css
@@ -13,6 +13,7 @@
   text-align: left;
   background: var(--navbar-background);
   overflow: auto;
+  overflow-x: hidden;
 }
 
 .Sidebar.collapsed {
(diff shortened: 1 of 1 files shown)


src/frontend/components/UI/Sidebar/index.css
```diff
@@ -60,6 +60,14 @@
 
 .Sidebar__item:hover {
   background-color: var(--navbar-active-background);
+  transition: 0.2s;
+  scale: 1.02;
+  width: 99%;
```
redromnon, line 65: If this isn't set, the hovered item will expand the width of the sidebar due to which a horizontal scrollbar will appear below the Heroic version text.

arielj: We can use `overflow-x: hidden` in the `.Sidebar` selector to avoid the need of these `width`s. It would be more maintainable.

arielj (approved):
I added one comment, otherwise it looks good

src/frontend/components/UI/Sidebar/index.css
redromnon: Alright,  I'll look into this.

redromnon: I've used `overflow-x: hidden` now.

arielj (approved):

PR #2264, opened 2022-12-26
[Refactor] Push status updates from the backend to the frontend

arielj:
This is another PR that I'm extracting from the original game status refactor PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1857

Since that PR got too out of date and it's hard to rebase it, I found it easier to extract smaller and easier-to-review PRs to eventually move all that code.

With these changes, status updates for uninstalling/moving/repairing games and installing/updating/uninstalling the EOSOverlay installation are pushed from the backend to the frontend, the same way other status updates are handled for game installation or game progress.

This makes it more consistent to always push updates from the backend and the updates are pushed closer to the source of actual action.

With this I was also able to remove the `handleGameStatus` function in the ContextProvider, to have less stuff there.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1027,46 +1027,63 @@ ipcMain.on('showItemInFolder', async (e, item) => showItemInFolder(item))
 ipcMain.handle(
   'uninstall',
   async (event, appName, runner, shouldRemovePrefix, shouldRemoveSetting) => {
+    sendFrontendMessage('gameStatusUpdate', {
+      appName,
+      runner,
+      status: 'uninstalling'
+    })
+
     const game = getGame(appName, runner)
 
     const { title } = game.getGameInfo()
 
+    let uninstalled = false
+
     try {
       await game.uninstall()
+      uninstalled = true
     } catch (error) {
       notify({
         title,
         body: i18next.t('notify.uninstalled.error', 'Error uninstalling')
       })
       logError(error, { prefix: LogPrefix.Backend })
-      return
     }
-    if (shouldRemovePrefix) {
-      const { winePrefix } = await game.getSettings()
-      logInfo(`Removing prefix ${winePrefix}`, {
-        prefix: LogPrefix.Backend
-      })
-      // remove prefix if exists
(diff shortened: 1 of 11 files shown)


src/backend/main.ts
```diff
@@ -1027,46 +1027,63 @@ ipcMain.on('showItemInFolder', async (e, item) => showItemInFolder(item))
...
-      const { winePrefix } = await game.getSettings()
-      logInfo(`Removing prefix ${winePrefix}`, {
-        prefix: LogPrefix.Backend
-      })
-      // remove prefix if exists
-      if (existsSync(winePrefix)) {
-        rmSync(winePrefix, { recursive: true })
+
+    if (uninstalled) {
+      if (shouldRemovePrefix) {
```
arielj, line 1054: there's nothing changing inside this `if`, I just added this code inside the if and github is messing up the diff

src/common/typedefs/ipcBridge.d.ts
```diff
@@ -156,7 +156,7 @@ interface AsyncIPCFunctions {
     shoudlRemoveSetting: boolean
   ) => Promise<void>
   repair: (appName: string, runner: Runner) => Promise<void>
-  moveInstall: (args: MoveGameArgs) => StatusPromise
+  moveInstall: (args: MoveGameArgs) => Promise<void>
```
arielj, line 159: We only call `moveInstall` in one place and we don't really use it's returned value

src/frontend/state/GlobalState.tsx
```diff
@@ -618,8 +618,7 @@ export class GlobalState extends PureComponent<Props> {
     })
 
     window.api.handleGameStatus(async (e: Event, args: GameStatus) => {
-      const { libraryStatus } = this.state
-      return this.handleGameStatus({ ...libraryStatus, ...args })
+      return this.handleGameStatus({ ...args })
```
arielj, line 621: we were passing `...libraryStatus` here but we were not really using those keys

Nocccer (approved):
LGTM

PR #2265, opened 2022-12-27
[Refactor] Load settings synchronously

arielj:
This PR is a refactor on the way we load the application settings to make it synchronous.

There are a few benefits of this:
- usually, sync code is easier to manage than async code (it's easy to miss awaits)
- because of the load being async, we were actually loading the config multiple times because promises were running in parallel until one of the promises finished and it was stored in memory, now we only read the files once and store the settings in memory, I noticed this problem while debugging this issue https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2262
- I was able to convert other functions that were async into sync too
- I was able to remove a magic ignore ts linter comment, handling config being undefined properly by making it protected

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -276,7 +276,7 @@ if (!gotTheLock) {
     // We can't use .config since apparently its not loaded fast enough.
     // TODO: Remove this after a couple of stable releases
     // Affects only current users, not new installs
-    const settings = await GlobalConfig.get().getSettings()
+    const settings = GlobalConfig.get().getSettings()
     const { language } = settings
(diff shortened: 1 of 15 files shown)


src/backend/main.ts
```diff
@@ -755,7 +755,9 @@ ipcMain.handle(
   }
 )
 
-ipcMain.handle('getUserInfo', LegendaryUser.getUserInfo)
+ipcMain.handle('getUserInfo', async () => {
+  return LegendaryUser.getUserInfo()
+})
```
arielj, line 760: I had to convert this one back to async

Nocccer (approved):
LGTM

PR #2267, opened 2022-12-27
[Download Manager] Some images fixes

arielj:
This PR fixes 3 issues I found in the download manager.

- When a game finishes downloading and a new one starts, the new game has the previous game's picture
- When a game finishes downloading, the next elements in the queue have the wrong image from the previous element in the list
- Some images have images with a different aspect ratio (like Far Cry 5)

Issues 1 and 2 only are noticeable if looking at the Downloads screen when a game finishes and there are elements in the queue (re-entering the Downloads page fixes the images)

Before the aspect-ratio fix:
image

After:
image

---

### DIFF
--- a/src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.tsx
+++ b/src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.tsx
@@ -60,7 +60,7 @@ const DownloadManagerItem = ({ element, current }: Props) => {
       }
     }
     getNewInfo()
-  }, [])
+  }, [element])
 
   const { art_cover, art_square } = gameInfo || {}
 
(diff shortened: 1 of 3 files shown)


src/frontend/screens/DownloadManager/components/DownloadManagerItem/index.tsx
```diff
@@ -60,7 +60,7 @@ const DownloadManagerItem = ({ element, current }: Props) => {
       }
     }
     getNewInfo()
-  }, [])
+  }, [element])
```
arielj, line 63: we need this so when the current element changes to the next in the list, it fetches the new gameInfo

src/frontend/screens/DownloadManager/index.tsx
```diff
@@ -119,8 +119,12 @@ export default React.memo(function DownloadManager(): JSX.Element | null {
           <div className="dmItemList">
             <DownloadManagerHeader time="queued" />
             {plannendElements.length > 0 ? (
-              plannendElements.map((el, key) => (
-                <DownloadManagerItem key={key} element={el} current={false} />
+              plannendElements.map((el) => (
+                <DownloadManagerItem
+                  key={el.params.appName}
```
arielj, line 124: using the element's appName as the key instead of the list index, so elements are updated properly

flavioislima (approved):
LGTM!
Thanks for fixing that!

PR #2271, opened 2022-12-27
[Tech] Only parse english on `yarn i18n`

Nocccer:
---

### DIFF
--- a/i18next-parser.config.js
+++ b/i18next-parser.config.js
@@ -19,7 +19,7 @@ module.exports = {
   defaultNamespace: 'translation',
   // Default namespace used in your i18next config
 
-  defaultValue: (locale, namespace, key) => key,
+  defaultValue: '',
   // Default value to give to empty keys
 
   failOnWarnings: false,
@@ -75,12 +75,6 @@ module.exports = {
   // Whether to use the keys as the default value; ex. "Hello": "Hello", "World": "World"
   // This option takes precedence over the `defaultValue` and `skipDefaultValues` options
 
-  verbose: true,
+  verbose: true
   // Display info about the parsing including some stat
-
-  i18nextOptions: {
-    fallbackLng: 'en',
-    returnEmptyString: false,
-    returnNull: false
-  }
 }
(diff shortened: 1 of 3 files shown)


i18next-parser.config.js
```diff
@@ -115,6 +75,12 @@ module.exports = {
   // Whether to use the keys as the default value; ex. "Hello": "Hello", "World": "World"
   // This option takes precedence over the `defaultValue` and `skipDefaultValues` options
 
-  verbose: true
+  verbose: true,
   // Display info about the parsing including some stat
+
+  i18nextOptions: {
```
arielj, line 81: I wonder if this configuration should be when we use i18next in the backend/frontend instead, since here it only applies to the parser/task.

I'll do some tests later today.

arielj (commented):
So, I was testing, and I think it's better to set some of this configuration in src/backend/main.ts and src/frontend/index.tsx when we initialize i18next.

If we set `returnEmptyString: false` and `returnNull: false` there, fallbacks work fine for all cases and we would be fine for anything that weblate does: missing translation, empty string as a translation, null as a translation will fallback to the `en` string, even missing defaults automatically fallback to the translation key so we don't need the `defaultValue` function.

So my suggestion is to only change the list of locales in this parser config and then set the 2 options to `false` in the files I mentioned above.

Nocccer:
> So, I was testing, and I think it's better to set some of this configuration in src/backend/main.ts and src/frontend/index.tsx when we initialize i18next.
> 
> If we set `returnEmptyString: false` and `returnNull: false` there, fallbacks work fine for all cases and we would be fine for anything that weblate does: missing translation, empty string as a translation, null as a translation will fallback to the `en` string, even missing defaults automatically fallback to the translation key so we don't need the `defaultValue` function.
> 
> So my suggestion is to only change the list of locales in this parser config and then set the 2 options to `false` in the files I mentioned above.

Yes good catch. I added your suggestion. Should be fine now.

arielj (approved):

PR #2272, opened 2022-12-27
[UI] Game page improvements

flavioislima:
Some UI changes and improvements on the gamepage:
- Use short description instead of the long one.
- Add Game release date
- Add game genre.

---

### DIFF
--- a/src/backend/gog/library.ts
+++ b/src/backend/gog/library.ts
@@ -650,7 +650,11 @@ export class GOGLibrary {
       art_square: verticalCover,
       cloud_save_enabled: cloudSavesEnabledGames.includes(String(info.id)),
       extra: {
-        about: { description: description, longDescription: '' },
+        about: {
+          description: description,
+          longDescription: '',
+          shortDescription: description
+        },
         reqs: [],
         storeUrl: `https://gog.com${info.url}`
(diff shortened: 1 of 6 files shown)


src/backend/gog/library.ts
```diff
@@ -650,7 +650,11 @@ export class GOGLibrary {
       art_square: verticalCover,
       cloud_save_enabled: cloudSavesEnabledGames.includes(String(info.id)),
       extra: {
-        about: { description: description, longDescription: '' },
+        about: {
+          description: description,
+          longDescription: '',
+          shortDescription: description
```
arielj, line 656: we are setting `description` for both `description` and `shortDescription` here, and we are doing the same for Epic games with `res.sandbox.configuration[0].configs.shortDescription`. Maybe we can just use `description` instead of adding this new key? or maybe remove `description` and only use `shortDescription`? having 2 keys with the same values will be confusing eventually

also, some of this information is cached so for some users `shortDescription` will be empty cause the cached information won't have this key


one thing I'm confused about is that we were setting `longDescription: ''` in gog and epic before these changes, I don't see how this new changes will make use of a short description since it seems to be what we were already using? hmmm

PR #2279, opened 2022-12-28
[UI/UX] Open game settings into a modal

flavioislima:
Based on some user feedback we got about the number of clicks to change configurations for a game, especially on Linux that this is more needed, I decided to simplify that and move the game settings to a modal.

This is good for several reasons:
- Easy access from the library or the game page without navigating to another page back and forth.
- Keeps it separated from the Heroic settings by default, so it is harder to get confused since this is another complaint from some people.
- Makes testing different configurations and settings more accessible.

image

---


flavioislima:
Forgot to deal with the game log settings as well, will fix it later.

src/frontend/screens/Settings/components/SettingsModal/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+    getSettings()
+  }, [appName])
+
+  if (!title) {
+    return <UpdateComponent />
+  }
+
+  // create setting context functions
+  const contextValues: SettingsContextType = {
```
arielj, line 48: most of this is duplicated with the Settings screen, since settings are kinda critical maybe we can move most of this into a `hasSettingsContext` hook to make sure we don't get this logic out of sync?

flavioislima: I was wondering on how to avoid this duplication. My first idea was to use React Portals to render this settings but won't work since it needed the context to be passed on. 
How do you think this context would work? I wonder if we put the settings context above APP if the portal would work or at least the values would be available imediatelly.

flavioislima: Done, created a hook that generate the context values and can be used anywhere.

flavioislima:
Created the hook and also changed the log button to show the game log on a modal as well:
image

Besides that fixed the inconsistency with the game settings on general settings.

public/locales/fr/translation.json
```diff
@@ -70,7 +70,7 @@
                 "title": "Erreur de mise-à-jour"
             },
             "wine-not-found": {
-                "invalid": "La version de Wine sélectionnée n'a pas été trouvée. Installez-la ou sélectionnez une version différente dans les paramètres du jeu{{newline}}Version : {{version}}{{newline}}Chemin : {{path}}",
+                "invalid": "La version de Wine sélectionnée n'a pas été trouvée. Installez-la ou sélectionnez une version différente dans les paramètres du jeu{{newline}}Version\u00a0: {{version}}{{newline}}Chemin\u00a0: {{path}}",
```
arielj, line 73: i18next is doing this weird thing here

src/frontend/components/UI/Collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+
+type Props = {
+  isOpen: boolean
+  summary: string
+  children: React.ReactNode
+  isColapsible?: boolean
```
arielj, line 7:
```suggestion
  isCollapsible?: boolean
```

arielj (commented):
I have a suggestion that I think could make the code more friendly:
- instead of adding isSettingsModalOpen and setIsSettingsModalOpen in the global context we can add a new nested context called `SettingsModalContext`
- instead of using `isSettingsModalOpen` and `setIsSettingsModalOpen` we can have more declarative functions like `showGameSettings(gameInfo)` and `showGameLogs(gameInfo)` and we could do `const {showGameSettings, showGameInfo} = useContext(SettingsModalContext)`
- the context could have another function called `closeModal` to be more declarative instead of passing `setIsSettingsModal(false)`


One minor issue I noticed: when you open the settings for a game and you change something like the wine version, after closing the modal the game page still shows the previous wine. It can be confusing for users thinking that the change was not applied.

It's great to finally have the settings without having to change screens all the time!

arielj:
Another issue I noticed: you can set environment variables in the modal and they show up, but if you close and open the modal they are not shown anymore (though they are still in the settings of the game, just not visible)

public/locales/fr/translation.json
flavioislima: yes, good catch

flavioislima:
@arielj ok, I will look into those issues. But I am not sure about the other changes. Having another context might be too much IMO. Also, because where it will be? since we are using the modal both on the library and on the game page, so it would need to have those as children, I am just afraid that this will cause unnecessary re-renders on the library once we change something inside of it.

flavioislima:
> One minor issue I noticed: when you open the settings for a game and you change something like the wine version, after closing the modal the game page still shows the previous wine. It can be confusing for users thinking that the change was not applied.

Yes, that's because the game page is not observing the settings changes, I am not sure on how to solve that.
Perhaps triggering a re-fetch of the changes when the modal is closed. I can try.

flavioislima:
@arielj about the env variables table. I can see that the variables are ok until they are sent to `TableInput`.
I can see them when calling this method: `getEnvironmentVariables()`

But then, on the table they come as an empty array, do you have any idea why this is happening?

In theory, I am using the same context values as when we have the settings page, so I am not sure what is going on in this case.

arielj:
> @arielj about the env variables table. I can see that the variables are ok until they are sent to `TableInput`. I can see them when calling this method: `getEnvironmentVariables()`
> 
> But then, on the table they come as an empty array, do you have any idea why this is happening?
> 
> In theory, I am using the same context values as when we have the settings page, so I am not sure what is going on in this case.

I'll check that later today!

arielj:
@flavioislima I found the issue with the env variables, I don't know how to properly fix it but I found a workaround.

The problem seems to be that when we open the settings modal, the first time it renders it doesn't have the game config in the SettingsContext values yet so `enviromentOptions` returns `[]`, that gets passed finally to the TwoColTableInput component and store in the `rowData` state. Then the context gets the game settings but the old value is still in the `rowData` state.

If we render the modal ONLY after `contextValues.config` is populated it works fine, something like:

```ts
  if (Object.keys(contextValues.config).length === 0) {
    return null
  }

  return (
    <Dialog ...
```

This also prevents a small flash where you can see empty fields while the config is being loaded for a moment and then they get populated.

flavioislima:
@arielj thanks for find that. I added this check on the settings hook and then on the modal I check if `contextValues` is null or not. Now it is working fine.  ⚔️

src/frontend/components/UI/Sidebar/components/SidebarLinks/index.tsx
```diff
@@ -234,45 +181,20 @@ export default function SidebarLinks() {
...
-              <NavLink
-                role="link"
-                data-testid="linkSync"
-                to={`/settings/${runner}/${appName}/sync`}
-                state={{ ...state, runner: state?.runner }}
-                className={classNames('Sidebar__item SidebarLinks__subItem', {
-                  ['active']: type === 'sync'
-                })}
-              >
-                <span>{t('settings.navbar.sync')}</span>
```
arielj, line 261: I think we are losing access to the sync saves config now? I can' t test this at the moment to check

flavioislima: That's totally true. I can't believe I missed that. Thanks for finding this.

Yes, I will need to render the save sync section here as well. I will move it to game settings.

src/backend/logger/__tests__/logfile.test.ts
```diff
@@ -115,7 +115,7 @@ describe('logger/logfile.ts', () => {
       // @ts-ignore replaceAll error
       `heroic-${date.toISOString().replaceAll(':', '_')}.log`
     )
-    date.setFullYear(2021)
```
arielj, line 118: maybe do `date.getFullYear()-1` here?

arielj:
I noticed there are 2 things missing if we want to keep them:
- this footer saying the settings are saved automatically and the appname (I found this info useful)
image
- the `right click` to copy the config to the clipboard and open the config file (I have to say this context menu to me feels counter intuitive, it's hard to discover so I'm ok not having it but maybe it can be like actual buttons in the footer close to the previous item)

One suggestion: put the content of `Wine Extensions` inside the `Wine` collapsible. Since `Wine Extensions` is affected by the wine version selected in the `Wine` section, if `Wine Extensions` is collapsed it's easy to miss that it changed. Maybe it's too much info in one section but maybe it's worth to try?

PR #2281, opened 2022-12-29
[UI/UX] Add filter for non-available games

flavioislima:
Added a new button to filter out non-available games.

image
image

---

### DIFF
--- a/src/frontend/screens/Library/components/GameCard/index.tsx
+++ b/src/frontend/screens/Library/components/GameCard/index.tsx
@@ -48,9 +48,10 @@ import UninstallModal from 'frontend/components/UI/UninstallModal'
 interface Card {
   buttonClick: () => void
   hasUpdate: boolean
-  forceCard?: boolean
   isRecent: boolean
   gameInfo: GameInfo
+  isAvailable?: boolean
+  forceCard?: boolean
 }
 
 const storage: Storage = window.localStorage
@@ -60,13 +61,11 @@ const GameCard = ({
   buttonClick,
   forceCard,
   isRecent = false,
-  gameInfo: gameInfoFromProps
+  gameInfo: gameInfoFromProps,
+  isAvailable
 }: Card) => {
   const [gameInfo, setGameInfo] = useState(gameInfoFromProps)
   const [showUninstallModal, setShowUninstallModal] = useState(false)
-  const [gameAvailable, setGameAvailable] = useState(
-    gameInfoFromProps.is_installed
-  )
   const [isLaunching, setIsLaunching] = useState(false)
 
   const { t } = useTranslation('gamepage')
@@ -96,6 +95,8 @@ const GameCard = ({
     thirdPartyManagedApp
   } = gameInfoFromProps
 
+  const gameAvailable = isAvailable
+
   // if the game supports cloud saves, check the config
   const [autoSyncSaves, setAutoSyncSaves] = useState(hasCloudSave)
   useEffect(() => {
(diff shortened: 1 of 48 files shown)


src/frontend/screens/Library/components/GameCard/index.tsx
```diff
@@ -121,6 +123,28 @@ const GameCard = ({
           runner
         })
         setGameAvailable(gameAvailable)
+        const nonAvailbleGames = storage.getItem('nonAvailableGames') || '[]'
```
arielj, line 126: hmmm it feels weird to me to store this information in the browser's storage, I feel like the backend should provide that information when listing games instead?

flavioislima: I thought about that but at the same time I did not want to change the library if that is what you mean. I think handling the list on frontend it is more intuitive.
What do you propose? use electron-store?

arielj: hmmm I think what it feels weird to me is that we are storing that information at all, when the source of truth is if the game folder is present or not

maybe a `window.api.listUnavailableGames` or something like that that checks if the game folder for each of the installed games is available and returns the list of games that are not available? we are already asking the backend if a game is available or not

if that's not possible I guess it's fine to store that either in the browser storage or in an electron store

flavioislima: I am actually using it as a temporary store just to make the filter work. Because the filter is handled on the frontend. If we want to hide the non-available games on the frontend, that is more dynamic than requesting the list on the frontend. 
Besides, the nature of this call is pretty volatile since it the person can just plug the drive and refresh the library and the state will change.

Also, when starting heroic the store is cleaned up since it will check again which games are available or not.

src/frontend/screens/Library/components/GamesList/index.tsx
```diff
@@ -26,8 +56,66 @@ const GamesList = ({
...
+    createGameCards()
+  }, [
+    library,
+    onlyInstalled,
+    layout,
+    gameUpdates,
+    isRecent,
+    nonAvailbleGamesArray,
+    showNonAvailable
+  ])
```
arielj, line 118: I think `useMemo` is a better hook to use here, this is kinda recreating what useMemo does but it will trigger an extra render when calling setGameCards

flavioislima: good point, fixed it

src/frontend/screens/Library/components/GamesList/index.tsx
```diff
@@ -26,8 +56,66 @@ const GamesList = ({
   onlyInstalled = false,
   isRecent = false
 }: Props): JSX.Element => {
-  const { gameUpdates } = useContext(ContextProvider)
+  const { gameUpdates, showNonAvailable } = useContext(ContextProvider)
   const { t } = useTranslation()
+  const [gameCards, setGameCards] = useState<JSX.Element[]>([])
+
+  useMemo(() => {
```
arielj, line 63: sorry, I just realized the createGameCards stuff is async and useMemo is not intended for async memoization, sorry for suggesting that, useEffect is the right one here :see_no_evil: 

with sync content we could do `const gameCards = useMemo(() => ....)` instead of using useState+useEffect, but we can't here

flavioislima: hmm, weird that it did not broke anything, maybe the status update could break using memo, not sure.
anyway, reverted to useEffect

PR #2286, opened 2022-12-30
[UX] Disable the option to delete the prefix if it's the default prefix

### LINKED ISSUE #2256: Disable `delete prefix` option if prefix is the default prefix
### Problem description

Currently, when uninstalling a game we can check the option to delete the prefix, even if the prefix is Heroic's default prefix (`~/.wine` for example).

Users might click it by mistake or because of not knowing what that actually does and if there are other games using the same prefix that will break them.

### Feature description

In the uninstall dialog, we should still show the option but disabled so it CAN'T be checked and we should also include a message explaining WHY it's disabled (like `This games uses your default prefix: <path>. Delete the folder manually if you know what you are doing` or something like that).

### Alternatives

An alternative is to NOT display the option to delete the prefix in that case (but hiding things from the user with no explanation might be confusing since you don't know why it's missing and you might thing it's a bug)

### Additional information

_No response_

arielj:
This PR fixes #2256 

If the game's prefix is the default prefix, we disable the checkbox and show a message explaining why it is disabled and tell the user to delete it manually if they know what they are doing.

image

---


src/frontend/components/UI/UninstallModal/index.tsx
```diff
@@ -23,42 +23,51 @@ const UninstallModal: React.FC<UninstallModalProps> = function (props) {
     useContext(ContextProvider)
   const [isWindowsOnLinux, setIsWindowsOnLinux] = useState(false)
   const [winePrefix, setWinePrefix] = useState('')
-  const [checkboxChecked, setCheckboxChecked] = useState(false)
-  const [settingcheckboxChecked, setsettingCheckboxChecked] = useState(false)
+  const [deletePrefixChecked, setDeletePrefixChecked] = useState(false)
+  const [deleteSettingsChecked, setDeleteSettingsChecked] = useState(false)
```
arielj, line 27: I renamed this variables to make them more clear

src/frontend/components/UI/UninstallModal/index.tsx
```diff
@@ -23,42 +23,51 @@ const UninstallModal: React.FC<UninstallModalProps> = function (props) {
...
+
+    setIsWindowsOnLinux(true)
+    setWinePrefix(gameSettings.winePrefix)
+    setDisableDeleteWine(gameSettings.winePrefix === defaultSettings.winePrefix)
   }
 
   useEffect(() => {
     checkIfWindowsOnLinux()
-  })
+  }, [])
```
arielj, line 70: this was missing the dependencies and was firing more than needed

flavioislima (approved):
LGTM

PR #2294, opened 2022-12-31
[Gamepad] Match sony devices by vendor id

imLinguin:
This changes a regex to match for vendor id for Sony controllers, that way all their devices will have Sony prompts

This fixes the issue with DS4 having Xbox prompts

### DIFF
--- a/src/frontend/components/UI/ControllerHints/index.tsx
+++ b/src/frontend/components/UI/ControllerHints/index.tsx
@@ -121,7 +121,7 @@ export default function ControllerHints() {
 
   useEffect(() => {
     // set the brand for the images to use
-    if (activeController.match(/sony|0ce6|PS3|PLAYSTATION|0268|'2563.*0523/i)) {
+    if (activeController.match(/sony|054c|PS3|PLAYSTATION|0268|'2563.*0523/i)) {
       setLayout('ps')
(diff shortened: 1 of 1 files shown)


arielj (approved):

src/frontend/components/UI/ControllerHints/index.tsx
```diff
@@ -121,7 +121,7 @@ export default function ControllerHints() {
 
   useEffect(() => {
     // set the brand for the images to use
-    if (activeController.match(/sony|0ce6|PS3|PLAYSTATION|0268|'2563.*0523/i)) {
```
arielj, line 124: maybe we can add 054c instead of replacing it? I don't remember why I put 0ce6 at some point there but maybe removing it breaks this for some users

imLinguin: the 0ce6 is specific to DS5 controller, I think it's save to delete when we have a vendor id instead

Etaash-mathamsetty: yeah a vendor ID is probably the best idea

arielj:
Should we have this PR against `stable`? it's a small change and it could go on the next stable

PR #2298, opened 2023-01-01
Disable Winemenubuilder when launching with Wine

CommandMC:
Winemenubuilder is responsible for adding application shortcuts & managing file associations. For most users, it it more known as the reason all their `.ini` files are suddenly opening in Wine's Notepad
Disabling it is, in my opinion, the more sensible option. It is, for example, also done by Lutris (see [their FAQ](https://lutris.net/faq#winemenubuilder))

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -268,9 +268,24 @@ function setupWineEnvVars(gameSettings: GameSettings, gameId = '0') {
   // Add WINEPREFIX / STEAM_COMPAT_DATA_PATH / CX_BOTTLE
   const steamInstallPath = join(flatPakHome, '.steam', 'steam')
   switch (wineVersion.type) {
-    case 'wine':
+    case 'wine': {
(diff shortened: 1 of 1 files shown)


arielj:
Is there an easy way to test this? for me wine never creates shortcuts for the games I install for example, and I couldn't say that installing a game changed my file associations but maybe I was not paying attention or I didn't install the "right" game to do that

Etaash-mathamsetty:
> Is there an easy way to test this? for me wine never creates shortcuts for the games I install for example, and I couldn't say that installing a game changed my file associations but maybe I was not paying attention or I didn't install the "right" game to do that
> 
> The description of the PR sounds reasonable

This depends on the version of wine/proton you use, something based on proton would definitely disable the file associations by default (wine-ge maybe?), but something like wine staging or even vanilla wine won't do this by default, so it's a good idea to always just disable them

flavioislima (approved):

PR #2299, opened 2023-01-01
[Fix] Log file deletion logic

CommandMC:
When generating a 1-month-ago date, we were handling month 0 in a special case (and then not correctly handling it). This is actually not necessary, since JS takes care of this on its own

---


arielj (approved):

arielj:
@CommandMC I just noticed this went into `main`, we should have this in `stable` too, do you want to open a new PR?

PR #2301, opened 2023-01-01
[Settings] Update Alternative binary settings correctly

arielj:
After the refactor of the settings, I introduced by mistake some setting variables changes and in the case of the alternative binaries there was even a problem with the logic (setting was changed with one key using the settings context and with a different key updating the configStore explicitly). The problem was that removing the binary didn't update the values properly.

This PR fixes that by removing that extra code and using the binary we actually display in the settings.

The problem before this PR was easy to reproduce:
- set an alternative binary
- check the logs that it's using the alt binary to check its version
- remove the alternative binary
- check the logs, it still uses the previous alt binary to check the version instead of the built in binary
- set the built in binary as the alt one
- check the logs, it now uses the built in one

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -465,9 +465,11 @@ function splitPathAndName(fullPath: string): { dir: string; bin: string } {
 }
 
 function getLegendaryBin(): { dir: string; bin: string } {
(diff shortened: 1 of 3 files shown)


redromnon (approved):
Tested. LGTM :+1:

flavioislima (approved):

PR #2308, opened 2023-01-04
[Feat] Add settings to Auto Update games and Ignore Game Updates

flavioislima:
This adds a new global setting to auto update games, so on the game card it checks if the game has update and if the setting is enabled and if the ignore game updates is disabled. Then it send the game to the queue.

I think it is redundant to add this code to the game page since the library is where we check for the game updates so I am checking only on the game card.

Another option that was added was to ignore game updates. This Will ignore the auto update feature and also when clicking play it won't show the update warning.

I did not add this setting to the global game settings since I think this should be set individually.

---


Nocccer (approved):
LGTM. I can't test because i don't have games installed or they don't have any updates.

src/frontend/screens/Settings/components/IgnoreGameUpdates.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { useContext } from 'react'
```
arielj, line 1: Maybe this setting makes more sense to be per-game?

src/frontend/screens/Library/components/GameCard/index.tsx
```diff
@@ -126,6 +126,25 @@ const GameCard = ({
...
+      }
+      const settings = await window.api.requestAppSettings()
+      const gameSettings = await window.api.requestGameSettings(appName)
+      // if has update check for the AutoUpdateGames setting and if true, call handleUpdate
+      if (settings.autoUpdateGames && !gameSettings.ignoreGameUpdates) {
+        handleUpdate()
+      }
+    }
+    handleAutoUpdate()
+  }, [hasUpdate])
```
arielj, line 146: should this be handled in the backend instead? this will do this check every time we open the library

I think it makes more sense to handle this feature when we check for all game updates at `ipcMain.handle('checkGameUpdates'` in main.tsx instead

src/frontend/helpers/library.ts
```diff
@@ -196,6 +196,16 @@ const launch = async ({
   showDialogModal
 }: LaunchOptions): Promise<{ status: 'done' | 'error' }> => {
   if (hasUpdate) {
+    const { ignoreGameUpdates } = await window.api.requestGameSettings(appName)
+
+    if (ignoreGameUpdates) {
+      return window.api.launch({
+        appName,
+        runner,
+        launchArguments: '--skip-version-check'
```
arielj, line 205: I understand gogdl does not support this flag, is it ok to send it anyway? should we ignore this setting for gog games if we don't have a way to skip updates?

flavioislima: True, forgot about this. Will fix.

src/frontend/screens/Library/components/GameCard/index.tsx
flavioislima: Hmm, we check only if a game has Update, but I will check if it is possible to do that from the backend only.

flavioislima:
@arielj changed the logic to the backend, can you test again?
@imLinguin on macOS (will try on other OSes), if I change the installed.json manually to test the updates, it runs the updates but never updates the file and the game keep showing that there is an update. I don't think this was introduced here since the manual update also has the same issue. Can you test it when you have the time?

src/frontend/screens/Library/components/GameCard/index.tsx
flavioislima: done

imLinguin:
Sure I'll test it later

imLinguin:
Just tested this on this branch, everything updates correctly

flavioislima:
> Just tested this on this branch, everything updates correctly (at least for GOG)

have you tested more than one game?
I have tested Gris and Death's Door and Gris has issues while the other doesn't.

I will test a few more.

imLinguin:
Of course I tested it with multiple ones. I can't reproduce it.

arielj:
I think I've asked this a few times but it's never clear to me, is there an easy way to force an older version of a game so it has updates? or how is the simplest way to test this? is there a game that is known to have constant updates?

imLinguin:
For GOG you can force it by editing versionEtag in game `installed.json`  
Not sure about Epic

flavioislima:
Yeah, it seems only GRIS had this issue for some reason, so I will continue with the merging.

PR #2311, opened 2023-01-04
[Feature] Refactor wiki game Info and add applegamingwiki

Nocccer:
Refactor:
- Make use of PCGamingWiki howlongtobeat id 
- Add AppleGamingWikiInfo
- Added test for all functions
- Make it one ipc call instead of three.
- Added one component in frontend called WikiGameInfo which contains the old components underneath.
image

---


redromnon (commented):
Tested here on Linux. Both HowLongToBeat and Game Scores display perfectly. Didn't test for CrossOver.

src/backend/wiki_game_info/wiki_game_info.ts
```diff
@@ -0,0 +1,67 @@
...
+    logInfo(`Getting ExtraGameInfo data for ${title}`, LogPrefix.ExtraGameInfo)
+
+    const pcgamingwiki = await getInfoFromPCGamingWiki(title, gogID)
+    const applegamingwiki = isMac
+      ? await getInfoFromAppleGamingWiki(title)
+      : null
+    const howlongtobeat = await getHowLongToBeat(
+      title,
+      pcgamingwiki?.howLongToBeatID ?? ''
+    )
```
arielj, line 48: maybe we can do this 3 requests in parallel instead of sequentially?

then we can await on a Promise.all

arielj: also, this is wrapped with a try/catch, if any of the requests fail this all fails? for example if for some reason howlongtobeat is down this will make the pcgamingwiki fail

Nocccer: Should not happen because all three methods are also wrapped in try/catch. The try/catch here is only for electron-store. Don't know if that can throw something.
If not we could remove the try/catch.

Nocccer:
@flavioislima @arielj i removed the hltb id because the howlongtobeat package is anyway broken for specific id's. Now all three calls are parallel.

Do you know if electron-store can throw something, else we can remove the try/catch in `wiki_game_info.ts`

arielj (commented):
Adding some comments while I test this feature.

I see in the logs that the request to get the info is done when the game page is opened, even if I never click on the show extra info. Would be great to only do it when I click on the option.


I understand the howlongtobeat info is in a detail/summary block for people that do not want to see it, but I think it should be fine to always show the score, now it's kinda annoying that you click there and you have to do more clicks to actually see any real info


I tested with `A Game Of Thrones: The Board Game Digital Edition` and the `Show extra info` option is there but when I open it there's no info to show:
image

I don't know if we should hide the option or fallback to a `we wouldn't find info` message


Maybe change the name of the option to just `Extra info`? for consistency, since we don't use `Show system requirements` for the other dialog


The circles for how long to beat change the cursor to `pointer` but when clicked nothing happens, I don't know if it should open HLTB's website?


We should use `button` tags instead of divs, buttons are better for accessibility


I feel the `Extra` title in the dialog is kinda useless, what do you think about showing the info like this:
image

arielj:
> Do you know if electron-store can throw something, else we can remove the try/catch in `wiki_game_info.ts`

I think it doesn't throw exceptions, I don't see anything in the docs.

arielj:
I pushed the changes.

I only didn't change the requests to pcgamingwiki done when opening the game page since now I see it's used to get the steamid for the protondb link.

arielj (approved):

PR #2314, opened 2023-01-05
[Feature] Custom title bar

Nocccer:
Via the pacakge https://www.npmjs.com/package/custom-electron-titlebar it is possible and easy to add a custom titlebar and get rid of the os provided one.

- Added support for font and theme
https://user-images.githubusercontent.com/61798668/210848893-e2763294-b7ba-4ecf-bde9-da90965cbb17.mp4
- Titlebar supports electron dev menu in yarn start
image

Todo:
- Add support for f11 again. Only working in development mode somehow.
- Add custom icons for minimize, maximize and close.

---


src/backend/main_window.ts
```diff
@@ -66,12 +70,16 @@ export const createMainWindow = () => {
     }
   }
 
+  // setup the titlebar main process
+  setupTitlebar()
```
arielj, line 74: I think it's important to make this optional

- with this custom titlebar we lose the features the window manager shows when you right click the title bar for example
- users may have their own window decoration style/personalizations they want to use instead (like text alignment, font, icons, etc)

flavioislima:
Tested here and it is not working on macOS:
image

Nocccer:
> Tested here and it is not working on macOS: image

Seems that only the button does not work.

Nocccer:
Closed because macos is not well supported.

flavioislima:
I think should be fine to skip macOS on that and support windows and linux only,

imLinguin:
On Mac we could add this without window handles, since we can still position the trafficLights and make it integrate nicely into the UI

Nocccer:
> On Mac we could add this without window handles, since we can still position the trafficLight and make it integrate nicely into the UI
> 
> https://www.electronjs.org/docs/latest/tutorial/window-customization#customize-the-traffic-light-position-macos

We could do that for all os like that.

imLinguin:
Yeah, I bet that's what that lib is doing under the hood

Nocccer:
> Yeah, I bet that's what that lib is doing under the hood

Maybe. It could also do more. So maybe it is easier to write our own solution if we want that.

imLinguin:
We can achieve that with Electron's magic https://www.electronjs.org/docs/latest/tutorial/window-customization#set-custom-draggable-region

Nocccer:
> We can achieve that with Electron's magic https://www.electronjs.org/docs/latest/tutorial/window-customization#set-custom-draggable-region

Nice works great. Even double click maximize it.

PR #2325, opened 2023-01-08
[Bug] Fix TypeError if input for getRegDataType is undefined

Nocccer:
Should fix this problem reported here:
https://discord.com/channels/812703221789097985/1061633792823001098

---

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -17,23 +17,23 @@ diverse, inclusive, and healthy community.
 Examples of behavior that contributes to a positive environment for our
 community include:
 
-* Demonstrating empathy and kindness toward other people
-* Being respectful of differing opinions, viewpoints, and experiences
-* Giving and gracefully accepting constructive feedback
-* Accepting responsibility and apologizing to those affected by our mistakes,
+- Demonstrating empathy and kindness toward other people
+- Being respectful of differing opinions, viewpoints, and experiences
+- Giving and gracefully accepting constructive feedback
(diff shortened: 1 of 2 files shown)


imLinguin:
Let's explore this issue more, maybe we should default to some type when it's not provided

CommandMC:
`reg` uses `REG_SZ` as a default type, I assume the same is intended here

imLinguin:
Here is the instruction it fails on

```json
{
            "install": {
                "action": "setRegistry",
                "arguments": {
                    "deleteSubkeys": true,
                    "root": "HKEY_LOCAL_MACHINE",
                    "saveGameData": true,
                    "subkey": "Software\\Microsoft\\Games\\Outwars"
                }
            },
            "languages": [
                "*"
            ],
            "name": "r01"
        },
```
Not sure what `saveGameData` means, but I think we should ignore that for now. This PR should suffice for now, in the future we will switch to `scriptinterpreter.exe` nevertheless

imLinguin (approved):

imLinguin:
We may also consider this case

```json

{
            "install": {
                "action": "setRegistry",
                "arguments": {
                    "root": "HKEY_LOCAL_MACHINE",
                    "subkey": "Software\\Microsoft\\Games\\Outwars\\1.0.0",
                    "valueData": "",
                    "valueName": "IS UninstallString",
                    "velueType": "string"
                }
            },
            "languages": [
                "*"
            ],
            "name": "r18"
        },

```
Where `valueData` is empty

PR #2327, opened 2023-01-09
[FIX] Improve Wine not available / not found handler

flavioislima:
I am trying to find a solution to this issue people complain sometimes when wine was not found or maybe it was deleted somehow, the game fails to launch and you need to go to the settings and change it back and forth, etc.
So my solution is pretty simple:
- On launch, if the game is not native, check if the wine is valid. if not, try the default one on the config. 
- If the one set as default is not valid somehow, query the list of wine, proton, crossover, etc. And get the first one.
- if none was found, then it will fail and show a message to the user.

Besides that, I added a few QOL improvements on the settings part.
I mean, we have a WHOLE class to deal with the settings and there is no way to SET a Single setting, everytime we change a setting in Heroic it rewrites the whole file again. So if we need to change a setting on the backend we need to query the whole list, then change the key inside that object, and then rewrite the settings again. That is not practical.

So I added the `setSetting` method to the class on the backend and also added the ipc call on the frontend. so we can now set one single setting on backend and frontend, making our life easier if we need to handle something on runtime.

Another small method I added was `pathExists`, so we can query on the frontend if a path exists and use it to update the UI or show some message to the user, etc.

---

### DIFF
--- a/src/frontend/screens/Settings/components/WineVersionSelector.tsx
+++ b/src/frontend/screens/Settings/components/WineVersionSelector.tsx
@@ -29,12 +29,25 @@ export default function WineVersionSelector() {
     getAltWine()
   }, [])
 
+  useEffect(() => {
+    // Fix setting in case wine path does not exists anymore
+    const updateWine = async () => {
+      const winePathExists = await window.api.pathExists(wineVersion.bin)
+      if (!winePathExists) {
+        const { wineVersion: defaultWine } =
+          await window.api.requestAppSettings()
+        setWineVersion(defaultWine)
+      }
+    }
+    updateWine()
+  }, [wineVersion, altWine])
+
   return (
     <SelectField
       label={
         isLinux
           ? t('setting.wineversion')
-          : t('setting.crossover-version', 'Crossover Version')
+          : t('setting.crossover-version', 'Crossover/Wine Version')
       }
       htmlId="setWineVersion"
       onChange={(event) =>
@@ -62,7 +75,6 @@ export default function WineVersionSelector() {
                   <li>~/.local/share/lutris/runners/wine</li>
                   <li>~/.var/app/com.valvesoftware.Steam (Steam Flatpak)</li>
                   <li>/usr/share/steam</li>
-                  <li>Everywhere on the system (CrossOver Mac)</li>
                 </i>
               </ul>
               <span>{t('help.wine.part2')}</span>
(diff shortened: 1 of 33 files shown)


imLinguin (approved):
Looks good

src/frontend/screens/Settings/components/WineVersionSelector.tsx
```diff
@@ -29,12 +30,24 @@ export default function WineVersionSelector() {
     getAltWine()
   }, [])
 
+  useEffect(() => {
+    const updateWine = async () => {
+      const winePathExists = await window.api.pathExists(wineVersion.bin)
+      console.log('winePathExists', winePathExists, wineVersion.bin)
```
CommandMC, line 36:
```suggestion
```

src/backend/utils.ts
```diff
@@ -871,6 +875,86 @@ export const spawnAsync = async (
...
+  const wineIsValid = await validWine(gameSettings.wineVersion)
+
+  logError(
+    `Wine version ${gameSettings.wineVersion.name} is not valid, trying another one.`,
+    LogPrefix.Backend
+  )
+  appendFileSync(
+    logFileLocation,
+    `Wine version ${gameSettings.wineVersion.name} is not valid, trying another one.`
+  )
```
CommandMC, line 911: These messages should be after the `wineIsValid` check. Right now, they're always printed out

src/backend/utils.ts
```diff
@@ -871,6 +875,86 @@ export const spawnAsync = async (
...
+
+  if (!wineIsValid) {
```
CommandMC, line 913: Could implement an early-return here to avoid one level of indentation
```ts
if (wineIsValid) {
  return true
}
// check if the default wine is valid now
const { wineVersion: defaultwine } = ...
```

CommandMC (changes requested):

CommandMC:
Prettier wants lists in Markdown to be defined with `-`, not `*`. So 
```markdown
- Item1
- Item2
- Item3
```
and not
```markdown
* Item1
* Item2
* Item3
```

src/backend/utils.ts
flavioislima: done

src/backend/utils.ts
flavioislima: fixed

src/frontend/screens/Settings/components/WineVersionSelector.tsx
flavioislima: fixed

CommandMC (approved):

PR #2328, opened 2023-01-10
[Fix] Actually search for WINEDLLOVERRIDES when disabling WineMenuBuilder

CommandMC:
In #2298, I had *intended* to append the WineMenuBuilder override to an already-set `WINEDLLOVERRIDES` variable if one is present. While doing that however, I was accidentally searching for a variable with the name "WINEMENUBUILDER" instead
That's now fixed

---


flavioislima (approved):
LGTM

PR #2340, opened 2023-01-13
Use the exe's dir as the working dir to run exe in prefix

arielj:
When using the `Run EXE in prefix` option, the file is not executed in the file's directory as the working dir. This is a problem because it can break executables that reference other files relative to the current working dir.

This PR does 2 things:
- update `runWineCommand` for games to have the same signature as the generic `runWineCommand`
- pass the exe's dir as the `startFolder` to run the commands

The way I'm testing this is with Rayman 2 The Great Escape from GOG (I think it was a giveaway). It has a `GXSetup.exe` file in the install folder that needs to run with the correct working dir or it fails.

---


imLinguin (approved):
Looks good

flavioislima (approved):
Cool! 👍🏽

PR #2342, opened 2023-01-14
[Tech] Util restructure

Nocccer:
This PR restructure `util.ts` and split it into multiple files.
This helps to stop growing of the file and make the code more readable and testable.

- All functions that are related to wine went to a new `utils.ts` file in the `backend/wine` folder
- All functions that are related to gog went to a new `utils.ts` file in the `backend/gog` folder
- All functions that are related to legendary went to a new `utils.ts` file in the `backend/legendary` folder
- The left over functions were splitted into:
  - abort
  - app
  - connection
  - error
  - filesystem
  - format
  - game
  - process
  - runtime
  - system
- To still import function via `import {xyz} from 'backend/utils'` a index.ts was added which export all files under utils folder.

If there is any function or subsection that you think does not fit let me know.

---


.github/workflows/codecheck.yml
```diff
@@ -17,3 +17,5 @@ jobs:
         run: yarn
       - name: Check Typescript syntax
         run: yarn codecheck
+      - name: Find dead code
```
Nocccer, line 20: Remove

src/backend/config.ts
```diff
@@ -29,10 +29,10 @@ import {
   getSteamCompatFolder,
   configStore
 } from './constants'
-import { execAsync } from './utils'
 import { execSync } from 'child_process'
 import { logError, logInfo, LogPrefix } from './logger/logger'
 import { dirname, join } from 'path'
+import { execAsync } from './utils/process/process'
```
arielj, line 35: if we create an `index.ts` inside the `process` (and the others) and export things we can import things without this duplication, like `import { execAsync } from './utils/process'`

even more, we could also have an index.ts inside `/utils` that exports everything from every file inside it so we can still do `import { execAsync } from './utils'` like it was before and keep the separation in files for testability/maintainability

Nocccer: Sounds good to me. Will work on this.

Nocccer:
Closing this. To much conflicts. Probably redo would make more sense.

PR #2343, opened 2023-01-15
[Bug] Disable esync and fsync for windows

Nocccer:
Not sure if this is true but it seems fsync and esync is not available for windows. Prove me wrong.

---

### DIFF
--- a/src/frontend/screens/Settings/components/EnableEsync.tsx
+++ b/src/frontend/screens/Settings/components/EnableEsync.tsx
@@ -4,11 +4,16 @@ import { ToggleSwitch } from 'frontend/components/UI'
 import useSetting from 'frontend/hooks/useSetting'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faCircleInfo } from '@fortawesome/free-solid-svg-icons'
+import { isWindows } from 'backend/constants'
 
 const EnableEsync = () => {
   const { t } = useTranslation()
 
   const [enableEsync, setEnableEsync] = useSetting('enableEsync', false)
+
+  if (isWindows) {
+    return <></>
+  }
 
   return (
     <div className="toggleRow">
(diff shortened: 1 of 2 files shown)


arielj (commented):
I commented on the commit but it makes more sense to comment here on the PR

we are already only showing this for linux non native games only https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/frontend/screens/Settings/sections/GamesSettings/index.tsx#L120

I do think it makes sense to move the code here but we should remove conditional in the other places then

src/frontend/screens/Settings/components/EnableEsync.tsx
```diff
@@ -4,12 +4,17 @@ import { ToggleSwitch } from 'frontend/components/UI'
...
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faCircleInfo } from '@fortawesome/free-solid-svg-icons'
+import { isWindows } from 'backend/constants'
 
 const EnableEsync = () => {
   const { t } = useTranslation()
 
   const [enableEsync, setEnableEsync] = useSetting('enableEsync', false)
 
+  if (isWindows) {
```
arielj, line 14: in the other code in the GameSettings we only show this config if linux and non native, we should do the same check here (like we do for fsync)

src/frontend/screens/Settings/components/EnableFsync.tsx
```diff
@@ -14,7 +15,7 @@ const EnableFsync = () => {
   const isLinux = platform === 'linux'
   const [enableFsync, setEnableFsync] = useSetting('enableFsync', false)
 
-  if (!isLinux || isLinuxNative) {
+  if (!isLinux || isLinuxNative || isWindows) {
```
arielj, line 18: `!isLinux` already covers the os being windows or mac, no need to for extra conditional here

Nocccer:
Ok is already checked in Gamesettings as @arielj pointed out.

PR #2348, opened 2023-01-16
Read GlobalConfig to get alternative binaries

arielj:
This was not working, they were read from an electron store but they were updated using GlobalConfig so things were in different files.

---


flavioislima (approved):
I think I have used the `configStore` in the past because the `getSettings` was async. But since it is not anymore then should work just fine now.

PR #2350, opened 2023-01-16
[Fix] Save sync fixes

flavioislima:
This moves the handling of the auto-save syncs to the backend. Instead of doing it in several different files on the frontend like we are doing today, it will just check the setting on the launch command and do the download and upload. This should fix an issue on steam deck gaming mode after closing the games, where the auto sync won't work since the frontend is killed after the launching.

It also should fix some cases where the save path input is not shown sometimes.
Also added a method to retry to get the saves path on click and cancel the searching for the save path.
Fixed the visual glitch (I can call that) when showing that it was searching for the saves path. Created a new component that shows a simple text with the progress and accepts a callback.
Removed the text that shows the save location for GOG games, I think that is too technical and have no use case for the end user unless for debugging. This was causing the layout to break and be different from the one on legendary saves.

---


src/backend/main.ts
```diff
@@ -973,6 +978,24 @@ ipcMain.handle(
 
     logInfo(`Launching ${title} (${game.app_name})`, LogPrefix.Backend)
 
+    if (autoSyncSaves && isOnline()) {
```
Nocccer, line 981: Shoudl this not be disabled for sideload games?

src/backend/main.ts
```diff
@@ -1067,6 +1087,31 @@ ipcMain.handle(
 
     await addRecentGame(game)
 
+    if (autoSyncSaves && isOnline()) {
```
Nocccer, line 1090: Same here for sideloaded ?

Nocccer (changes requested):
LGTM. I have some question if we catching sideload games.

src/backend/main.ts
flavioislima: But there is any way to activate this setting on sideloaded games? I don't think so.

src/backend/main.ts
flavioislima: Same answer

PR #2352, opened 2023-01-17
[UX] Prevent old search to step on new search

arielj:
I noticed this bug that, if you type in the search field a few characters relatively fast (not really that fast), after a moment the search result was the one from the first character press and not the last. This was because the new letter updated things and a the useEffect of GameList was triggering without cleaning up the previous useEffect.

---


flavioislima (approved):
I cannot reproduce the issue on the. main branch but on this PR things works as well so I am fine with it.

PR #2362, opened 2023-01-18
[Feat] Enable Wine Manager on macOS

flavioislima:
This PR adds the ability to install Wine-crossover and Wine-Staging versions for macOS, making it even more simple for macOS user to play Windows games.

image

image

---

### DIFF
--- a/src/backend/config.ts
+++ b/src/backend/config.ts
@@ -167,13 +167,13 @@ abstract class GlobalConfig {
       return wineSet
     }
 
-    const winePaths: string[] = []
+    const winePaths = new Set<string>()
 
     // search for wine installed on $HOME/Library/Application Support/heroic/tools/wine
     const wineToolsPath = `${heroicToolsPath}/wine/`
     if (existsSync(wineToolsPath)) {
       readdirSync(wineToolsPath).forEach((path) => {
-        winePaths.push(join(wineToolsPath, path))
+        winePaths.add(join(wineToolsPath, path))
       })
     }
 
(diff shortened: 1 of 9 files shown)


src/backend/config.ts
```diff
@@ -166,35 +166,49 @@ abstract class GlobalConfig {
...
+          readFileSync(infoFilePath, 'utf-8')
+        ) as PlistObject
+        const version = info['CFBundleShortVersionString'] || ''
+        const name = info['CFBundleName'] || ''
+        const wineBin = join(winePath, '/Contents/Resources/wine/bin/wine64')
+        if (existsSync(wineBin)) {
+          wineSet.add({
+            ...this.getWineExecs(wineBin),
+            lib: `${winePath}/Contents/Resources/wine/lib`,
+            lib32: `${winePath}/Contents/Resources/wine/lib`,
```
Nocccer, line 202: Would not suggest adding the some folder for `lib` and `lib32`. This will add the folder twice to `LD_LIBRARY_PATH`

src/backend/wine/manager/utils.ts
```diff
@@ -32,12 +32,11 @@ async function updateWineVersionInfos(
   logInfo('Updating wine versions info', LogPrefix.WineDownloader)
   if (fetch) {
     logInfo('Fetching upstream information...', LogPrefix.WineDownloader)
+    const repositorys = isMac
```
Nocccer, line 35: Did this mistake aswell :D
I thought i fixed that in my PR.

```suggestion
    const repositories = isMac
```

Nocccer (approved):
LGTM. I have 2 suggestion which are not breaking anything.

src/backend/config.ts
flavioislima: Wine on mac does not support 32bit games so we need to use both, because otherwise the verifyWine check will fail.

src/backend/wine/manager/utils.ts
flavioislima: I wanted to fix but now I want to trademark it. 😂 
Lets fix it another time.

Issue #2330, opened 2023-01-11
Show a short explanation of what Sync Saves: Unsupported means


arielj:
### Problem description

Currently it's not clear for users where do we get that information and if there's anything that can be done for it to be supported.

### Feature description

When sync saves are unsupported we can show an `(i)` icon with a message like `This games does not support Epic's cloud saves. This information is provided by the game developers. Some games do implement their own cloud save system`

Changing `Epic's` for `GOG's` depending on the game.

(any suggestion for how to make this message more clear is welcomed)

### Alternatives

Another option is to still enable the `Sync Saves` section in the settings and display that message there, there's more space to put more text.

flavioislima:
Some games on GOG also does not support so maybe lets make the message less specific.

enzoIto:
Is still open? I would like to pick this issue

Nocccer:
> Is still open? I would like to pick this issue

I think it is still open. Feel free to pick it :)

PR #2366, opened 2023-01-19
[UI/Help] Changing unsupported sync save message

enzoIto:
This PR changes the `sync save` unsupported message as request here #2330

Thought about how to summarize the message to not just specify a platform. Feel free to give any sugestion!

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -354,7 +354,10 @@ export default React.memo(function GamePage(): JSX.Element | null {
                     }}
                   >
                     <b>{t('info.syncsaves')}:</b>{' '}
-                    {t('cloud_save_unsupported', 'This game does not support cloud saves')}
+                    {t(
+                      'cloud_save_unsupported',
+                      'This game does not support cloud saves'
+                    )}
                   </div>
                 )}
                 {!is_installed && !isSideloaded && !notSupportedGame && (
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -39,7 +39,8 @@ import { install } from 'frontend/helpers/library'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import {
   faTriangleExclamation,
-  faEllipsisV
+  faEllipsisV,
+  faCircleInfo
 } from '@fortawesome/free-solid-svg-icons'
 import { hasProgress } from 'frontend/hooks/hasProgress'
 import ErrorComponent from 'frontend/components/UI/ErrorComponent'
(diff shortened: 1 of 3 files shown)


Nocccer:
I think what @arielj wanted is a helper icon next to the message. So the message stays the same only a helper icon is added with more information.

Here is a example:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/f462e8fd571b335ca42668512b3c77f1a2de52f1/src/frontend/screens/Settings/components/AutoDXVK.tsx#L55-L61

enzoIto:
> I think what @arielj wanted is a helper icon next to the message. So the message stays the same only a helper icon is added with more information.
> 
> Here is an example:
> 
> https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/f462e8fd571b335ca42668512b3c77f1a2de52f1/src/frontend/screens/Settings/components/AutoDXVK.tsx#L55-L61

Oh ok, i'm gonna do that.

enzoIto:
@Nocccer this could be a dumb question, but how do i test it? I just implemented as the example above, but i would like to test the UI. Do I need to install some game that is cloud save unsupported to see the message?

arielj:
You can probably remove this conditionals so the message is always shown, just don't commit the changes

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -354,7 +355,14 @@ export default React.memo(function GamePage(): JSX.Element | null {
                     }}
                   >
                     <b>{t('info.syncsaves')}:</b>{' '}
-                    {t('cloud_save_unsupported', 'Unsupported')}
+                    <FontAwesomeIcon
+                      className="helpIcon"
+                      icon={faCircleInfo}
+                      title={t(
+                        'cloud_save_unsupported',
```
Nocccer, line 362:
```suggestion
                        'help.cloud_save_unsupported',
```

Nocccer (changes requested):
I would add a help parent to the key else it will be added to the root path of the translation file.
Also run `yarn i18n` locally once and push the updated translation file for `en`.

Everything else looks good. Tahnks for contributing.

src/frontend/screens/Game/GamePage/index.tsx
Nocccer: Ok i tested it and i found more problems.
This should fix it:
```typescript
                    {t('cloud_save_unsupported', 'Unsupported')}
                    <FontAwesomeIcon
                      className="helpIcon"
                      icon={faCircleInfo}
                      title={t(
                        'help.cloud_save_unsupported',
                        'This game does not support cloud saves. This information is provided by the game developers. Some games do implement their own cloud save system'
                      )}
                    />
```

enzoIto: Thanks!

Nocccer (approved):

flavioislima:
@enzoIto can you add a screenshot on how it looks now?

enzoIto:
> @enzoIto can you add a screenshot on how it looks now?

image

flavioislima:
@enzoIto cool, thanks for that  👍🏽

PR #2367, opened 2023-01-19
[UI/UX] Improve moving install feedback + Fixes and Refactor

flavioislima:
- Add moving progress for each file.
- Split moving functions and refactor them so we can use the same for both GOG and Epic.
- Use `robocopy` on windows instead of `mv`
- Use `rsync` on mac and linux when available, if not, use `mv` (without progress)

image

image

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -605,17 +605,13 @@ export default React.memo(function GamePage(): JSX.Element | null {
     }
 
     if (isMoving) {
-      if (file) {
-        return `${t('status.moving-files', 'Moving file')}  '${file}':  ${
-          percent ? `${percent}` : '...'
-        }`
-      }
-
-      if (percent) {
+      if (file && percent) {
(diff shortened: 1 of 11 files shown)


src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -605,7 +607,20 @@ export default React.memo(function GamePage(): JSX.Element | null {
...
+        return `${t('status.moving-files', 'Moving file')}  '${file}':  ${
+          percent ? `${percent}` : '...'
+        }`
+      }
+
+      if (percent) {
+        return `${t(
+          'status.moving',
+          'Moving Installation, please wait'
+        )}:  ${percent}`
```
CommandMC, line 620: Might want to do this with a translation variable, I don't think every language is going to have this `<message>: <percent>` form

flavioislima: Check now it that was what you meant.

CommandMC (approved):

PR #2372, opened 2023-01-22
[UI] Fix Incorrect Sideload settings and Refactor nativeGame

redromnon:
This PR fixes the wrong Settings options displayed for sideloaded games with the install platform as native (Linux/macOS).
It also refactors the way how games are identified as native using `window.api.isNative()`

---

### DIFF
--- a/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
+++ b/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
@@ -47,8 +47,7 @@ type Props = {
 export default function GamesSettings({ useDetails = true }: Props) {
   const { t } = useTranslation()
   const { platform } = useContext(ContextProvider)
-  const { isDefault, gameInfo, isMacNative, isLinuxNative } =
-    useContext(SettingsContext)
+  const { isDefault, gameInfo } = useContext(SettingsContext)
   const [wineVersion] = useSetting('wineVersion', defaultWineVersion)
   const isLinux = platform === 'linux'
   const isCrossover = wineVersion?.type === 'crossover'
@@ -57,11 +56,13 @@ export default function GamesSettings({ useDetails = true }: Props) {
   const [nativeGame, checkNativeGame] = useState(false)
 
   useEffect(() => {
-    window.api
-      .isNative({ appName: gameInfo?.app_name!, runner: gameInfo?.runner! })
-      .then((nativeGame) => {
-        checkNativeGame(nativeGame)
-      })
+    if (gameInfo) {
+      window.api
+        .isNative({ appName: gameInfo?.app_name, runner: gameInfo?.runner })
+        .then((nativeGame) => {
+          checkNativeGame(nativeGame)
+        })
+    }
   }, [])
 
   return (
--- a/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
+++ b/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
@@ -66,7 +66,7 @@ export default function GamesSettings({ useDetails = true }: Props) {
       }
       getIsNative()
     }
-  })
+  }, [])
 
   return (
     <>
--- a/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
+++ b/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
@@ -47,7 +47,7 @@ type Props = {
 export default function GamesSettings({ useDetails = true }: Props) {
   const { t } = useTranslation()
   const { platform } = useContext(ContextProvider)
-  const { isDefault, gameInfo } = useContext(SettingsContext)
+  const { isDefault, gameInfo, isLinuxNative } = useContext(SettingsContext)
   const [wineVersion] = useSetting('wineVersion', defaultWineVersion)
   const isLinux = platform === 'linux'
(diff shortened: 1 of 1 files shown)


src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
```diff
@@ -141,7 +141,9 @@ export default function SideloadDialog({
       art_cover: imageUrl,
       is_installed: true,
       art_square: imageUrl,
-      canRunOffline: true
+      canRunOffline: true,
+      is_linux_native: gameInfo.is_linux_native ?? false,
+      is_mac_native: gameInfo.is_mac_native ?? false
```
redromnon, line 146: Any different and better approach for this? 
Will defaulting it to false cause any issues? I didn't encounter any.

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -47,18 +47,23 @@ type Props = {
...
   const isCrossover = wineVersion?.type === 'crossover'
   const hasCloudSaves = gameInfo?.cloud_save_enabled
 
-  const nativeGame =
-    isWin ||
-    (isMacNative && gameInfo?.install.platform === 'Mac') ||
-    (isLinuxNative && gameInfo?.install.platform === 'linux')
+  const [nativeGame, checkNativeGame] = useState(false)
+
+  useEffect(() => {
```
flavioislima, line 58: Ideally we should use a async function here because otherwise we will won't block the rendering and the value will change after the component is loading.
Also the pattern for the useState is not good. if the value is `nativeGame` the setter should be `setNativeGame`.

```
  useEffect(() => {
  const getIsNative = async () => {
  if (gameInfo){
   const isNative = await window.api...
   setNativeGame(isNative)
  }
  getIsNative()
  }
```

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -47,18 +47,23 @@ type Props = {
 export default function GamesSettings({ useDetails = true }: Props) {
   const { t } = useTranslation()
   const { platform } = useContext(ContextProvider)
-  const { isDefault, gameInfo, isMacNative, isLinuxNative } =
-    useContext(SettingsContext)
+  const { isDefault, gameInfo, isLinuxNative } = useContext(SettingsContext)
```
flavioislima, line 50: Can't we remove `isLinuxNative` ?

flavioislima (changes requested):

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
redromnon: `isLinuxNative` is required for fixing the cloud save sync displaying for GOG Linux games. That fix was pushed to the Small Fixes PR. It displayed an error whenever I removed it.

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
redromnon: Noted.

redromnon: Using async is causing the bug to occur again, i.e. displaying Wine options for Linux Sideloaded games.
```ts
useEffect(() => {
    const getIsNative = async () => {
      if (gameInfo) {
        const isNative = await window.api.isNative({
          appName: gameInfo?.app_name,
          runner: gameInfo?.runner
        })
        setNativeGame(isNative)
      }
      getIsNative()
    }
  })
```

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
redromnon: It is needed that to fix the GOG Linux games cloud saves bug. Otherwise, it will show that setting even with using `window.api.isNative`

I pushed this fix to the now merged Small Fixes PR.

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
redromnon: Just needed to move the `gameInfo` if check above and it works -
```ts
useEffect(() => {
    if (gameInfo) {
      const getIsNative = async () => {
       ............
    }
  }
```

redromnon: If the above looks good to you, I'll push this.

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -47,18 +47,26 @@ type Props = {
...
+      const getIsNative = async () => {
+        const isNative = await window.api.isNative({
+          appName: gameInfo?.app_name,
+          runner: gameInfo?.runner
+        })
+        setNativeGame(isNative)
+      }
+      getIsNative()
+    }
+  })
```
flavioislima, line 69: One last thing, you need to add a empty array here otherwise it will run infinitely.
Everytime you add a useEffect always add an empty array or an array with the dependencies for it to check if they change then the useEffect will trigger again. In this case an empty array is fine:

```suggestion
  }, [])
```

PR #2376, opened 2023-01-22
Try to solve Download/Disk speed not getting picked up consistently

CommandMC:
As discussed on Discord:
When reading out install/update progress from the output of Legendary (and I assume GOGDL as well), we were sometimes not getting the Download/Disk write speed. This is caused by Python buffering the output to stdout, and us only sending progress updates if everything (ETA, download percent, as mentioned download/disk speed, etc.) is in *one* message. This is often not the case
I've attempted to fix this issue by sending `Partial<InstallProgress>` messages to the frontend instead. It *should* be able to handle those just fine (updating the properties that were included, and leaving those that weren't alone), however this does not work. It seems like the `hasProgress` hook's `progress` state variable is not actually getting saved (?), returning to its default value whenever a re-render happens

Maybe someone else can tell me what's going wrong here

---

### DIFF
--- a/src/backend/gog/games.ts
+++ b/src/backend/gog/games.ts
@@ -22,7 +22,8 @@ import {
   InstallArgs,
   InstalledInfo,
   InstallPlatform,
-  WineCommandArgs
+  WineCommandArgs,
+  InstallProgress
 } from 'common/types'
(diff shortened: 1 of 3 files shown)


CommandMC:
Closing as this was now implemented in #2496

PR #2384, opened 2023-01-27
[UI] New Icons + new Default Theme + CSS Variables Overhaul

flavioislima:
Screenshots:
will add soon

---


LithRakoon:
This is very pretty, but i have some concern about users with sensitive eyes. this colour palate could cause eyestrain, it think it would be better if the colors were a little more muted.

flavioislima:
> This is very pretty, but i have some concern about users with sensitive eyes. this colour palate could cause eyestrain, it think it would be better if the colors were a little more muted.
>   

Which parts exactly?

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -49,12 +49,11 @@ export default function GamesSettings({ useDetails = true }: Props) {
   const { platform } = useContext(ContextProvider)
   const { isDefault, gameInfo, isLinuxNative } = useContext(SettingsContext)
   const [wineVersion] = useSetting('wineVersion', defaultWineVersion)
+  const [nativeGame, setNativeGame] = React.useState(false)
```
arielj, line 52: is this change needed? we use `useContext` and `useEffect` directly, for consistency I would argue to use `useState` directly without `React.`

src/frontend/components/UI/Sidebar/index.scss
```diff
@@ -42,6 +51,7 @@
   cursor: pointer;
   font-size: var(--text-md);
   color: var(--navbar-accent);
+  color: var(--navbar-inactive, var(--navbar-accent));
```
arielj, line 54: this color is overriding the previous declaration, I see this repeated pattern of using 2 colors in other parts of this file too which looks incorrect

also, related to this change, themes need to be updated to use these new variables, for example this is how Marine's sidebar look now:
image

All themes now have the same active/inactive colors of the default theme instead of the actual theme's accent

src/frontend/styles/_buttons.scss
```diff
@@ -40,33 +40,43 @@ button {
 }
 
 .button.is-primary {
-  background: var(--primary);
+  background: var(--primary-button, var(--primary));
 }
 
 .button.is-primary:hover {
-  background: var(--primary-hover);
+  background: var(--primary-button, var(--primary-hover));
```
arielj, line 47: if `--primary-button` is defined then the hovered and non-hovered backgrounds will be the same

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -60,7 +86,7 @@
         max-height: 0px;
         overflow: hidden;
         right: 0;
-        top: 100%;
+        top: 112%;
```
arielj, line 89: with this change you can't hover from the button to the menu, since the space between it will count as un-hovering the button, hiding the menu (in the main branch you can currently hover the menu without clicking the button)

src/frontend/components/UI/SearchBar/index.css
```diff
@@ -3,8 +3,9 @@
   width: 100%;
   position: relative;
   display: inline-flex;
-  background: var(--input-background);
-  border-radius: var(--space-3xs);
+  background: var(--search-bar-background, var(--input-background));
+  border-radius: var(--space-md);
+  border: 2px solid var(--search-bar-border, var(--input-backgroundd));
```
arielj, line 8: just personal opinion, but this border looks a bit off in some themes other than the new default, in dracula it feels `blurry`

maybe it can be added only in the new theme instead of being a default of the element?

I also think the color is too dark and the contrast is not great

LithRakoon:
> > This is very pretty, but i have some concern about users with sensitive eyes. this colour palate could cause eyestrain, it think it would be better if the colors were a little more muted.
> 
> Which parts exactly?
 
the blue accent colors are too oversaturated/neon, this will easily cause eyestrain for people for example me at night.
i think if they were more muted but still have the same hue that they would be pretty nice.
the blues for example here:
image
image

arielj (commented):
there are some issues when changing themes:

some elements now are using the default theme colors instead of the theme's colors, check this:
image

the sidebar text, the magnifying glass, the add game, and the play button, they are using the colors from the new theme instead of the colors from Zombie

Nord Dark looks broken:
image
image
image

titles, toggle labels, and the epic logo in the card are blank on dark background

flavioislima:
@arielj ok, will check this out its being complicated to adapt the new themes to the new variables format, also the custom themes are not working properly as well, even though I can follow the variables to the point where I can see that it is using the ones defined on the css, but still for some reason they do not load. Not all of them.

arielj:
> @arielj ok, will check this out its being complicated to adapt the new themes to the new variables format, also the custom themes are not working properly as well, even though I can follow the variables to the point where I can see that it is using the ones defined on the css, but still for some reason they do not load. Not all of them.

maybe we can make a simpler PR just adding the new theme with the current available variables instead of changing the css variables? then after the 2.6.0 release we can take more time to change variables, adapt all themes, etc

if something specific is needed for the new theme we can always do that with more css selectors to target specific elements anyway

flavioislima:
> > @arielj ok, will check this out its being complicated to adapt the new themes to the new variables format, also the custom themes are not working properly as well, even though I can follow the variables to the point where I can see that it is using the ones defined on the css, but still for some reason they do not load. Not all of them.
> 
> maybe we can make a simpler PR just adding the new theme with the current available variables instead of changing the css variables? then after the 2.6.0 release we can take more time to change variables, adapt all themes, etc
> 
> so the refactor doesn't delay the release
> 
> if something specific is needed for the new theme we can always do that with more css selectors to target specific elements anyway

Yes, but the problem is that everything is a mess today. We have several variables that are not even being used, we do not have a default place with the colors and so on. So in the future if we want to introduce the feature to create a theme from inside heroic the way that it is today would be impossible. 

I will see what I can do to this, will try to fix the issues, this is still WIP, if I cannot fix it until tomorrow I will split it into different PRs. 

There are other improvements that I did on this one as well that I will see how I will port to a different PR.

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
flavioislima: The change is needed, will fix the import

arielj:
I agree that a lot of things can be refactored and improved there, we have a mess, I'm just saying that maybe it's better to do that refactor after the release, it's a big change this close to a release

src/frontend/components/UI/Sidebar/index.scss
flavioislima: Fixed it, I did not add all variables like you said, I added the default theme on the Themes.css, removed it from colors.css. Also fixed a few variables to match the correct style.

flavioislima: image

src/frontend/styles/_buttons.scss
flavioislima: That's intentional, on the new theme now we add a Box shadow on the primary and success buttons instead of changing the background. It is a way of differentiating from the install button. It was not my idea, it is on the new design system on figma for the next UI but I think it is cool so I added it here on this design as well.
I think Billie changed that because for some people the color of the button won't change due to visual impairment. I have some issues myself with colors that are too close to each other so adding this box shadow, makes it is easier to see that we are hovering it.

image

image

src/frontend/screens/Game/GamePage/index.scss
flavioislima: Ok, found a fix now, it is working fine now and we kept the spacing.

src/frontend/components/UI/SearchBar/index.css
flavioislima: I changed it to show only when on focus, also it is not a border anymore but a box-shadow.

flavioislima:
Merging this one.

also @LithRakoon changed the colors like you said, should be better now.

PR #2385, opened 2023-01-27
[Tech] Refactor Download Dialog + Cache GOG Install Info

flavioislima:
Basically the title.
cache the GOG install info by platform and also simplifies the Download Dialog that was doing some weird stuff to get the `installInfo` and the platform list.

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -99,6 +99,8 @@ export default React.memo(function GamePage(): JSX.Element | null {
   const [gameAvailable, setGameAvailable] = useState(true)
 
   const isWin = platform === 'win32'
+  const isLinux = platform === 'linux'
+  const isMac = platform === 'darwin'
   const isSideloaded = runner === 'sideload'
 
   const isInstalling = status === 'installing'
@@ -144,10 +146,21 @@ export default React.memo(function GamePage(): JSX.Element | null {
   useEffect(() => {
     const updateConfig = async () => {
       if (gameInfo) {
-        const { install } = gameInfo
+        const {
+          install,
+          is_linux_native = undefined,
+          is_mac_native = undefined
+        } = { ...gameInfo }
 
-        if (runner !== 'sideload' && !notSupportedGame && install.platform) {
-          getInstallInfo(appName, runner, install.platform)
+        const installPlatform =
+          install.platform || (is_linux_native && isLinux)
+            ? 'linux'
+            : is_mac_native && isMac
+            ? 'Mac'
+            : 'Windows'
+
+        if (runner !== 'sideload' && !notSupportedGame) {
+          getInstallInfo(appName, runner, installPlatform)
             .then((info) => {
               if (!info) {
                 throw 'Cannot get game info'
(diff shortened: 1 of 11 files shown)


src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -144,10 +146,21 @@ export default React.memo(function GamePage(): JSX.Element | null {
...
-
-        if (runner !== 'sideload' && !notSupportedGame && install.platform) {
-          getInstallInfo(appName, runner, install.platform)
+        const {
+          install,
+          is_linux_native = undefined,
+          is_mac_native = undefined
+        } = { ...gameInfo }
+
+        const installPlatform =
```
flavioislima, line 155: Needed to bring back this logic because it was making the getInstallInfo never returns anything. It was previously changed on some recent PR.

src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx
```diff
@@ -218,7 +211,7 @@ export default function DownloadDialog({
   }
 
   useEffect(() => {
-    const getIinstallInfo = async () => {
+    const getIinstInfo = async () => {
```
Nocccer, line 214: Typo ?

```suggestion
    const getInstallInfo = async () => {
```

src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx
```diff
@@ -264,24 +257,9 @@ export default function DownloadDialog({
         return
       }
     }
-    getIinstallInfo()
+    getIinstInfo()
```
Nocccer, line 260: Typo ?

```suggestion
    getInstallInfo()
```

Nocccer (changes requested):
LGTM but found possible typo

src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx
flavioislima: Not a typo, the problem is that we already have a getInstallInfo from import.
So in lack of a better name, I intentionally changed this one.

src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx
flavioislima: same as above

src/frontend/screens/Library/components/InstallModal/index.tsx
```diff
@@ -136,8 +136,8 @@ export default React.memo(function InstallModal({
           setPlatformToInstall(e.target.value as InstallPlatform)
         }
       >
-        {availablePlatforms.map((p) => (
-          <option value={p.value} key={p.value}>
+        {availablePlatforms.map((p, i) => (
+          <option value={p.value} key={i}>
```
arielj, line 140: is this change needed? in general something unique from the element is better than an index in a collection for the key (if, for any reason, the elements in the array change order -not sure if it can actually happen in this case-, the key points to an incorrect element, we've had issues with that in other places)

I'd argue to always use something that is unique from the current element unless there's a good reason not to

flavioislima: They wont change the order since it is a hardcoded object. But after I have removed the useState related with the default platform, it was stuck on the first value. So I changed that, although other things might have fixed it and I just did not tested this change again. Anyway, I don't think this causes any harm like I said since it is a hardcoded object.

PR #2408, opened 2023-02-03
[Fix] Run game update check when Heroic starts

CommandMC:
We weren't checking for game updates when Heroic was launched. Now we are again!
See commit messages for details

I don't feel entirely confident about this fix (after all, the `library` check was *probably* there for a reason), but nothing seemed to explode when removing it

---

### DIFF
--- a/src/frontend/state/GlobalState.tsx
+++ b/src/frontend/state/GlobalState.tsx
@@ -421,7 +421,7 @@ class GlobalState extends PureComponent<Props> {
     }
 
     let updates = this.state.gameUpdates
-    if (checkUpdates && library) {
+    if (checkUpdates) {
       try {
         updates = await window.api.checkGameUpdates()
       } catch (error) {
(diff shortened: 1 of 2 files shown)


src/frontend/state/GlobalState.tsx
```diff
@@ -421,7 +421,7 @@ class GlobalState extends PureComponent<Props> {
     }
 
     let updates = this.state.gameUpdates
-    if (checkUpdates && library) {
+    if (checkUpdates) {
```
flavioislima, line 424: maybe we can just remove these checks and then call the checkGameUpdate on every refresh for runner 'all' or 'legendary'

flavioislima (approved):
LGTM

src/frontend/state/GlobalState.tsx
CommandMC: GOG Games can also have updates

Other than that, I believe this system is currently working fine. Why change it?

PR #2414, opened 2023-02-03
Readme: add instructions for fixing app is damaged error

MichaelGDev48:
This adds instructions for users on fixing app is damaged error.
---


MichaelGDev48:
This readme instruction should be removed when the binaries are codesigned by the team (can be done for free) till then this will help users

flavioislima:
Thanks for that!

PR #2407, opened 2023-02-02
Update issue template with the new Heroic version

arielj:
Removed the `beta` since the latest stable is newer than any old beta.

Added the `2.6.1` in the selector so that information is not lost.

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #2416, opened 2023-02-03
[Fix] Fixes for 2.6 release

flavioislima:
---

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -48,7 +48,7 @@ body:
   - type: markdown
     attributes:
       value: |-
-        ## Latest stable version of Heroic is 2.5.2. Please, update Heroic before reporting a bug if you have an older version.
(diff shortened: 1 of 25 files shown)


arielj (approved):
I tested things like the gog downloads for linux, the download size, themes

Things look good to me :+1: (some theme's still have small issues, but I don't think they are critical)

I think we should include an update of gogdl too since many users are having issues downloading games (I made a PR last night for it, I'll wait for linguin to review it)

One small thing to add, can you include an edit of the github issue template mentioning the `latest stable` to be `2.6.2`? like this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2407 (but that one is going to be outdated already)

flavioislima:
Yes, sure. Will do that on this PR and wait for your fix on gogdl but I think we should release the hotfix today since we have some critical stuff broken.

arielj:
> Yes, sure. Will do that on this PR and wait for your fix on gogdl but I think we should release the hotfix today since we have some critical stuff broken.

one thing we can do, if the fix is not ready by the time you want to release, we can downgrade gogdl to the commit before the latest, that one works fine

PR #2427, opened 2023-02-05
[Fix] Update Discord rich presence icon

TabulateJarl8:
Since the icon was updated in a recent release, I have added it as an asset in the Discord developer portal and updated the code in Heroic to reflect this new icon.

image

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -588,7 +588,7 @@ function constructAndUpdateRPC(gameName: string): RpcClient {
   client.updatePresence({
     details: gameName,
     instance: true,
-    largeImageKey: 'icon',
+    largeImageKey: 'icon_new',
     large_text: gameName,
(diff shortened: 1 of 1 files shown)


Etaash-mathamsetty:
haven't tested, but seems good to me

Nocccer:
Can you add this maybe to my PR? So we don't have two PR for discord rich presence.

TabulateJarl8:
How might I go about doing that? Should I just change the branch I'm merging into to be the branch your PR is on or should I do something else?

Nocccer:
> How might I go about doing that? Should I just change the branch I'm merging into to be the branch your PR is on or should I do something else?

Sorry for the late reply. Just let it as it is. I already merged my PR.

Nocccer (approved):
Checked the icon name. LGTM. Thanks for that.

Issue #2284, opened 2022-12-29
Option to dismiss the controller-friendly layout / revert to mouse-friendly layout


arielj:
### Problem description

When using Heroic on a system with both mouse and game controller inputs, if the controller gets activated Heroic changes the layout to be more `controller-friendly` removing some buttons.

If you don't want to use that layout for any reason (like you moved the controller by mistake to play a game and not intentionally to control Heroic), you have to either unplug the controller OR restart Heroic.

### Feature description

Ideally, detect when the user interacts using a mouse and revert the layout to the mouse friendly one.

If not possible, show something in the UI to revert back to the `mouse-friendly` layout. Maybe in the bottom bar with the controller hints we can have a button with a `mouse` icon and a legend `revert to original layout` or something like that.

I'm not sure for how long this reverted change should last in this case. Should it work until heroic is closed? should moving the controller again trigger the controller-friendly ui again? I don't know what's best. Safest/easiest is to start with just the button revert the layout until the controller is used again

PR #2430, opened 2023-02-06
[UX] Remove controller hints when using mouse

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2284

When using the steam deck, it's really common to switch between mouse/gamepad controller layout, there's an issue that if you prefer to use the mouse layout to controller heroic but, for any reason, you use the gamepad layout (like right after exiting or before starting a game) then you are locked in that heroic interface (without the settings/play buttons).

This PR changes that by detecting when the mouse moves and revering the layout.

Users can move back and forth between layouts by just using the different controller layouts and Heroic adapts.

---

### DIFF
--- a/src/frontend/helpers/gamepad.ts
+++ b/src/frontend/helpers/gamepad.ts
@@ -411,6 +411,15 @@ export const initGamepad = () => {
 
     currentController = controllerIndex
     dispatchControllerEvent(gamepad.id)
+
+    window.addEventListener(
+      'mousemove',
+      () => {
+        currentController = -1
+        dispatchControllerEvent('')
+      },
+      { once: true }
+    )
   }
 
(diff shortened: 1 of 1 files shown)


src/frontend/helpers/gamepad.ts
```diff
@@ -411,6 +411,15 @@ export const initGamepad = () => {
...
     currentController = controllerIndex
     dispatchControllerEvent(gamepad.id)
+
+    window.addEventListener(
+      'mousemove',
+      () => {
+        currentController = -1
+        dispatchControllerEvent('')
+      },
+      { once: true }
```
arielj, line 421: only listen for the first mouse movement after a controller is activated, we don't need to listen to that constantly

Nocccer (approved):
LGTM

PR #2435, opened 2023-02-07
[EPIC] Login with preload script

arielj:
This PR is a replacement of https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/1762

I think I found a simpler approach. Instead of overriding the epic functions, the injected JS just adds an `onload` function that, when the expected json response is present, will send a message to the host (the react app) with the code.

It is similar to the current login method but we don't rely on `find-in-page` and we don't mess with the user's clipboard.

The only thing I'm not sure about is that I was not able to make the `contextBridge.exposeInMainWorld` code work so I can't seem to hide the ipcRenderer under a function, but I'm only using it in a callback so I think it's still isolated?

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "heroic",
-  "version": "2.6.2",
+  "version": "2.6.3",
   "private": true,
   "main": "build/electron/main.js",
   "homepage": "./",
@@ -28,7 +28,8 @@
       "build/icon.icns",
       "build/win_icon.ico",
       "build/icon-dark.png",
-      "build/icon-light.png"
+      "build/icon-light.png",
+      "build/webviewPreload.js"
     ],
     "protocols": [
       {
--- a/public/webviewPreload.js
+++ b/public/webviewPreload.js
@@ -1,0 +1,9 @@
+const { ipcRenderer } = require('electron')
+
+window.onload = () => {
+  const content = document.body.innerText
+  if (content.match(/"authorizationCode":/)) {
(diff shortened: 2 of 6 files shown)


derrod:
I think you forgot the preload script.

But it's also worth noting that the reason to emulate the launcher was to allow for third-party logins (Google, Nintendo, Facebook, etc.) which may block Electron itself (Google does at least).

arielj:
> I think you forgot the preload script.

oh right, I think it's gitignored, I'll check that

> But it's also worth noting that the reason to emulate the launcher was to allow for third-party logins (Google, Nintendo, Facebook, etc.) which may block Electron itself (Google does at least).

hmmm good point, I only tested the normal email+password login, I'll see what happens when I try to log in using one of the other methods

arielj:
@derrod not sure if I'm missing something, I did try to login using Gmail for example and it works since after the Google login process it takes me to the same page with the authorization code json response (I have to login to Gmail inside Heroic, but that's how it's currently in the stable release too)

I added the preload js file to the PR

Nocccer (approved):
Tested it with google login and it worked.

flavioislima (changes requested):
Tested here, codewise looks good but I think heroic cannot find the preload file after packaging the build. I get stuck here (notice that I did bump the version number to be sure it is your build):


image

flavioislima (approved):
It was a simple fix to make the preload to work now, everything works fine after that :D

PR #2444, opened 2023-02-10
[FIX] Fix various issues with wine management

Etaash-mathamsetty:
this fixes various issues with wine and how esync and fsync are handled, makes winetricks and winecfg open more consistently, enabling and disabling dxvk and vkd3d now always work, no matter if fsync and esync are enabled, it just works significantly better and is less broken

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -881,11 +881,11 @@ ipcMain.handle('requestSettings', async (event, appName) => {
 })
 
 ipcMain.handle('toggleDXVK', async (event, { appName, action }) =>
-  DXVK.installRemove(
-    await GameConfig.get(appName).getSettings(),
-    'dxvk',
-    action
-  )
+  GameConfig.get(appName)
+    .getSettings()
+    .then(async (gameSettings) =>
+      DXVK.installRemove(gameSettings, 'dxvk', action)
+    )
 )
 
 ipcMain.on('toggleVKD3D', (event, { appName, action }) => {
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -589,7 +589,6 @@ ipcMain.handle('callTool', async (event, { tool, exe, appName, runner }) => {
             wait: false
           })
         : game.runWineCommand({
-            gameSettings,
             commandParts: ['winecfg']
           })
       break
@@ -604,7 +603,6 @@ ipcMain.handle('callTool', async (event, { tool, exe, appName, runner }) => {
               wait: false
             })
           : game.runWineCommand({
-              gameSettings,
               commandParts: [exe],
               startFolder: workingDir
             })
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -584,7 +584,7 @@ ipcMain.handle('callTool', async (event, { tool, exe, appName, runner }) => {
     case 'winecfg':
       isSideloaded
         ? runWineCommand({
-            gameSettings: gameSettings,
+            gameSettings,
             commandParts: ['winecfg'],
             wait: false
           })
@@ -598,13 +598,13 @@ ipcMain.handle('callTool', async (event, { tool, exe, appName, runner }) => {
         const workingDir = path.parse(exe).dir
         isSideloaded
           ? runWineCommand({
-              gameSettings: gameSettings,
+              gameSettings,
               commandParts: [exe],
               startFolder: workingDir,
               wait: false
             })
           : game.runWineCommand({
-              gameSettings: gameSettings,
+              gameSettings,
               commandParts: [exe],
               startFolder: workingDir
             })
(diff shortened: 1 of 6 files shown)


Etaash-mathamsetty:
this is a nice fix for the wineboot -u in dxvk code for now, it's more ideal to get rid of it entirely, but I am not sure how I am supposed to do that, so I will leave it for later

flavioislima (changes requested):

src/backend/main.ts
```diff
@@ -579,29 +579,32 @@ ipcMain.handle('callTool', async (event, { tool, exe, appName, runner }) => {
...
   switch (tool) {
     case 'winetricks':
-      await verifyWinePrefix(gameSettings)
       await Winetricks.run(wineVersion, winePrefix, event)
       break
     case 'winecfg':
       isSideloaded
         ? runWineCommand({
-            gameSettings,
+            gameSettings: gameSettings,
```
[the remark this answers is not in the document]
Etaash-mathamsetty: I thought it looked more organized with it in there

Nocccer (changes requested):
I see we passing winePrefix and winePath which in my eyes are part of the gamesettings.
So instead parsing winePrefix, winePath and gamesettings, we could probably only send gameSettings. Is this possible?

Etaash-mathamsetty:
> I see we passing winePrefix and winePath which in my eyes are part of the gamesettings. So instead parsing winePrefix, winePath and gamesettings, we could probably only send gameSettings. Is this possible?

I knew you guys were going to ask for this, I was just too lazy and was hoping you won't notice :sweat_smile:

Etaash-mathamsetty:
idk what's going on with the re request review stuff

Nocccer (commented):
Codewise looks good but i still wonder if this doe snot break somehting? Does the runWineCommand know which prefix to choose and where the `wineBin` is?

I think @CommandMC should review that.

CommandMC:
> Does the runWineCommand know which prefix to choose and where the `wineBin` is?

Yes, `runWineCommand` takes care of everything (and also handles things like spaces or other special characters correctly, as it uses `spawn` instead of `exec`)

src/backend/main.ts
```diff
@@ -879,12 +880,20 @@ ipcMain.handle('requestSettings', async (event, appName) => {
...
 )
 
-ipcMain.on('toggleVKD3D', (event, { winePrefix, winePath, action }) => {
-  DXVK.installRemove(winePrefix, winePath, 'vkd3d', action)
+ipcMain.on('toggleVKD3D', (event, { appName, action }) => {
+  GameConfig.get(appName)
+    .getSettings()
+    .then((gameSettings) => {
+      DXVK.installRemove(gameSettings, 'vkd3d', action)
+    })
```
CommandMC, line 896: This is nitpicky, but maybe make these have the same format (so either use `.then` in both, or `await` in both)? Otherwise, someone looking at this in the future might wonder why they're structured differently

Etaash-mathamsetty: the reason they are different is because the toggleDXVK has to return a Promise<Boolean> and DXVK.InstallRemove returns that Promise<Boolean>

CommandMC: You could still change either one to use the format of the other. As an example:
```patch
Index: src/backend/main.ts
IDEA additional info:
<+>UTF-8
diff --git a/src/backend/main.ts b/src/backend/main.ts
--- a/src/backend/main.ts	(revision Staged)
+++ b/src/backend/main.ts	(date 1677780018583)
@@ -881,11 +881,9 @@
 })
 
 ipcMain.handle('toggleDXVK', async (event, { appName, action }) =>
-  DXVK.installRemove(
-    await GameConfig.get(appName).getSettings(),
-    'dxvk',
-    action
-  )
+  GameConfig.get(appName)
+    .getSettings()
+    .then((settings) => DXVK.installRemove(settings, 'dxvk', action))
 )
 
 ipcMain.on('toggleVKD3D', (event, { appName, action }) => {
```

CommandMC: The patch I've posted there passes `yarn codecheck`. It still returns a `Promise<boolean>`, as it should.

Are you sure you haven't added curly brackets around everything after the `=>`?
```ts
() => false
```
is different from
```ts
() => { false }
```
after all (one returns `false`, the other `void`)

Etaash-mathamsetty: > The patch I've posted there passes `yarn codecheck`. It still returns a `Promise<boolean>`, as it should.
> 
> Are you sure you haven't added curly brackets around everything after the `=>`?
> 
> ```ts
> () => false
> ```
> 
> is different from
> 
> ```ts
> () => { false }
> ```
> 
> after all (one returns `false`, the other `void`)

thanks for that, fixed

CommandMC (approved):

Nocccer (approved):

PR #2448, opened 2023-02-12
[UI/UX] IMprovements and fixes on Gamepage, Dialogs and more

flavioislima:
- Changelog dialog: fixed link to previous URLs opening inside the main window, requiring a restart when on game mode in the SteamDeck.
- Sort by status item now will work as a filter to show only installed games instead of just sorting them by install status.
- Fix: return back to the library's previous position when hitting the back button.
- Added a border on the sideload image.
- Show a spinner icon when adding the game on the sideload dialog.
- Library header will be sticky now, making it easier to access the filters
- Decreased the animation for the gamecards to 100ms instead of 200ms.
- Fix error message on the native dialog box since it was totally broken when using special characters
- Show a error dialog for all other legendary errors to make it easy to debug.
- Created a generic popover element to allow using a popup for any icon or text when needed.
- Moved HLTB, Scores, Game Install info and time played to a popup to show when clicked on the respective text. (why not using hover? because when navigating with the joystick it was opening the elements what is not desired, so clicking here makes more sense imo.
- Other styles fixes. 

---


Etaash-mathamsetty:
I feel like there should be more of an indication that you can hover over that text and actually see the how long to beat and whatnot, I will test this later

flavioislima:
> I feel like there should be more of an indication that you can hover over that text and actually see the how long to beat and whatnot, I will test this later 

The problem is that we got requests to not show these informations all the time. On hover is easy to see. Yes, some people don't like them. So clicking should be more fine.

Also when using the joystick, pressing the button to show is better than showing it when selected.

src/frontend/screens/Library/index.tsx
```diff
@@ -86,15 +86,24 @@ export default React.memo(function Library(): JSX.Element {
   useLayoutEffect(() => {
     const scrollPosition = parseInt(storage?.getItem('scrollPosition') || '0')
 
-    if (listing.current !== null && scrollPosition > 0) {
-      listing.current.scrollTo(0, scrollPosition)
-    }
+    listing.current?.addEventListener('scroll', () => {
```
arielj, line 89: would be better to extract this callback as a function like:

```ts
const storeScroll = () => {
  ...
}

listing.current?.addEventListener('scroll', storeScroll)
```

that way you can then do `listing.current?.removeEventListener('scroll', storeScroll)` without the function duplication

flavioislima: done

src/frontend/components/UI/DialogHandler/components/MessageBoxModal/index.tsx
```diff
@@ -17,8 +17,24 @@ interface MessageBoxModalProps {
   type: DialogType
 }
 
+function decodeHTML(html: string) {
```
arielj, line 20: would be nice to add a comment on this for the future, I think it makes sense now because I can see the PR but it's kinda confusing without context

src/frontend/components/UI/DialogHandler/components/MessageBoxModal/index.tsx
```diff
@@ -17,8 +17,24 @@ interface MessageBoxModalProps {
...
+        <p />
+      </span>
+    )
+  })
+}
+
 const MessageBoxModal: React.FC<MessageBoxModalProps> = function (props) {
   const { t } = useTranslation()
+
+  const message = decodeHTML(props.message)
```
arielj, line 36: maybe too performance nit picky, but this can probably be memoized so it doesn't run on every render and only of porps.message changes

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -348,22 +386,28 @@ export default React.memo(function GamePage(): JSX.Element | null {
...
                 {is_installed && !showCloudSaveInfo && (
-                  <div
+                  <p
                     style={{
                       color: '#F45460'
                     }}
+                    className="iconWithText"
                   >
-                    <b>{t('info.syncsaves')}:</b>{' '}
+                    <CloudOff />
```
arielj, line 408: I'm not sure why, but this icon gets focused when moving with a controller and can be clicked too
image

src/frontend/components/UI/PopoverComponent/index.tsx
```diff
@@ -0,0 +1,47 @@
+import Popover from '@mui/material/Popover/Popover'
+import React, { useState } from 'react'
+import './index.scss'
+
+interface PopoverComponentProps {
+  item: React.ReactElement
+  children: React.ReactElement | React.ReactNode | React.ReactNode[]
+}
+
+const PopoverComponent: React.FC<PopoverComponentProps> = ({
```
arielj, line 10: When I open one of these popovers with a controller I'm trapped in it, I don't know how to close it. I can't move to an element outside of the popover and pressing `B` goes back in history instead of closing it. The only way to close it currently is clicking outside

src/frontend/screens/Game/GamePage/index.tsx
arielj: Same happens with download size and install size
image

src/frontend/components/UI/PopoverComponent/index.scss
```diff
@@ -0,0 +1,12 @@
...
+  color: var(--text-primary);
+  border: 1px solid var(--border-color);
+  border-radius: 4px;
+  box-shadow: 0 0 0 1px var(--border-color);
+  padding: 16px;
+  width: fit-content;
+  max-width: 500px;
+  overflow-y: auto;
+  overflow-x: hidden;
+}
```
arielj, line 12: This CSS is affecting the context menus too, it has extra padding now for example:
image

src/frontend/components/UI/PopoverComponent/index.tsx
flavioislima: This is a bit tricky to solve, will I need to add a method on the controller side for it?

flavioislima: Ok, found a way. Clicking again, or pressing the open button again should work to close it.

arielj:
I have to admit that I don't like the change of the installed icon acting as a filter and not as a sort now. I personally prefer to have see all with the installed first and the rest after.

I this this will also trigger a lot of reports because if you had that configured as `show installed first` in 2.6.2, after you update heroic suddenly only the installed games show up and it's impossible to know that the icon that did one thing before changed its behavior because it all looks the same. I can imagine users reporting this as a problem.

What is the reason behind that change? users complained about it or was it more aimed to prevent the issue we had with rendering libraries with a lot of games? I'd argue to revert that change

flavioislima:
@arielj we receive feedback of people that would like a filter to show only the installed game. We had that before and we have removed it.

About people reporting issues, just a reminder that on new releases heroic shows the changelog and this information will be there so unless people can't read, they will know that.

I think we could have both but not the way that the filters are organized because then it's another icon to add.

If we add the multiselect filter in the future like we have on the new design, then that would be solved.

The advantage imo is that you can have a cleaner view especially if you open heroic from the game mode on the deck with only the installed games.

arielj:
seeing how people ignore the big red message saying fortnite does not work on linux due to the anticheat I doubt people will read the change log properly though, some people will but I image it's going to be an easy miss for most users

I personally prefer the current behavior but it's just personal opinion, it's hard to know if people prefer the current behavior because people that prefer it didn't express that because it's what it is

I guess we can wait an see

I'll put the filter/sorting based on the new design higher in my TODO

flavioislima:
If you want to work on the filter, then I will revert the change here. 
so when you add the filter it will be more complete and avoid confusion like you said

arielj:
everything else :+1: to me, I've been really busy to work on any heroic stuff but my only concern was with the filtering, it's fine to merge this

arielj (approved):

PR #2449, opened 2023-02-14
Sanitize shortcut names

andyvorld:
Sanitize game title for shortcut creation.

Previously games like `Ruined King: A League of Legends Story` will silently fail and create an invalid shortcut named `Ruined King` in windows, due to not filtering reserved characters.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,6 +158,7 @@
     "react-markdown": "^8.0.3",
     "react-router-dom": "^6.3.0",
     "recharts": "^2.1.14",
+    "sanitize-filename": "^1.6.3",
     "shlex": "^2.1.2",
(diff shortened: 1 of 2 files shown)


Etaash-mathamsetty:
this should only be done on windows, there is no such restriction on linux and mac os

Etaash-mathamsetty:
seems like we need this sanitation for wineprefixes on linux/macOS rather than for shortcuts, so I guess everyone has a use for it

flavioislima (approved):
LGTM
I think it is good to have this for all OSes and not for Windows only since it covers some issues we could have on linux and mac as well.

flavioislima:
Thanks for the contribution @andyvorld

PR #2452, opened 2023-02-15
[Fix] Fix auto enabling of anti cheat runtimes

Etaash-mathamsetty:
title, previous code didn't work at all lol (I could have sworn it did)


flavioislima (approved):
LGTM.

i think it was not working before because the settings were being changed only on that scope and not the actual game setting. This new approach will work.

PR #2460, opened 2023-02-17
Render an empty div for game cards until they inter the viewport

arielj:
This is an alternative to https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2447

Instead of pagination, it renders all the library elements but for the ones that are outside of the viewport it renders an empty div until the element has to be displayed. Then it uses the IntersectionObserver to actually render the real content.

I'll add some comments in the PR.

There are still improvements that can be made (because of how react works, all hooks has to be defined before rendering the empty div, so there's extra unnecessary work there, also cards are re-rendered multiple times) but I'll leave those things for another PR.

You can see the empty divs using devtools and how they are updated when scrolling:
image

---

### DIFF
--- a/src/frontend/screens/Library/components/GameCard/index.css
+++ b/src/frontend/screens/Library/components/GameCard/index.css
@@ -10,6 +10,7 @@
   flex-direction: column;
   border-radius: var(--space-3xs);
   animation: fade-in 0.2s ease-in;
+  aspect-ratio: 173/275;
 }
 
 .gameCard:focus-within {
--- a/src/frontend/screens/Library/components/GameCard/index.css
+++ b/src/frontend/screens/Library/components/GameCard/index.css
@@ -13,6 +13,10 @@
   aspect-ratio: 173/275;
 }
 
+.gameCard.gamepad {
+  aspect-ratio: 3/4;
+}
+
 .gameCard:focus-within {
   outline: -webkit-focus-ring-color auto 1px;
 }
@@ -102,7 +106,7 @@
   padding: var(--space-xs-fixed);
 }
 
-.gameCard > .icons.gamepad {
+.gameCard.gamepad > .icons {
   display: none;
 }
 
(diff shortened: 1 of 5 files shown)


src/frontend/screens/Library/components/GameCard/index.css
```diff
@@ -10,6 +10,11 @@
   flex-direction: column;
   border-radius: var(--space-3xs);
   animation: fade-in 0.2s ease-in;
+  aspect-ratio: 173/275;
+}
+
+.gameCard.gamepad {
+  aspect-ratio: 3/4;
```
arielj, line 17: these are the current ratios, but I'm hardcoding them so the empty divs use the correct space

src/frontend/screens/Library/components/GameCard/index.tsx
```diff
@@ -63,6 +63,24 @@ const GameCard = ({
   isRecent = false,
   gameInfo: gameInfoFromProps
 }: Card) => {
+  const [visible, setVisible] = useState(false)
+
+  useEffect(() => {
```
arielj, line 68: start as invisible, make visible when the `visible-cards` event is fired and this card is in the list of visible cards

src/frontend/screens/Library/components/GameCard/index.tsx
```diff
@@ -344,7 +372,7 @@ const GameCard = ({
         />
       )}
       <ContextMenu items={items}>
-        <div className={wrapperClasses}>
+        <div className={wrapperClasses} data-app-name={appName}>
```
arielj, line 375: using this data attribute for the intersection observer

src/frontend/screens/Library/components/GamesList/index.tsx
```diff
@@ -26,62 +26,51 @@ const GamesList = ({
   onlyInstalled = false,
   isRecent = false
 }: Props): JSX.Element => {
-  const { gameUpdates, showNonAvailable } = useContext(ContextProvider)
+  const { gameUpdates } = useContext(ContextProvider)
   const { t } = useTranslation()
-  const [gameCards, setGameCards] = useState<JSX.Element[]>([])
```
arielj, line 31: this useState and the useEffect changing it was causing extra re-renders

I moved this back to the return jsx instead of using a useState+useEffect

flavioislima (approved):
Lgtm

PR #2461, opened 2023-02-17
[Fix] Kill leftover wine processes after stop button is pressed

Etaash-mathamsetty:
it may be dangerous, but we kill the game process anyways, so I don't think it matters too much if we kill the wineserver too, this would kill leftover wine processes (e.g game launchers) in the prefix when the kill button is pressed, which is crucial when side loading games. 

### DIFF
--- a/src/backend/gog/games.ts
+++ b/src/backend/gog/games.ts
@@ -891,7 +891,7 @@ class GOGGame extends Game {
   public async stop(): Promise<void> {
     const pattern = isLinux ? this.appName : 'gogdl'
     killPattern(pattern)
-    if (!this.isNative()) {
+    if (!this.isNative() && isLinux) {
       const gameSettings = await this.getSettings()
       await shutdownWine(gameSettings)
     }
--- a/src/backend/gog/games.ts
+++ b/src/backend/gog/games.ts
@@ -888,8 +888,16 @@ class GOGGame extends Game {
   // Could be removed if gogdl handles SIGKILL and SIGTERM for us
   // which is send via AbortController
   public async stop(): Promise<void> {
+    const gameSettings = await this.getSettings()
     const pattern = isLinux ? this.appName : 'gogdl'
     killPattern(pattern)
+    if (!this.isNative())
+      await runWineCommand({
+        gameSettings,
+        commandParts: ['wineboot', '-k'],
+        wait: true,
+        protonVerb: 'waitforexitandrun'
+      })
   }
 }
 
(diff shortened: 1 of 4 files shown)


arielj:
is it safe to do this? I mean, what if there's another wine process running outside of what heroic starts? like some other app that the user is using. will this kill those too?

Etaash-mathamsetty:
> is it safe to do this? I mean, what if there's another wine process running outside of what heroic starts? like some other app that the user is using. will this kill those too? or does it only kill the wine running on a specific prefix?

this only kills wine on the current prefix that the game uses

src/backend/gog/games.ts
```diff
@@ -888,8 +888,16 @@ class GOGGame extends Game {
   // Could be removed if gogdl handles SIGKILL and SIGTERM for us
   // which is send via AbortController
   public async stop(): Promise<void> {
+    const gameSettings = await this.getSettings()
     const pattern = isLinux ? this.appName : 'gogdl'
     killPattern(pattern)
+    if (!this.isNative())
+      await runWineCommand({
```
arielj, line 895: would be nice to extract this `await runWineCommand(...` block into a function like killPattern so we don't have to repeat the same code in 3 different places

something like:

```ts
public async stop(): Promise<void> {
    const pattern = isLinux ? this.appName : 'gogdl'
    killPattern(pattern)
    if (!this.isNative()) {
      const gameSettings = await this.getSettings()
      shutdownWine(gameSettings)
    }
}
```

and the same for all stores, and you can define the `shutdownWine` function where `killPattern` is defined in backend utils

arielj (approved):
I tested this with a Ubisoft game that always leaves some Ubisoft processes open after closing the game that I have to close manually with the terminal. Using the stop button now kills those processes too :+1:

arielj:
the only things I don't know is those `wait: true` and `protonVerb` arguments, maybe someone else knows better if they are correct

or can you explain what are they form?

imLinguin:
It's not needed for GOG games, since gogdl registers itself as a `child subreaper`. There is no way for any executable to "escape" from the process tree

Etaash-mathamsetty:
> It's not needed for GOG games, since gogdl registers itself as a `child subreaper`. There is no way for any executable to "escape" from the process tree

I'll remove it from the gog one then

Etaash-mathamsetty:
> the only things I don't know is those `wait: true` and `protonVerb` arguments, maybe someone else knows better if they are correct
> 
> or can you explain what are they form?

It's basically just saying to wait for the wine command to finish executing

imLinguin (approved):

PR #2464, opened 2023-02-18
[UX] Install UbisoftConnect when running UbisoftConnect games

arielj:
This PR adds a new feature to install UbisoftConnect automatically in a prefix when running games that require it.

It takes care of:
- download the ubisoft installer and cache it if needed
- install the launcher and arial fonts using winetricks
- only does that when the prefix is updated

I still have to test this on a Mac since I don't know if the `winetricks arial` command actually works there. If it doesn't then I'll have to find a way to install arial without using winetricks (I tried downloading the files from https://github.com/pushcx/corefonts and running them in the prefix, but I think I'm missing some steps)

Not the best video, but you can see at the end it launches the launcher and it shows the user what it's doing:

https://user-images.githubusercontent.com/188464/219869368-b45cba60-66eb-454c-afd5-449b8b8d8eb2.mp4

EDIT: this does NOT install the launcher on Windows, I have no way to test that so I didn't try to implement it. And it's also easy for Windows users to simply install UbisoftConnect manually anyway so I think it's not that critical

EDIT2: tested on mac, seems like wine crossover and wine staging don't have the proton updated that fixed ubisoft connect, so the launcher crashes, but we can do nothing about it

EDIT3: one thing I'm NOT able to test is legendary's activation command so I didn't include it here, since my accounts are already linked I don't know how to test that, but I think it will be easy to add in the future if needed inside the same setup function after the launcher is installed

---


public/locales/en/gamepage.json
```diff
@@ -190,6 +191,7 @@
         "syncingSaves": "Syncing Saves",
         "this-game-uses-third-party": "This game uses third party launcher and it is not supported yet",
         "totalDownloaded": "Total Downloaded",
+        "ubisoft": "Installing Ubisoft",
```
arielj, line 194: new status messages so we make what's happening more clear for the user

src/backend/games.ts
```diff
@@ -46,6 +47,7 @@ abstract class Game {
   abstract update(): Promise<{ status: 'done' | 'error' | 'abort' }>
   abstract isNative(): boolean
   abstract runWineCommand(args: WineCommandArgs): Promise<ExecResult>
+  abstract runner: Runner
```
arielj, line 50: I added this for games so we can do `game.runner` instead of having to pass a `runner` string around

src/backend/gog/games.ts
```diff
@@ -508,6 +511,12 @@ class GOGGame extends Game {
       `Launch Command: ${fullCommand}\n\nGame Log:\n`
     )
 
+    sendFrontendMessage('gameStatusUpdate', {
+      appName: this.appName,
+      runner: this.runner,
+      status: 'playing'
+    })
```
arielj, line 518: before this status, the status is now `launching`, so we change it to `playing`

src/backend/legendary/games.ts
```diff
@@ -868,6 +871,12 @@ class LegendaryGame extends Game {
       )
     }
 
+    sendFrontendMessage('gameStatusUpdate', {
+      appName: this.appName,
+      runner: this.runner,
+      status: 'playing'
+    })
```
arielj, line 878: before this status, the status is now `launching`, so we change it to `playing`

src/backend/tools.ts
```diff
@@ -331,10 +331,11 @@ export const Winetricks = {
       )
     }
   },
-  run: async (
+  runWithArgs: async (
```
arielj, line 334: this allows us to run winetricks without its UI

src/frontend/screens/Library/components/GameCard/index.tsx
```diff
@@ -191,12 +191,15 @@ const GameCard = ({
       )
     }
     if (isInstalled) {
+      const disabled =
+        isLaunching ||
+        ['syncing-saves', 'launching', 'ubisoft'].includes(status!)
```
arielj, line 196: disable stop button while launching and installing ubisoft stuff too

src/common/types.ts
```diff
@@ -165,6 +165,7 @@ export type Status =
   | 'notSupportedGame'
   | 'notInstalled'
   | 'installed'
+  | 'ubisoft'
```
flavioislima, line 168: maybe instead of `ubisoft` we could use something like: 'preparing', 'setup' or something like this because then we can use it for other things as well.

flavioislima (approved):
LGTM!
that is pretty useful since there is a lot of Ubi games on Epic now and I think it will work just fine.

Left just a few comments on what could be improved but overall looks good.

src/common/types.ts
arielj: I was actually thinking about having more statuses here, so we can be more verbose to the user (like "preparing prefix" for when wineboot is running), maybe we could have another more generic `preparing` that we can use for gog's setup?

I think having more states make it more clear for the user (and also for us to know when something is failing)

src/backend/legendary/setup.ts
```diff
@@ -0,0 +1,104 @@
...
+  try {
+    await downloadIfNotCached(cachedUbisoftInstallerPath, UBISOFT_INSTALLER_URL)
+
+    await game.runWineCommand({
+      commandParts: [cachedUbisoftInstallerPath, '/S']
+    })
+
+    return true
+  } catch {
+    logWarning(
```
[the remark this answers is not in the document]
arielj: just pushed the error message if the launcher was not installed successfully :+1:

flavioislima (approved):
Looks good, feel free to merge it after fixing the conflicts 👍🏽

PR #2473, opened 2023-02-19
Fix non-available status for games just installed

arielj:
This PR fixes a bug in a maybe uncommon use case: if you install a game and, without restarting heroic, remove its folder, after refreshing the library the game does NOT show up as not available.

The issue was that, when a game is installed, the status in the frontend's state is set to `done` and that was preventing the `notAvailable` status to be considered.

---


flavioislima (approved):
Lgtm

Issue #1485, opened 2022-06-16
Cannot load big GOG libraries (2400+ entries)


shakeyourbunny:
### Describe the bug

Heroic Launcher seems to have really troube loading bigger game libraries, like mine, which has on GOG 2400+ entries.

Maybe it still works and is doing the initial scan, but there is no indication what it is doing.

### Add logs

( don't know where the logs is )

### Steps to reproduce

- select GOG account
- enter credentials

- Heroic Launcher seems to hang until the task is killed.

### Expected behavior

Getting the games into the library.

### System Information

- OS: SteamOS 3.2, Suse Tumbleweed (current iteration of the week, usually there is a distribution update once a week)
- Heroic Version: current

### Additional information

Perhaps add a progress dialog, in what stages the scan is and how much it has progressed.

If you do a parse of the web site library, provide on what page the scan is and on what game; also on game x / num games, just like lgogdownloader.

This information could be provided in a non-modal window or something.

imLinguin:
Progress bar is something to consider then. You can find logs under Settings > Logs. 

2300+ games - that's really big library 😯

shakeyourbunny:
I tried the same procedure on my (beefier desktop computer) and did load after looking for 20 min on the spinning cursor (mind you my internet connection is 200 MBit downwards, so that should not a problem).

````
(12:58:22) INFO:    [Legendary]:       Running command : C:\Users\pulaski\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary --version
(12:58:24) INFO:    [Legendary]:       Legendary location: C:\Users\pulaski\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary
(12:58:24) INFO:    [Gog]:             GOGDL location: C:\Users\pulaski\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\gogdl
(12:58:24) INFO:    [Backend]:         

Heroic Version: 2.4.0-beta Caesar Clown
Legendary Version:  0.20.26 Dark Energy
OS: Microsoft Windows 10 Pro KERNEL: 10.0.19044 ARCH: x64
CPU: AMD Ryzen 5 3600 6-Core Processor @3.6 
RAM: Total: 31.93 GiB Available: 19.36 GiB
GRAPHICS: GPU0: AMD Radeon RX 6700 XT VRAM: 12272MB DRIVER:  


[...]

PR #2480, opened 2023-02-21
[GOG] Fetch images in the background to speed up initial library render

arielj:
This should improve the user experience for users loading the GOG library for the first time like https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1485

Instead of waiting for the images to be fetched before showing the library, this PR makes it so the GOG library is initially rendered with a fallback Heroic image and it triggers the images loading in the background.

Also, it splits the number of requests that are done in chunks of 20 requests to make them in parallel so for example a list of 100 games will be split into 5 chunks of 20 and the duration of the process is only the sum of the slowest request of each chunk instead of the sum of 100 requests.

After everything is processed, the information is stored and the frontend is notified to reload the library.

The tradeoff is that we'll see all GOG games with a fallback image for a while (I thought about notifying the frontend after every chunk so the images are updated progressively and not all at the end, but it was getting kinda complex and I'd prefer to do that in another PR as an enhancement after this one is merged)

Easiest way to test/qa this is to delete the gog_store/library.json file and then open Heroic.

---

### DIFF
--- a/src/backend/gog/library.ts
+++ b/src/backend/gog/library.ts
@@ -1,3 +1,4 @@
+import { sendFrontendMessage } from './../main_window'
 import axios, { AxiosError, AxiosResponse } from 'axios'
 import { GOGUser } from './user'
 import {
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
Looks good. Tested here and works fine!

PR #2481, opened 2023-02-22
Remove fsrhack toggle

Etaash-mathamsetty:
it doesn't do anything anymore :/


redromnon:
Why would you discard it?
I can confirm it works fine with Heroic 2.6.2 - I specifically use it for Rocket League.

imLinguin:
That's right, that wasn't ever removed from any GE release. Of course except the one after upstream wine rebase

Etaash-mathamsetty:
> Why would you discard it? I can confirm it works fine with Heroic 2.6.2 - I specifically use it for Rocket League.

it's enabled by default, so the toggle no longer does anything since it never sets the env variable to 0
simple state table:
toggle off: enabled by default
toggle on: enabled by env variable

it's always enabled regardless, so let's throw it out

redromnon:
I'm not sure; @flavioislima Any thoughts on this?

flavioislima:
> I'm not sure; @flavioislima Any thoughts on this?

Hmm, we have discussed that on our discord but I'm not 100% sure.
We know that some wine versions is enabling this by default now, but there is no way to control the fsr sharpness without using env variables though. I think we should keep it for now and get back to it in the future.

flavioislima:
We can reopen that in the future if we change our minds about this. Closing for now.

Nocccer:
What you think about checking if the enviroment variable is set and if not we set it to 1. With this way people can set it to anything via the enviroment variable setting. The wine underneath will override it anyway if it needs something different. We can remove the setting with this solution added to `setupEnvVars` after `LD_PRELOAD`:

```typescript
  if (!process.env.WINE_FULLSCREEN_FSR && !ret.WINE_FULLSCREEN_FSR) {
    ret.WINE_FULLSCREEN_FSR = ''1"
  }
  
  if (!process.env.WINE_FULLSCREEN_FSR_STRENGTH && !ret.WINE_FULLSCREEN_FSR_STRENGTH) {
    ret.WINE_FULLSCREEN_FSR_STRENGTH = ''2"
  }

```

PR #2493, opened 2023-02-25
[FIX] don't require wineboot for wine installations.

Etaash-mathamsetty:
This lets me use my custom wine builds again! (after symlinking wineserver)

### DIFF
--- a/src/backend/config.ts
+++ b/src/backend/config.ts
@@ -393,23 +393,16 @@ abstract class GlobalConfig {
   }
 
   /**
-   * Checks if a Wine version has Wineboot/Wineserver executables and returns the path to those if they're present
+   * Checks if a Wine version has the Wineserver executable and returns the path to those if they're present
    * @param wineBin The unquoted path to the Wine binary ('wine')
-   * @returns The quoted paths to wineboot and wineserver, if present
+   * @returns The quoted paths to wineserver, if present
    */
-  public getWineExecs(wineBin: string): {
-    wineboot: string
-    wineserver: string
-  } {
+  public getWineExecs(wineBin: string): { wineserver: string } {
     const wineDir = dirname(wineBin)
-    const ret = { wineserver: '', wineboot: '' }
+    const ret = { wineserver: '' }
     const potWineserverPath = join(wineDir, 'wineserver')
     if (existsSync(potWineserverPath)) {
       ret.wineserver = potWineserverPath
-    }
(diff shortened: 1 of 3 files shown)


flavioislima (commented):
I think this will fail if wineboot does not exists:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/9212526c7a6f3be470d26751c0d567dffc1765bc/src/backend/launcher.ts#L475

Etaash-mathamsetty:
> I think this will fail if wineboot does not exists:
> 
> https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/9212526c7a6f3be470d26751c0d567dffc1765bc/src/backend/launcher.ts#L475

if the wine installation is valid running wine wineboot --init will work fine

src/backend/config.ts
```diff
@@ -393,24 +393,17 @@ abstract class GlobalConfig {
   }
 
   /**
-   * Checks if a Wine version has Wineboot/Wineserver executables and returns the path to those if they're present
+   * Checks if a Wine version has the Wineserver executable and returns the path to those if they're present
```
CommandMC, line 396:
```suggestion
   * Checks if a Wine version has the Wineserver executable and returns the path to it if it's present
```

src/backend/config.ts
```diff
@@ -393,24 +393,17 @@ abstract class GlobalConfig {
...
    * @param wineBin The unquoted path to the Wine binary ('wine')
-   * @returns The quoted paths to wineboot and wineserver, if present
+   * @returns The quoted paths to wineserver, if present
```
CommandMC, line 398:
```suggestion
   * @returns The quoted path to wineserver, if present
```

src/backend/config.ts
```diff
@@ -393,24 +393,17 @@ abstract class GlobalConfig {
...
    */
-  public getWineExecs(wineBin: string): {
-    wineboot: string
-    wineserver: string
-  } {
+  public getWineExecs(wineBin: string): { wineserver: string } {
```
CommandMC, line 400: Somewhat unnecessary now to return an object with just one key
Might as well just return the path directly

Etaash-mathamsetty: this required more code changes, and I just found it simpler to leave it the way it is, but I can it change if needed

flavioislima (approved):
LGTM

flavioislima (approved):
LGTM

PR #2495, opened 2023-02-25
[Fix]: Improve pc/applegamingwiki accuracy and add gamesdb info

Nocccer:
- Use search api for pcgamingwiki and applegamingwiki to find more games where title did not exactly match
- Get wine rating for applegamingwiki
- Get steamid from gamesdb
- Fix a bug where wiki info were stored into config store instead his own 

---


flavioislima (approved):
LGTM 🥇 
TEsted here and the bug was fixed. Now everything is being stored into wikigaming.json.

flavioislima:
Needed to revert this one, it broke the other stores. So it is not writing the configStore, for instance, this makes the epic games show as logged out every time Heroic is restarted since GlobalState reads the username from the store.

PR #2496, opened 2023-02-26
[Fix]: Inaccuracy of disk and down speed

Nocccer:
This fixes the problem with the inaccuracy of the download and disk speed.
The problem was that either legendary or child_process does not write/read stdout at the same time. Means we get actually two calls to `onInstallOrUpdateOutput` with one time the `eta` and `bytes` and the other time with the `disk` and `down` speed. We did not handled it and just returned earlier if `eta` or `bytes` is empty. The fix just stores the last values of the progress and send the progress if all values are fetched.

Applied the same strategy to gog aswell.
Here a video which shows that the change should fixed it:

https://user-images.githubusercontent.com/61798668/221420102-3c0265b2-1bf5-45dd-8c39-a4bb58e9765d.mp4

---

### DIFF
--- a/src/backend/gog/games.ts
+++ b/src/backend/gog/games.ts
@@ -219,7 +219,7 @@ class GOGGame extends Game {
     const bytesMatch = data.match(/Downloaded: (\S+) MiB/m)
     const bytes =
       bytesMatch && bytesMatch?.length >= 2
-        ? bytesMatch[1]
+        ? `${bytesMatch[1]}MiB`
         : this.lastProgress.bytes
 
     // parse log for download speed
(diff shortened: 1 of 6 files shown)


flavioislima (approved):
Lgtm

Nocccer:
> Lgtm

I will try another solution because the current one calls the frontend two times with the same values. You can see that in the video disk and down speed are printed twice in the graph.

Nocccer:
@flavioislima i fixed the double send and also increased the points we show in the graph from 10 to 20

flavioislima:
> @flavioislima i fixed the double send and also increased the points we show in the graph from 10 to 20

Cool, can you just change MiB to MB? I know I know but most people have no idea what MiB is.

PR #2501, opened 2023-02-27
[Fix] EGL sync on macOS/Linux

CommandMC:
This PR started out by wanting to fix a logic error with our EGL sync implementation on macOS/Linux. Sparing the details (feel free to look at the old code yourself), it was never possible to actually initiate a Sync. This is now solved by just running a sync/un-sync whenever the path is changed.

Along the way, I found that we have no way to easily express a "path selector" settings component, so I've added it with the `PathSelectionBox` and went ahead and used that with the new EGL sync & everywhere else applicable. As a small bonus, this also makes all of those settings visually consistent (some were using MUI icons while others used Fontawesome ones).

I currently have no (convenient) way of testing this on a clean install or on OSs other than Linux. Could someone else help out there?

---


src/frontend/components/UI/TextInputField/index.tsx
```diff
@@ -4,7 +4,7 @@ import ContextProvider from 'frontend/state/ContextProvider'
 import './index.css'
 
 interface TextInputFieldProps {
-  htmlId: string
+  htmlId?: string
```
arielj, line 7: we should always have an html id so we can pair the input with the corresponding label properly, do you need this change for anything specific?

src/frontend/screens/Settings/sections/SyncSaves/legendary.tsx
```diff
@@ -107,37 +104,13 @@ export default function LegendarySyncSaves({
         />
       ) : (
         <>
-          <TextInputWithIconField
-            htmlId="inputSavePath"
```
arielj, line 111: this htmlId prop is missing in the change

src/frontend/components/UI/PathSelectionBox/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+        filters: pathDialogFilters,
+        defaultPath: pathDialogDefaultPath
+      })
+      .then((path) => {
+        if (path) onPathChange(path)
+      })
+  }
+
+  return (
+    <>
```
arielj, line 69: we don't need this extra fragment

src/frontend/components/UI/ThemeSelector/index.tsx
```diff
@@ -68,37 +59,6 @@ export const ThemeSelector = () => {
...
   }, [])
 
-  // on input change only update the state
-  const onThemesPathChanged = (e: ChangeEvent<HTMLInputElement>) => {
-    if (themesPath !== e.target.value) {
-      setThemesPath(e.target.value)
-    }
-  }
-
-  // only save config on input blur when editing manually
```
arielj, line 78: we lost this feature in the refactor (without this change the config gets saved for every key stroke instead of when blurring the field

src/frontend/screens/Library/components/InstallModal/WineSelector/index.tsx
```diff
@@ -90,22 +88,14 @@ export default function WineSelector({
...
-            <TextInputWithIconField
+            <PathSelectionBox
+              type="directory"
+              onPathChange={setWinePrefix}
+              path={winePrefix}
+              pathDialogTitle={t('box.wineprefix', 'Select WinePrefix Folder')}
               label={t('install.wineprefix', 'WinePrefix')}
               htmlId="setinstallpath"
-              placeholder={winePrefix}
-              value={winePrefix.replaceAll("'", '')}
```
arielj, line 97: is this call to `replaceAll` important? it's lost in the refactor

src/frontend/screens/Settings/components/AltGOGdlBin.tsx
```diff
@@ -27,53 +24,24 @@ const AltGOGdlBin = () => {
...
       )}
       htmlId="setting-alt-gogdl"
+      type="file"
+      onPathChange={setAltGogdlBin}
+      path={altGogdlBin}
       placeholder={t(
         'placeholder.alt-gogdl-bin',
         'Using built-in GOGDL binary...'
       )}
-      value={altGogdlBin.replaceAll("'", '')}
```
arielj, line 54: same here with this `replaceAll`

src/frontend/screens/Settings/components/DefaultInstallPath.tsx
```diff
@@ -12,31 +10,16 @@ const DefaultInstallPath = () => {
...
-    <TextInputWithIconField
+    <PathSelectionBox
+      type="directory"
+      onPathChange={setDefaultInstallPath}
+      path={defaultInstallPath}
+      pathDialogTitle={t('box.default-install-path')}
+      pathDialogDefaultPath={defaultInstallPath}
       label={t('setting.default-install-path')}
       htmlId="default_install_path"
-      value={defaultInstallPath?.replaceAll("'", '')}
```
arielj, line 30: looks like we do this `replaceAll` in most of them?

src/frontend/components/UI/TextInputField/index.tsx
CommandMC: I saw that some uses of `TextInputField` (like the one in Legendary's save sync page) *only* specified a htmlId & not a label. That seemed redundant to me (without a label, there's nothing to link)

I should've used union types here (so that if you specify either htmlId or label, you also have to specify the other

src/frontend/screens/Library/components/InstallModal/WineSelector/index.tsx
CommandMC: I don't see why we would have this, as it doesn't actually change the underlying value (that is already stripped of some problematic characters when it's read from the settings) and besides, anything using it should be able to handle special characters anyways (at least everything that uses `spawn` instead of `exec`)

src/frontend/components/UI/TextInputField/index.tsx
CommandMC: Testing this further, it seems that React does not like me using Unions here

Returning this to how it was before then

src/frontend/screens/Library/components/InstallModal/WineSelector/index.tsx
arielj: yeah, I don't really know why this was added in the first place (maybe somebody knows of a specific game that needs it but I don't)

arielj (approved):
I'm not sure exactly how to test this since I've never used this feature, but code looks good to me :+1:

src/frontend/screens/Settings/components/EgsSettings.tsx
```diff
@@ -1,156 +1,83 @@
...
+        htmlId="syncToggle"
+        value={egsPath}
+        handleChange={handleSync}
+        title={t('setting.egs-sync')}
+        disabled={isSyncing}
+      />
+    )
   }
 
+  // On macOS/Linux, we have to give Legendary the path to the Wineprefix
```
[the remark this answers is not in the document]
CommandMC: It will not, Legendary only supports EGL-Sync with the Windows version of the EGL

src/frontend/components/UI/ToggleSwitch/index.tsx
```diff
@@ -6,7 +6,7 @@ import './index.css'
 interface Props {
   htmlId: string
   handleChange: ChangeEventHandler<HTMLInputElement>
-  value: boolean
+  value: unknown
```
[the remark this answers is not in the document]
CommandMC: Only improves convenience a bit. For EGL-Sync, we store a string value that dictates if the switch is toggled, so instead of passing `value={Boolean(eglPath)}`, you can just pass `value={eglPath}` directly (and it takes care of converting it to a Boolean)

We probably also have other uses of this component where the same could be applied, I haven't really checked (and even then, that might be better in a separate PR)

PR #2513, opened 2023-03-03
[Enhancement]: Provide Wine-ge/Proton-ge latest version via wine manager

Nocccer:
This PR adds a Wine-GE and PRoton-Ge latest version which is updateable via the wine-manager.
The PR also cleans up some test and fixes some small issues.

- Fixed issue where a error was thrown if one fetch call of one github repository failed. Now it shows a error, so other fetches can still succeed.

---

### DIFF
--- a/src/backend/wine/manager/downloader/utilities.ts
+++ b/src/backend/wine/manager/downloader/utilities.ts
@@ -52,6 +52,21 @@ async function fetchReleases({
 
           releases.push(release_data)
         }
+
+        // add latest to list
+        releases.unshift({
+          version: releases[0].version.replace(
+            /[0-9]+[-|.]+[0-9]+/m,
+            '-latest'
+          ),
+          type: releases[0].type,
+          date: releases[0].date,
+          download: releases[0].download,
+          downsize: releases[0].downsize,
+          disksize: releases[0].disksize,
+          checksum: releases[0].checksum
+        })
+
         resolve(releases)
       })
       .catch((error) => {
--- a/src/backend/wiki_game_info/gamesdb/__tests__/utils.test.ts
+++ b/src/backend/wiki_game_info/gamesdb/__tests__/utils.test.ts
@@ -28,12 +28,18 @@ describe('getInfoFromGamesDB', () => {
   })
 
   test('catches axios throws', async () => {
-    jest.spyOn(axios, 'get').mockRejectedValueOnce(new Error('Failed'))
+    jest.spyOn(axios, 'get').mockRejectedValueOnce({
+      response: {
+        data: {
+          error_description: 'Failed'
+        }
+      }
+    })
 
     const result = await getInfoFromGamesDB('Jotun', 'Grouse', 'legendary')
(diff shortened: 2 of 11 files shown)


Etaash-mathamsetty:
I made a feature request for this in protonup-qt, and wanted to add it here too, but I guess you beat me to it lol thanks

Etaash-mathamsetty:
seems to be working

src/backend/wine/manager/downloader/utilities.ts
```diff
@@ -52,6 +52,21 @@ async function fetchReleases({
 
           releases.push(release_data)
         }
+
+        // add latest to list
+        releases.unshift({
+          version: releases[0].version.replace(
+            /[0-9]+[-|.]+[0-9]+/m,
+            '-latest'
```
CommandMC, line 60: I wonder if writing `${type}-latest` (so for example "Wine-GE-latest") would be easier to read here

Nocccer: Depends if we are fine that some information is missing:
- With regex:
image

- Your suggestion:
image

I don't mind which to take.

CommandMC: Well, the fact that Wine-GE is based on Proton's Wine source is not something the general person has to know, and everyone who does likely already knows this
I just see the potential confusion with GE-Proton there

CommandMC:
Something more general: Currently, when a user is using this Wine version, there's no indication in the logs *which* exact version they're using (what I mean is: You just see "Wine-GE-Proton-Latest", while what "Latest" means depends on when the user last installed/updated it). Maybe something could be added in that regard (one approach would be adding an optional `version` property to `WineInstallation` & setting that to the tag name of the release)

Nocccer:
> Something more general: Currently, when a user is using this Wine version, there's no indication in the logs _which_ exact version they're using (what I mean is: You just see "Wine-GE-Proton-Latest", while what "Latest" means depends on when the user last installed/updated it). Maybe something could be added in that regard (one approach would be adding an optional `version` property to `WineInstallation` & setting that to the tag name of the release)

The property `version` is actually the name. I got the version tag into the name without big changes. Problem is that if there is a update, the version is already the new version even before you updaed it. Other ways seems to not be solved easy. We could introduce bugs in the logic of wine manager. Maybe we can refactor that in another pr ?

image

Nocccer:
@CommandMC we need to fix the the information about the version in another pr. Everything i try breaks stuff. We should refactor the wine-manager anyway to a tool manager and downlaod dxvk and vkd3d other them aswell.

Etaash-mathamsetty:
Seems to work well for me, I think once this is merged, we can make a PR that automatically installs wine if not found

flavioislima:
> Seems to work well for me, I think once this is merged, we can make a PR that automatically installs wine if wine is not found

Yes, good point. This is planned and we should prioritise that.

flavioislima (approved):
Lgtm

Issue #2488, opened 2023-02-23
Parse gog .info files to list addtional executables.  


Bitwolfies:
### Problem description

GOG games tend to come with several exe's used to configure the game, or outright other parts of a game, and manually dragging them to the run in prefix button is slow and cumbersome.

### Feature description

GOG Galaxy has a solution to this problem, with the additional executables section.  
![Screenshot_20230223_020959](https://user-images.githubusercontent.com/65789901/220877741-fd982b42-2f9a-4298-936e-22b27b7cd96a.png)

Thankfully, these options are all listed with the games .info file, no external db required.
![Screenshot_20230223_020903](https://user-images.githubusercontent.com/65789901/220877546-fc653855-4214-4e20-b0b5-3349b0d28a3d.png)
It would be great if Heroic could list these options within its gui.

imLinguin:
That's something I was planning to do. Probably in form of popup window, play task select on game page, or option in settings to also allow for this feature when running from shortcut or protocol

PR #2515, opened 2023-03-03
[GOG] Support launch options

imLinguin:
This PR enables launch options for GOG games like in Legendary, except here we are loading the file every time when we open game page

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1474,8 +1474,8 @@ ipcMain.handle('syncGOGSaves', async (event, gogSaves, appName, arg) =>
   getGame(appName, 'gog').syncSaves(arg, '', gogSaves)
 )
 
-ipcMain.handle('readGOGGameInfoFile', async (event, appName: string) =>
-  GOGLibrary.get().readInfoFile(appName)
+ipcMain.handle('getGOGLaunchOptions', async (event, appName: string) =>
+  GOGLibrary.get().getLaunchOptions(appName)
 )
 
 ipcMain.handle(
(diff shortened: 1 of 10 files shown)


imLinguin:
Closes #2488

src/backend/main.ts
```diff
@@ -1474,6 +1474,10 @@ ipcMain.handle('syncGOGSaves', async (event, gogSaves, appName, arg) =>
   getGame(appName, 'gog').syncSaves(arg, '', gogSaves)
 )
 
+ipcMain.handle('readGOGGameInfoFile', async (event, appName: string) =>
+  GOGLibrary.get().readInfoFile(appName)
+)
```
CommandMC, line 1479: This is purely my opinion, but the Frontend just being able to read the whole .info file sounds a little... extreme? I believe we've also had some talks about doing more logic in the Backend in the past

Factoring this in a way that the Frontend just asks for the launch options and the Backend then does that work (reading out the file, filtering out tasks) might be *cleaner*

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -158,6 +161,37 @@ export default React.memo(function GamePage(): JSX.Element | null {
...
+                  .then((infoFile) => {
+                    const newLaunchOptions: LaunchOption[] = []
+                    infoFile?.playTasks.forEach((task, index) => {
+                      if (
+                        task.type === 'FileTask' &&
+                        !task?.isHidden &&
+                        task.category !== 'document'
+                      ) {
+                        newLaunchOptions.push({
+                          name: task?.name || infoFile.name,
```
CommandMC, line 178: Is it possible for a task's `name` property to be undefined?

If this is correct, could you also adapt the `TaskBase` interface?

src/backend/main.ts
imLinguin: I was thinking about it too, and wanted to hear your opinion on that. I'll move it to backend then

CommandMC (approved):

PR #2518, opened 2023-03-04
[UI] title on ETA

TornadoRadon:
Just added a title on the estimated time on the downloading page. Without a title, it's hard to understand what this time means.

---


Nocccer:
Can you add a image please how it looks.

arielj (approved):
looks good to me, my only concern is that it's not accessible with a controller, but we have that problem in many places where we use the title attribute and we have to figure out a standard, I think that can be done in the future

PR #2525, opened 2023-03-08
[Tech] Update just about all of our dependencies

CommandMC:
As the title says: Everything that's possible to update is now updated

Packages left out:
- `axios` - Produces some weird `Adapter 'http' is not available in the build` errors when updated
- `vite-plugin-electron` - Honestly at a loss to how, but updating this completely breaks the build. It seems like it's trying to run files intended for the Frontend in the Backend process?

`unimported` was introduced as an alternative to the somewhat-deprecated `ts-prune`. Thanks for the help @Nocccer

---


package.json
```diff
@@ -122,57 +122,57 @@
...
+    "systeminformation": "^5.17.12",
     "ts-prune": "^0.10.3",
-    "tslib": "^2.4.0"
+    "tslib": "^2.5.0"
   },
   "scripts": {
     "start": "vite",
     "codecheck": "tsc --noEmit",
-    "find-deadcode": "ts-prune --error",
+    "find-deadcode": "ts-prune | sed -e '/satisfies/,+1d'",
```
Nocccer, line 175: Can't trigger the error in vscode. So i would asume it has something todo with webstorm and need to be configured there. So this should be reverted in my eyes.

.github/workflows/lint.yml
```diff
@@ -20,4 +20,4 @@ jobs:
       - name: Prettier code.
         run: yarn prettier
       - name: Find dead code
-        run: yarn find-deadcode
+        run: test "$(yarn find-deadcode | tee "$(tty)" | wc -l)" -lt 4
```
Nocccer, line 23: Why we need this ?

Nocccer (changes requested):
LGTM but left some open questions.

package.json
CommandMC: TS-Prune does not support the `satisfies` operator. This is not a configuration issue, it's simply because the thing is virtually dead (but no other alternatives exist, as far as I can tell)
You won't see any false-positifes right now since we don't use that operator anywhere right now. Try adding something like
```ts
export const myObject = {} satisfies Record<string, unknown>
```
anywhere, you'll see that TS-Prune will complain about 3 things in that line:
- The actual variable (`myObject`), which is valid
- `satisfies`
- The token directly after `satisfies` (in this case `Record`)

.github/workflows/lint.yml
CommandMC: As I've explained in the commit message: By piping the output of ts-prune into `sed`, we no longer get the behavior that any error makes the workflow fail. This command essentially restores that behavior

package.json
Nocccer: Ok we don't use the `satisfies` keyword yet.
I discovered this [pacakge](https://www.npmjs.com/package/unimported) from the ts-prune repo where someone had the same problem. We might just switch to this and remove ts-prune ?

CommandMC: I'm a little confused here, I've already tried out that package, discovered that it doesn't seem to work right, and documented those findings on Discord ([here](https://discord.com/channels/812703221789097985/812703424995000361/1083052712641429565)). You even replied to that message

I don't think this issue is going to be in ts-prune forever. If it's really as simple as updating one of their dependencies, *someone* there will get around to it at some point, and then we can revert this mess

Nocccer: Oh ok forgot that. 
Testing here and it works for me. We just need to setup the config file correctly.
My `.unimportedrc.json` so far that gives reasonable results:

```json
{
  "rootDir": "./src",
  "preset": "node",
  "entry": [
    "src/**/*"
  ],
  "extensions": [
    ".js",
    ".jsx",
    ".ts",
    ".tsx",
    ".d.ts"
  ],
  "ignorePatterns": [
    "**/node_modules/**"
  ],
  "ignoreUnimported": [],
  "ignoreUnused": [],
  "ignoreUnresolved": []
}
```

Result:
```
  | 66 unresolved imports

   1 │ common/types
   2 │ common/types/gog
   3 │ common/types/legendary
   4 │ common/types/electron_store
   5 │ backend/utils/aborthandler/aborthandler
   6 │ backend/constants
   7 │ backend/main_window
   8 │ backend/wiki_game_info/wiki_game_info
   9 │ common/types/epic-graphql
  10 │ backend/logger/logger
  11 │ backend/tools
  12 │ backend/dialog/dialog
  13 │ backend/launcher
  14 │ backend/wiki_game_info/gamesdb/utils
  15 │ ../../tsconfig
  16 │ backend/__tests__/skip
  17 │ backend/updater
  18 │ backend/electron_store
  19 │ backend/config
  20 │ frontend/assets/epic-logo.svg
  21 │ frontend/assets/gog-logo.svg
  22 │ frontend/assets/heroic-icon.svg
  23 │ frontend/components/UI
  24 │ frontend/components/UI/StoreFilter
  25 │ frontend/state/ContextProvider
  26 │ frontend/helpers/electronStores
  27 │ frontend/helpers
  28 │ frontend/types
  29 │ frontend/components/UI/WebviewControls
  30 │ frontend/components/UI/Dialog
  31 │ frontend/assets/settings_icon_alt.svg
  32 │ frontend/assets/heroic_card.jpg
  33 │ frontend/components/UI/PopoverComponent
  34 │ frontend/screens/Library/components
  35 │ frontend/helpers/library
  36 │ frontend/hooks/hasProgress
  37 │ frontend/components/UI/ErrorComponent
  38 │ frontend/components/UI/Anticheat
  39 │ frontend/components/UI/StoreLogos
  40 │ frontend/hooks/hasStatus
  41 │ frontend/components/UI/WikiGameInfo/components/HowLongToBeat
  42 │ frontend/components/UI/WikiGameInfo/components/GameScore
  43 │ frontend/components/UI/UninstallModal
  44 │ frontend/assets/down-icon.svg
  45 │ frontend/assets/play-icon.svg
  46 │ frontend/assets/stop-icon.svg
  47 │ frontend/assets/stop-icon-alt.svg
  48 │ frontend/components/UI/ActionIcons
  49 │ frontend/screens/WineManager/components/WineItem
  50 │ frontend/components/UI/ExternalLinkDialog
  51 │ frontend/components/UI/LanguageSelector
  52 │ frontend/components/UI/ThemeSelector
  53 │ frontend/hooks/useSetting
  54 │ frontend/helpers/gamepad
  55 │ frontend/components/UI/TwoColTableInput
  56 │ frontend/screens/Settings/SettingsContext
  57 │ frontend/components/UI/ProgressDialog
  58 │ frontend/components/UI/Collapsible/Collapsible
  59 │ frontend/components/UI/TextWithProgress
  60 │ frontend/hooks/useSettingsContext
  61 │ frontend/components/UI/ToggleSwitch
  62 │ frontend/assets/info_icon.svg
  63 │ frontend/assets/denied_icon.svg
  64 │ frontend/assets/rounded_checkmark_icon.svg
  65 │ frontend/components/UI/FormControl
  66 │ frontend/components/UI/UpdateComponent

 │ 5 unused dependencies

   1 │ @fontsource/cabin
   2 │ @fontsource/rubik
   3 │ source-map-support
   4 │ ts-prune
   5 │ tslib
```

Nocccer: Ok i think i am close to solve this. Can i push it ?

.unimportedrc.json
```diff
@@ -0,0 +1,15 @@
+{
+  "rootDir": "./src",
+  "preset": "node",
+  "entry": ["src/**/*"],
```
CommandMC, line 4: Adding every single source file as an entrypoint here effectively disables the tool altogether

package.json
CommandMC: Unimported does not do quite the same thing as ts-prune, as it only checks for unimported *files*. I could still add new unused types to `common/types.ts` (for example) without it erroring out

Nocccer: Ok yes i see. Threfore unimported tells you which third party modules are not used.The problem i see is that ts-prune is kind of deprecated and a new maintained tool would be nice

.unimportedrc.json
Nocccer: Ok will resolve this.

package.json
CommandMC: Sure, but we're actively sacrificing functionality when switching to Unimported

Don't get me wrong, I'd absolutely prefer an alternative, but isn't the main point of having these tools that you get per-export analysis, instead of just per-file? It's easy to leave some variable in a file exported, but a whole *file* being unused sounds unlikely

Nocccer (approved):

PR #2528, opened 2023-03-09
[Shortcuts] Use icon from api for gog shortcuts

imLinguin:
This is simple addition that will load icon from API instead of using squashed cover:

Before:  
obraz

After:  
obraz

It may be necessary to manually delete icon from icons directory so new ones can be pulled.

**Note** squashed covers will still be used as fallback if for some reason API call fails

### DIFF
--- a/src/backend/shortcuts/utils.ts
+++ b/src/backend/shortcuts/utils.ts
@@ -58,18 +58,17 @@ async function getIcon(appName: string, gameInfo: GameInfo | SideloadGame) {
     mkdirSync(heroicIconFolder)
   }
 
-  let image = gameInfo.art_square.replaceAll(' ', '%20').replace('{ext}', 'png')
+  // By default use vertical image - art_square in jpg format
+  let image = gameInfo.art_square.replaceAll(' ', '%20').replace('{ext}', 'jpg')
+  let icon = `${heroicIconFolder}/${appName}.jpg`
 
   if (gameInfo.runner === 'gog') {
     const productApiData = await GOGLibrary.getProductApi(appName)
-    if (productApiData) {
-      if (productApiData.data.images?.icon) {
-        image = 'https:' + productApiData.data.images?.icon
-      }
+    if (productApiData && productApiData.data.images?.icon) {
+      image = 'https:' + productApiData.data.images?.icon
+      icon = `${heroicIconFolder}/${appName}.png` // Allow transparency
     }
   }
-
-  const icon = `${heroicIconFolder}/${appName}.png` // Use png's since we need support for transparency
 
   if (!checkImageExistsAlready(icon)) {
     downloadImage(image, icon)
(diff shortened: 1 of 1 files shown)


redromnon:
The existing icons use jpg format. There might be a case where the user has forgotten they already have a shortcut and try to add it anyway from the game page' game actions menu.

This will result in the new/modified shortcut that uses the wrong image's path with a png extension since a jpg image exists and it fails to show the icon.

redromnon:
> It may be necessary to manually delete icon from icons directory so new ones can be pulled.
I think you're right.

redromnon:
> It may be necessary to manually delete icon from icons directory so new ones can be pulled.

I think you're right.

Nocccer (changes requested):
I would find it better if the art_square holds the correct image. We check runners here again. We should avoid that as much as possible. The gameInfo should contain all correct informations no matter which runner.

imLinguin:
The issue with making it generic is that Epic doesn't provide such images. Also `art_square` is a left over name from Epic API, which is a vertical image as far as I know.

Nocccer:
> The issue with making it generic is that Epic doesn't provide such images. Also `art_square` is a left over name from Epic API, which is a vertical image as far as I know.

Epic provides everything like gog does. Else the add to steam feature had never worked for both. I would say we could refactor the gameinfo images and give them proper naming. Something like a new type:

```typescript
interface Images {
  banner?: string
  icon?: string
  portrait?: string
}

interface GameInfo {
  // ...
  images: Images
  // ...
}
```

imLinguin:
What I mean is epic doesn't have icons like GOG. 

we are limited to those fields in it's case

```json
"art_cover": "https://cdn1.epicgames.com/item/df2da503f2074f078f8090da3c27ec47/EGS_BioShockInfiniteCompleteEdition_MassMediaGames_S1_2560x1440-bf29199cfe7a76f62965f582571024f6_2560x1440-bf29199cfe7a76f62965f582571024f6",
"art_square": "https://cdn1.epicgames.com/item/df2da503f2074f078f8090da3c27ec47/EGS_BioShockInfiniteCompleteEdition_MassMediaGames_S2_1200x1600-bcf5e88c8afb572ec00d3ed231ec26e5",
```

Nocccer:
> ```json
> https://cdn1.epicgames.com/item/df2da503f2074f078f8090da3c27ec47/EGS_BioShockInfiniteCompleteEdition_MassMediaGames_S2_1200x1600-bcf5e88c8afb572ec00d3ed231ec26e5
> ```

I understand. It still does not make my idea impossible. For Epic icon and portrait is the same url for gog this differ. The good thing about is that we can use just this information without cehcking which runner we are pulling images. Gameinfo should be generic and the runner libs should care about correct image urls.

Nocccer (approved):
Ok lets approve for now and i will start the refactor of it today.

CommandMC:
I mean, nothing prevents you from pushing your changes onto this branch as well

imLinguin:
Exactly

imLinguin:
I'll try to add the structure like you suggested @Nocccer

Nocccer:
> I'll try to add the structure like you suggested @Nocccer 

Sure. Else i will do but didn't find time yet.

arielj:
@Nocccer should we merge this one and do the refactor later or do you think it's better to wait? I'm confused with the latest messages of the conversation

Nocccer:
Merge and refactor later

Issue #2526, opened 2023-03-09
Not all games are shown


a-cristi:
### Describe the bug

Using version 2.6.2 on SteamDeck I can not see all of my Epic Games games in the list of games, nor do they appear as suggestions in the search bar.

I'll try to give some context, in case it helps. First I noticed this when trying to install Borderlands 3 and searching for `Borderlands` got me no results. I have Borderlands 2, 3, and The Pre-Sequel, as well as DLC for them in my library so something should have shown up. I fixed this by logging out of my Epic account and logging back in.

I managed to install Borderlands 3 like this, but after it was installed, the game page reported that the game is not installed, and once I went back to the library screen the game was no longer there. Again, I logged out and logged back in and the game was there, but DLC was not listed when searching for it. Since then I have not managed to workaround this issue.

Periodically, I have to log out and log back in in order to see all the games. Sometimes, right after I open the launcher I can see all the games, but after a few seconds a lot of them disappear. Sometimes there is enough time until the games disappear to actually start the game, but after that the launcher still does not see the game as being in my library.

This is not a BL3 issue. After a fresh log in the launcher reports that I have 417 games, when the issue appears it reports only 229. I also have my GOG account linked, but at the moment I do not know if this issue also affects those games.

### Add logs

Logs collected after a startup in which not all games were present. The logs also include the moment I did a log out and log back in for my Epic Games account:

```text
(12:23:13) INFO:    [Connection]:       Connectivity: check-online
(12:23:13) INFO:    [Connection]:       Pinging external endpoints
(12:23:13) INFO:    [Legendary]:        Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary --version
(12:23:13) INFO:    [Legendary]:        Legendary location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(12:23:13) INFO:    [Gog]:              GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(12:23:13) INFO:    [Gog]:              GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(12:23:13) WARNING: [Backend]:          Failed to register protocol with OS.
(12:23:13) INFO:    [Connection]:       Connectivity: online
(12:23:13) INFO:    [Gog]:              Getting data about the user
[...]

PR #2533, opened 2023-03-11
[Epic] Prevent issue if platform is undefined in game metadata

arielj:
This should fix https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2526.

Seems like there must be some game that has `undefined` as the platform info instead of an array, this should prevent that from making it crash.

---


arielj:
@CommandMC looks like the problem are some DLCs, I'm not sure what's that game yet

CommandMC:
https://store.epicgames.com/en-US/p/south-park-the-fractured-but-whole--from-dusk-till-casa-bonita lists the DLC as supporting Windows.
Whether that's derived from the main game item or just a default value, I cannot say.

Defaulting to the platforms the main game supports sounds fine

src/backend/legendary/library.ts
```diff
@@ -1,3 +1,4 @@
+import { LegendaryInstallPlatform } from './../../common/types/legendary'
```
CommandMC, line 1:
```suggestion
import { LegendaryInstallPlatform } from 'common/types/legendary'
```
You might want to change `JavaScript -> Preferences: Import Module Specifier` to "shortest" (or "non-relative") in your VSCode settings

arielj: weird, I already had that in `shortest`, I'll check again next time it auto imports something

CommandMC: I also remember having issues with this back when I was using VSCode. We probably have to explicitly tell it about those import paths *somewhere*

CommandMC (approved):

PR #2534, opened 2023-03-11
[Fix]: Get rid of wrong default wine prefix "~/.wine"

Nocccer:
This PR replaces all `~/.wine` with a new default path to `~/Heroic/Prefixes/default`. This should solve a problem people have currently on steam deck, where `~/.wine` is not writeable.

---


src/backend/config.ts
```diff
@@ -621,8 +621,8 @@ class GlobalConfigV0 extends GlobalConfig {
         name: userName
       },
       wineCrossoverBottle: 'Heroic',
-      winePrefix: isWindows ? defaultWine : `${userHome}/.wine`,
-      wineVersion: isWindows ? {} : defaultWine
+      winePrefix: isWindows ? {} : join(heroicDefaultWinePrefix, 'default'),
```
imLinguin, line 624: Is this empty object for windows intended? It could lead to unexpected behavior, since we expect winePrefix as a string

imLinguin (approved):
Looks good, left one small comment, more of a question than an actual issue

CommandMC:
With `persist` now being added to the Flatpak build, is `~/.wine` really still read-only/discarded?
Not disputing the validity of this PR (that path definitely is a more sensible location), just wondering about that

imLinguin:
It should be mapped to sandboxed location

src/backend/config.ts
Nocccer: You are right. Setting a empty string makes more sense

src/frontend/screens/Settings/components/AutoDXVK.tsx
```diff
@@ -13,8 +12,8 @@ const AutoDXVK = () => {
     'autoInstallDxvk',
     false
   )
-  const home = configStore.get('userHome', '')
-  const [winePrefix] = useSetting('winePrefix', `${home}/.wine`)
+  const [defaultWinePrefix] = useSetting('defaultWinePrefix', '')
+  const [winePrefix] = useSetting('winePrefix', defaultWinePrefix)
```
arielj, line 16: I have a question about this, I see `winePrefix` is having `defaultWinePrefix` as the fallback/default value

this default value before this change was `~/.wine` (a prefix), after this change it would be the value of the `defaultWinePrefix` config, but I see that is set here https://github.dev/Heroic-Games-Launcher/HeroicGamesLauncher/blob/690f522155940d2fc6dff3339a53f38002ee37d9/src/backend/config.ts#L609-L610 with the value of the `heroicDefaultWinePrefix` variable, which is `~/Games/Heroic/Prefixes` (the folder where we add prefixes, not a prefix), shouldn't we be using `~/Games/Heroic/Prefixes/default` here too instead?

src/backend/game_config.ts
```diff
@@ -270,7 +271,8 @@ class GameConfigV0 extends GameConfig {
       if (isMac) {
         defaultSettings.wineCrossoverBottle = wineCrossoverBottle
       } else if (isLinux) {
-        defaultSettings.winePrefix = winePrefix || `${userHome}/.wine`
+        defaultSettings.winePrefix =
+          winePrefix || join(heroicDefaultWinePrefix, 'default')
```
arielj, line 275: maybe we can add `join(heroicDefaultWinePrefix, 'default')` as a constant in constants.ts?

src/frontend/screens/Settings/components/AutoDXVK.tsx
arielj: I think we probably want to use `/Games/Heroic/Prefixes/default` as the defaultWinePrefix fallback instead of `''`, a `''` as a value means that we could be setting an empty string as the winePrefix path in some edge case

Nocccer: You are right. Anyway if the defaultWinePrefix is empty, someone really messed up because it is set by a constant, so i didn't added a fallback for the prefix directory.

src/frontend/screens/Settings/components/WinePrefix.tsx
```diff
@@ -17,9 +16,11 @@ const WinePrefix = () => {
 
   const isWin = platform === 'win32'
 
-  const home = configStore.get('userHome', '')
-  const [winePrefix, setWinePrefix] = useSetting('winePrefix', `${home}/.wine`)
   const [defaultWinePrefix] = useSetting('defaultWinePrefix', '')
+  const [winePrefix, setWinePrefix] = useSetting(
+    'winePrefix',
+    defaultWinePrefix
```
arielj, line 22: I think this one is missing the `+ /'default'`?

arielj (approved):
Looks good to me!

PR #2537, opened 2023-03-12
[Tech] Make absolute import paths in tests working again

Nocccer:
---


tsconfig.json
```diff
@@ -27,5 +27,5 @@
     "typeRoots": ["./node-modules/@types", "src/common/typedefs"]
   },
   "include": ["src"],
-  "exclude": ["vite.config.ts", "**/__tests__/**", "**/__mocks__/**"]
+  "exclude": ["vite.config.ts", "**/__mocks__/**"]
```
CommandMC, line 30: You could just get rid of the whole `exclude` key here, along with deleting `tsconfig.eslint.json` (and pointing ESLint to this config file instead)

Nocccer:
@CommandMC can you help me with that lint problem i can't solve it. I added `vite.config.ts` in `include` or `exclude`. Does not matter the lint problem still exist.

CommandMC:
That error *does* go away if you remove the `include` key completely, but it turns out that `tsc` does not like trying to *build* that config file
Seems like you'd need two tsconfig files after all, my bad

Nocccer:
> That error _does_ go away if you remove the `include` key completely, but it turns out that `tsc` does not like trying to _build_ that config file Seems like you'd need two tsconfig files after all, my bad

I removed the include section, still the error persist. I ended up adding vite.config.ts to eslintignore file. I think we don't need to lint that file.

PR #2542, opened 2023-03-16
[Performance] Do not store all the output of commands in memory

arielj:
This PR fixes an issue where we were storing all the output of gogdl/legendary/wine/etc in memory for the duration of those processes. That means that for any process with extensive output we were storing maybe megabytes and megabytes of strings for no real use.

This PR changes that to only store the last 50 message (50 is a kinda arbitrary number thinking that we would care about the last messages of an output if needed (like to detect some error message).

This should have a beneficial impact in memory usage, specially when running games that output a lot of data when running through Wine-GE or for long installations (I've had installations extending for hours and we were storing a lot of text every single second).

In some cases the number of lines was so high that converting them to a string when we called `.join()` would actually result in an error exceeding the valid string length!!

This is a bit too specific to test, but you can change the places I changed in this PR to add a `console.log(stdout.join())` and same for stderr at the end of any of those methods to see how much we were storing in memory before and after.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -17,7 +17,8 @@ import {
   searchForExecutableOnPath,
   quoteIfNecessary,
   errorHandler,
-  removeQuoteIfNecessary
+  removeQuoteIfNecessary,
+  memoryLog
 } from './utils'
 import {
   logDebug,
@@ -602,8 +603,8 @@ async function runWineCommand({
       )
     }
 
-    const stdout: string[] = []
-    const stderr: string[] = []
+    const stdout = memoryLog()
+    const stderr = memoryLog()
 
     child.stdout.on('data', (data: string) => {
       if (options?.logFile) {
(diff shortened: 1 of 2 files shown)


imLinguin (approved):
Looks good

arielj:
> LGTM. Do you think it is possible to have a benchmark of before/after these changes, maybe installing a big game like GTAV or Cyberpunk2077. Would be a good thing to show on the release notes, people love performance improvement fire

I'm doing some checks and the output during installation is not really that much, for a 60GB game the output is like... 3Mb of text (which is a lot of text, but not a problem), it's not a big change, but some games that constantly print things to the stdout (I tested far cry 3 with wine-ge for example that prints a fixme message all over the place) this can mean even gigabytes of memory (memory just keeps going up and up over time since the game keeps outputting text for some missing wine stuff)

I think it's hard to put it on concrete numbers since it can go from saving a few kilobytes to whatever a game would output for an extended period of time

Also when using WINEDEBUG you can get A LOT of output depending on the config, I just opened absolutedrift with some debug env variable and I could easily get 900Mb of memory usage in like a minute for the process that starts the game (not the game exe but the node process)

I don't know if it would fall in the category of memory leak, it kinda does?

arielj:
I can't really get a clear benchmark, I hope some of the comments there serve the purpose for anyone interested on how it affects memory usage and could be linked in the release notes

PR #2549, opened 2023-03-17
[Tech] Refactor Protocol to make it more extensible

flavioislima:
This PR changes the protocol and how we create the shortcuts to make them more extensible in the future.
Previously, we did not pass the runner on the protocol and Heroic had to guess it by looping through all runners to find the correct one. This was not ideal.

With this PR, we will include the runner in the protocol for future shortcuts, making them more efficient.
We also added a method to handle the older shortcuts so they can still work.

Additionally, we implemented the Install modal that shows up when trying to launch a game that is not installed instead of installing it with the default options.

---


src/backend/protocol.ts
```diff
@@ -2,91 +2,177 @@ import { dialog } from 'electron'
...
+ * Parses a url into a tuple of [Command, Runner?, string?]
+ * @param url The url to parse
+ * @returns A tuple of [Command, Runner?, string?]
+ * @example
+ * parseUrl('heroic://ping')
+ * // => ['ping', undefined, undefined]
+ * parseUrl('heroic://launch/gog/123')
+ * // => ['launch', 'gog', '123']
+ * parseUrl('heroic://launch/legendary/123')
+ * // => ['launch', 'legendary', '123']
```
arielj, line 71: maybe add an example of the old launch command? I understand it would be `heroic://launch/123` => `['launch', undefined, '123']`

src/backend/protocol.ts
```diff
@@ -2,91 +2,177 @@ import { dialog } from 'electron'
...
+ * @example
+ * findGame('gog', '123')
+ * // => { app_name: '123', title: '123', is_installed: true, runner: 'gog' ...}
+ * findGame('legendary', '123')
+ * // => { app_name: '123', title: '123', is_installed: true, runner: 'legendary' ...}
+ **/
+async function findGame(
+  runner: Runner | undefined,
+  arg: string | undefined = ''
+): Promise<GameInfo | SideloadGame | null> {
```
arielj, line 164: maybe we can return `null` at the beginning if arg is `''`, we don't really have to loop through the runners for a game that will never exist

src/backend/protocol.ts
```diff
@@ -2,91 +2,177 @@ import { dialog } from 'electron'
...
+  // If the runner is specified, only search for that runner
+  const runnersToSearch = runner ? [runner, 'sideload'] : RUNNERS
```
arielj, line 166: why are we always appending the `sideload` runner? the comment says we'll only check the specified one so it's not clear why we try sideload of the game can't be found in the specified runner (I imagine that check with the sideload runner will always fail anyway in that case)

src/backend/protocol.ts
flavioislima: done

src/backend/protocol.ts
flavioislima: done

src/backend/protocol.ts
flavioislima: ok, refactor that part now to be more clear.

Nocccer (approved):

PR #2555, opened 2023-03-18
[Tech] Downgrade back to Vite 3

CommandMC:
This resolves issues related to `check-disk-space`. We'll ultimately have to find *some* solution to this, as staying on Vite 3 doesn't sound like a good idea.

---


arielj (approved):
thanks! I can confirm this fixes the issue

I agree that we have to find a solution for the future

PR #2557, opened 2023-03-18
[Fix] Return an empty string when save path calculation fails

CommandMC:
This PR includes two minor improvements to our save sync code, hopefully reducing the amount of questions we get about it
- As mentioned in the title, instead of returning the save path with variables (`{AppData}/SomeApp`, for example) when the actual path can't be computed, an empty string is returned instead
- The `Promise.race` in the IPC function was removed. Apart from the reason outlined in the commit message, it is also the only way *I* could test the first change (on a somewhat slow laptop, everything takes a little longer)

To test this change:
1. Open your `~/.config/legendary/installed.json` and locate a game with a `save_path` (if you don't have one, "Loop Hero" is quick to install). Delete that property, then mark the file as read-only.
2. Start Heroic, head to the same game; try to compute the save path. You'll get an error message (about Legendary not being able to write to `installed.json`), and you'll see that the save input box is empty.
3. Make sure to reset the permissions on your `installed.json` afterwards!

There is probably also an easier way to test this, although I couldn't find anything consistent

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -73,7 +73,6 @@ import {
   getLatestReleases,
   getShellPath,
   getCurrentChangelog,
-  wait,
   checkWineBeforeLaunch
 } from './utils'
 import {
@@ -1515,14 +1514,8 @@ ipcMain.handle(
 
 ipcMain.handle(
   'getDefaultSavePath',
-  async (event, appName, runner, alreadyDefinedGogSaves) => {
-    return Promise.race([
-      getDefaultSavePath(appName, runner, alreadyDefinedGogSaves),
-      wait(15000).then(() => {
-        return runner === 'gog' ? [] : ''
-      })
-    ])
-  }
+  async (event, appName, runner, alreadyDefinedGogSaves) =>
+    getDefaultSavePath(appName, runner, alreadyDefinedGogSaves)
 )
 
 // Simulate keyboard and mouse actions as if the real input device is used
--- a/src/backend/save_sync.ts
+++ b/src/backend/save_sync.ts
@@ -42,7 +42,11 @@ async function getDefaultSavePath(
 
 async function getDefaultLegendarySavePath(appName: string): Promise<string> {
   const game = getGame(appName, 'legendary')
-  const { save_path } = game.getGameInfo()
+  const { save_folder, save_path } = game.getGameInfo()
+  logInfo(
+    ['Computing save path for save folder', save_folder],
+    LogPrefix.Legendary
+  )
   if (save_path) {
     logDebug(
       ['Legendary has a save path stored, discarding it:', save_path],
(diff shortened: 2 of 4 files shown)


arielj:
> * As mentioned in the title, instead of returning the save path with variables (`{AppData}/SomeApp`, for example) when the actual path can't be computed, an empty string is returned instead

does this mean that the input field is empty in those cases? because I think it's usually useful to see the path with the placeholders for support

I've never really used the save syncs but if that's the case maybe we could still show the path with the placeholders but a warning saying that the path is not usable because it was not resolved properly and disabling the button to sync?

I prefer telling the user `hey, this is wrong, look for help` than not showing something because then they don't know what actually happened

CommandMC:
Yup, the input field will just be empty then, which will also make the Frontend display "Save folder not found, please select it manually"

I was originally thinking that we could just grab the path from PCGW, but that path is incorrect for a few games. I'll make it log the path with variables, that should still make it accessible *enough* for support

arielj:
Do you know what would happen if the save path calculation gets stuck? I know we had that issue that it was never ending and that's why the race/timeout was added. I wonder what's the implication of removing it (I was not too involved in the change that added this timeout so I don't know how to test what happens when it's not there).

CommandMC:
In the event that this does happen (which I cannot reproduce; it *does* take a long time sometimes but it never gets truly stuck), you can cancel the process in the Frontend

CommandMC:
Apart from the fact that I cannot consistently reproduce it truly getting stuck / returning nothing (at least not without altering configuration files as described in the PR description):
A fixed timeout will always have issues. In my case, a full prefix initialization with Legendary being ran afterwards takes >20 seconds. A modern PC might be able to do it in ~10, while even slower ones (factoring in HDD speeds and slow processors, potentially also with other processes needing resources) might take >30.
If we'd set the timeout to, for example, 40 seconds, that would *work* for everyone, but would hardly help on modern systems. If we'd instead set it to 15, that might help on modern systems, but would completely break it on anything slower

My point, as noted above, is that leaving this up to the user ("Oh, this is taking rather long, let me abort that and ask about it") will be more robust than us assuming some arbitrary timeframe

arielj:
> My point, as noted above, is that leaving this up to the user ("Oh, this is taking rather long, let me abort that and ask about it") will be more robust than us assuming some arbitrary timeframe

One question about this, let's say the calculation is taking too long or got stuck in an endless calculation, what happens if the user closes the settings? will that kill the process that is doing that calculation or will that stay running in the background?

because it that stays in the background and stopping it depends on the user clicking `cancel` it can be a problem (would be a wine process running in the background preventing other wine processes from running)

maybe it already cancels the calculation when closing the dialog but just asking to be sure, because I imagine that re-opening the settings dialog will trigger a new calculation if the old one was stuck, it won't re-use the current one for the user to cancel it

CommandMC:
Yup, that has always been a problem (and this isn't the only IPC function affected). If I have the time and motivation for it over the weekend, I might look into how far we can modify Electron's IPC system further to make every function cancelable (imagine a handler function for `ipcMain.handle` not just getting an implicit event, but also an AbortController that the frontend can then use to, well, abort on-demand)

arielj (approved):
this looks fine to me, I have that concern with the cancellation of the calculation but I see that was a possible issue before too

though I'll wait for @flavioislima to comment again since he had comments too

Issue #2568, opened 2023-03-25
Scroll bars scroll down continuously when controller support enabled


Richard-Payne:
### Describe the bug

Whenever a scrollbar is visible in the app, it scroll down with no input from me.  Inputting scroll up commands with my mouse scroll wheel can force it to scroll up but as soon as I stop inputs, it starts going down again.  

I found, on a hunch, that turning off the controller support fixed the issue. 

I have a Steam Controller and a Logitech G29 Wheel plugged in.   Both work perfectly in games.  Neither seem to operate the Heroic UI as far as I can tell.  Something in the controller support code is reading a continuous down input from somewhere, but I'm not sure where.

### Add logs

```text
(18:59:06) INFO:    [Connection]:       Connectivity: check-online
(18:59:06) INFO:    [Connection]:       Pinging external endpoints
(18:59:06) INFO:    [Legendary]:        Running command: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary --version
(18:59:06) INFO:    [Legendary]:        Legendary location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(18:59:06) INFO:    [Gog]:              GOGDL location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(18:59:06) INFO:    [Gog]:              GOGDL location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(18:59:06) WARNING: [Backend]:          Protocol already registered.
(18:59:06) INFO:    [Connection]:       Connectivity: online
(18:59:06) INFO:    [Backend]:          User Not Found, removing it from Store
(18:59:06) INFO:    [Gog]:              Running command: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --version
(18:59:06) INFO:    [Backend]:          AreWeAntiCheatYet data downloaded
(18:59:06) INFO:    [Backend]:          Loading Screen Ready
(18:59:06) INFO:    [Backend]:          Frontend Ready
(18:59:06) INFO:    [Backend]:          Checking for new Heroic Updates
(18:59:06) INFO:    [Backend]:          Checking for current version changelog
(18:59:06) INFO:    [Backend]:          

Heroic Version: 2.6.2 Trafalgar Law
Legendary Version:  0.20.32 Dark Energy (hotfix #6)
GOGdl Version: 0.6

[...]

PR #2570, opened 2023-03-26
[Fix] Ignore Logitech's G29

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2568.

Logitech's G29 is a wheel and it's causing continuous scroll because the browser maps some axes at 1.0 and we don't have a layout for a wheel to map the buttons.

For now the simplest solution is to ignore the wheel as a device to control the UI. If we want to actually support this we'd have to come up with a specific layout for it.

---


Etaash-mathamsetty (approved):
LGTM

PR #2571, opened 2023-03-26
[Fix regression] Set correct horizontal image url for GOG games

arielj:
I noticed a regression caused by my refactor to load GOG images async. The `art_cover` was being set incorrectly, missing the protocol and the extension based on the previous code https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2480/files#diff-b81a7a1b73f7704fde02ec68ba60851c261a0139471d91fce4893769f9844330L637

This only happens in the main branch.

If you have this issue already in your configuration files you might need to delete the gog_store/library.json and store_cache/gog_library.json files so it gets updated (I guess clearing heroic cache should fix that too)

---

### DIFF
--- a/src/backend/gog/library.ts
+++ b/src/backend/gog/library.ts
@@ -678,7 +678,7 @@ export class GOGLibrary {
       store_url: `https://gog.com${info.url}`,
       developer: '',
       app_name: String(info.id),
-      art_cover: info.image,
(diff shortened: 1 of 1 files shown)


Etaash-mathamsetty (approved):
LGTM

flavioislima (approved):

PR #2576, opened 2023-03-27
[FIX] Fix sideload statuses

Etaash-mathamsetty:
Whenever you started playing a sideloaded game, the status would be stuck on launching, this patch fixes that so that it now says playing!

### DIFF
--- a/src/backend/sideload/games.ts
+++ b/src/backend/sideload/games.ts
@@ -179,6 +179,12 @@ export async function launchApp(appName: string): Promise<boolean> {
         }
       }
 
+      sendFrontendMessage('gameStatusUpdate', {
+        appName: appName,
+        runner: 'sideload',
+        status: 'playing'
+      })
+
       const commandParts = shlex.split(launcherArgs ?? '')
       await callRunner(
         commandParts,
@@ -214,6 +220,12 @@ export async function launchApp(appName: string): Promise<boolean> {
       const globalSettings = await GameConfig.get('global').getSettings()
       gameSettings.wineVersion = globalSettings.wineVersion
     }
+
+    sendFrontendMessage('gameStatusUpdate', {
+      appName: appName,
+      runner: 'sideload',
+      status: 'playing'
+    })
 
     await runWineCommand({
       commandParts: [executable, launcherArgs ?? ''],
--- a/src/backend/sideload/games.ts
+++ b/src/backend/sideload/games.ts
@@ -179,12 +179,6 @@ export async function launchApp(appName: string): Promise<boolean> {
         }
       }
 
-      sendFrontendMessage('gameStatusUpdate', {
-        appName: appName,
-        runner: 'sideload',
-        status: 'playing'
-      })
-
       const commandParts = shlex.split(launcherArgs ?? '')
       await callRunner(
         commandParts,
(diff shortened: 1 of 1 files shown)


flavioislima:
Have you debugged why this line is not working?
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/690f522155940d2fc6dff3339a53f38002ee37d9/src/backend/main.ts#L1021

Etaash-mathamsetty:
> Have you debugged why this line is not working?
> 
> https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/690f522155940d2fc6dff3339a53f38002ee37d9/src/backend/main.ts#L1021
> 
> Because we have this status handler in one place for all runners so should work just fine. I also cannot replicate this issue. Can you give more details on how to replicate it?

that is only for launching, not playing

flavioislima:
> > Have you debugged why this line is not working?
> > 
> > https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/690f522155940d2fc6dff3339a53f38002ee37d9/src/backend/main.ts#L1021
> > 
> > Because we have this status handler in one place for all runners so should work just fine. I also cannot replicate this issue. Can you give more details on how to replicate it?
> 
> that is only for launching, not playing
> absolutely none of my sideloaded games show playing ever, only launching, which means that this code does work

I don't understand what you mean. Launching is playing. Are you maybe using some of those pretty specific wine versions you build?

Etaash-mathamsetty:
> I understand I introduced this problem when I added the Ubisoft status, the status label stays `Launching` and never changes to `Playing` for sideloaded games, it doesn't affect the playing it's just visual if I understand it right

correct

Etaash-mathamsetty:
> > > Have you debugged why this line is not working?
> > > https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/690f522155940d2fc6dff3339a53f38002ee37d9/src/backend/main.ts#L1021
> > > 
> > > Because we have this status handler in one place for all runners so should work just fine. I also cannot replicate this issue. Can you give more details on how to replicate it?
> > 
> > 
> > that is only for launching, not playing
> > absolutely none of my sideloaded games show playing ever, only launching, which means that this code does work
> 
> I don't understand what you mean. Launching is playing. Are you maybe using some of those pretty specific wine versions you build?

no it's stuck on the gray launching button and never switches to playing

flavioislima:
> > I understand I introduced this problem when I added the Ubisoft status, the status label stays `Launching` and never changes to `Playing` for sideloaded games, it doesn't affect the playing it's just visual if I understand it right
> 
> correct, I have no idea when the issue was introduced, but it might have been this because I don't think it's an issue with 2.6.2

Ah ok, that makes sense then. Now we have launching and playing.

src/backend/sideload/games.ts
```diff
@@ -197,6 +197,12 @@ export async function launchApp(appName: string): Promise<boolean> {
         }
       )
 
+      sendFrontendMessage('gameStatusUpdate', {
```
arielj, line 200: one thing I'm confused here, shouldn't this change right BEFORE the callRunner call instead of AFTER?

because if `callRunner` is awaiting until you exit the game then the status will be `Launching` all the time until you exit the game, then it would change to `playing`

Etaash-mathamsetty: I'll revert the commit that I added that changed the order

arielj (approved):
looks good!

PR #2578, opened 2023-03-28
[Tech] Refactor Game Managers

BrettCleary:
This PR refactors the legendary, gog, and sideload store managers into a consistent module level interface.

This reduces complexity and decouples the store managers from code that consumes it through abstraction and allows for further extensibility of game managers to other stores in the future. 

This also removes the overly complex object oriented singleton approach which only serves to remove the need to pass appName to each function. A module level functional approach is simpler, easier to test, and accomplishes the same objectives of code reuse and consistent interface.

Sideload Game info has been added into GameInfo which makes sense because the frontend shouldn't be concerned with implementation level backend details like what store/runner the game is using

Fixes a bug with the default install path setting

Fixes a bug where successive downloads would remain queued and not install

## Tested on Windows
(Launch Proton tested on SteamDeck)
### Epic Games

- [x] Login (alternative)
- [x] Logout 
- [x] Install
- [x] Import Games
- [x] Launch Native
- [x] Launch Proton
- [x] Move Game
- [x] Change Install Path
- [x] Uninstall

### GOG

- [x] Login
- [x] Logout 
- [x] Install
- [x] Import
- [x] Launch Native
- [x] Launch Proton
- [x] Move Game
- [x] Change Install Path
- [x] Uninstall

### Sideload

- [x] Add Game
- [x] Launch Game
- [x] Remove Game

## Further Work
Refactor `gog/user.ts` and `legendary/user.ts` to use the same module level functional type
- I would prefer to do this on a separate branch

Fix electron module mocking for tests
- Electron should be automocked with only some of the methods overridden instead of exposing a new interface that only mocks some of the electron methods
- This might be best handled on a separate branch as well

---


flavioislima (approved):
Thats a lot of stuff but in general looks good.
Tested basically everything on a new install and current install and everything seems to work just fine.

I think this approach is fine and finally, we are moving out of OOP and all boilerplate it was bringing to the code.

Thanks for that

flavioislima:
I am merging this one to avoid more conflicts in the future.

PR #2580, opened 2023-03-28
[Testing] Wrap all Legendary commands inside a Mutex to execute them sequentially

CommandMC:
This adds a `Mutex` around all Legendary commands, meaning they'll get executed one-by-one. **This will have issues** (for example, downloading a game will lock the mutex, which will lead to the user being unable to then launch other games in the meantime). It *should* solve the "constantly getting logged out" issue though

---

### DIFF
--- a/src/backend/legendary/library.ts
+++ b/src/backend/legendary/library.ts
@@ -42,6 +42,7 @@ import { installStore, libraryStore } from './electronStores'
 import { callRunner } from '../launcher'
 import { dirname, join } from 'path'
 import { isOnline } from '../online_monitor'
+import { Mutex } from 'async-mutex'
 
 /**
  * Legendary LegendaryLibrary.
@@ -631,19 +632,23 @@ export class LegendaryLibrary {
   public hasGame = (appName: string) => this.allGames.has(appName)
 }
 
+const legendaryMutex = new Mutex()
+
 export async function runLegendaryCommand(
   commandParts: string[],
   abortController: AbortController,
   options?: CallRunnerOptions
 ): Promise<ExecResult> {
   const { dir, bin } = getLegendaryBin()
-  return callRunner(
-    commandParts,
-    { name: 'legendary', logPrefix: LogPrefix.Legendary, bin, dir },
-    abortController,
-    {
-      ...options,
-      verboseLogFile: legendaryLogFile
-    }
+  return legendaryMutex.runExclusive(async () =>
+    callRunner(
+      commandParts,
+      { name: 'legendary', logPrefix: LogPrefix.Legendary, bin, dir },
+      abortController,
+      {
+        ...options,
+        verboseLogFile: legendaryLogFile
+      }
+    )
   )
 }
--- a/package.json
+++ b/package.json
@@ -153,6 +153,7 @@
     "@mui/material": "^5.11.12",
     "@node-steam/vdf": "^2.2.0",
     "@shockpkg/icon-encoder": "^2.1.3",
+    "async-mutex": "^0.4.0",
     "axios": "^0.26.1",
     "check-disk-space": "^3.3.1",
     "classnames": "^2.3.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2292,6 +2292,13 @@ async-exit-hook@^2.0.1:
   resolved "https://registry.yarnpkg.com/async-exit-hook/-/async-exit-hook-2.0.1.tgz#8bd8b024b0ec9b1c01cccb9af9db29bd717dfaf3"
   integrity sha512-NW2cX8m1Q7KPA7a5M2ULQeZ2wR5qI5PAbw5L0UOMxdioVk9PMZ0h1TmyZEkPYrCvYjDlFICusOu1dlEKAAeXBw==
 
+async-mutex@^0.4.0:
+  version "0.4.0"
+  resolved "https://registry.yarnpkg.com/async-mutex/-/async-mutex-0.4.0.tgz#ae8048cd4d04ace94347507504b3cf15e631c25f"
+  integrity sha512-eJFZ1YhRR8UN8eBLoNzcDPcy/jqjsg6I1AP+KvWQX80BqOSW1oJPJXDylPUEeMr2ZQvHgnQ//Lp6f3RQ1zI7HA==
+  dependencies:
+    tslib "^2.4.0"
+
 async@^3.2.3:
   version "3.2.4"
   resolved "https://registry.yarnpkg.com/async/-/async-3.2.4.tgz#2d22e00f8cddeb5fde5dd33522b56d1cf569a81c"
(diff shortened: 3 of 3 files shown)


Etaash-mathamsetty:
I feel like this solution is a little too extreme, there might be a better way, I don't think every legendary command will cause a logout and so maybe we should only lock this mutex for certain commands

Bananaman:
@Etaash-mathamsetty No, every command needs a lock.

But... the lock is only critical in the first few seconds of the command, when the token is being refreshed. If multiple commands try to refresh the exact same expired token at the same time, that's when the 1st command refreshes the expired token, followed by the 2nd or later commands immediately failing due to the expired token being invalid (already refreshed and deleted from Epic's servers), thus causing the "you are logged out" error.

So the locking should definitely be for every call to Legendary, BUT, with a caveat:

We can release the lock after something like 10 seconds if the command isn't done yet. But I am not sure how to implement that with this mutex/locking library.

Bananaman:
> I feel like, that problem could be solved by implementing locks in legendary. One for auth files and other for library (another issue with concurrent downloads, unrelated to Heroic)

Damn why didn't I think about that. Yes, Legendary itself should implement this, locking the `user.json` auth file while tokens are being refreshed, and having all other Legendary processes wait for the update before they read the new one's contents and do their own work. Totally, that is the correct solution.

Etaash-mathamsetty:
> @Etaash-mathamsetty No, every command needs a lock.
> 
> But... the lock is only critical in the first few seconds of the command, when the token is being refreshed. If multiple commands try to refresh the exact same expired token at the same time, that's when the 1st command refreshes the expired token, followed by the 2nd or later commands immediately failing due to attempting to refresh the exact same expired token that's now fully invalid (already refreshed and fully deleted from Epic's servers by the 1st command), thus causing the "you are logged out" error.
> 
> So the locking should definitely be for every call to Legendary, BUT, with a caveat:
> 
> We can release the lock after something like 10 seconds if the command isn't done yet. But I am not sure how to implement that with this mutex/locking library.

seems like it should be done on the legendary side then?

Bananaman:
Yeah this is definitely a Legendary issue instead. Legendary needs to wait for lock access before it proceeds with its commands.

It should use a good locking library which uses OS-level locking features for robustness and to avoid race conditions. And if Legendary starts and detects that `user.json` is locked, it should wait for access, then read the latest contents (new token) from there, so that the new token is immediately used by all waiting Legendary instances.

But as mentioned here, there's currently plans for a heavy-handed global lock in Legendary which will break a lot of features: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2580#issuecomment-1487313218

flavioislima:
I think we could have our own fork of Legendary with the changes because there are things that are important to Heroic but do not makes sense to Legendary. 

Heroic uses Legendary but Legendary is its own thing. And Rodney already added several features to improve our integration here.

Maybe @derrod can give his two cents on this.

src/backend/legendary/library.ts
```diff
@@ -631,19 +632,23 @@ export class LegendaryLibrary {
...
   const { dir, bin } = getLegendaryBin()
-  return callRunner(
-    commandParts,
-    { name: 'legendary', logPrefix: LogPrefix.Legendary, bin, dir },
-    abortController,
-    {
-      ...options,
-      verboseLogFile: legendaryLogFile
-    }
+  return legendaryMutex.runExclusive(async () =>
```
flavioislima, line 643: I think instead of blocking everything, we could use a timeout here to avoid blocking other stuff.
Because I imagine that this is to avoid doing several requests to refresh the list of games due to re-render on some react component. But If we completely block the process, Heroic will be on an infinite loop when refreshing the library if a game is being downloaded if I understood this correctly.

Maybe implement this like on this example:
https://github.com/DirtyHairy/async-mutex#limiting-the-time-waiting-for-a-mutex-or-semaphore-to-become-available

derrod:
> I think we could have our own fork of Legendary with the changes because there are things that are important to Heroic but do not makes sense to Legendary.
> 
> Heroic uses Legendary but Legendary is its own thing. And Rodney already added several features to improve our integration here.
> 
> Maybe @derrod can give his two cents on this.

I've been meaning to add a lock to Legendary for a long time, but I've admittedly neglected the project somewhat in recent months. I've been pretty busy getting everything ready for https://github.com/obsproject/obs-studio/releases/tag/29.1.0-beta1!

I think adding it to Legendary itself is the best way, but I also cannot guarantee any ETA on that.

I would implement it as a lockfile in the config directory that contains the current PID, if Legendary is invoked it would check the file and the PID contained within to see if another legendary process is still running and abort if that is the case.

src/backend/legendary/library.ts
CommandMC: Problem is that we can't know if/when Legendary writes to its "sensitive" files.

As an example, every command that requires authentication might write to `user.json` at any point of its runtime.
Right now, I believe all of them authenticate at the start, so theoretically a timeout would work, but any change to that in Legendary would then break this system, potentially without us knowing before we release & some unlucky users with exactly the right system configuration are affected

Bananaman:
> I think adding it to Legendary itself is the best way, but I also cannot guarantee any ETA on that.
> 
> I would implement it as a lockfile in the config directory that contains the current PID, if Legendary is invoked it would check the file and the PID contained within to see if another legendary process is still running and abort if that is the case.

Hi @derrod, thank you for the guidance on this issue and the great work you do on Legendary!

Since Legendary is written in Python, I would strongly suggest avoiding homebrew PID locks, and instead using this cross-platform locking library which uses OS-feature file locks:

- Project: https://github.com/tox-dev/py-filelock
- Docs: https://py-filelock.readthedocs.io/en/latest/

It's a very Pythonic library, is easy to use, and does locking in the best way for every platform (using every platform's own, best per-platform OS-kernel locking methods). The locks are per-process, meaning that all threads within 1 Python process have access to the same lock, just in case Legendary is multi-threaded.

With this, the following becomes possible:

- Before reading or writing to `user.json` or the library JSON: Wait to acquire the lock.
- - If reading had to wait: Read the latest version of the file (whatever the other process put in it).
- - If writing had to wait: Read the latest version of the file from disk, add the changes that were about to be written, and write them to disk.
- Furthermore, any action that uses the authentication tokens could look at the timestamp of `user.json` and automatically load the latest token from disk if the modification timestamp has changed from what Legendary last saw when it originally read the data.

This would solve all the issues of processes desyncing their data.

But an initial implementation could be a lot more coarse than this granular approach, just to get the ball rolling.

arielj:
@CommandMC should we keep this PR open? the issue we had with the login was fixed and this is probably not needed?

CommandMC:
The root issue is of course still present, although I agree in that this PR will never be the solution. This will have to be added into Legendary itself

PR #2585, opened 2023-03-30
Call `legendary list` only once when refreshing the library

arielj:
This might fix the login issues that seems to be caused by 2 calls to `legendary list` happening at the same time.

We were doing 2 calls: one to refresh the library and, if checking updated, one to process the updates.

With this change we only do 1: if we are checking updates we don't do the first one (since checking updates now calls the same method), if we are NOT checking updates then we only do the first one.

---

### DIFF
--- a/src/backend/legendary/library.ts
+++ b/src/backend/legendary/library.ts
@@ -41,7 +41,7 @@ import {
 import { installStore, libraryStore } from './electronStores'
 import { callRunner } from '../launcher'
 import { dirname, join } from 'path'
-import { isOnline } from '../online_monitor'
+import { isOnline } from 'backend/online_monitor'
 
 /**
  * Legendary LegendaryLibrary.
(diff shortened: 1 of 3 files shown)


src/backend/legendary/library.ts
```diff
@@ -247,42 +247,7 @@ export class LegendaryLibrary {
...
-    }
-
-    if (res.error) {
-      logError(
-        ['Failed to check for game updates:', res.error],
-        LogPrefix.Legendary
-      )
-      return []
-    }
+    this.getGames(true)
```
arielj, line 250: we can reuse `this.getGames` that does all the online checks and legendary call and also updates the list of games

src/frontend/state/GlobalState.tsx
```diff
@@ -410,6 +410,15 @@ class GlobalState extends PureComponent<Props> {
   ): Promise<void> => {
     console.log('refreshing')
 
+    let updates = this.state.gameUpdates
```
arielj, line 413: moved this call to checkGameUpdates to the top, so it updates the library info before using it in case this is the call to `legendary list`

Nocccer (approved):
LGTM

flavioislima (approved):
LGTM

arielj:
don't merge this! I think I've found one issue

arielj:
ok now it's fine, I only had changed the `listUpdateableGames` for Legendary refresh the library and I was missing the change for GOG so the GOG library was not updating properly when heroic opens

PR #2586, opened 2023-03-31
[Tests] Add automated tests for window.api

BrettCleary:
### Summary
This branch runs Playwright in a Docker container to automate tests of window.api methods and enable other e2e tests. icpMain handlers can be called directly as well for hard to reach preload scripts.

### Background Info
Playwright is an e2e testing framework with experimental electron support https://www.electronjs.org/docs/latest/tutorial/automated-testing#using-playwright 
- Spectron was previously popular but has been deprecated and is no longer maintained

We currently use jest to do module level unit tests of internal only code on the backend. 
There are significant limitations when using jest to test/mock window.api methods that utilize external code (e.g. electron) methods.

We would like the ability to conduct e2e testing and isolated behavioral testing of the backend api interface in dev mode and in the packaged app to improve QA and reduce time spent manually testing the app.

### Rationale
Testing the window.api methods is the best place to expand automated testing within Heroic in the short term. These tests will isolate the backend and ensure that window.api methods are behaviorally correct. This also allows us to refactor implementation details in the backend without having to update tests, as long as the interface remains the same. 
- A recent example of this would be the store manager refactor which could have used much of the same tests written previously since the window.api interface didn't change much but the implementation was dramatically overhauled

We are also free to change frontend features without having to refactor window.api tests.

Debugging issues and developing full stack features is much simpler, more robust, and faster. Instead of developing and debugging full stack manually, we can write tests first to ensure the backend is working before beginning to develop the frontend. This can help to decouple frontend and backend development

With high test coverage, most of the issues will be constrained to the frontend only as well

### Further Work 
Implement tests for each of the window.api methods.
Explore epic/gog dummy accounts perhaps added as github action env variable secrets to test api methods requiring login

---


Nocccer:
Good stuff. Anyway last time i wanted to write e2e test via spectron i could not test something because you need to login into epic or gog and this can't be done automatically, because of captcha. Maybe playwright can solve this ?

BrettCleary:
> Good stuff. Anyway last time i wanted to write e2e test via spectron i could not test something because you need to login into epic or gog and this can't be done automatically, because of captcha. Maybe playwright can solve this ?
> 
> Update: Ok the captcha should be not the problem. The problem is that locally our epic/gog account is used. It better should use a dummy or simple account to avoid damage your own account config. This also might solve the problem on the buildbot where no account is logged in already. So probably a kind of container/sandbox is needed to start the test isolated and with fresh configs.
> 
> Update2: We probably need to run e2e tests with docker to avoid local config changes.

Great idea. I just dockerized the tests and created github actions for them

I think we should work on setting up the dummy epic/gog account login and writing tests for those particular api methods on a different branch. What do you think @Nocccer ?

flavioislima (approved):
Pretty nice stuff!

PR #2593, opened 2023-04-03
[FIX] GOG Fixes

flavioislima:
- Fix an issue with GOG games keep asking for an update after an update since the ETAG was not being updated on installed.json file.
- Fix an issue where when changing a GOG game path on macOS games, it was not adding the .app folder to the path.
- Fix an issue with importing native GOG macOS games, where we could not select the .app file, only the folder.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "heroic",
-  "version": "2.6.3",
+  "version": "2.7.0",
   "private": true,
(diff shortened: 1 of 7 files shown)


Nocccer (approved):
Lgtm. Maybe linguin can still look over it?

imLinguin (approved):
Looks good

PR #2601, opened 2023-04-05
README: Remove the instruction for removing attribute to app

MichaelGDev48:
This is no longer needed since 2.7 as it is now properly signed with an apple developer certificate

adjusted readme to reflect new changes
---

- [ x] Tested the feature and it's working on a current and clean install.
- [ x] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ x] Created / Updated Tests (If necessary)
- [ x] Created / Updated documentation (If necessary)


flavioislima (approved):
Good catch! Forgot that when I added the signing workflow.

PR #2612, opened 2023-04-07
[Fix]: Blank screen on startup

Nocccer:
- [x] Fixed frontend error where length was called on undefined
- [x] Fixed frontend error where `epic.library` in `src/frontend/components/UI/SearchBar/index.tsx` was not iterateable.

Extra:
- Updated flatpak files
- Removed unessecary if statement (probably merge fault)

---


src/frontend/components/UI/SearchBar/index.tsx
```diff
@@ -24,15 +24,15 @@ export default React.memo(function SearchBar() {
...
-    const library = new Set(
-      [...epic.library, ...gog.library, ...sideloadedLibrary]
-        .filter(Boolean)
-        .map((g) => g.title)
-        .sort()
-    )
-    return [...library].filter((i) =>
-      new RegExp(fixFilter(filterText), 'i').test(i)
-    )
+    const library = [
```
flavioislima, line 27: We are using Set right now to avoid duplicating items in case the person has the same game on Epic and Gog since we are using the title.

I think you can solve this issue by just doing the `epic?.library`, etc. Because we are already removing the invalid items when doing the `filter(Boolean)` later. I think that should do the trick.

Nocccer: I added back the Set.

PR #2628, opened 2023-04-15
[Tech] Update to Vite 4

CommandMC:
TODO:
- [X] Bundle size of the `main` application is rather large (500KB before update, ~4.5MB now), find out why this is happening if possible. Probably something to do with Vite 4 itself, along with `vite-plugin-electron` changes
     **Update**: This is most likely caused by *some dependency* of ours including extra code when not built for the browser. Since this same option also makes other packages *work*, we can't just disable it either. In the grand scheme of things, +4MB isn't that bad, considering a final bundle for any platform can easily approach 100MB
- [X] `src/backend/logger/ipc_handler.ts` seems to be imported/ran twice when running the development build (not happening in production-built version), causing Electron to error about trying to handle the same event twice. No idea at all why this is happening  
     **Update**: Seems like this is caused by Vite getting confused about our circular import issues (which to be fair, I don't blame it for). So that looks like it's *also* better to be untangled in another PR

---


src/backend/wine/manager/downloader/__tests__/main/getter.test.ts
```diff
@@ -63,23 +63,4 @@ describe('Main - GetAvailableVersions', () => {
...
-      })
-      .catch(() => {
-        throw Error('No error should be thrown!')
-      })
-
-    expect(axios.default.get).not.toBeCalled()
-    expect(console.warn).toBeCalledWith(
-      'Unknown and not supported repository key passed! Skip fetch for -1'
-    )
-  })
```
CommandMC, line 84: This situation can't ever happen in practice since Typescript forces us to make sure we're passing valid repositories to the function

src/frontend/components/UI/Sidebar/index.tsx
```diff
@@ -8,7 +8,7 @@ import { DMQueueElement } from 'common/types'
 
 import { ReactComponent as HeroicIcon } from 'frontend/assets/heroic-icon.svg'
 
-let sidebarSize = localStorage.getItem('sidebar-width') || 240
+let sidebarSize = Number(localStorage.getItem('sidebar-width')) || 240
```
CommandMC, line 11: Not sure if this is correct here. Node's type definitions say that `getItem` only ever returns a string, but we want a number here

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "ignoreDeprecations": "5.0",
```
CommandMC, line 3: I plan on tackling this in another PR, since that'll involve quite a bit of work

vite.config.ts
```diff
@@ -1,53 +1,61 @@
...
   },
   plugins: [
     react(),
     electron([
       {
         entry: 'src/backend/main.ts',
-        vite: electronViteConfig
+        vite: {
+          build: {
+            outDir: 'build/main',
```
CommandMC, line 38: We have to separate the `main` and `preload` scripts now since `vite-plugin-electron` assumes that a script has finished building once the folder it should be in is created (at least it seems like that, haven't checked the code).

If we don't build to separate directories, it builds one script, then thinks both are done and starts Electron, which promptly fails

arielj:
> * [x]  `src/backend/logger/ipc_handler.ts` seems to be imported/ran twice when running the development build (not happening in production-built version), causing Electron to error about trying to handle the same event twice. No idea at all why this is happening
>   **Update**: Seems like this is caused by Vite getting confused about our circular import issues (which to be fair, I don't blame it for). So that looks like it's _also_ better to be untangled in another PR

I don't know if it's related to this because I think what I see happens earlier than that, but I see duplicated messages when starting the app:
image

I also see this message that I don't see in the main branch:

```

**Comment by arielj** [comment]
I do see build time and file sizes are pretty big compared to vite 3

the preload.js with vite 3:
```
✓ 9 modules transformed.
build/electron/preload.js   18.32 KiB / gzip: 3.57 KiB
built in 69ms.
```

With vite 4:
```
✓ 385 modules transformed.
build/preload/preload.js  199.56 kB
built in 7311ms.
```

I agree that 4mb is not that much in a 100mb file, but I wonder what's the impact of that when the browser has to parse that, it's a lot of javascript.

CommandMC:
> I see duplicated messages when starting the app

You're building the main backend (`src/backend/main.ts`) and the preload (`src/backend/preload.ts`), so you'll also get two messages. *Ideally* we'd keep the preload in its own directory (`src/preload/`) to signify this disconnected-ness, but I generally don't like to make architectural changes for no reason

> I wonder what's the impact of that when the browser has to parse that, it's a lot of javascript

I personally haven't noticed any impact on startup times.
I've talked to the maintainer of our Electron plugin about this, the root cause of this issue is that *some package* we're using is inserting a lot of code when built for NodeJS (as opposed to building for a web browser). The fact that packages are even build for this (correct) scenario is a new addition in the plugin, and without it, certain other dependencies (like `axios`) wouldn't work.
I'm not sure how this worked with our older version of the plugin, my best guess would be that we hit just the right combination of Vite + Electron plugin + axios versions. Of course, being locked to those specific versions is not a good idea

I plan on combing through our dependency list to find these problematic packages and possibly find/write replacements, but that would balloon this PR up to "nobody-wants-to-review-it" range pretty fast

CommandMC:
Superseded by #3218

PR #2629, opened 2023-04-16
Refactor legendary setup

Etaash-mathamsetty:
~~created this PR early when wine-ge 8.x (or proton 8) eventually releases~~
it's released


guihkx:
(I just found this PR 'by accident', so please disregard this comment if I don't make any sense).

If I understand correctly, this is only applicable for Proton 8, right? But doesn't Heroic allow changing the Proton version freely? What would happen if the user manually downgraded their Proton version and tried to run Ubisoft Connect?

Etaash-mathamsetty:
> (I just found this PR 'by accident', so please disregard this comment if I don't make any sense).
> 
> If I understand correctly, this is only applicable for Proton 8, right? But doesn't Heroic allow changing the Proton version freely? What would happen if the user manually downgraded their Proton version and tried to run Ubisoft Connect?

proton ships fonts already, so that's not a problem, and users should be using wine-ge 8.0 when that releases. (it will be pretty well regression tested since valve is taking forever to release proton 8)

Etaash-mathamsetty:
this became relevant way sooner than expected :0

Etaash-mathamsetty:
this is now relevant! (wine-ge 8.0 has released)

arielj:
Maybe we should add a conditional for now to ignore this if wine 8? I know after the proton 8 release there was an issue with Elden Ring, I if it's better to just ignore the installation if 8 and if anybody needs to use 7 for any reason (who knows) this would still work

flavioislima:
I think it is fine to install that on wine 8, it is just a font anyway. I don't think it will cause any harm and adding a `wine --version` check will slow down the process as well.

Perhaps once Wine-GE is more stable in the future we can get back to this but for now this is too hasty imo.

Etaash-mathamsetty:
ok tested with wine-ge (it needed WINE_FSYNC_SIMULATE_SCHED_QUANTUM=1 to run), and I can confirm that it works without installing the arial font

Etaash-mathamsetty:
I think this can get merged now, considering wine-ge can run ubisoft connect flawlessly now ootb

Etaash-mathamsetty:
it's been a few months, I think we could potentially merge this after 2.9.1

arielj (approved):
Looks good to me, I tested installing/playing a ubisoft game and fallout new vegas.

PR #2563, opened 2023-03-21
[FIX] Graphics card info can't always be resolved

Kek5chen:
When starting a game, or starting the launcher, it tries to resolve the GPU and driver information using the systeminformation library's graphics() call (aswell as cpu(), mem() and osInfo() which work unreasonably, but i do see an issue not checking if they run indefinitely too).

Fixes #2562

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [x] Tested the feature and it's working on a current and clean install.
- [x] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #2638, opened 2023-04-18
Fix missing game settings info in logs

arielj:
This PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2563/files introduced a regression. Because we are calling `writeFileSync` in the 2 places, the second call (system info) replaces the content of the file that had the content of the first call (the game settings).

Now it uses writeFileSync for the game settings and then appendFileSync for the system info to not override the content.

The game settings are now the first part of the file, before the 2.7 release the system info was first. This is not that easy to avoid since the system info fetch is async on purpose cause it can get stuck, so we can't wait for it to be the beginning of the file.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -955,7 +955,7 @@ ipcMain.handle(
 
     systemInfo.then((systemInfo) => {
       if (systemInfo === '') return
-      writeFileSync(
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
LGTM

PR #2659, opened 2023-04-28
[UX] Add Keyboard global shortcuts

flavioislima:
This pull request adds several hotkeys to improve the user experience of our application. The new hotkeys are as follows:

CommandOrControl+R: Reloads the app
CommandOrControl+Q: Quits the app
CommandOrControl+Shift+I: Opens the dev tools
CommandOrControl+K: Opens the Heroic settings screen on the frontend
CommandOrControl+L: Opens the library screen
CommandOrControl+J: Opens the downloads screen

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -19,7 +19,8 @@ import {
   powerSaveBlocker,
   protocol,
   screen,
-  clipboard
+  clipboard,
+  globalShortcut
 } from 'electron'
 import 'backend/updater'
 import { autoUpdater } from 'electron-updater'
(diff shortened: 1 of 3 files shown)


arielj (approved):
Tested on linux and it works fine :+1:

I have 2 questions but I don't think they are blockers:
- Maybe Ctrl+D can open the Downloads screen?
- I wonder if there's a way to add menu entries for the library, downloads, and settings shortcuts so then users can discover the shortcuts from the menu like you can discover things like quit/reload/devetools

image

Or maybe an idea for the future to show this information somewhere inside heroic.

src/backend/main.ts
```diff
@@ -397,6 +398,36 @@ if (!gotTheLock) {
...
+    // hotkey to open the library screen on frontend
+    globalShortcut.register('CommandOrControl+L', () => {
+      sendFrontendMessage('openScreen', '/library')
+    })
+
+    // hotkey to open the downloads screen on frontend
+    globalShortcut.register('CommandOrControl+J', () => {
+      sendFrontendMessage('openScreen', '/download-manager')
+    })
+
```
Nocccer, line 430: Can we add this maybe in a seperated file?

Nocccer (changes requested):
LGTM. To avoid filling up `main.ts` again, is it possible to move the shortcuts to another file/folder?

src/backend/main.ts
flavioislima: Unfortunately not, I tried moving it to other places but it needs to be done inside the `app.WhenReady` method, see documentation:
https://www.electronjs.org/docs/latest/api/global-shortcut/

flavioislima:
@ariel, yes we need to make them more discoverable and we can expand that in the future. For now I think as a first implementation is fine though.

PR #2672, opened 2023-05-04
[FIX] fix Keyboard shortcuts being activated globally rather than within heroic

Etaash-mathamsetty:
Unfortunately, we lose the custom shortcuts we created (but we retain the default ones, like reload, dev tools, quit, etc), but at least I don't have heroic stealing focus every time I press control-r

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -19,8 +19,7 @@ import {
   powerSaveBlocker,
   protocol,
   screen,
-  clipboard,
-  globalShortcut
+  clipboard
 } from 'electron'
 import 'backend/updater'
 import { autoUpdater } from 'electron-updater'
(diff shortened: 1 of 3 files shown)


src/backend/main.ts
```diff
@@ -398,36 +396,6 @@ if (!gotTheLock) {
...
-    })
-
-    // hotkey to open the library screen on frontend
-    globalShortcut.register('CommandOrControl+L', () => {
-      sendFrontendMessage('openScreen', '/library')
-    })
-
-    // hotkey to open the downloads screen on frontend
-    globalShortcut.register('CommandOrControl+J', () => {
-      sendFrontendMessage('openScreen', '/download-manager')
```
arielj, line 428: I wonder if we can add these shortcuts in the frontend shortcuts.ts file https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/frontend/helpers/shortcuts.ts

that uses event listeners of the frontend so we can listen the combinations without having to deal with the menu or depending on electron features

arielj: we can always call electron functions when needed through window.api to still support the quit/devtools/reload commands

arielj: I made another PR with that idea since it's reusing things removed from this PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2684

we can keep all the new shortcuts while also making them local without relying on electron accelerators

Issue #2680, opened 2023-05-05
Download progress is not reported


outphase:
### Describe the bug

I am currently downloading The Beast Inside via GOG, and while the game is in fact being downloaded, HGL is not reporting any progress at all. The UI is still functional otherwise.

### Add logs

```text
(13:10:10) INFO:    [Legendary]:        Refreshing library...
(13:10:10) INFO:    [Legendary]:        Refreshing Epic Games...
(13:10:10) INFO:    [Connection]:       Connectivity: check-online
(13:10:10) INFO:    [Connection]:       Pinging external endpoints
(13:10:10) INFO:    [Legendary]:        Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary --version
(13:10:10) INFO:    [Legendary]:        Legendary location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(13:10:10) INFO:    [Gog]:              GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(13:10:10) INFO:    [Gog]:              GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(13:10:10) INFO:    [Legendary]:        Game list updated, got 343 games & DLCs
(13:10:10) WARNING: [Backend]:          Failed to register protocol with OS.
(13:10:10) INFO:    [Gog]:              Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/outphase/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json --version
(13:10:10) INFO:    [Connection]:       Connectivity: online
(13:10:10) INFO:    [Gog]:              Checking if login is valid
(13:10:10) INFO:    [Gog]:              Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/outphase/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json auth
(13:10:10) INFO:    [Legendary]:        Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary list --third-party
(13:10:10) INFO:    [Backend]:          Loading Screen Ready
(13:10:10) INFO:    [Frontend]:         Refreshing Library
[...]

PR #2683, opened 2023-05-06
Fix installation/update progress reports to the frontend

arielj:
This PR fixes a regression in 2.7.1 where the installation progress is not reported properly to the frontend.

Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2680

---

### DIFF
--- a/src/backend/storeManagers/gog/games.ts
+++ b/src/backend/storeManagers/gog/games.ts
@@ -148,12 +148,14 @@ interface tmpProgressMap {
   [key: string]: InstallProgress
 }
 
-const defaultTmpProgress = {
-  bytes: '',
-  eta: '',
-  percent: undefined,
-  diskSpeed: undefined,
-  downSpeed: undefined
+function defaultTmpProgress() {
+  return {
+    bytes: '',
+    eta: '',
+    percent: undefined,
+    diskSpeed: undefined,
+    downSpeed: undefined
+  }
 }
 const tmpProgress: tmpProgressMap = {}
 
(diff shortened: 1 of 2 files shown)


src/backend/storeManagers/gog/games.ts
```diff
@@ -165,48 +167,59 @@ export function onInstallOrUpdateOutput(
...
 ) {
   if (!Object.hasOwn(tmpProgress, appName)) {
-    tmpProgress[appName] = defaultTmpProgress
+    tmpProgress[appName] = defaultTmpProgress()
   }
+  const progress = tmpProgress[appName]
+
   // parse log for percent
-  const percentMatch = data.match(/Progress: (\d+\.\d+) /m)
+  if (!progress.percent) {
```
arielj, line 175: these checks were there in the original code, we need them to make sure we update only specific fields of the progress object

src/backend/storeManagers/gog/games.ts
```diff
@@ -148,12 +148,14 @@ interface tmpProgressMap {
   [key: string]: InstallProgress
 }
 
-const defaultTmpProgress = {
-  bytes: '',
-  eta: '',
-  percent: undefined,
-  diskSpeed: undefined,
-  downSpeed: undefined
+function defaultTmpProgress() {
```
arielj, line 151: these default object was inlined before and extracted here, but it needs to be inside a function so we always return a new object

the current code without a function passes the same object around and it's getting updated, it doesn't stay i this initial state

flavioislima (approved):
Tested here and the progress is working again.

Issue #2676, opened 2023-05-04
Hotkeys are captured globally on macOS.


astrochili:
### Describe the bug

It looks like [these hotkeys](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2659) are captured globally on macOS.

Being in the Internet browser, for example, Cmd+R or Cmd+Q affect Heroic, but not the browser. It's absolutely not okay.

### Steps to reproduce

Press Cmd+Q (or other hotkey) when Heroic's window isn't active.

### Expected behavior

Shortcuts works in the current context and nothing happens in Heroic.

### Heroic Version

Version 2.7.1 (2.7.1)

### System Information

- macOS Ventura

astrochili:
A pull request by @Etaash-mathamsetty already exists: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2672

Etaash-mathamsetty:
> A [pull request](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2672) by @Etaash-mathamsetty already exists, but looks like it just disables the new functionality.

It doesn't it retains some of the shortcuts while removing some of the heroic specific ones. While it's possible to add them back, let's fix the bug first

Issue #2671, opened 2023-05-03
New Keyboard Shortcuts Are Global and Overriding Other Programs


WillPresley:
### Describe the bug

After the 2.7.1 update, I noticed that Ctrl-R wasn't working to refresh the tab in Chrome anymore. After peeking in Process Explorer to see what was firing up when I used the keyboard shortcut, I noticed it was Heroic Launcher. This doesn't seem to be intended behavior, as I would expect those shortcuts to only work with the program focused. If it was intended, it would be great to have a toggle for Local/Global or the ability to change/disable them, because I'd much rather not have them as global shortcuts.

### Add logs

```text
(14:32:08) INFO:    [Legendary]:        Refreshing library...
(14:32:08) INFO:    [Legendary]:        Refreshing Epic Games...
(14:32:08) INFO:    [Connection]:       Connectivity: check-online
(14:32:08) INFO:    [Connection]:       Pinging external endpoints
(14:32:08) INFO:    [Legendary]:        Running command: C:\Users\XXXXX\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary --version
(14:32:08) INFO:    [Legendary]:        Legendary location: C:\Users\XXXXX\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary
(14:32:08) INFO:    [Gog]:              GOGDL location: C:\Users\XXXXX\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\gogdl
(14:32:08) INFO:    [Gog]:              GOGDL location: C:\Users\XXXXX\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\gogdl
(14:32:08) INFO:    [Legendary]:        Game list updated, got 263 games & DLCs
(14:32:08) WARNING: [Backend]:          Protocol already registered.
(14:32:08) INFO:    [Connection]:       Connectivity: online
(14:32:08) INFO:    [Gog]:              Checking if login is valid
(14:32:08) INFO:    [Gog]:              Running command: C:\Users\XXXXX\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\gogdl --auth-config-path C:\Users\XXXXX\AppData\Roaming\heroic\gog_store\auth.json auth
[...]

PR #2684, opened 2023-05-06
[Fix] Handle shortcuts from the frontend without depending on menu

arielj:
This PR is an adaptation of the current globalShortcuts feature, instead of using globalShorcuts that are causing issues, it uses an event listener in the frontend and based on the different combinations it sends a message to the backend to be processed

This fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2676 and fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2671

---


flavioislima (approved):
Cool, way better now. Tested here and it is working just fine!

PR #2688, opened 2023-05-07
[FIX] Incorrect response on Epic Login

flavioislima:
Downgraded Electron to v23.1.2 since the version 23.1.3 had some regression on the webview component causing it to return incorrect response. 

The problem is that this issue is still occorruing on version 24.2 so we will need to ask for help from the electron team or find another workaround for it in the future.

For now, I tested more than 10 times and with this version the login always work. When i upgrade it, the login always fails.

---

### DIFF
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -197,7 +197,7 @@ export default function WebView() {
         partition="persist:epicstore"
         src={startUrl}
         allowpopups={trueAsStr}
-        useragent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/200.0"
+        useragent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/111.0"
         {...(preloadPath ? { preload: preloadPath } : {})}
       />
       {showLoginWarningFor && (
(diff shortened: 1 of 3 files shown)


src/frontend/screens/WebView/index.tsx
```diff
@@ -197,7 +197,7 @@ export default function WebView() {
         partition="persist:epicstore"
         src={startUrl}
         allowpopups={trueAsStr}
-        useragent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/200.0"
+        useragent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/111.0"
```
flavioislima, line 200: Decided to use a version that exists instead of a future number to avoid any other issue. But in my tests this does not make any difference.

arielj:
@flavioislima should we close this PR? since the issue seems to not be the electron version? I understand that after this change it's still failing sometimes

flavioislima:
> @flavioislima should we close this PR? since the issue seems to not be the electron version? I understand that after this change it's still failing sometimes

As a electron change version, yes, but them we need another one to investigate the issue with epic login.

arielj:
I'll close this then to avoid confusion, the electron version seems to be fine and the useragent doesn't seem to cause the issue

PR #2689, opened 2023-05-07
[FIX] On Logout, do Not clear All Cache

flavioislima:
I noticed that when logging out from Legendary the GOG cache was also being cleaned up and this is causing the GOG library to be clear after that and a full refresh was needed to reload the GOG lib.

---


imLinguin (approved):
Looks good!

arielj (approved):
looks good to me

PR #2693, opened 2023-05-09
Fix issue with epic login/browserview not saving state between route changes

katuak:
This PR disables the use of `webviewTag` and migrates to BrowserView. This has some benefits:
- Better performance
- Fixes an issue with the epic games login
- Moves the logic away from the frontend


src/backend/browserview.ts
```diff
@@ -0,0 +1,146 @@
...
+  URLchanged: { (): void }[] = []
+
+  constructor(
+    identifier: IBrowserViewIdentifier,
+    { initialURL }: IBrowserViewOptions
+  ) {
+    super()
+    this.view = new BrowserView()
+    this.initialURL = initialURL
+    this.view.webContents.on('did-navigate-in-page', () => {
```
aitor-gomila, line 21: Not completely sure about this approach. It has many issues. 
1. Creates two different URLchanged variables, for backend and for frontend
2. RemoteBrowserView is intended to be a mere reflection of NodeBrowserView, so you can create many instances safely. In this case it will not work because the callbacks are in the frontend.
3. However putting them on backend isn't very useful, because you can't send methods via IPC.
@flavioislima

Issue #2678, opened 2023-05-05
Native linux games failing to install


3vilg3nius:
### Describe the bug

after the 2.7.1 update when installing native linux games the install will fail, windows games install fine.

### Add logs

```text
(15:29:50) INFO:    [Gog]:              Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/john/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json auth

(15:29:50) DEBUG:   [Backend]:          {"free":409338982400,"diskSize":488983355392,"message":"381.23 GiB / 455.4 GiB","validPath":true}

(15:29:50) INFO:    [Gog]:              Getting game metadata: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/john/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json info 1150392214 --token <redacted> --lang=en-US --os windows

(15:29:52) DEBUG:   [Backend]:          {"free":409338982400,"diskSize":488983355392,"message":"381.23 GiB / 455.4 GiB","validPath":true}

(15:30:02) DEBUG:   [Backend]:          {"free":87987875840,"diskSize":471416549376,"message":"81.95 GiB / 439.04 GiB","validPath":true}

(15:30:03) INFO:    [Gog]:              Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/john/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json auth

(15:30:03) INFO:    [Gog]:              Getting game metadata: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/john/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json info 1150392214 --token <redacted> --lang=en-US --os linux

(15:30:04) ERROR:   [Gog]:              stdout = false and res.abort = false in getInstallInfo

(15:30:04) INFO:    [DownloadManager]:  DUSK  was added to the download queue.

[...]

PR #2695, opened 2023-05-10
[Fix] Linux native games failing to add to launcher

av-dx:
Due to the error in obtaining the installation info for them, Linux-native (GOG) games were being installed on the system but not added to the Launcher as reported in #2678 . This workaround gets the Windows install info instead, proceeding with successful installation.

---


arielj (approved):
thanks! tested here and it works as expected

PR #2697, opened 2023-05-10
Simplify access of main `BrowserWindow` and consume less resources

aitor-gomila:
This PR replaces `getMainWindow()` with `mainWindow`, which simplifies code and calls to `mainWindow`. It also helps keep just 1 version of it instead of multiple references.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -145,10 +145,10 @@ app.commandLine?.appendSwitch('remote-debugging-port', '9222')
 const { showOpenDialog } = dialog
 const isWindows = platform() === 'win32'
 
-async function initializeWindow(): Promise<BrowserWindow> {
+async function initializeWindow(): Promise<void> {
   createNecessaryFolders()
   configStore.set('userHome', userHome)
-  const mainWindow = createMainWindow()
+  createMainWindow()
 
   if ((isSteamDeckGameMode || isCLIFullscreen) && !isCLINoGui) {
     logInfo(
@@ -160,7 +160,7 @@ async function initializeWindow(): Promise<BrowserWindow> {
       ],
       LogPrefix.Backend
     )
-    mainWindow.setFullScreen(true)
+    mainWindow!.setFullScreen(true)
   }
 
   setTimeout(() => {
@@ -170,30 +170,30 @@ async function initializeWindow(): Promise<BrowserWindow> {
 
   GlobalConfig.get()
 
-  mainWindow.setIcon(icon)
+  mainWindow!.setIcon(icon)
   app.setAppUserModelId('Heroic')
   app.commandLine.appendSwitch('enable-spatial-navigation')
 
-  mainWindow.on('close', async (e) => {
+  mainWindow!.on('close', async (e) => {
     e.preventDefault()
 
     if (!isCLIFullscreen && !isSteamDeckGameMode) {
       // store windows properties
-      configStore.set('window-props', mainWindow.getBounds())
+      configStore.set('window-props', mainWindow!.getBounds())
     }
 
     const { exitToTray } = GlobalConfig.get().getSettings()
 
     if (exitToTray) {
       logInfo('Exitting to tray instead of quitting', LogPrefix.Backend)
-      return mainWindow.hide()
+      return mainWindow!.hide()
     }
 
     handleExit()
   })
 
   if (isWindows) {
-    detectVCRedist(mainWindow)
+    detectVCRedist(mainWindow!)
   }
 
   if (!app.isPackaged) {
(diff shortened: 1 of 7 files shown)


src/backend/main.ts
```diff
@@ -170,30 +170,30 @@ async function initializeWindow(): Promise<BrowserWindow> {
 
   GlobalConfig.get()
 
-  mainWindow.setIcon(icon)
+  mainWindow!.setIcon(icon)
```
Nocccer, line 173: remove the !

src/backend/main.ts
```diff
@@ -170,30 +170,30 @@ async function initializeWindow(): Promise<BrowserWindow> {
...
   app.setAppUserModelId('Heroic')
   app.commandLine.appendSwitch('enable-spatial-navigation')
 
-  mainWindow.on('close', async (e) => {
+  mainWindow!.on('close', async (e) => {
```
Nocccer, line 177: remove the !

src/backend/main.ts
```diff
@@ -170,30 +170,30 @@ async function initializeWindow(): Promise<BrowserWindow> {
...
-      configStore.set('window-props', mainWindow.getBounds())
+      configStore.set('window-props', mainWindow!.getBounds())
     }
 
     const { exitToTray } = GlobalConfig.get().getSettings()
 
     if (exitToTray) {
       logInfo('Exitting to tray instead of quitting', LogPrefix.Backend)
-      return mainWindow.hide()
+      return mainWindow!.hide()
```
Nocccer, line 189: remove `!`

src/backend/main.ts
```diff
@@ -170,30 +170,30 @@ async function initializeWindow(): Promise<BrowserWindow> {
...
     }
 
     handleExit()
   })
 
   if (isWindows) {
-    detectVCRedist(mainWindow)
+    detectVCRedist(mainWindow!)
```
Nocccer, line 196: remove `!`

src/backend/main.ts
Nocccer, line 1: Never use `!` operator. Remove all of them you added.

aitor-gomila:
It's highly unlikely that mainWindow is undefined, since it is defined after 'ready' event of electron. In case it's not, what do we do? What about `mainWindow?`? (silently ignores/errors instead of panicking)

Nocccer:
> It's highly unlikely that mainWindow is undefined, since it is defined after 'ready' event of electron. In case it's not, what do we do? What about `mainWindow?`? (silently ignores/errors instead of panicking)

"It's highly unlickely" is still a check worth. That is in my opinion just bad programming. Don't trust that it never be undefined or null. 

Also if someone changes the behavior of creating the window all other calls of `mainWindow!` will be not detected that they can be undefined by the type check.

arielj:
I don't know if I understand this PR, it's removing something that was added on purpose avoid working with a global variable and as a consequence of this change you have to add all the extra `!` which feels like a typing system code smell (you can't trust your type).

I don't see the benefit of reverting an abstraction, what's the comparison before and after of the resources? is this really necessary? sounds like a micro-optimization and it's hard to see what do we gain

there are way more important things to optimize to really have an impact on resource

aitor-gomila:
> 

Actually, this is benefitial for the type system. Window could return null but we weren't taking that into account. Exporting a global variable directly is the way it was intended in JS

aitor-gomila:
Writing down a function every single time someone needs to call mainWindow doesn't sound good, what do you propose?

Nocccer:
> > 
> 
> Actually, this is benefitial for the type system. Window could return null but we weren't taking that into account. Exporting a global variable directly is the way it was intended in JS

I don't mind the variable instead the function but you need to make sure that it is not null before using. This is needed for both ways. 

Btw what do you think about making the mainWindow a singleton? So it can't be null if you ask for the handler.

aitor-gomila:
> I don't mind the variable instead the function but you need to make sure that it is not null before using. This is needed for both ways.

Agreed

> Btw what do you think about making the mainWindow a singleton? So it can't be null if you ask for the handler.

I don't know, there's cases where it will be null and there's nothing to do because it's an electron thing. I would just abort and log it

arielj:
> Actually, this is benefitial for the type system. Window could return null but we weren't taking that into account.

I think we were, the type checker was not complaining about it so it was handled, for example in one of the files it's replaced to now use `mainWindow!` it was doing `const mainWindow = getMainWindow()!` which handles it in the same way in one place instead of in every use of mainWindow (in that specific handle callback mainWindows MUST 100% exists because that event comes from the frontend so I was adding the `!` at the beginning). In other places it's doing `mainWindow?.` to also ignore null possibilities.

> Exporting a global variable directly is the way it was intended in JS

I don't know where this statements comes from, you can export whatever you want, what do you mean it's the intended way in JS? is it kind of a language official statement?

> Writing down a function every single time someone needs to call mainWindow doesn't sound good, what do you propose?

I propose not changing it, I don't see the problem. At least from the description of the PR this is not really simplifying the use (you have to add more `!` in more places) and it's not clear what's the resources usage benefit (you say `It also helps keep just 1 version of it instead of multiple references` and I don't see the problem with having local references to the same variable, I understand it's not copying the mainWindow instance every time, it's just adding a local reference).

personally I prefer the locality of having the variable defined at the beginning of the function instead of a global variable


but that is more of my opinion, whatever everyone prefers I'm ok with, I just feel this PR is more an opinion too rather than an actual objective benefit, at least I can't see it

aitor-gomila:
> I just feel this PR is more an opinion too rather than an actual objective benefit, at least I can't see it

Fair

PR #2705, opened 2023-05-14
[GOG]: Get game builds only from main branch

imLinguin:
This PR addresses https://github.com/Heroic-Games-Launcher/heroic-gogdl/issues/34  
Here we are making sure we use builds from main branch only. This is temporal change and branching will be further customizable in slowly progressing gogdl refactor. 

### DIFF
--- a/src/backend/storeManagers/gog/library.ts
+++ b/src/backend/storeManagers/gog/library.ts
@@ -17,7 +17,8 @@ import {
   GOGGameDotIdFile,
   GOGClientsResponse,
   GamesDBData,
-  Library
+  Library,
+  BuildItem
 } from 'common/types/gog'
 import { basename, join } from 'node:path'
 import { existsSync, readFileSync } from 'graceful-fs'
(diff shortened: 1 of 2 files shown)


flavioislima (approved):
LGTM
Not sure how to test this though.

imLinguin:
Stardew Valley is great example, builds from compat branch were picked before this change. Other game like that is the one mentioned in the gogdl issue. If you own those you can test them

src/backend/storeManagers/gog/library.ts
```diff
@@ -676,7 +677,9 @@ export async function getMetaResponse(
         'If-None-Match': etag
       }
     : undefined
-  const metaUrl = buildData.data?.items[0]?.link
+  const metaUrl =
+    buildData.data?.items.find((build: BuildItem) => !build.branch)?.link ||
```
arielj, line 681:
```suggestion
    buildData.data?.items?.find((build: BuildItem) => !build.branch)?.link ||
```

do we need a `?` here? not sure, maybe we don't (I'm always confused what happens when chaining `?` operators)

imLinguin: Maybe we should, I'm not sure how the response looks like when there are no builds available (that's the edge case but it doesn't hurt I guess.

PR #2717, opened 2023-05-19
[Feature] Download DXVK 1.10.3 if no Vulkan 1.3 support is detected

CommandMC:
First and foremost: Yes, this idea is *borrowed from* Lutris' most recent release. I have however been meaning to do something like this for a long time, since I own several PCs affected by this as well

A lot of older GPUs don't support Vulkan 1.3, which DXVK versions > 1.X.X requires. For those users, the "Auto-Install DXVK" function was entirely useless, as the DXVK Heroic installs will never work on their system.
Now, Heroic will query the user's GPUs for their Vulkan support and install the older 1.10.3 version of DXVK if no Vulkan 1.3 support is detected.

Internally, this is done by:
- accepting not just a string as a download URL for tools, but also a function returning a string
- DXVK's url function querying Vulkan support and choosing either the latest release or 1.10.3 based on that
- new utility functions being added to interface with Vulkan directly

Vulkan functions are called using a helper binary (https://github.com/imLinguin/vulkan-helper-rs)

---


imLinguin (approved):
Wow that was fast. Looks good to me

CommandMC:
The tests look strange to me, it seems like Jest doesn't reset the module's state between runs (even with `resetModules` enabled)

src/backend/tools.ts
```diff
@@ -43,7 +48,36 @@ export const DXVK = {
...
+          }
+          if (any_gpu_supports_version([1, 2, 0])) {
+            logInfo(
+              'The GPU(s) in this system only support Vulkan 1.2, falling back to DXVK 1.10.3',
+              LogPrefix.DXVKInstaller
+            )
+            return 'https://api.github.com/repos/doitsujin/dxvk/releases/tags/v1.10.3'
+          }
+          logWarning(
+            'No GPU with Vulkan 1.2 support found, DXVK will not work',
```
Etaash-mathamsetty, line 74: this should be changed to vulkan 1.1
https://github.com/doitsujin/dxvk/wiki/Driver-support#dxvk-1103

Etaash-mathamsetty (approved):
looks fine other than this

src/backend/tools.ts
```diff
@@ -43,7 +48,36 @@ export const DXVK = {
       },
       {
         name: 'dxvk',
-        url: 'https://api.github.com/repos/doitsujin/dxvk/releases/latest',
+        url: () => {
```
arielj, line 51: maybe this function can be moved outside the `const tools = [...` so we would just set url as a string, and then we don't need the `const download_url = typeof tool.url === 'string' ? tool.url : tool.url()` line if it's always a string

arielj: if you do that you can even do something like `if (download_url) { tools.push({name: 'dxvk', url: dxvkUrl, etc...}) }` so you can handle the case when DXVK cannot be used (we have the FIXME saying we can't handle that case)

CommandMC: Simply not downloading DXVK isn't quite enough, the frontend/other parts of the backend would still try to install it (which would then fail, since the folder isn't there)

arielj: oh I see

arielj: I think my first comment still applies though, to not have to do the if typeof === 'string' conditional

CommandMC: I've moved it into its own function with aebb5fe now. This is definitely clearer, thanks for the suggestion

PR #2718, opened 2023-05-19
[Fix] `AppSettings` type mismatches

CommandMC:
- `userInfo`'s structure was completely different to how it should be like, and it seems like it wasn't used anywhere, so removing it felt like the best option
- The `@ts-expect-error` around the object was moved, this way it won't hide errors like this in the future
- Some other smaller issues were corrected

This is also one of the many steps we'll have to take to untangle our circular dependency issues, as this `LegendaryUser` import currently leads to this import chain:
```
src/backend/constants.ts
-> src/backend/config.ts
-> src/backend/storeManagers/legendary/user.ts
-> src/backend/utils.ts
-> src/backend/storeManagers/legendary/library.ts
-> src/backend/launcher.ts
-> src/backend/tools.ts
...
```

---

### DIFF
--- a/src/backend/config.ts
+++ b/src/backend/config.ts
@@ -5,7 +5,7 @@ import {
   readdirSync,
   writeFileSync
 } from 'graceful-fs'
-import { homedir, userInfo as user } from 'os'
+import { homedir } from 'os'
 import { parse as plistParse, PlistObject } from 'plist'
 
 import {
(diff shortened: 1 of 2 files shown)


CommandMC:
I can't find any usages of `customWinePaths` in places that aren't gated by an `if (!isWindows)`, so I'm *guessing* this is fixed. Still, I'll give this another test on Windows in a few days

flavioislima:
@CommandMC any reason this one has not merged since it was approved in May?

CommandMC:
I wanted to test this on Windows one more time, but then never got around to it. I'll fix up the conflicts & re-test soon:tm:

PR #2719, opened 2023-05-20
[FIX] Require DXVK to enable VKD3D-proton

Etaash-mathamsetty:
~~Works fine, but the checkbox that shows if vkd3d is enabled doesn't become false when dxvk is disabled, requiring the user to leave the modal and reopen it for the checkbox to change state. I have no idea how to fix this :(~~

vkd3d-proton now requires dxvk


Etaash-mathamsetty:
so I used 100% of my brain power and fixed the last issue with this PR

CommandMC (approved):

Issue #2674, opened 2023-05-04
HL 2.5.0.beta1 -> 2.8.0 Unable to Login to Epic: Captcha answers always wrong.


PadTrick:
### Describe the bug

i try to login into my epic account, but regardless what i do, the given answers are always wrong.

### Add logs

```text
no log
```

### Steps to reproduce

try to login

### Expected behavior

to login

### Heroic Version

Latest Stable

### System Information

Arch

arielj:
that sometimes happens to me but even on my browser, I have to try a few times until it changes the prompt of the captcha and then it works fine

can you confirm if after trying 3 or 4 times it changes the prompt and then it works?

PadTrick:
tried it over 20 times on heroic launcher. in browser it works flawless

flavioislima:
This is my experience and it is working just fine. the problem is that those captchas are becoming more and more difficult now:

![2023-05-05_09-52-47 (1)](https://user-images.githubusercontent.com/26871415/236404505-f947bf86-ccda-43d6-bc2c-38703679edf0.gif)

Closing since we can't reproduce the issue.

PadTrick:
strange, for me the only way to get it was to login into browser and solve the captcha there.
i even had the exact same captcha and i solved it, where HL told me that it was wrong.

i'm currently using 2nd login method which doesnt login, but the games in my lib are atleast working.

maybe it's a strange behavior of the captcha idk

flavioislima:
Yes, the problem is not the captcha. Looking further into it, I can see it return status 400, bad request, after the login.

So my theory is that epic found a way of blocking heroic.

Tried several things and could not fix it so far.

I'm reopening this issue.

PadTrick:
Ok, now i cant even play any game anymore.
I'm loggin into epic, using 2nd method, because normal method doesnt work.

An error has occurred! Try to Logout and Login on your Epic account.
gamemodeauto:[cli] INFO: Logging in...
Traceback (most recent call last):
File "legendary/cli.py", line 3061, in <module>File "legendary/cli.py", line 2976, in main
File "legendary/cli.py", line 592, in launch_game
File "legendary/core.py", line 187, in login
ValueError: No saved credentials
[9048] Failed to execute script 'cli' due to unhandled exception!

arielj:
I can't reproduce this, I tried with a clean config and I was able to login first try with the captcha, maybe it was a temporary thing?

[...]

PR #2726, opened 2023-05-24
[FIX] Second try to fix Epic login

flavioislima:
Since some people said that Firefox on Linux was working fine, I changed the user agent to use it on the Login Screen only.
Tested on macOS:
- With default user agents: always fail
- With Firefox on Linux user agent: Captcha fail once and pass the second time. No reponse error message.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -259,6 +259,8 @@ if (!gotTheLock) {
   app.whenReady().then(async () => {
     initStoreManagers()
     initOnlineMonitor()
+    app.userAgentFallback =
+      'Mozilla/5.0 (Windows NT 10.0; WOW64; rv:70.0) Gecko/20100101 Firefox/70.0'
 
     getSystemInfo().then((systemInfo) => {
       if (systemInfo === '') return
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -70,17 +70,6 @@ export default function WebView() {
   const isEpicLogin = runner === 'legendary' && startUrl === epicLoginUrl
   const [preloadPath, setPreloadPath] = useState('')
 
-  // Small hack here to use the user agent of firefox on linux for the epic login screen
-  // and firefox on windows for the rest of the pages
-  // this is to test if the epic login improves
-  let userAgent =
-    'Mozilla/5.0 (Windows NT 10.0; rv:78.0) Gecko/20100101 Firefox/85.0'
-  if (isEpicLogin) {
-    // firefox on linux for login screen
-    userAgent =
-      'Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/113.0'
-  }
-
   useEffect(() => {
     let mounted = true
     const fetchLocalPreloadPath = async () => {
@@ -208,7 +197,6 @@ export default function WebView() {
         partition="persist:epicstore"
         src={startUrl}
         allowpopups={trueAsStr}
-        useragent={userAgent}
         {...(preloadPath ? { preload: preloadPath } : {})}
       />
       {showLoginWarningFor && (
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -70,6 +70,17 @@ export default function WebView() {
   const isEpicLogin = runner === 'legendary' && startUrl === epicLoginUrl
   const [preloadPath, setPreloadPath] = useState('')
 
+  // Small hack here to use the user agent of firefox on linux for the epic login screen
+  // and firefox on windows for the rest of the pages
+  // this is to test if the epic login improves
+  let userAgent =
+    'Mozilla/5.0 (Windows NT 10.0; rv:78.0) Gecko/20100101 Firefox/85.0'
+  if (isEpicLogin) {
+    // firefox on linux for login screen
+    userAgent =
+      'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:85.0) Gecko/20100101 Firefox/85.0'
+  }
+
   useEffect(() => {
     let mounted = true
     const fetchLocalPreloadPath = async () => {
@@ -197,7 +208,7 @@ export default function WebView() {
         partition="persist:epicstore"
         src={startUrl}
         allowpopups={trueAsStr}
-        useragent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/200.0"
+        useragent={userAgent}
         {...(preloadPath ? { preload: preloadPath } : {})}
       />
       {showLoginWarningFor && (
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -78,7 +78,7 @@ export default function WebView() {
   if (isEpicLogin) {
     // firefox on linux for login screen
     userAgent =
-      'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:85.0) Gecko/20100101 Firefox/85.0'
+      'Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/113.0'
   }
 
   useEffect(() => {
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -197,6 +197,7 @@ export default function WebView() {
         partition="persist:epicstore"
         src={startUrl}
         allowpopups={trueAsStr}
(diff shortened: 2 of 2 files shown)


Arcitec:
I tried with this artifact: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/actions/runs/5073127293?pr=2726

The actions I did:

- Tried the latest Flathub version's login. Failed, as expected.
- Uninstalled the app and then installed the build artifact from this PR.
- Tried again. Sadly, captcha always fail. This is the most frustrating captcha ever invented.

image

- So I used Alternative Login, opened latest Linux Firefox (Flatpak) manually.
- Created a Private Window to be 100% sure I don't involve any weird cookies that would give me "more trust".
- I signed in. Didn't even get asked for a captcha here either, as explained in the previous thread.
- Voila.

Perhaps the user-agent of Electron still includes some junk that they detect (like perhaps an "Electron/X.Y.Z" identifier is part of the agent).

I don't think there's anything wrong in the Chrome/Chromium/Electron browser engine itself, because Chrome on Windows and Android work perfectly. So I don't think it's the browser engine itself that's the issue here.

Arcitec:
According to https://ipecho.net/extra my current Firefox version's user agent (the working browser) is:

```
Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/113.0
```

Arcitec:
> I will change, lets see if it makes any diffence

:( I tried the latest build. Still getting captcha failed every time.

We need to somehow verify that the user agent is being sent exactly like that. In past cases I have seen Electron throw in its own "Electron/X.Y.Z" into the final string.

Perhaps do a local test to make "Epic Login" go to "ipecho.net/extra" and ensure that the agent is 100% identical to this:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2726#issuecomment-1562032133

If that turns out to be the case, it would mean that they are detecting something else about the environment. Like for instance, is there some way to ensure that the Electron iframe is 100% sandboxed and cannot detect that it is running inside Electron?

I remember seeing that Electron has contexts for iframes where it's possible to set sandboxing attributes to ensure that the frame cannot see the host environment.

But of course... Chrome/Brave fails on Linux even when using the real browsers, with zero extensions, zero cookies, zero settings tweaks... So

Summary again since there's a lot of info here:

- IP: Ruled out. Works with other browsers on same IP.
- Cookies: Ruled out. The browsers that work also work in private tabs (no cookies).
- Extensions/browser tweaks: Ruled out. I have tried with blank browser profiles with zero tweaks.
- Works: Firefox on Linux. Chrome/Brave on Windows. Chrome on Android.
- Fails: Chrome/Brave on Linux. Heroic of course.
- Could the issue be Electron (Chromium)-related?: Maybe, but why does Chrome/Brave on Windows work then...
- The fact that Firefox works is the reason I think they are looking at User Agent popularity... Almost every Linux user uses it.

Arcitec:
Hmmm... I have some good progress.

I came up with my own test via "[User Agent Switcher](https://chrome.google.com/webstore/detail/user-agent-switcher-for-c/djflhoibgkdhkhhcedjiklpkjnoahfmg)" for Chrome. An official extension by Google.

- Firefox results:

```
<h3>Sent Info:</h3>
User-Agent | Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/113.0
Accept-Language | en-US,en;q=0.5
```

- Chrome headers:

```
User-Agent | Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/113.0
Accept-Language | en-US,en;q=0.9
```

I set the extension to spoof on every website. So every site thinks I am using Firefox, with that 100% identical user agent.

Used a private browser tab. It still gives me the captcha and fails it.

So that's fun. Perhaps Chromium/Electron engine is screwed up after all? But why does the same browser engine work on Windows...

I will try some other user agents, such as Chrome on Windows, to see if any of them work.

Arcitec:
## Okay we can stop trying. I have found the issue.

### Test 1: Linux

- I installed a fresh, latest version of Google Chrome. Zero addons or tweaks.
- My user agent (default for this latest version): `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/113.0.0.0 Safari/537.36`
- Trying captcha: It gives me captcha, fails, gives me another popup immediately, then fails the entire thing with the typical red "Incorrect response" banner at Epic.
- So I went to https://www.useragents.me/ to get the latest user agent for Windows.
- I used "[User Agent Switcher](https://chrome.google.com/webstore/detail/user-agent-switcher-for-c/djflhoibgkdhkhhcedjiklpkjnoahfmg)" to make my whole browser spoof the latest Chrome for Windows (note it's the same version as I use, but just for Windows instead): `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/113.0.0.0 Safari/537.36`
- Trying captcha: I get an infinite captcha loop. As soon as I finish a captcha the hCaptcha popup itself says "Please try again" in red text in the bottom right (at the "Verify" button), and immediately gives me another captcha. I completed about 20 captchas before giving up. It never gives me the Epic error. Instead it gives me a literal infinite loop.
- **That is our first clue...*

### Test 2: Windows

- Latest Chrome installed in a fresh Windows 11 Virtual Machine.
- Exact same IP, and mere 5 minutes after I had spammed 20 failed captchas in a row from this IP.
- Completely fresh install of Chrome. Zero changes or tweaks, Fresh profile.
- I use the default user agent, no tweaks (no extensions): `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/113.0.0.0 Safari/537.36`
- As you can see, that's the exact agent I was spoofing via Linux Chrome five minutes ago!
- Tried to login to Epic games.
- Instant login. No captcha. No errors. Just instantly welcome!
- **This is our second clue...*

### So what is going on:

- hCaptcha is detecting that we have the Linux version of Chrome/Chromium.
- It hates us.
- If we spoof the user agent, it still knows that we are actually the Linux (or Mac) version of Chrome/Chromium and hates us even more.
- There is some method they are using for detecting Linux/Mac browsers.
- Because despite being the EXACT same Chrome version, EXACT same fresh profiles, it STILL hates the Linux version of it but instantly allows the Windows version of it.
- But is hCaptcha really the bad guys here? Yes. It's not Epic's fault? No it's not.
- First test: Try the the hardest enterprise security here: https://nopecha.com/demo/hcaptcha
- Linux: Chrome with Linux OR Windows user agent - gets infinite captcha loop, even in incognito tabs without previous cookies.
- Windows: Chrome with default Windows user agent - succeeds after 2-3 captcha challenges (it presents 3 "please try again".

So the final answer is this: hCaptcha has some severe hatred against Linux and Mac users. It is able to detect non-Windows browsers REGARDLESS of User Agent. And it hates them.

We need to just complain to hCaptcha.

arielj:
I don't think they do something to detect linux, if that were the case nobody would be able to log in and I can login on linux with no issues, there must be something else going on.

I can log in with 3 different browsers on my Mint machine and also from within heroic.

It would be great if we could compare the requests done in the linux and the windows machines, to see if they are indeed the same or something is different

flavioislima:
I removed the user-agent flag now so it will use the default one, let's try to see if it improves or not.

Arcitec:
> I removed the user-agent flag now so it will use the default one, let's try to see if it improves or not.

Tried it now. It's been 16 hour since my intense testing, so any IP-based hatred would have cooled off by now. Went into Heroic's new build. Tried to login. Got the hardest captcha. It began to loop failures, but after 3 failures it stopped looping and simply gave me the beautiful, familiar screen:

image

I will remind any new readers: This is from the exact same computer and IP that *ALWAYS* gets *ZERO CAPTCHA, INSTANT LOGIN* whenever I use *ANY* web browser on Windows instead. Same IP. Same browser. Same fresh profile. Zero extensions. Zero tweaks.

hCaptcha hates Linux.

So yeah, [let's contact hCaptcha](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2674#issuecomment-1563637671).

slynobody:
(temporary?) solution: change webview-agent *as well* as main heroic browser agent to 'Mozilla/5.0 (Windows NT 10.0; WOW64; rv:70.0) Gecko/20100101 Firefox/70.0' 

to test without compiling: ctrl-shift-i, see current main agent: navigator.userAgent, change: network conditions; webview: elements / body / app / webview / Webview__webview: simply change to above)

would be great to have both the above (system-agnostic) as default though, esp for the planned side-loaded-apps-feature (to prevent similar errors on other side-loaded-apps / sites working with cloudflare, also with hcaptcha etc.).

PR #2734, opened 2023-05-27
[Feat] Add DLCs manager for Epic Games

flavioislima:
This PR adds the ability to:
- Select individual DLCs to be installed for Epic games, instead of installing all or nothing.
- Adds a new item on the Game Page SubMenu to open a DLC manager dialog where is possible to install and uninstall available DLCs.
- Works only for Epic for now, GOG keeps the current behavior until GogDL adds this feature as well.
- When selecting the DLCs on the install dialog, they will be added to the download Queue just after the main game.

image

image

image

image

---

### DIFF
--- a/src/backend/downloadmanager/downloadqueue.ts
+++ b/src/backend/downloadmanager/downloadqueue.ts
@@ -179,6 +179,12 @@ function getQueueInformation() {
 
 function cancelCurrentDownload({ removeDownloaded = false }) {
   if (currentElement) {
+    if (currentElement.params.installDlcs) {
+      const dlcsToRemove = currentElement.params.installDlcs as string[]
+      for (const dlc of dlcsToRemove) {
+        removeFromQueue(dlc)
+      }
+    }
     if (isRunning()) {
       stopCurrentDownload()
     }
--- a/src/frontend/components/UI/DLCList/DLC/index.tsx
+++ b/src/frontend/components/UI/DLCList/DLC/index.tsx
@@ -1,0 +1,60 @@
+import React, { useEffect, useState } from 'react'
+import { DLCInfo } from 'common/types/legendary'
+import './index.scss'
+import { useTranslation } from 'react-i18next'
+import { getGameInfo, install } from 'frontend/helpers'
+import { Runner } from 'common/types'
+import UninstallModal from 'frontend/components/UI/UninstallModal'
+
+type Props = {
+  dlc: DLCInfo
+  runner: Runner
+}
+
+const DLC = ({ dlc, runner }: Props) => {
+  const { title, app_name } = dlc
+  const { t } = useTranslation('gamepage')
+  const [isInstalled, setIsInstalled] = useState(false)
+  const [showUninstallModal, setShowUninstallModal] = useState(false)
+
+  useEffect(() => {
+    const checkInstalled = async () => {
+      const dlcInfo = await getGameInfo(app_name, runner)
+      console.table(dlcInfo)
+      const installed = dlcInfo.is_installed
+      setIsInstalled(installed)
+    }
+    checkInstalled()
+  }, [dlc, runner])
+
+  function mainAction() {
+    if (isInstalled) {
+      setShowUninstallModal(true)
+    } else {
+      console.log('installing')
+    }
+  }
+
+  return (
+    <>
+      {showUninstallModal && (
+        <UninstallModal
+          appName={app_name}
+          runner={runner}
+          onClose={() => setShowUninstallModal(false)}
+          isDlc
+        />
+      )}
+      <div className="dlcItem">
+        <span className="title">{title}</span>
+        <span className="action" onClick={() => mainAction()}>
+          {isInstalled
+            ? t('dlc.uninstall', 'Uninstall')
+            : t('dlc.install', 'Install')}
+        </span>
+      </div>
+    </>
+  )
+}
+
+export default React.memo(DLC)
--- a/src/frontend/components/UI/DLCList/DLC/index.tsx
+++ b/src/frontend/components/UI/DLCList/DLC/index.tsx
@@ -3,27 +3,31 @@ import { DLCInfo } from 'common/types/legendary'
 import './index.scss'
 import { useTranslation } from 'react-i18next'
 import { getGameInfo } from 'frontend/helpers'
-import { Runner } from 'common/types'
+import { GameInfo, Runner } from 'common/types'
 import UninstallModal from 'frontend/components/UI/UninstallModal'
 
 type Props = {
   dlc: DLCInfo
   runner: Runner
+  mainAppInfo: GameInfo
+  onClose: () => void
 }
 
-const DLC = ({ dlc, runner }: Props) => {
+const DLC = ({ dlc, runner, mainAppInfo, onClose }: Props) => {
   const { title, app_name } = dlc
   const { t } = useTranslation('gamepage')
   const [isInstalled, setIsInstalled] = useState(false)
   const [showUninstallModal, setShowUninstallModal] = useState(false)
+  const [dlcInfo, setDlcInfo] = useState<GameInfo | null>(null)
 
   useEffect(() => {
     const checkInstalled = async () => {
-      const dlcInfo = await getGameInfo(app_name, runner)
-      if (!dlcInfo) {
+      const info = await getGameInfo(app_name, runner)
+      if (!info) {
         return
       }
-      const installed = dlcInfo.is_installed
+      setDlcInfo(info)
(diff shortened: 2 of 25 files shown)


src/frontend/screens/Library/components/InstallModal/DownloadDialog/DLCDownloadListing.tsx
```diff
@@ -0,0 +1,84 @@
...
+    }
+    setDlcsToInstall(newDlcsToInstall)
+
+    if (newDlcsToInstall.length === DLCList.length) {
+      setInstallAllDlcs(true)
+    }
+
+    if (newDlcsToInstall.length < DLCList.length) {
+      setInstallAllDlcs(false)
+    }
```
CommandMC, line 51:
```suggestion
    setInstallAllDlcs(newDlcsToInstall.length === DLCList.length)
```

src/frontend/components/UI/DLCList/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { t } = useTranslation()
+
+  return (
+    <div className="dlcList">
+      <div className="dlcHeader">
+        <span className="title">{t('dlc.title', 'Title')}</span>
+        <span className="size">{t('dlc.size', 'Size')}</span>
+        <span className="actions">{t('dlc.actions', 'Actions')}</span>
+      </div>
+      {dlcs.length > 0 &&
```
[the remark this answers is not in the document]
flavioislima: So, at first it was only accessible if we have DLCs, but the problem is that there is a delay to check for the DLCs since we need to wait for the `legendary info` command to finish. So then the menu was hidden at first and then was appearing. to avoid this behavior I decided to show it for all games and check for the owned DLCs afterwards.

src/frontend/components/UI/DLCList/DLC/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+  runner: Runner
+  mainAppInfo: GameInfo
+  onClose: () => void
+}
+
+const DLC = ({ dlc, runner, mainAppInfo, onClose }: Props) => {
+  const { title, app_name } = dlc
+  const { libraryStatus, showDialogModal } = useContext(ContextProvider)
+  const { t } = useTranslation('gamepage')
+  const [isInstalled, setIsInstalled] = useState(false)
```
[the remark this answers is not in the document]
flavioislima: The problem is that the information about the install will only be available after the useEffect finishes it. But I will try.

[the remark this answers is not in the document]
flavioislima: I think worked fine, sent a fix.

src/frontend/components/UI/DLCList/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+const DlcList = ({ dlcs, runner, mainAppInfo, onClose }: Props) => {
+  const { t } = useTranslation()
+
+  return (
+    <div className="dlcList">
+      <div className="dlcHeader">
+        <span className="title">{t('dlc.title', 'Title')}</span>
+        <span className="size">{t('dlc.size', 'Size')}</span>
+        <span className="actions">{t('dlc.actions', 'Actions')}</span>
+      </div>
```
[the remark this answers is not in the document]
flavioislima: make sense, done

src/frontend/screens/Library/components/InstallModal/DownloadDialog/DLCDownloadListing.tsx
flavioislima: done

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -179,6 +179,12 @@ function getQueueInformation() {
 
 function cancelCurrentDownload({ removeDownloaded = false }) {
   if (currentElement) {
+    if (currentElement.params.installDlcs) {
+      const dlcsToRemove = currentElement.params.installDlcs as string[]
```
CommandMC, line 183: Not sure if we can currently get here with GOG games, but
```suggestion
    if (Array.isArray(currentElement.params.installDlcs)) {
      const dlcsToRemove = currentElement.params.installDlcs
```
might be safer

CommandMC (approved):

src/frontend/components/UI/DLCList/DLC/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+          isDlc
+        />
+      )}
+      <div className="dlcItem">
+        <span className="title">{title}</span>
+        {refreshing ? '...' : <span className="size">{size(dlcSize)}</span>}
+        {showInstallButton && (
+          <span className="action" onClick={() => mainAction()}>
+            <FontAwesomeIcon
+              icon={isInstalled ? faTrash : faCloudArrowDown}
```
arielj, line 113: I think we should use the same download icon we use in the library or the wine manager for consistency. Here it looks like this:

image

but in the library it's:

image

arielj (commented):
I'm testing installing the alien game with its only DLC and there's no indication in the downloads that it includes the DLC

image

I think it's because since it's all the DLCs it has and it's just passing the `--with-dlcs` flag then it's not adding the DLCs as extra downloads which makes sense to me but maybe we can add a `(with all DLCs)` after the name in these cases?

like: `Alien: Isolation (with all DLCs)`

src/frontend/screens/Game/GameSubMenu/index.tsx
```diff
@@ -213,6 +215,8 @@ export default function GamesSubmenu({
     return <CircularProgress className="link button is-text is-link" />
   }
 
+  const showDlcsItem = onShowDlcs && runner === 'legendary'
```
arielj, line 218: maybe we can also check if the game is installed here?

when I open the dialog for a game that is not installed the spinner never goes away and it never shows the install size

image

the dlc shows fine in the install dialog

arielj (commented):
one issue I noticed is that it's installing all DLCs if none is selected

I tried installing The Messenger without picking the `picnic` DLC and the install command has the `--with-dlcs` and then the when I open the DLCs dialog it shows as installed

arielj:
another issue I noticed (but this one shouldn't be a blocker) is that you can click a DLC in the completed list in the download manager and it takes you to the specific DLC's details page instead of the parent's page

for example, I installed the messenger DLC individually so I have the `picnic panick` entry in the download manager, I click that and it takes me to the picnic panic game details page but its `play` button doesn't work, then I go back and launch the base the messenger game and it works ootb

I think we should redirect to the DLC's parent if the game we are opening the GamePage is actually a DLC to avoid issue (I think many of the three dots menu options will fail or do weird stuff for DLCs -like moving the game-)

I think it's not a blocker (can be handled outside this PR) but would be nice if it's easy to fix

flavioislima:
@ariel, those issues with the messenger and alien isolation are for Epic games? GOG we still pass the flag but for epic games, we won't. In my tests here, Dying Light Enhanced edition, for instance, has 2 DLCs, if I select All DLCs it adds the 2 to the queue as expected.

But for GOG games, it will pass the flag to install all.

src/frontend/screens/Game/GameSubMenu/index.tsx
flavioislima: ah ok, makes sense. Will fix

flavioislima: image

src/frontend/components/UI/DLCList/DLC/index.tsx
flavioislima: image

flavioislima: fixes

flavioislima:
> another issue I noticed (but this one shouldn't be a blocker) is that you can click a DLC in the completed list in the download manager and it takes you to the specific DLC's details page instead of the parent's page
> 
> for example, I installed the messenger DLC individually so I have the `picnic panick` entry in the download manager, I click that and it takes me to the picnic panic game details page but its `play` button doesn't work, then I go back and launch the base the messenger game and it works ootb
> 
> I think we should redirect to the DLC's parent if the game we are opening the GamePage is actually a DLC to avoid issue (I think many of the three dots menu options will fail or do weird stuff for DLCs -like moving the game-)
> 
> I think it's not a blocker (can be handled outside this PR) but would be nice if it's easy to fix

For now I will disable the click on DLCs and also remove the action icons:
image

arielj (approved):
looking good! all my comments are addressed and the feature works correctly for me :+1:

PR #2739, opened 2023-05-29
[Feat] Add support for Sideloading Browser Apps and Games

flavioislima:
This PR adds the support to add a URL for a sideload App. 

Extending even more the capabilities of Heroic since now it can be used for several other purposes from launching Xcloud, Nvidia GeForceNow, or even Video Streaming services like Netflix, HBO Max, or YouTube. I am sure this will be useful to someone to make devices like the Steam Deck even more flexible.

By default, all browser apps will be launched in fullscreen. We can add a setting for it in the future if people need it. For now, let us keep like this.

image

image

image

---

### DIFF
--- a/src/backend/storeManagers/storeManagerCommon/games.ts
+++ b/src/backend/storeManagers/storeManagerCommon/games.ts
@@ -35,11 +35,14 @@ const openNewBrowserGameWindow = async (
   browserUrl: string,
   abortController: AbortController
 ): Promise<boolean> => {
+  const mainUrlName = browserUrl.split('://')[1].split('/')[0]
+
   return new Promise((res) => {
     const browserGame = new BrowserWindow({
       icon: icon,
       fullscreen: true,
       webPreferences: {
+        partition: `persist:${mainUrlName}`,
         webviewTag: true,
         contextIsolation: true,
         nodeIntegration: true
--- a/src/backend/storeManagers/storeManagerCommon/games.ts
+++ b/src/backend/storeManagers/storeManagerCommon/games.ts
@@ -35,14 +35,14 @@ const openNewBrowserGameWindow = async (
   browserUrl: string,
   abortController: AbortController
 ): Promise<boolean> => {
-  const mainUrlName = browserUrl.split('://')[1].split('/')[0]
+  const hostname = new URL(browserUrl).hostname
 
   return new Promise((res) => {
     const browserGame = new BrowserWindow({
       icon: icon,
       fullscreen: true,
       webPreferences: {
-        partition: `persist:${mainUrlName}`,
+        partition: `persist:${hostname}`,
         webviewTag: true,
         contextIsolation: true,
         nodeIntegration: true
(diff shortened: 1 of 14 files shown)


src/backend/storeManagers/storeManagerCommon/games.ts
```diff
@@ -33,30 +32,30 @@ export function logFileLocation(appName: string) {
 }
 
 const openNewBrowserGameWindow = async (
-  browserUrl: string
+  browserUrl: string,
+  abortController: AbortController
 ): Promise<boolean> => {
+  const mainUrlName = browserUrl.split('://')[1].split('/')[0]
```
imLinguin, line 38: How about using `URL()` constructor here instead of this nasty splitting

src/backend/storeManagers/storeManagerCommon/games.ts
```diff
@@ -33,30 +32,30 @@ export function logFileLocation(appName: string) {
...
-        encodeURIComponent(browserUrl)
-      : `file://${path.join(
-          buildDir,
-          './index.html?view=BrowserGame&browserUrl=' +
-            encodeURIComponent(browserUrl)
-        )}`
-
-    browserGame.loadURL(url)
+    browserGame.loadURL(browserUrl)
     setTimeout(() => browserGame.focus(), 200)
```
imLinguin, line 53: maybe instead of timeout use ready event to avoid race conditions

imLinguin (changes requested):
Looks good overall, that feature seems nice. However I'm missing a way to close the window properly, sure there is a AbortController like for games. I just don't see how one could close the game in gamescope session without killing whole process tree using Quit Game from Deck overlay. (Just a thought, killing the process isn't that bad after all)

Also left small comments for you to consider

src/backend/storeManagers/storeManagerCommon/games.ts
flavioislima: Cool, will check this out.

src/backend/storeManagers/storeManagerCommon/games.ts
flavioislima: sounds good

flavioislima:
> Looks good overall, that feature seems nice. However I'm missing a way to close the window properly, sure there is a AbortController like for games. I just don't see how one could close the game in gamescope session without killing whole process tree using Quit Game from Deck overlay. (Just a thought, killing the process isn't that bad after all)
> 
>

Have you tried?
I will test on the Steam deck later but based on how HyperPlay is behaving, it works just fine there.

imLinguin:
That was just a thought I had. I'll test the feature later

imLinguin (approved):

slynobody:
thank you! maybe just a little early to ask: is each webapp isolated from each other and/or could you give them / alter specific user-data-dirs? (user facing options would also be great to alter eachs specific ://flags/ as well as cookie / cache-deletion policy)

flavioislima:
> thank you! maybe just a little early to ask: is each webapp isolated from each other and/or could you give them / alter specific user-data-dirs? (user facing options would also be great to alter eachs specific ://flags/ as well as cookie / cache-deletion policy)

For now they are isolated from each other. I think it's safer this way.

imLinguin:
It's worth noting they are isolated based on the domain of provided URL

PR #2742, opened 2023-05-30
Fix GOG download size in download manager

arielj:
When adding GOG games to the download queue, the game size shows up as `??`.

The issue is that we were passing `Windows` instead of `windows` as the platform.

---

### DIFF
--- a/src/backend/storeManagers/gog/library.ts
+++ b/src/backend/storeManagers/gog/library.ts
@@ -16,7 +16,6 @@ import {
   GogInstallInfo,
   GOGGameDotIdFile,
   GOGClientsResponse,
-  GogInstallPlatform,
   GamesDBData,
   Library,
   BuildItem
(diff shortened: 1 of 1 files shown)


imLinguin:
Same thing is also happening for Linux native (not sure if that's intended), which is probably related to not existing linux platform in gogdl info command.
I'm unable to test Mac too

src/backend/storeManagers/gog/library.ts
```diff
@@ -377,6 +377,8 @@ export async function getInstallInfo(
   installPlatform = 'windows',
   lang = 'en-US'
 ): Promise<GogInstallInfo | undefined> {
+  installPlatform = installPlatform.toLowerCase()
```
flavioislima, line 380:
```suggestion
  installPlatform = installPlatform === 'Mac' ? 'osx' : installPlatform.toLowerCase()
```

this fixes mac as well

flavioislima (approved):

PR #2746, opened 2023-05-30
[Feat] Use Legendary API to get SDL List (Fix Fortnite install and more)

flavioislima:
For a long time Rodney from Legendary had this API that we could consult the list of Selective downloads, fixes, and more. But we were just hardcoding the list instead of getting from there and every time it changed we needed to update on our end as well.

This is fixed here, now, when opening the install modal for an Epic game we check on the API if there is any SDL available for it and list it right away. So if Rodney updates on the API it gets updated automatically on Heroic as well.

image

image

Other Small fixes:
- Fix the Install dialog showing a Browser Icon and install platform
- Fix error dialog showing for shared memory errors (since they are handled already)

---


derrod:
As a note, `https://heroic.legendary.gl/v1/version.json` has a `game_overrides.sdl_config.<app name>` key that indicates whether or not an SDL file exists and what version it is. Legendary itself uses this for caching SDL files and to avoid unnecessary requests to see whether or not one exists. So you could, for example, fetch the version information once on startup and then use that to determine if an SDL file needs to be downloaded without doing a request every time.

src/backend/storeManagers/legendary/library.ts
```diff
@@ -634,3 +637,42 @@ export async function runRunnerCommand(
     }
   )
 }
+
+export async function getLegendaryApi(): Promise<ResponseDataLegendaryAPI> {
+  const response = await axios.get<ResponseDataLegendaryAPI>(
```
arielj, line 642: maybe we should wrap this in a try/catch? if for some reason the request fails this will throw an exception

src/backend/storeManagers/legendary/library.ts
```diff
@@ -634,3 +637,42 @@ export async function runRunnerCommand(
...
+    'https://heroic.legendary.gl/v1/version.json'
+  )
+  return response.data
+}
+
+export async function getGameSdl(
+  appName: string
+): Promise<SelectiveDownload[]> {
+  // get data from https://heroic.legendary.gl/v1/sdl/appName.json using axios
+  const response = await axios.get<Record<string, SelectiveDownload>>(
```
arielj, line 652: same here with the try/catch

flavioislima:
> As a note, `https://heroic.legendary.gl/v1/version.json` has a `game_overrides.sdl_config.<app name>` key that indicates whether or not an SDL file exists and what version it is. Legendary itself uses this for caching SDL files and to avoid unnecessary requests to see whether or not one exists. So you could, for example, fetch the version information once on startup and then use that to determine if an SDL file needs to be downloaded without doing a request every time.

Yes, I will add some caching to it.
We are just requesting it when the person clicks the install button. But I think I can cache it and setup a timer to expire.

Issue #2730, opened 2023-05-26
Add an option to see download speed in megabits instead of megabytes


sTiKyt:
### Problem description

It's very frustrating to see megabyte speeds while the rest of my environment uses megabits, it's confusing

### Feature description

Add an option to see download speed in megabits instead of megabytes

ram1660:
I would like to take this issue.

PR #2749, opened 2023-05-30
Download and disk click toggle for megabyte and megabit.

ram1660:
In this issue #2730 a user asked to add the ability to view download and disk speeds as Megabits because of his computer is in Megabits format.
I implemented a click event on the disk speeds and download speeds. Each click will change the values from Megabytes to Megabits.

---

- [x] Tested the feature, and it's working on a current and clean install.
- [x] Tested the main App features, and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)

### DIFF
--- a/src/frontend/screens/DownloadManager/components/ProgressHeader/index.tsx
+++ b/src/frontend/screens/DownloadManager/components/ProgressHeader/index.tsx
@@ -11,9 +11,14 @@ interface Point {
   disk: number
 }
 
-const roundToNearestHundredth = function (val: number | undefined) {
+const roundToNearestHundredthByte = function (val: number | undefined) {
   if (!val) return 0
   return Math.round(val * 100) / 100
+}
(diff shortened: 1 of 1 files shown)


arielj:
I think it would be better to add a checkbox to make it more obvious, it will be hard for people to actually discover that's a thing you can do. Also a checkbox is more accessible than clicking a div.

This is not getting persisted, right? I think users that want this will probably be annoyed to have to change this back every time they open the downloads screen.

Finally, there will be a disconnection between the download screen and other places displaying these units, we would have to change this everywhere.

Maybe a setting will be better, and then we can use that everywhere and also it's persisted.

arielj:
hey @ram1660 are you still working on this?

I'll close the PR since it's been stale for many months, if you finish the changes it can be reopened, just doing some cleanup

Issue #2721, opened 2023-05-20
Clicking "Download games without HTTPS" does nothing


wjdp:
### Describe the bug

Clicking "Download games without HTTPS" does nothing

![image](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/1690934/ff3b684e-d361-462f-bf3b-876aadc92c3c)

### Add logs

```text
(22:35:15) INFO:    [Legendary]:        Refreshing library...
(22:35:15) INFO:    [Legendary]:        Refreshing Epic Games...
(22:35:15) INFO:    [Connection]:       Connectivity: check-online
(22:35:15) INFO:    [Connection]:       Pinging external endpoints
(22:35:15) INFO:    [Legendary]:        Running command: C:\Users\Will\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary --version
(22:35:15) INFO:    [Legendary]:        Legendary location: C:\Users\Will\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary
(22:35:15) INFO:    [Gog]:              GOGDL location: C:\Users\Will\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\gogdl
(22:35:15) INFO:    [Gog]:              GOGDL location: C:\Users\Will\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\gogdl
(22:35:15) INFO:    [Legendary]:        Game list updated, got 143 games & DLCs
(22:35:15) WARNING: [Backend]:          Protocol already registered.
(22:35:15) INFO:    [Connection]:       Connectivity: online
(22:35:15) INFO:    [Gog]:              Running command: C:\Users\Will\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\gogdl --auth-config-path C:\Users\Will\AppData\Roaming\heroic\gog_store\auth.json --version
(22:35:15) INFO:    [Backend]:          Loading Screen Ready
(22:35:15) INFO:    [Frontend]:         Refreshing Library
(22:35:15) WARNING: [Backend]:          listUpdateableGames not implemented on Sideload Library Manager
(22:35:15) INFO:    [Gog]:              Found 0 game(s) to update
(22:35:15) INFO:    [Backend]:          Frontend Ready
(22:35:16) INFO:    [Backend]:          Checking for new Heroic Updates
[...]

PR #2755, opened 2023-06-02
[Bug] Fix "no https" toggle not working

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2721

The onClick handler was always setting this to `false`

---


Nocccer (approved):
Good catch

imLinguin:
Not sure if that applies here, but maybe we should make it clear this feature works for Epic only. GOG CDNs redirect to https automatically.

Issue #2594, opened 2023-04-03
Remove focus after game ends


JakubFojtik:
### Feature description

Can you please remove the launcher window focus after closing a game? Or tell me where in source i could toggle it myself?
Is it intentional or a side effect?

theli-ua:
It is indeed annoying for heroic to pop up from tray after you close the game

Zorrototo:
Please an update on that, is it intended? Can you add a setting to disable that behavior if it is intended?

arielj:
A fix for this will be included in the next release

PR #2758, opened 2023-06-02
[UX] Do not force focus on main window after closing game

arielj:
This fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2594

We were forcing the focus on the main window after a game was closed which was not great UX.

The are many scenarios when starting a game:
- game was started with no GUI (in this case the autofocus was already ignored)
- game was started from the try (in this case the autofocus was already ignored)
- game was started from the gui (the windows is back on focus after closing a game anyway since it was the last window with focus, even after removing these lines)
- main window is minimized/close while a game is loading (in this cases the window was forced to gain focus after close, even if the user explicitly move it out of the way, not it's not)
- sometimes a game takes a long time to actually end after exiting, if you move to another window to do something else then heroic would steal the focus after an undefined amount of time

With the current change we don't force the focus, if the window is the first screen after closing a game it still regains focus automatically anyway.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -958,7 +958,6 @@ ipcMain.handle(
     })
 
     const mainWindow = getMainWindow()
-    const showAfterClose = mainWindow?.isVisible()
     if (minimizeOnLaunch) {
       mainWindow?.hide()
     }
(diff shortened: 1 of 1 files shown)


imLinguin (approved):
It annoyed me a bit too

PR #2759, opened 2023-06-02
[GOG]: Switch to Galaxy way of loading library

imLinguin:
This is a test of trying to replicate how GOG Galaxy loads the library.  

We achieve this using galaxy-library endpoint which contains games from all connected integrations (updated by client), and GOG (updated internally). We filter out all the gog games and use those.  
If we decided to go hardcore we could maybe support those integrations in the future in some way.   

Of course this can be further extended in terms of caching and deciding whether to fetch new data for games.

One proposition is to verify if we have up to date data when opening the game page.  
~~For now I'm marking it as wip (library pagination is still to be done)~~

Additionally the galaxy-library endpoint provides us with certificates that somehow in pair with GOG access token allow us to get data from gamesdb that would return 404 otherwise.  
Additionally I reduced IO traffic when pulling the library thanks to allowing to temporarily offload the CacheStore to memory, and then sync it to drive after all operations.  

This PR also adds new field `installable` to GameInfo. Preventing Cannot get game info errors for goodie packs.

### DIFF
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -133,7 +133,7 @@
     },
     "label": {
         "game": {
-            "not-installable-game": "",
+            "not-installable-game": "Game is NOT Installable",
             "third-party-game": "Third-Party Game NOT Supported"
         },
         "launching": "Launching",
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -133,6 +133,7 @@
     },
     "label": {
         "game": {
+            "not-installable-game": "",
             "third-party-game": "Third-Party Game NOT Supported"
         },
         "launching": "Launching",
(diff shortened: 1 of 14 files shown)


imLinguin:
This PR also removes duplicated `getGamesdbData` function from `wiki_game_info` utils

imLinguin:
New `not installable` feature:

image

image
image

flavioislima (commented):
Looks cool, will do some testing

flavioislima (approved):
Tested here and everything looks good:

- [x] List games
- [x] Install
- [x] Launch
- [x] Uninstall

During the cache update, I got some missing/wrong information but then it updated it automatically after a few seconds. I think that is expected.

flavioislima:
@imLinguin I think this one is fine to merge if you want to.

Issue #2575, opened 2023-03-26
Search suggestions/autocomplete shows DLCs but then picking them shows no results


arielj:
### Describe the bug

When using the search, a dropdown with suggestions/autocomplete appears. The results include DLCs. When selecting a DLC, since DLCs are not actually displayed in the library, the library shows no results.

![image](https://user-images.githubusercontent.com/188464/227811318-fdaca678-3b67-4872-ada0-92e5f5e599e0.png)

![image](https://user-images.githubusercontent.com/188464/227811332-eca29bfe-728c-4bd2-8cfb-1fe5483af18d.png)

### Add logs

```text
Nothing relevant
```

### Steps to reproduce

1. Search for something that could be part of a DLC name
2. Select the DLC in the suggestions

### Expected behavior

DLCs should not show up as suggestions.

### Heroic Version

Latest Stable

### System Information

Not relevant

amanx123:
Would like to work on this. Assign me please

arielj:
hey @amanx123, any update on this? if you can't work on this I can unassign you in case somebody else wants to take this

arielj:
@amanx123 I included a fix for this in a PR with other fixes too, so no need to work on it anymore

arielj:
This got fixed in 2.9.1, DLCs don't show up as suggestions

Issue #2756, opened 2023-06-02
Error in backend after adding sideloaded app


arielj:
### Describe the bug

After adding an app with Add Game, this errors show up in the logs related to extra game info and shortcuts

### Add logs

```text
(00:47:56) INFO:    [Backend]:          Adding shortcuts for undefined
(00:47:56) ERROR:   [ExtraGameInfo]:    Was not able to get ExtraGameInfo data for undefined TypeError: Cannot read properties of undefined (reading 'replaceAll')
    at removeSpecialcharacters (/home/ariel/dev/oss/HeroicGamesLauncher/build/electron/main.c42fbcef.js:7981:15)
    at getWikiGameInfo (/home/ariel/dev/oss/HeroicGamesLauncher/build/electron/main.c42fbcef.js:2555:13)
    at addNonSteamGame (/home/ariel/dev/oss/HeroicGamesLauncher/build/electron/main.c42fbcef.js:2691:26)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
(00:47:56) WARNING: [Shortcuts]:        Couldn't find a icon for undefined with: TypeError: Cannot read properties of undefined (reading 'replaceAll')
    at getIcon$1 (/home/ariel/dev/oss/HeroicGamesLauncher/build/electron/main.c42fbcef.js:2139:35)
    at addNonSteamGame (/home/ariel/dev/oss/HeroicGamesLauncher/build/electron/main.c42fbcef.js:2744:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
(00:47:56) INFO:    [Shortcuts]:        Prepare Steam images for undefined
(node:219585) UnhandledPromiseRejectionWarning: Error: Input must be string
    at sanitize (/home/ariel/dev/oss/HeroicGamesLauncher/node_modules/sanitize-filename/index.js:41:11)
    at module.exports [as default] (/home/ariel/dev/oss/HeroicGamesLauncher/node_modules/sanitize-filename/index.js:54:16)
    at shortcutFiles (/home/ariel/dev/oss/HeroicGamesLauncher/build/electron/main.c42fbcef.js:3020:40)
    at addShortcuts$3 (/home/ariel/dev/oss/HeroicGamesLauncher/build/electron/main.c42fbcef.js:2935:35)
[...]

PR #2771, opened 2023-06-03
[UX/UI] Some fixes in search autocomplete

arielj:
This PR includes a fix for https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2575 and some other fixes I noticed while fixing that:

- if we had the same game in 2 stores, there was no way to select which one to open and Epic had priority (this was not an issue when clicking only changed the results), I added a `(GOG)` or `(Epic)` or `(sideload)` suffix
- we were ordering and then filtering, which is slower than first filtering and then ordering since then we need to order less elements
- we were converting the list of games to a list of titles to render it and then on click we were fetching the game again, now it just uses the list of games to render it and passes the game as an argument on click to avoid that extra work
- since now we only have unique games, no need to convert everything into a set and then back to an array, so now it's all a single array
- it now ignores DLCs from the search result, which created issues (before it was producing the error in the issue linked, after the element changed to open the game page it was opening the DLC page which doesn't really work)

image

---


flavioislima:
Can you add a small fix, probably one line, to the shortcuts creator? I think Dlcs after being installed are cresting shortcuts and being added to steam and desktop if these options are on on the settings.

So just need to check if it's a dlc before adding the shortcuts. 👍

arielj:
> Can you add a small fix, probably one line, to the shortcuts creator? I think Dlcs after being installed are cresting shortcuts and being added to steam and desktop if these options are on on the settings.
> 
> So just need to check if it's a dlc before adding the shortcuts. +1

I'll do that in a different PR, there's also an error related to shortcuts for sideloaded apps and that change will make more sense there https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2756

src/frontend/components/UI/SearchBar/index.tsx
```diff
@@ -30,19 +35,19 @@ export default React.memo(function SearchBar() {
   const list = useMemo(() => {
     // Set can't handle spread of undefined. Leading to
```
flavioislima, line 36: I think you can remove these comments now.

flavioislima (approved):
LGTM

src/frontend/components/UI/SearchBar/index.tsx
arielj: done!

arielj:
hey @flavioislima I can't merge this, I think it's because I skipped CI in the last commit (because it was just removing a comment), maybe you can merge since you are the owner

flavioislima:
> hey @flavioislima I can't merge this, I think it's because I skipped CI in the last commit (because it was just removing a comment), maybe you can merge since you are the owner

Done

Nocccer:
> hey @flavioislima I can't merge this, I think it's because I skipped CI in the last commit (because it was just removing a comment), maybe you can merge since you are the owner

FYI: You can retrigger Checks INSIDE the pr unser Checks or you use the vscode GitHub Action extension

Issue #2733, opened 2023-05-26
Ability to disable logging globally


salihmarangoz:
### Problem description

My logs folder became larger than 10GB.

### Feature description

A checkbox to disable saving logs to the disk, globally.

### Alternatives

Checking log folder size and rotating may be an option.

PR #2773, opened 2023-06-03
[Logs] New option to disable logs

arielj:
This should fix https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2733

- added an option to `Disable logs`. I put the option under Advanced so users are more careful when turning this toggle ON.
- added a new `forceLog` option when calling the log functions that will ignore this setting
- in many places we are writing/appending directly to a log file, added a check around those (but ideally we should move some of that into `log*` calls)
- some things are always logged just in case
- added a warning message in the logs when logs are disabled, so if somebody share a weird looking log we can see if the reason is logs being disabled

The benefit of this is to avoid 2 things:
- writing to disk can be resource intensive, specially on slow disks and for games with a lot of output
- even for games with not much output, if playing for long periods of time the logs can grow really big

---


Nocccer:
Just an Idea for another pr. We could add a selektive Play Button where people can choose `play` and `play (Logs)`.

flavioislima (approved):
LGTM

salihmarangoz:
Thanks!

Issue #2117, opened 2022-12-01
[UI] Adjust the UI to Right-to-left Languages.


Ali-x98:
### Describe the bug

When using right-to-left languages (e.g., Arabic, Persian, Hebrew), the UI should be mirrored to the right side of the screen, along with all texts and UI elements.

There was an attempt to fix this in #868, but the UI is still not completely mirorred.

Comparing English and Persian (since Arabic is still not in the selection even though it's 100% complete):
![image](https://user-images.githubusercontent.com/25479373/205084994-84950969-ab5e-4d3e-84ca-4689c2627272.png)
![image](https://user-images.githubusercontent.com/25479373/205084304-69b25aed-d69c-425a-b2da-cc1fbf6ff16f.png)

The side bar should be on the right, and the text inside should be written right-to-left.

This image from [Apple](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/SupportingRight-To-LeftLanguages/SupportingRight-To-LeftLanguages.html) demonstrates this well:
![image](https://user-images.githubusercontent.com/25479373/205087969-55a3b61d-fd99-43e8-86cd-86829bfdb2fb.png)

### Add logs

```text
-
```

### Steps to reproduce

1. Go to Setting
2. Click on the 'Choose App Language' box
3. Select any RTL language (I think Persian is currently the only one available).

### Expected behavior

UI should be mirrored to the right side.

### System Information

- OS [Fedora 37]: 
- Heroic Version [2.5.0]:

flavioislima:
Good, we will look into it. And also add Arabic to the list.

PR #2775, opened 2023-06-03
[UI/UX] Adapt layout properly to RTL languages

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2117

Apart from the sidebar that was in the wrong side, some screens like the game page were totally broken and were unusable (the 3 dots menu opened outside the viewport for example).

I replaced almost all used of physical properties with logical properties:
- `text-align: left/right` with `start/end`
- divided the `margin` shorthand into `margin-block` and `margin-inline` when needed, and simplified to 3 values when not needed
- same with the `padding` shorthand
- replaced the individual `padding-left/right` and `margin-left/right` with `padding-inline-start/end` and `margin-inline-start-end`
- moved some elements to the opposite side when needed (back button, store logo, some text, etc)
- allow resizing sidebar when it's on the right

replacing physical properties with logical ones should generate no changes in the LTR languages, I tried to check most of the changes manually but there were too many.

We should start defaulting to logical properties always and most things will "just work".

I won't put screenshots for every change because this affects almost all pages I think.

image

image

image

I imagine there are still places where the direction is not correct but I don't know that much about RTL languages.

I also think there might be more languages that are RTL but I don't to change that without confirmation of users using that language: https://lingohub.com/academy/best-practices/ltr-languages-list (malaysian, hungarian, croatian, and more show up in that list)

---

### DIFF
--- a/src/frontend/App.css
+++ b/src/frontend/App.css
@@ -125,7 +125,7 @@ svg.disabled {
   line-height: var(--space-lg);
   font-family: var(--secondary-font-family);
   cursor: pointer;
-  text-align: left;
+  text-align: start;
 }
 
 .react-contextmenu-item:hover {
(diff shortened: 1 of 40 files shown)


flavioislima (approved):
Cool, that looks like a great improvement for these languages since we only had it for the settings screen 👍🏽

flavioislima (approved):
Cool, that looks like a great improvement for these languages since we only had it for the settings screen 👍🏽

PR #2778, opened 2023-06-04
[BUG] Don't add shortcut for DLCs. Fix adding sideload app shortcuts

arielj:
This PR fixes two issues:

- Heroic was adding shortcuts for DLCs, but in general launching a DLC directly wouldn't work.
- Shortcuts creation/deletion failed for sideloaded apps

This fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2756

---

### DIFF
--- a/src/backend/storeManagers/sideload/games.ts
+++ b/src/backend/storeManagers/sideload/games.ts
@@ -104,10 +104,11 @@ export async function uninstall({
   const old = libraryStore.get('games', [])
   const current = old.filter((a: GameInfo) => a.app_name !== appName)
 
+  const gameInfo = getGameInfo(appName)
   const {
     title,
     install: { executable }
-  } = getGameInfo(appName)
+  } = gameInfo
   const { winePrefix } = await getSettings(appName)
 
   if (shouldRemovePrefix) {
@@ -125,7 +126,7 @@ export async function uninstall({
 
   notify({ title, body: i18next.t('notify.uninstalled') })
 
-  removeShortcuts(appName)
+  removeShortcutsUtil(gameInfo)
 
   sendFrontendMessage('gameStatusUpdate', {
     appName,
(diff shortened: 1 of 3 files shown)


src/backend/storeManagers/sideload/games.ts
```diff
@@ -125,7 +126,7 @@ export async function uninstall({
 
   notify({ title, body: i18next.t('notify.uninstalled') })
 
-  removeShortcuts(appName)
+  removeShortcutsUtil(gameInfo)
```
arielj, line 129: we already have the gameInfo here, no need to call the function that looks for the gameInfo

src/backend/storeManagers/sideload/library.ts
```diff
@@ -59,7 +60,7 @@ export function addNewApp({
     current[gameIndex] = { ...current[gameIndex], ...game }
   } else {
     current.push(game)
-    addShortcuts(app_name)
+    addShortcuts(game)
```
arielj, line 63: we already have the gameInfo here, no need to call the function that looks for the gameInfo

src/backend/storeManagers/sideload/library.ts
```diff
@@ -27,7 +27,8 @@ export function addNewApp({
     title,
     install: {
       executable,
-      platform
+      platform,
+      is_dlc: false
```
arielj, line 31: mark it as not DLC just to be explicit since now we check if `is_dlc` is true to ignore the shortcuts

imLinguin (approved):
Looks good

flavioislima (approved):
Looks good, thanks for fixing that 👍🏽

flavioislima (approved):
Looks good, thanks for fixing that 👍🏽

PR #2792, opened 2023-06-11
[FIX] Switch to sniper runtime

Etaash-mathamsetty:
People who recently installed steam on Linux will have only the sniper runtime installed, therefore we should switch to it by default. 


redromnon:
Maybe another alternative is to detect if the sniper runtime is installed first and then switch to that?

For instance, my system still has the older scout runtime since I didn't use Steam for quite some time. So launching a Linux native game from Heroic will fail to launch if I enable the checkbox unless I download the runtime from Steam. But I'm pretty sure many other users may not have the same case as me.

Another way is to display a dialog asking the user to install sniper runtime in Steam.

Etaash-mathamsetty:
> Maybe another alternative is to detect if the sniper runtime is installed first and then switch to that?
> 
> For instance, my system still has the older scout runtime since I didn't use Steam for quite some time. So launching a Linux native game from Heroic will fail to launch if I enable the checkbox unless I download the runtime from Steam. But I'm pretty sure many other users may not have the same case as me.
> 
> Another way is to display a dialog asking the user to install sniper runtime in Steam.

Ok seems fair

flavioislima:
The Sniper runtime only works with version 8?
Perhaps we should have a method to get all available runtimes like we do for the Wine versions.

Then, depending on the Proton version we select the correct runtime. If it does not exists, simply ignore the runtime setting.

Etaash-mathamsetty:
> The Sniper runtime only works with version 8? Perhaps we should have a method to get all available runtimes like we do for the Wine versions.
> 
> Then, depending on the Proton version we select the correct runtime. If it does not exists, simply ignore the runtime setting.

that's going to be very hard how are we supposed to detect if a certain version of proton is proton 8? And some proton-tkg builds use soldier even on the 8.x series

imLinguin:
Doesn't tool manifest indicate that?

src/backend/launcher.ts
```diff
@@ -121,16 +124,32 @@ async function prepareLaunch(
     gameSettings.useSteamRuntime &&
     (isNative || gameSettings.wineVersion.type === 'proton')
   if (shouldUseRuntime) {
+    let nonNativeRuntime: SteamRuntime['type'] = 'soldier'
+    if (!isNative) {
+      try {
+        const parentPath = normalize(join(gameSettings.wineVersion.bin, '..'))
+        const requiredAppId = VDF.parse(
+          readFileSync(join(parentPath, 'toolmanifest.vdf'), 'utf-8')
+        )['manifest']['require_tool_appid']
```
imLinguin, line 133: Why dont just use dot syntax here
```suggestion
        ).manifest?.require_tool_appid
```

imLinguin (approved):
Looks good, one small syntax thing to consider

src/backend/launcher.ts
```diff
@@ -121,16 +124,32 @@ async function prepareLaunch(
...
+        if (requiredAppId === 1628350) nonNativeRuntime = 'sniper'
+      } catch (error) {
+        logError('Failed to parse toolmanifest.vdf', LogPrefix.Backend)
```
CommandMC, line 136: Logging the error if something goes wrong is generally a good idea. E.g.:
```suggestion
        logError(['Failed to parse toolmanifest.vdf:', error], LogPrefix.Backend)
```

src/backend/launcher.ts
```diff
@@ -121,16 +124,32 @@ async function prepareLaunch(
...
       return {
         success: false,
         failureReason:
           'Steam Runtime is enabled, but no runtimes could be found\n' +
           `Make sure Steam ${
-            isNative ? 'is' : 'and the "SteamLinuxRuntime - Soldier" are'
+            isNative
+              ? 'is'
+              : `and the SteamLinuxRuntime - ${
+                  nonNativeRuntime === 'sniper' ? 'Sniper' : 'Solider'
```
CommandMC, line 151:
```suggestion
                  nonNativeRuntime === 'sniper' ? 'Sniper' : 'Soldier'
```

src/backend/utils.ts
```diff
@@ -633,10 +633,15 @@ async function searchForExecutableOnPath(executable: string): Promise<string> {
   }
 }
 async function getSteamRuntime(
-  requestedType: 'scout' | 'soldier'
+  requestedType: 'scout' | 'sniper' | 'soldier'
```
CommandMC, line 636: While you're touching this, might as well make this a little easier to work with:
```suggestion
  requestedType: SteamRuntime['type']
```

src/backend/launcher.ts
```diff
@@ -121,16 +124,32 @@ async function prepareLaunch(
     gameSettings.useSteamRuntime &&
     (isNative || gameSettings.wineVersion.type === 'proton')
   if (shouldUseRuntime) {
+    let nonNativeRuntime: SteamRuntime['type'] = 'soldier'
+    if (!isNative) {
+      try {
```
CommandMC, line 129: A comment explaining what's being done here (and why it's done) might be helpful to future readers

PR #2802, opened 2023-06-15
[Fix] Cannot read properties of undefined (reading 'is_dlc')]

Nocccer:
This PR makes sure that the frontend does not throw if something acces undefined object which contains `is_dlc`.
Every backend `getGameInfo` returns know an invalid `GameInfo` instead `null`.

---


arielj:
should we change the type of `.install` to be possibly undefined? because if this can be undefined it would be easy to run into this issue again in the future if typescript does not warn us about it and we forget

Nocccer:
> should we change the type of `.install` to be possibly undefined? because if this can be undefined it would be easy to run into this issue again in the future if typescript does not warn us about it and we forget

I tried another approach. Maybe that fixes it aswell. If you think the undefined way is better, let me know.

flavioislima (approved):
LGTM

PR #2803, opened 2023-06-16
Set Legendary's config path to a Heroic-specific location

CommandMC:
Legendary's configuration will now be stored in a Heroic-specific location. This avoids issues where users change their `config.ini` and in turn change Heroic's behavior.
Existing configuration data will be migrated automatically, meaning anyone testing this PR *should* not notice anything out of the ordinary.

The original issue I've aimed to solve here was reported on our Discord ([here](https://discord.com/channels/812703221789097985/1093506682673446983)): When launching a game, we used to run `legendary launch <AppName> --json --offline` to get Legendary's original launch configuration for the game. This is problematic on macOS systems without CrossOver installed, as Legendary will crash when trying to auto-detect a version.
With us now controlling the config file, we simply don't have to check for user modifications at all.

---

### DIFF
--- a/src/backend/constants.ts
+++ b/src/backend/constants.ts
@@ -39,10 +39,8 @@ const currentGlobalConfigVersion: GlobalConfigVersion = 'v0'
 const flatPakHome = env.XDG_DATA_HOME?.replace('/data', '') || homedir()
 const userHome = homedir()
 const configFolder = app.getPath('appData')
-const legendaryConfigPath = isLinux
-  ? join(configFolder, 'legendary')
-  : join(userHome, '.config', 'legendary')
 const appFolder = join(configFolder, 'heroic')
+const legendaryConfigPath = join(appFolder, 'legendaryConfig', 'legendary')
 const configPath = join(appFolder, 'config.json')
 const gamesConfigPath = join(appFolder, 'GamesConfig')
 const toolsPath = join(appFolder, 'tools')
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -858,31 +858,6 @@ export async function launch(
     )
   }
 
-  // Log any launch information configured in Legendary's config.ini
-  const { stdout } = await runLegendaryCommand(
-    ['launch', appName, '--json', '--offline'],
-    createAbortController(appName)
-  )
-
-  appendFileSync(
-    logFileLocation(appName),
-    "Legendary's config from config.ini (before App's settings):\n"
-  )
-
-  try {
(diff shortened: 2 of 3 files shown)


arielj:
I wonder if we want to do this or not, because there are some settings that can be configured in Legendary's `.ini` file that we can't set in Heroic (like a pre-launch script) and maybe users are using that and if we duplicate this configure somewhere else this will be confusing if they go and try to edit this (they'll try to edit the original files without knowing we copied them inside heroic's settings)

Maybe it's safer to just catch that error so it's not showed to the user and skip the config logging?

That way we remove the error for the user without duplicating Legendary's settings.

It sounds like it's a bug in Legendary that it fails in those cases and it would go away eventually when fixed, and we can ignore it since it's just something we run to have more logs.

flavioislima:
> I wonder if we want to do this or not, because there are some settings that can be configured in Legendary's `.ini` file that we can't set in Heroic (like a pre-launch script, max memory, preferred cdn) and maybe some users are using that and if we duplicate this configure somewhere else this will be confusing if they go and try to edit things (they'll edit the original file without knowing we copied them inside heroic's settings).
> 
> Maybe it's safer to just catch that error so it's not showed to the user and skip the config logging?
> 
> That way we remove the error for the user without duplicating Legendary's settings.
> 
> It sounds like it's a bug in Legendary that it fails in those cases and it would go away eventually when fixed, and we can ignore it since it's just something we run to have more logs.

I believe we should do this change because it not only fixes the issue on Mac but also makes the launching faster.
I really don't think it is common for Heroic users to use Legendary-specific configurations. Also, the features you mentioned, like running a script before the game, it is on our Roadmap and I want to work on it soon. 

Rodney actually developed this feature of using a specific configuration because I asked to a long time ago since it was causing issues with Heroic. This also makes the usage of Legendary within Heroic more predictable since we know exactly the configuration we will use instead of dealing with some modified config in Legendary that can conflict with another config Heroic passes.

flavioislima (approved):
LGTM

CommandMC:
Linux applications are supposed to respect `XDG_CONFIG_HOME` (see [the XDG Base Directory Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables)), with Legendary adding support for it in [this commit](https://github.com/derrod/legendary/commit/b6a701708b2585ac394bbedf452e5b89fd250727).
The only other applications that could honor this would be other Linux-native applications, but Legendary will never launch one of those, so this should be fine

butterbutts:
It seems like the Epic Games alternative login method still uses the old .config directory for legendary. Not sure if this is supposed to be the case.

PR #2808, opened 2023-06-19
Fix getGameInfo and prefix when uninstalling

arielj:
I started fixing the uninstall dialog but it actually fixes two issues:

- the uninstall dialog was failing to get the correct getGameInfo (the `installedGames` variable was not properly initialized) and then it was also skipping the set of the prefix for non-native games (so it was not showing the prefix to uninstall)
- when fixing the getGameInfo api function, this also fixed that issue when sometimes the GamePage component would show a game as NOT installed when the game is actually installed and displayed just fine in the library

@flavioislima I'm not sure why the `if (!isNative || isDLC)` check there included the native value so it would be great if you can review that, I don't know why that was there in the first place.

---

### DIFF
--- a/src/backend/storeManagers/legendary/library.ts
+++ b/src/backend/storeManagers/legendary/library.ts
@@ -57,6 +57,7 @@ const library: Map<string, GameInfo> = new Map()
 
 export async function initLegendaryLibraryManager() {
   loadGamesInAccount()
+  refreshInstalled()
 }
 
 /**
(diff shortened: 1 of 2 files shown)


flavioislima (approved):
LGTM

flavioislima:
> @flavioislima I'm not sure why the if (!isNative || isDLC) check there included the native value so it would be great if you can review that, I don't know why that was there in the first place.

I also do not remember to be honest. I may have introduced there by mistake or was testing something when I was implementing the DLCs manager and forgot to revert.

Issue #2810, opened 2023-06-20
[Bug]: Applications shortcut creation cannot be disabled on Mac


arielj:
### Describe the bug

For Windows and Linux we have these options to configure where to create shortcuts automatically:
![image](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/188464/793dda53-3e24-4b55-a1cf-c5f4158ed49c)

We don't provide these options (or an equivalent) for Mac, so shortcuts in `Applications` are always created.

### Add logs

```text
I don't have logs.
```

### Steps to reproduce

1. Install a game on a Mac
2. The game appears in the Applications screen
3. Go to the settings, there's no way to disable this feature compared to windows or linux

### Expected behavior

We should have a setting (probably reusing an existing one with a different copy on the mac) to be able to disable the automatic creation.

### Heroic Version

Latest Stable

### System Information

Mac

### Additional information

We only consider windows and linux to support shortcuts, but that is not the case anymore: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/frontend/screens/Settings/components/Shortcuts.tsx#L14

We always create the mac shortcut without considering this setting like we do for the other OSes: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/shortcuts/shortcuts/shortcuts.ts#L108

PR #2811, opened 2023-06-20
[Mac] Allow disabling automatic Applications shortcut creation

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2810

With this PR we can configure if shortcuts for games should be added to Launchpad on Mac, so there's better feature parity between OSes.

---

### DIFF
--- a/src/backend/shortcuts/shortcuts/shortcuts.ts
+++ b/src/backend/shortcuts/shortcuts/shortcuts.ts
@@ -105,7 +105,9 @@ Categories=Game;
       break
     }
     case 'darwin': {
-      await generateMacOsApp(gameInfo)
+      if (addStartMenuShortcuts || fromMenu) {
+        await generateMacOsApp(gameInfo)
+      }
       break
     }
   }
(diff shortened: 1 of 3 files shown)


src/backend/shortcuts/shortcuts/shortcuts.ts
```diff
@@ -160,7 +163,7 @@ function shortcutFiles(gameTitle: string) {
     }
     case 'darwin':
       menuFile = join(userHome, 'Applications', `${gameTitle}.app`)
-      desktopFile = join(userHome, 'Applications', `${gameTitle}.app`)
+      desktopFile = menuFile
```
arielj, line 166: just to make it more obvious it's the same path (I actually wanted to return `null` here instead since we don't add desktop shortcuts, but it was getting annoying with the types in other places if this is potentially null)

src/backend/shortcuts/shortcuts/shortcuts.ts
```diff
@@ -172,7 +175,8 @@ async function generateMacOsApp(gameInfo: GameInfo) {
 
   logInfo('Generating macOS shortcut', LogPrefix.Backend)
 
-  const appShortcut = shortcutFiles(gameInfo.title)[0]!
+  // shortcutFiles => [desktop, menu] on mac, we don't add desktop shortcut
+  const appShortcut = shortcutFiles(gameInfo.title)[1]!
```
arielj, line 179: replaced this to use `[1]` instead of `[0]` so it's more clear that this is not meant to be a desktop shortcut

PR #2822, opened 2023-06-24
[Bug] Fix popover inside settings dialog

arielj:
This PR fixes an issue where the popover component was not visible when used inside a modal.

Because of how the MaterialUI `Popover` component works, it was added to the document body, and it was always behind the `dialog` element and it was impossible to bring it to the front.

I tried many options of the Popover component to put it inside the dialog which partially fixed this but then the positioning was broken.

At the end I decided to go with a simpler solution implementing a similar functionality to have more control and to avoid all the weird things the MUI component was doing (adding styles, extra elements, scrolling the container, etc).

I tried this in:
- the settings as a normal page
- in game settings in a dialog
- in the `installation information` `time information` etc links in the game page
- the accessibility page

I think I covered all the places but please let me know if something looks wrong.

An example of this working where it was not working before:
image

---

### DIFF
--- a/src/frontend/components/UI/PopoverComponent/index.scss
+++ b/src/frontend/components/UI/PopoverComponent/index.scss
@@ -1,12 +1,17 @@
-.MuiPaper-root.MuiPopover-paper.popover-paper {
-  background-color: var(--background);
-  color: var(--text-primary);
-  border: 1px solid var(--border-color);
-  border-radius: 4px;
-  box-shadow: 0 0 0 1px var(--border-color);
-  padding: 16px;
-  width: fit-content;
-  max-width: 500px;
-  overflow-y: auto;
-  overflow-x: hidden;
+.popover-wrapper {
+  position: relative;
+  .popover {
+    background-color: var(--background);
+    color: var(--text-primary);
+    border: 1px solid var(--border-color);
+    border-radius: 4px;
+    box-shadow: 0 0 0 1px var(--border-color);
+    padding: 16px;
+    width: fit-content;
+    max-width: 500px;
+    overflow-y: auto;
+    overflow-x: hidden;
+    position: absolute;
+    z-index: 2;
+  }
 }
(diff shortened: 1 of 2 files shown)


src/frontend/components/UI/PopoverComponent/index.scss
```diff
@@ -1,12 +1,17 @@
...
+    color: var(--text-primary);
+    border: 1px solid var(--border-color);
+    border-radius: 4px;
+    box-shadow: 0 0 0 1px var(--border-color);
+    padding: 16px;
+    width: fit-content;
+    max-width: 500px;
+    overflow-y: auto;
+    overflow-x: hidden;
+    position: absolute;
```
arielj, line 14: this is only adding these last 2 props and the relative for the parent

src/frontend/components/UI/PopoverComponent/index.tsx
```diff
@@ -11,46 +10,45 @@ const PopoverComponent: React.FC<PopoverComponentProps> = ({
...
+  }
+
+  useEffect(() => {
     if (open) {
-      setAnchorEl(null)
-      return
-    }
+      // add a click listener to close the popover when clicking outside
+      const callback = (event: MouseEvent) => {
+        if (!wrapper.current!.contains(event.target as HTMLElement)) {
```
arielj, line 24: if clicked outside the wrapper, close the popover

src/frontend/components/UI/PopoverComponent/index.tsx
```diff
@@ -11,46 +10,45 @@ const PopoverComponent: React.FC<PopoverComponentProps> = ({
...
+      }
 
-    setAnchorEl(event.currentTarget)
-  }
+      document.addEventListener('click', callback)
 
-  const handleClose = () => {
-    setAnchorEl(null)
-  }
+      return () => {
```
arielj, line 31: when the popover is closed, remove the listener

redromnon (approved):
Works for me :+1:

PR #2824, opened 2023-06-24
[Feature]: Provide Protondb information on game page

kohend:
Support in getting compatibility information from ProtonDB
Tests pending approval of basic logic.

---

### DIFF
--- a/src/backend/wiki_game_info/protondb/utils.ts
+++ b/src/backend/wiki_game_info/protondb/utils.ts
@@ -1,0 +1,35 @@
+import { ProtonDBInfo, ProtonCompatibility } from 'common/types'
+import axios, { AxiosError } from 'axios'
+import { logInfo, logError, LogPrefix } from 'backend/logger/logger'
+
+export async function getInfoFromProtonDB(
+  steamID: string
+): Promise<ProtonCompatibility | null> {
+  if (steamID === '') {
+    return null
+  }
+  const url = `https://www.protondb.com/api/v1/reports/summaries/${steamID}.json`
+
+  const response = await axios
+    .get<ProtonDBInfo>(url, { headers: {} })
+    .catch((error: AxiosError) => {
+      logError(
+        [
+          `Was not able to get ProtonDB data for ${steamID}`,
+          error.response?.data.error_description
+        ],
+        LogPrefix.ExtraGameInfo
+      )
+      if (error.response?.status === 404) {
+        return null
+      }
+      throw new Error('connection error', { cause: error })
+    })
+
+  if (!response) {
+    return null
+  }
+  logInfo(`ProtonDB data for ${steamID} ${response.data}`)
+
+  return { level: response.data.tier }
+}
--- a/src/backend/wiki_game_info/protondb/utils.ts
+++ b/src/backend/wiki_game_info/protondb/utils.ts
@@ -1,11 +1,12 @@
 import { ProtonDBInfo, ProtonCompatibility } from 'common/types'
 import axios, { AxiosError } from 'axios'
-import { logInfo, logError, LogPrefix } from 'backend/logger/logger'
+import { logDebug, logError, LogPrefix } from 'backend/logger/logger'
 
 export async function getInfoFromProtonDB(
   steamID: string
 ): Promise<ProtonCompatibility | null> {
   if (steamID === '') {
+    logDebug('No SteamID, not getting ProtonDB info')
     return null
   }
   const url = `https://www.protondb.com/api/v1/reports/summaries/${steamID}.json`
@@ -27,9 +28,11 @@ export async function getInfoFromProtonDB(
     })
 
   if (!response) {
+    logDebug('No response when getting ProtonDB info')
     return null
   }
-  logInfo(`ProtonDB data for ${steamID} ${response.data}`)
+  const resp_str = JSON.stringify(response.data)
+  logDebug(`ProtonDB data for ${steamID} ${resp_str}`)
 
   return { level: response.data.tier }
 }
--- a/src/backend/wiki_game_info/protondb/utils.ts
+++ b/src/backend/wiki_game_info/protondb/utils.ts
@@ -1,4 +1,4 @@
-import { ProtonDBInfo, ProtonDBCompatibilityInfo } from 'common/types'
+import { ProtonDBCompatibilityInfo } from 'common/types'
 import axios, { AxiosError } from 'axios'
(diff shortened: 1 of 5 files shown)


src/common/types.ts
```diff
@@ -598,6 +598,18 @@ export interface GamesDBInfo {
   steamID: string
 }
 
+export interface ProtonCompatibility {
```
Nocccer, line 601: Better call it `ProtonDBInfo`

src/common/types.ts
```diff
@@ -598,6 +598,18 @@ export interface GamesDBInfo {
...
+  level: string
```
Nocccer, line 602: Can you also fetch steam deck compatibility?

src/backend/wiki_game_info/wiki_game_info.ts
```diff
@@ -48,12 +49,16 @@ export async function getWikiGameInfo(
         isMac ? getInfoFromAppleGamingWiki(title) : null
       ])
 
+    const protondb = await getInfoFromProtonDB(
```
Nocccer, line 52: Don't call this with empty string better Just call it if steamID is not empty.

src/backend/wiki_game_info/protondb/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+        [
+          `Was not able to get ProtonDB data for ${steamID}`,
+          error.response?.data.error_description
+        ],
+        LogPrefix.ExtraGameInfo
+      )
+      if (error.response?.status === 404) {
+        return null
+      }
+      throw new Error('connection error', { cause: error })
```
Nocccer, line 27: Don't throw anything. Just Log a message and return null. With this you complete Break heroic in offline mode

Nocccer (changes requested):
Really good job. Left some comments. If you don't want to write tests i can do that for you. Whats left is also the frontend. We could do this maybe in a seperated PR.

src/backend/wiki_game_info/wiki_game_info.ts
kohend: What should it be set as if SteamID is empty?

src/common/types.ts
kohend: Not from ProtonDB, so it'll be a different PR.

src/common/types.ts
kohend: ProtonDBInfo is the JSON that the API returns, should we keep it all? I don't think most is relevant.

src/common/types.ts
Nocccer: Seems they not provide it via api yet but would be cool if they would soon. Is there somewhere documentation about the api?

src/common/types.ts
Nocccer: I just want to rename it more generic because maybe later it does contain information about compatibility if protondb decide to extend there api.

src/backend/wiki_game_info/protondb/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+  steamID: string
+): Promise<ProtonDBCompatibilityInfo | null> {
+  if (steamID === '') {
+    logDebug('No SteamID, not getting ProtonDB info')
+    return null
+  }
+  const url = `https://www.protondb.com/api/v1/reports/summaries/${steamID}.json`
+
+  const response = await axios
+    .get<ProtonDBInfo>(url, { headers: {} })
```
Nocccer, line 15: To use `ProtonDBInfo` as name for heroic probably make this type used in axio some kind of temp type if possible.
Also will this break if protondb decide to add a new field from one day to another? Better fetch without type and check which fields are there?

src/backend/wiki_game_info/wiki_game_info.ts
Nocccer: First i thought about just not call `getInfoFromProtonDB` if `steamID` is empty, but that will not print any warning or error message that the information could not be fetched because of missing `steamID`. Easiest thing would be to allow passing `string | undefined` instead `string`:

```typescript
// protondb/utils.ts
export async function getInfoFromProtonDB(
  steamID: string | undefined
): Promise<ProtonDBCompatibilityInfo | null> {
  if !steamID {
    logDebug('No SteamID, not getting ProtonDB info')
    return null
  }
  // ...
 }
 
// wiki_game_info.ts

// ...
const protondb = await getInfoFromProtonDB(gamesdb.steamID ?? pcgamingwiki.steamID)
// ...
```

Nocccer (changes requested):
I found one improvement and i have one concern about making the response of protondb api a type. I don't know if axios crashes if the response does not match the type. Can happen if the api response changes.

kohend:
I want to add the SteamDeck compatibility support fetch and then add the frontend support for both, but I preferred to have both APIs in separate commits since they are mostly separate file paths, and thus separate features.

Nocccer:
> I want to add the SteamDeck compatibility support fetch and then add the frontend support for both, but I preferred to have both APIs in separate commits since they are mostly separate file paths, and thus separate features.

Feel free to make more small PR's. Easier for us to review fast.

src/backend/wiki_game_info/protondb/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { ProtonDBInfo, ProtonDBCompatibilityInfo } from 'common/types'
+import { ProtonDBCompatibilityInfo } from 'common/types'
```
Nocccer, line 1: I don't like the name `ProtonDBCompatibilityInfo` for me `ProtonDBInfo` fits much more.
Just my opinion.

src/backend/wiki_game_info/protondb/utils.ts
```diff
@@ -12,7 +12,7 @@ export async function getInfoFromProtonDB(
   const url = `https://www.protondb.com/api/v1/reports/summaries/${steamID}.json`
 
   const response = await axios
-    .get<ProtonDBInfo>(url, { headers: {} })
```
Nocccer: I don't have knowledge about axios with types. My question in the first place was if it crashes if the response does not match the type or if it just ignores not fitting json values? Maybe you have experience here you can share?

Nocccer (approved):
LGTM. Still left some questions. Let's arrange this in another PR. Will merge this now.

src/backend/wiki_game_info/protondb/utils.ts
kohend: I just made it non-typed, it's much safer, and it handles a case where the field we want doesn't exist.

PR #2826, opened 2023-06-29
[BUILD]: attempt to fix snap

imLinguin:
This is solely an **attempt** to fix some of the issues related to snap. We'll see how it goes. Any snap users are welcome to test it.

To test, checkout this branch. Follow development setup instructions from our README, then run

```
yarn dist:linux snap
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
       ]
     },
     "snap": {
+      "base": "core20",
       "allowNativeWayland": true,
       "category": "Games",
       "synopsis": "An Open Source Launcher for GOG and Epic Games",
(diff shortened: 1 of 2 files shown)


imLinguin:
Updating electron-builder fixes invalid base key. However this changes behavior of afterSign scripts, since we don't do signing in electron-builder it skips that step now.  

What's event worse electron-builder's allowNativeWayland flag is broken, when set to true it sets the `DISABLE_WAYLAND` env var to empty string, which prevents further build since snapcraft expects it to be a number or not defined at all.

I officially surrender in this PR. If we really want to support snap we should setup snapcraft.yml for that. We should discuss snap support more.

PR #2831, opened 2023-06-30
[Feat] Add support for Amazon Games

Gustash:
### Amazon Games Support

This PR adds support for Amazon Games.

It uses [Nile](https://github.com/imLinguin/nile) to interface with your Amazon Games library.

2023-06-30-215042_hyprshot

I have tested on Linux, macOS and Windows. All the basic features have been seen as working, as noted in the TODO list below.

EDIT:

Also tested the .flatpak artifact on the Steam Deck, all features are working there too.

TODO:

- [x] Display list of games
- [x] Filter by Amazon Games only
- [x] Game Info page
- [x] Installing game
- [x] Running game
- [x] Download progress 
- [x] Authentication
- [x] Game Updates
- [x] Importing (?)

---


Gustash:
Authentication is currently blocked until https://github.com/imLinguin/nile/issues/27 is resolved

derrod:
Currently this appears to not detect if a user has already logged in manually via the CLI. Otherwise seems to work fine for me (on Windows).

imLinguin:
@derrod this is due to sandboxed config of Nile. Same we do with legendary now

derrod:
> @derrod this is due to sandboxed config of Nile. Same we do with legendary now

The config file is migrated still, so Nile was logged in when running from Heroic. But it still showed as logged out and clicking the login button just did nothing. Also noteworthy that even deleting the Nile config didn't work as it would then copy the global one again, thus getting into a bad state again.

imLinguin:
Ah yes. I encountered same issue. Heroic doesn't detect that Nile is logged in already

Gustash:
> Currently this appears to not detect if a user has already logged in manually via the CLI. Otherwise seems to work fine for me (on Windows).

This has been resolved now

flavioislima:
That is pretty cool. Thanks for that.
One comment, not related though, is that we need to figure it out a better way of having those store filters since now it is not that practical. We should change it to be a select with the options instead.

I will test as soon as possible on all platforms.

Gustash:
> That is pretty cool. Thanks for that.

I'm happy to contribute ☺️

> One comment, not related though, is that we need to figure it out a better way of having those store filters since now it is not that practical. We should change it to be a select with the options instead.

Is this something you want me to do on this PR, or would it be more relevant in a separate PR after this?

src/common/types.ts
```diff
@@ -96,7 +98,7 @@ export interface ExtraInfo {
 export type GameConfigVersion = 'auto' | 'v0' | 'v0.1'
 
 export interface GameInfo {
-  runner: 'legendary' | 'gog' | 'sideload'
+  runner: 'legendary' | 'gog' | 'sideload' | 'nile'
```
flavioislima, line 101: We should use `Runner` here at some point.

flavioislima (commented):
Testing here. 
One thing that can be improved: 
if there are only windows games on Prime, then we could hide the OS filter. Right now it disables macOS but keeps the browser button, Browser should actually only show when OTHER is selected.
Clicking on it while the amazon is selected does nothing.

flavioislima:
@Gustash 
Let's do the other stuff I said on another PR.
For now just update the nile binaries and fix the conflicts and then it should be ready to merge 👍🏽

Gustash:
@flavioislima I've fixed the merge conflicts. I'm just waiting on Linguin fixing subprocess reaping in nile for Windows and macOS, as it currently doesn't work for killing games from Heroic.

Gustash:
Force stopping games is now working properly for Linux and macOS. On Windows it still doesn't work, but I found that gogdl also doesn't force stop the game on Windows. Since Nile shares pretty much a 1-to-1 mechanism for subprocesses as gogdl, this is not an isolated issue and has probably existed on Windows for a while.

I say we get this PR merged and deal with gogdl and nile not stopping games on Windows in a separate PR.

src/backend/storeManagers/nile/games.ts
```diff
@@ -0,0 +1,591 @@
...
+      ...wineEnvVars
+    }
+
+    const { bin: wineExec, type: wineType } = gameSettings.wineVersion
+
+    // Fix for people with old config
+    const wineBin =
+      wineExec.startsWith("'") && wineExec.endsWith("'")
+        ? wineExec.replaceAll("'", '')
+        : wineExec
```
imLinguin, line 382: I think this part is not needed, as there is no old config for Amazon games

imLinguin (approved):
Left small comment, but overall it looks good to me.

Amazing work ⚔️

src/backend/storeManagers/nile/games.ts
Gustash: Are we sure that global configs can't trigger this case?

imLinguin: hmm, maybe you're right

PR #2841, opened 2023-07-05
[FIX] Allow mangohud to find its config file again

Etaash-mathamsetty:
Basically title, without this patch it wasn't finding the config file because we were setting XDG_CONFIG_HOME to something different thanks to another PR. 


flavioislima:
@Etaash-mathamsetty we can merge this after fixing the conflicts

PR #2847, opened 2023-07-07
[FIX] Speed up the disabling of DXVK and VKD3D.

Etaash-mathamsetty:
Removes a whole lot of unnecessary "garbage" and hacks that didn't need to be there in the first place

### DIFF
--- a/src/backend/tools.ts
+++ b/src/backend/tools.ts
@@ -200,7 +200,7 @@ export const DXVK = {
         })
       }
 
-      logInfo('Restoring old dlls', LogPrefix.DXVKInstaller)
+      logInfo('Removing DLL overrides', LogPrefix.DXVKInstaller)
 
       // unregister the dlls on the wine prefix
       dlls.forEach(async (dll) => {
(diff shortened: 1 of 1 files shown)


src/backend/tools.ts
flavioislima, line 1: are you sure it's safe to remove the `restoreDlls` command part?

Etaash-mathamsetty: yes, I have thoroughly tested it

src/backend/tools.ts
```diff
@@ -200,34 +200,7 @@ export const DXVK = {
...
-          if (existsSync(`${dllPath}.bak`)) {
-            const removeDll = `rm ${dllPath64}`
-            exec(removeDll)
-          }
-        })
-        resolve(true)
-      })
-      // run wineboot -u restore the old dlls
-      const restoreDlls = ['wineboot', '-u']
       logInfo('Restoring old dlls', LogPrefix.DXVKInstaller)
```
CommandMC, line 203: This isn't really happening anymore, is it? Might want to adapt the message to say something like "Removing DLL overrides"

flavioislima (approved):

PR #2848, opened 2023-07-08
[UI]: Divide GamePage component into multiple components

arielj:
I started working in the redesign and I quickly found that the GamePage component was really complex and hard to work with because it kept growing and growing over time.

To be able to work more easily in the redesign, I'm extracting the different elements into individual components so we can move them around as needed without breaking functionality in the future.

Now the GamePage component fetches some info and passes that to the children, some by props and some by a context I added.

This is similar to what we have in the Settings screen.

Everything I tested worked fine, but would love to have more eyes on this since it affects the complete GamePage component.

---


flavioislima (approved):
LGTM

Tested here and everything seems to be working fine.

Nocccer (approved):
Merged master because i couldn't retrigger workflow.
Can be merged if workflow this time succed

flavioislima:
Merging this one to avoid future conflicts

PR #2866, opened 2023-07-16
[FIX]: Libraries shadowing

imLinguin:
fixes shadowing of some libraries. Thanks @Etaash-mathamsetty for finding this bug


Etaash-mathamsetty:
The issue is not fully fixed, but at least WRC works now :) 
mangohud still doesn't work unfortunately

imLinguin:
This PR should also fix issues with Steam runtime in Heroic flatpak

arielj:
I'm not sure how to review this one cause I don't understand what's the issue it's fixing, can you provide more context in the description @imLinguin ?

cause I also see the branch is named `update-nile` but it's doing other things

imLinguin:
This solves multiple issues with gogdl and nile shadowing the libraries of subprocesses. It also moves all wrappers to be added into `--wrapper` parameter, instead wapping our tools in them (which solves some issues with Steam runtimes in Flatpak too)

Added Nile to gamesdb mapping, fixed how settings detect native game (they checked if native version is available).

arielj:
got it, I understand the update for gogdl and nile are new binaries supporting the `--wrapper` option, right?

changes look good to me, at least what I understand of the code, I've never had the issue so I'm not sure it to actually test a before/after but I trust you and @Etaash-mathamsetty tested this already and works

maybe somebody else with a better understanding about wrappers can also review the code

src/backend/config.ts
```diff
@@ -320,7 +320,9 @@ class GlobalConfigV0 extends GlobalConfig {
       },
       wineCrossoverBottle: 'Heroic',
       winePrefix: isWindows ? '' : defaultWinePrefix,
-      wineVersion: defaultWine
+      wineVersion: defaultWine,
+      enableEsync: true,
+      enableFsync: true
```
arielj, line 325: this is changing the default to be `true` right? is this intentional?

I'm fine with that, just want to be sure this was the idea

imLinguin:
Yes, there are also improvements for removing part from LD_LIBRARY_PATH which is added by pyinstaller (essentially solution for library shadowing)

src/backend/config.ts
imLinguin: Yeah, I find it weird that we didn't have esync and fsync as default

flavioislima (approved):
LGTM

flavioislima:
It is working fine, although now you have conflicts to fix 😢

imLinguin:
will solve them tomorrow. Then I'll also update nile to handle being offline, it bails out currently (story for another PR)

flavioislima:
> will solve them tomorrow. Then I'll also update nile to handle being offline, it bails out currently (story for another PR)

Sure, no problem 👍🏽

PR #2867, opened 2023-07-19
[FIX] Use older vkd3d when vulkan 1.1 is detected

Etaash-mathamsetty:
title

### DIFF
--- a/src/backend/tools.ts
+++ b/src/backend/tools.ts
@@ -49,7 +49,7 @@ export const DXVK = {
     const tools = [
       {
         name: 'vkd3d',
-        url: 'https://api.github.com/repos/Heroic-Games-Launcher/vkd3d-proton/releases/latest',
+        url: getVkd3dUrl(),
         extractCommand: 'tar -xf',
         os: 'linux'
       },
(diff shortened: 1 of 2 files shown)


Etaash-mathamsetty:
i think we might have to upload a 2.6 binary to the heroic fork, because the one in the official fork is a .tar.zst or something

src/backend/tools.ts
```diff
@@ -529,3 +529,39 @@ function getDxvkUrl(): string {
...
+ */
+function getVkd3dUrl(): string {
+  if (!isLinux) {
+    return ''
+  }
+
+  if (any_gpu_supports_version([1, 3, 0])) {
+    const instance_version = get_vulkan_instance_version()
+    if (instance_version && semverLt(instance_version.join('.'), '1.3.0')) {
+      // FIXME: How does the instance version matter? Even with 1.2, newer DXVK seems to work fine
```
CommandMC, line 545:
```suggestion
      // FIXME: How does the instance version matter? Even with 1.2, newer VKD3D seems to work fine
```
:^)

src/backend/utils/graphics/vulkan/index.ts
```diff
@@ -1,22 +1,26 @@
 import { vulkanHelperBin } from 'backend/constants'
-import { spawnSync } from 'child_process'
+import { SpawnSyncReturns, spawnSync } from 'child_process'
 import { gte as semverGte } from 'semver'
 
 type VulkanVersion = [maj: number, min: number, patch: number]
 
+let instance_version: SpawnSyncReturns<string> | null = null
+let physical_versions: SpawnSyncReturns<string> | null = null
```
CommandMC, line 8: Could use something like
```suggestion
let instance_version: ReturnType<typeof spawnSync> | null = null
let physical_versions: ReturnType<typeof spawnSync> | null = null
```
here to always have the correct type & avoid the import

CommandMC:
Not sure if we can release a 2.6 version in our repo, as that'll make *that* release the latest one (unless GitHub compares SemVer versions to determine which release is the latest, which AFAIK it doesn't do)

Etaash-mathamsetty:
> Not sure if we can release a 2.6 version in our repo, as that'll make _that_ release the latest one (unless GitHub compares SemVer versions to determine which release is the latest, which AFAIK it doesn't do)

release 2.6 and then release 2.9 afterward lol

imLinguin:
@CommandMC I'm pretty sure we can release it, when creating a release we just need to make sure we disable `set as latest release`. Then GitHub will use semantic versioning to determine latest one

imLinguin (approved):

CommandMC (approved):

PR #2873, opened 2023-07-19
[FIX] Install dxvk and vkd3d on new prefixes if enabled

Etaash-mathamsetty:


CommandMC:
Do we consider DXVK/VKD3D to be part of a "complete prefix"? As in, should we consider a prefix to be *unusable* if those tools aren't installed while they're enabled in the settings?

- If we do, I'd say this should happen in `verifyWinePrefix` instead
- If we don't, `runWineCommand` shouldn't do this at all & the caller should instead decide when it wants to take care of these operations

I'm conflicted here, although I'd favor doing it in `verifyWinePrefix`

Etaash-mathamsetty:
> Do we consider DXVK/VKD3D to be part of a "complete prefix"? As in, should we consider a prefix to be _unusable_ if those tools aren't installed while they're enabled in the settings?
> 
>     * If we do, I'd say this should happen in `verifyWinePrefix` instead
> 
>     * If we don't, `runWineCommand` shouldn't do this at all & the caller should instead decide when it wants to take care of these operations
> 
> 
> I'm conflicted here, although I'd favor doing it in `verifyWinePrefix`

1. we can't do it in verifyWinePrefix because we should not need the gameSettings just to verify the wine prefix
2. The caller cannot do it because that would cause lots of code duplication and the prefix might not exist yet in many places.

CommandMC:
`verifyWinePrefix` already takes `GameSettings`, it just happens to not look at anything but the Wine version & prefix. If you do want to force DXVK/VKD3D to be installed if they're enabled, installing them in the function preparing the prefix would make more sense to me (rather than the function that "just" runs a given Wine command)

On (3):
DXVK and VKD3D are already auto-installed on launch for all games that are ran with Wine (using one function for all storefronts), see

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/297bf24308df9058be778b4e6c0185cceb3a5f6a/src/backend/launcher.ts#L229-L237

Testing this locally, it works exactly as designed (it always tries to install those components, even if they're already installed, but that's *fine*)

Etaash-mathamsetty:
seems like prepareWineLaunch wasn't run for sideloads

src/backend/main.ts
```diff
@@ -1585,22 +1583,11 @@ ipcMain.handle('getFonts', async (event, reload) => {
...
-    if (runner === 'gog' && updated) {
-      await setup(appName)
-    }
-    if (runner === 'nile' && updated) {
-      await nileSetup(appName)
-    }
-    if (runner === 'legendary' && updated) {
-      await setupUbisoftConnect(appName)
-    }
+    await prepareWineLaunch(runner, appName)
```
CommandMC, line 1590: `runWineCommandOnGame` now already runs `prepareWineLaunch`. Wouldn't this run it twice?

src/backend/tools.ts
```diff
@@ -352,7 +358,8 @@ export const Winetricks = {
       const linuxEnvs = {
         ...process.env,
         WINEPREFIX: winePrefix,
-        PATH: `${winepath}:${process.env.PATH}`
+        PATH: `${winepath}:${process.env.PATH}`,
+        ...setupWineEnvVars(gameSettings, appName)
```
CommandMC, line 362: Now that you're adding `setupWineEnvVars`, shouldn't we also have `setupEnvVars` here?

CommandMC (approved):

flavioislima (approved):

PR #2877, opened 2023-07-21
[FIX] Update wine manager on startup

Etaash-mathamsetty:


src/backend/main.ts
```diff
@@ -192,6 +193,16 @@ async function initializeWindow(): Promise<BrowserWindow> {
     }
   }, 2500)
 
+  if (!isWindows) {
```
arielj, line 196: maybe we can do this also only when opening heroic directly? Now if you open a game from a shortcut it will update the wine versions in the background, and if you open heroic later it will do that again

I think adding an extra `&& !isCLINoGui` condition in the if is enough?

arielj (approved):
looks good to me

src/backend/main.ts
```diff
@@ -192,6 +193,16 @@ async function initializeWindow(): Promise<BrowserWindow> {
     }
   }, 2500)
 
+  if (!isWindows && !isCLINoGui) {
+    setTimeout(async () => {
+      try {
+        await updateWineVersionInfos(true)
+      } catch (error) {
+        logError(error, LogPrefix.Backend)
```
Nocccer, line 201: I think there is a LogPrefix for the wine-manager

Etaash-mathamsetty: this is how it's run in refreshWineVersionInfo, so I see no issue

flavioislima (approved):

Issue #2858, opened 2023-07-13
Heroic opens wine explorer window after clicking play with game porting toolkit


blackxfiied:
### Describe the bug

I know the problem, it's becuase of the --wrapper argument. instead of referencing the "gameportingtoolkit" script in /usr/local/bin, heroic references gameportingtoolkit's wine64 (wine 7.7)

here's what heroic tries to execute according to logs (game launched is pikuniku.)
LD_PRELOAD= /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary launch 41869934302e4b8cafac2d3c0e7c293d --language en --wrapper "/usr/local/Cellar/game-porting-toolkit/1.0.2/bin/wine64 /Users/blackxfiied/my-game-prefix" --no-wine

this wont work becuase its trying to reference a folder to wine, which will just open the folder in file explorer

what does work is
LD_PRELOAD= /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary launch 48619f5656da446abccb727b78cda644 --language en --wrapper "/usr/local/bin/gameportingtoolkit /Users/blackxfiied/my-game-prefix" --no-wine

becuase the game poprting toolkit script executes it instead of feeding it directly to wine

### Add logs

```text
(10:25:56) INFO:    [Legendary]:        Running command: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary --version
(10:25:56) INFO:    [Legendary]:        Legendary location: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary
(10:25:56) INFO:    [Gog]:              GOGDL location: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/gogdl
(10:25:56) INFO:    [Connection]:       Connectivity: check-online
(10:25:56) INFO:    [Connection]:       Pinging external endpoints
(10:25:56) INFO:    [Backend]:          DRM module staus {
  "oimompecagnajdejgnnjijobebaeigek": {
    "name": "Widevine Content Decryption Module",
    "status": "new",
    "version": "4.10.2557.0"
  }
}
(10:25:56) WARNING: [Backend]:          Protocol already registered.
(10:25:57) INFO:    [Backend]:          Loading Screen Ready
(10:25:57) INFO:    [Frontend]:         Refreshing undefined Library
[...]

Issue #2832, opened 2023-07-01
File Explorer opening on launch, game not opening


kbdevs:
### Describe the bug

Hey everyone, I'm trying to launch a game through heroic and its just launching a file explorer of my prefix folder, any ideas what could I do to fix this?

(Some games have worked on windows version steam)

### Add logs

```text
(18:38:37) INFO:    [Backend]:          Adding shortcuts for PC Building Simulator

(18:38:37) INFO:    [DownloadManager]:  Finished Installation of ab277c0995e945d2b2c50c46883627f1

(18:38:37) INFO:    [DownloadManager]:  ab277c0995e945d2b2c50c46883627f1 added to download manager finished.

(18:38:37) INFO:    [DownloadManager]:  ab277c0995e945d2b2c50c46883627f1 removed from download manager.

(18:38:37) INFO:    [Frontend]:         Refreshing legendary Library

(18:38:37) INFO:    [Legendary]:        Refreshing library...

(18:38:37) INFO:    [Legendary]:        Refreshing Epic Games...

(18:38:37) INFO:    [Legendary]:        Game list updated, got 7 games & DLCs

(18:38:37) INFO:    [Backend]:          Stopping Power Saver Blocker

(18:38:37) DEBUG:   [Legendary]:        Using cached install info

(18:38:38) INFO:    [Legendary]:        Checking if EOS Overlay is enabled: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary eos-overlay info

(18:38:38) DEBUG:   [Legendary]:        Using cached install info

(18:38:38) DEBUG:   [Legendary]:        Using cached install info

(18:38:38) DEBUG:   [Legendary]:        Using cached install info

(18:38:38) INFO:    [Legendary]:        Running command: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary list --third-party

(18:39:26) INFO:    [Backend]:          Launching PC Building Simulator (ab277c0995e945d2b2c50c46883627f1)

(18:39:26) INFO:    [Legendary]:        Running command: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/legendary --version

(18:39:26) INFO:    [Backend]:          Checking if wine version exists: GPTK Wine (DX11/DX12 Only) - wine-7.7

(18:39:26) DEBUG:   [Legendary]:        Using cached install info

[...]

PR #2879, opened 2023-07-21
[FIX] Use gameportingtoolkit exectutable rather than wine to launch games

Alexandre2006:
Previously, when launching games using Apple's new Game Porting Toolkit, Heroic would launch by finding GPTK's copy of wine. This would cause file explorer to open instead of the game (as stated in #2832 and #2858).

With this fix, heroic uses the "gameportingtoolkit" executable to launch games, apps, and all wine commands (except for winetricks).

---

I've been able to run a few games using the Game Porting Toolkit, and have had no issues with running winecfg, winetricks, or installing, moving, verifying, and uninstalling games.

I have however noticed that many games do not work with the Game Porting Toolkit, although this is most likely a compatibility issue between the games and the toolkit (I've had trouble running many UE4/5 games).

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -597,7 +597,10 @@ async function runWineCommand({
     commandParts.unshift(prefix)
   }
 
-  const wineBin = wineVersion.bin.replaceAll("'", '')
+  const wineBin =
+    isGPTK && wineVersion.gptkBin !== null
+      ? wineVersion.gptkBin!.replaceAll("'", '')
+      : wineVersion.bin.replaceAll("'", '')
 
   logDebug(['Running Wine command:', commandParts.join(' ')], LogPrefix.Backend)
 
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -591,6 +591,12 @@ async function runWineCommand({
     commandParts.unshift(protonVerb)
   }
 
+  const isGPTK = wineVersion.type === 'toolkit'
+  if (isGPTK) {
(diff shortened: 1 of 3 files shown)


pakkographic:
I'm not sure what you're referring to, but there is a fix for GPTK running the wine explorer instead of the game: #2880, if you'd like to execute GPTK, just replace the WinePrefix path with your GPTK prefix path.

Alexandre2006:
#2832 offered a fix as to how to how to fix explorer being launched instead of the game by launching programs the way Apple intended, by using the gameportingtoolkit executable rather than running the programs through GPTK's copy of wine @pakkographic.

This fix worked flawlessly for me, however #2880 seems to work as well. I think both PRs are fine solutions to the problem.

Artoria2e5:
For what is worth, the `gameportingtoolkit` executable is really bare-bones. Here's its entire content:

```zsh
#!/bin/zsh
# Copyright (c) 2023 Apple Inc. All right reserved.

if [ -z "$1" ];  then
	echo "Usage: $0 <wine-prefix-path> <executable>"
fi

exe_path="cmd.exe"
if [ ! -z "$2" ]; then
	exe_path="$2"
fi

MTL_HUD_ENABLED=1 WINEESYNC=1 WINEPREFIX="$1" `brew --prefix game-porting-toolkit`/bin/wine64 "$exe_path" 2>&1 | grep "D3DM"
```

It's not even that well written: it does not allow you to give any extra argument to the exe provided, and gamers throw in extra arguments a lot for flags like `-dx12`!

PR #2880, opened 2023-07-21
[Fix] Treat GPTK Wine like regular Wine

CommandMC:
When launching GPTK, Heroic simply uses the underlying `wine64` binary. This binary is a (mostly) normal Wine binary, meaning it's meant to be launched with the WINEPREFIX environment variable set to the desired Wineprefix & a single argument (the executable to run). `gameportingtoolkit` on the other hand is the wrapper around Wine, not taking in a WINEPREFIX & instead requiring two arguments (the prefix, then the executable to run)
We were running `wine64`, but were passing it parameters intended for `gameportingtoolkit`. I'm not sure how this ever worked

While touching the responsible function (getWineFlags) I've also made it a little more type-safe & removed the unused `gameSettings` parameter

Note that I cannot test this myself, as I don't have a device running macOS, however there were plenty of people with this issue on our Discord & here (#2858)

---

### DIFF
--- a/src/backend/storeManagers/gog/games.ts
+++ b/src/backend/storeManagers/gog/games.ts
@@ -503,7 +503,7 @@ export async function launch(
         ? wineExec.replaceAll("'", '')
         : wineExec
 
-    wineFlag = [...getWineFlags(wineBin, gameSettings, wineType)]
+    wineFlag = getWineFlags(wineBin, wineType)
   }
 
   const commandParts = [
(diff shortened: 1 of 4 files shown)


pakkographic:
It works with this build of Heroic, It launches the game instead of the wine explorer, good job!

siim0n:
Hi, I can confirm this works on Sonoma, on a MBP M1 Pro, good job :)

pakkographic (approved):

arielj:
@flavioislima I think you are the best person to review this since you implemented it?

I do see comments saying it works so I'm fine with that too

flavioislima (approved):
That should work.
The problem is that I have removed the support for the scripts but forgot to update part of the code after it. Thanks for fixing it.

madr7d:
Thanks for your answer :) . However, after forcing WINEPREFIX, I launch the game and nothing happens, the "play now" button reappears, which is annoying.

PR #2882, opened 2023-07-22
[GOG]: fix obtaning manifest for updates check

imLinguin:
Due to recent CDN issues, this change switches to updated `_version=2` API, which provides fallback CDNs.

### DIFF
--- a/src/backend/storeManagers/gog/library.ts
+++ b/src/backend/storeManagers/gog/library.ts
@@ -25,7 +25,13 @@ import { basename, join } from 'node:path'
 import { existsSync, readFileSync } from 'graceful-fs'
 import { app } from 'electron'
 
-import { logError, logInfo, LogPrefix, logWarning } from '../../logger/logger'
+import {
+  logDebug,
+  logError,
+  logInfo,
+  LogPrefix,
+  logWarning
+} from '../../logger/logger'
 import { getGOGdlBin, getFileSize } from '../../utils'
 import { gogdlLogFile } from '../../constants'
 import {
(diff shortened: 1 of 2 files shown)


src/backend/storeManagers/gog/library.ts
```diff
@@ -587,21 +593,37 @@ export async function getMetaResponse(
...
-  const metaResponse = await axios.get(metaUrl, {
-    headers,
-    validateStatus: (status) => status === 200 || status === 304
-  })
-  return { status: metaResponse.status, etag: metaResponse.headers.etag }
+  const metaUrls =
+    buildData.data?.items?.find((build: BuildItem) => !build.branch)?.urls ||
+    buildData.data?.items[0]?.urls
+
+  for (const metaUrl of metaUrls) {
```
arielj, line 607: from the previous line, looks like `metaUrls` can be undefined, maybe add a `[]` as an extra fallback (`... || []`)? or this will fail when undefined

imLinguin: I can add a fallback, however there should always be at least one build since this function is called for installed games.

arielj: oh ok, so maybe there's something wrong in the types? all the `?` makes me think a lot of things there can be undefined
 but from your comment it looks like they can't really be undefined

arielj (approved):
This works fine for me, I left a comment about the types but shouldn't be a blocker anyway

flavioislima (approved):
LGTM

PR #2884, opened 2023-07-23
[UI] Redesign Game page

arielj:
I started the redesign of the game page. It's under a feature flag in Settings > Advanced.

Still a lot to do, but it's looking fine I think:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/188464/3e9549d8-6f52-453e-b708-8a85680e9c18

(this video is a bit outdated, there are a few more fixes after this screen recording)

It's not final but I think it's good enough to review and maybe merge. Since it's behind a feature flag we can merge smaller PRs until we are happy with the status, to avoid a big massive PR for the complete redesign and to avoid conflicts.

---


src/frontend/App.tsx
```diff
@@ -19,13 +19,15 @@ import ExternalLinkDialog from './components/UI/ExternalLinkDialog'
...
+  const { isSettingsModalOpen, isRTL, experimentalFeatures } =
+    useContext(ContextProvider)
 
   return (
     <div
       id="app"
       className={classNames('App', {
-        isRTL
+        isRTL,
+        oldDesign: !experimentalFeatures.enableNewDesign
```
arielj, line 30: add `oldDesign` class so we can apply different CSS to components

src/frontend/screens/Game/GamePage/components/AppleWikiInfo.tsx
```diff
@@ -47,7 +47,7 @@ const AppleWikiInfo = ({ gameInfo }: Props) => {
       }}
     >
       <WineBar />
-      {t('info.apple-gaming-wiki', 'AppleGamingWiki Rating')}:{' '}
+      <b>{t('info.apple-gaming-wiki', 'AppleGamingWiki Rating')}:</b>
```
arielj, line 50: add `b` tag so the different elements that we list are consistent in markup

src/frontend/screens/Game/GamePage/components/CloudSavesSync.tsx
```diff
@@ -44,8 +44,7 @@ const CloudSavesSync = ({ gameInfo }: Props) => {
           className="iconWithText"
         >
           <CloudQueue />
-          <b>{t('info.syncsaves')}</b>
-          {': '}
+          <b>{t('info.syncsaves')}:</b>
```
arielj, line 47: move the `:` inside the `b` for consistency with other elements

src/frontend/screens/Game/GamePage/components/InstalledInfo.tsx
```diff
@@ -57,6 +59,61 @@ const InstalledInfo = ({ gameInfo }: Props) => {
       wineVersion.type === 'crossover' ? wineCrossoverBottle : winePrefix
   }
 
+  const info = (
```
arielj, line 62: moved this outside of the return so I can return it with different wrappers depending on the design we need

Nyabinary:
Wonder when the redesign and all related changes are going be merged

src/frontend/components/UI/Sidebar/index.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useRef, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
```
arielj, line 1: removing `New Shiny Feature`

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -1,409 +1,731 @@
...
-  justify-content: stretch;
-  position: relative;
-
-  .titleWrapper {
-    display: flex;
-    position: sticky;
-    top: 0px;
-    padding-top: 2rem;
-    margin-top: 2rem;
+.App.oldDesign {
```
arielj, line 1: this diff looks super broken here, but I basically moved all the original style inside `.App.oldDesign` and then added new style after this for the new design, but github is mixing everything

flavioislima:
Testing this one on a clean install and found some bugs:
With the feature disabled the old Design is totally broken and renders only the game image:
Screenshot_20230929_134651

With the feature enabled, everything is just too big on 4k screen. We should have some limits to the sizes of the elements.
Screenshot_20230929_134816

imLinguin:
Looks like I broke the first one welp

imLinguin:
Old design should be fixed now.

I also removed this cutoff under the image by adding background to whole component on the left

obraz

vs

obraz

arielj:
would be great if we can get a suggestion from @biliesilva on how to handle such big screens, they'll look even emptier for sideloaded apps with no description

imLinguin:
We could potentially expand the page with things like achievements, screenshots, trailers and other metadata

biliesilva:
Let´s keep the same proportion and left orientation of the components until we have specific components for different breakpoints. Like this: 
Library


> would be great if we can get a suggestion from @biliesilva on how to handle such big screens, they'll look even emptier for sideloaded apps with no description

biliesilva:
Also, the uninstall button in this theme is getting more attention than the play button and this could lead the user to click on it accidentally. We could use a more neutral color.

arielj:
I pushed a few changes:
- limit size of the content for big screens (we might need some tweaks, layout changes a bit depending if the game installed or not, if it has anticheat or not, etc)
- improved the anticheat style for big screens
- use CachedImage for the background so it works when offline

Would love to merge this at this current state, the PR is already pretty big and it's under the experimental flag anyway, will make it easy to work on individual fixes now that the main skeleton for this page is done

And then I can start working on the library screen.

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -1,409 +1,780 @@
...
   }
 
-  .developer {
-    padding: var(--space-md-fixed) 0;
-    color: var(--text-default);
-  }
+  .mainInfo,
+  .tabContent {
+    outline: 3px solid #8a8a8a;
+    outline-offset: -4px;
```
imLinguin, line 458: Shouldn't the offset here be `-3px` instead? With background-color that I added on .mainInfo card, it looks like the background is sticking out of the outline

arielj: I know I had to do that because some visual glitch (the border was leaking at to bottom below the gradient)

I'm not on my usual computer for a few days and I can't reproduce the issue on the one I'm using now. I'll try to check reproduce the problem I had and see if I can put it back to be -3

flavioislima (approved):
I could still find some minor bugs but looks better and the original design is working fine now. So let's merge it to collect some feedback.

flavioislima:
So, just so we won't forget to the next iteration of this page.
The layout is still a bit weird:

image


And some of the elements that are clickable could have a hover effect so its more evident that they can be clicked. I also wonder if we should align the elements of the page in the middle for big screens. Middle-left to be clear. Just like the current design does.

But lets see what other people think about it as well :)

PR #2886, opened 2023-07-23
[Fix] Library refresh issues

arielj:
This PR includes 3 fixes:

- a different fix for the issue where people got logged out from legendary (because the `list --third-party` command ran twice in parallel). The previous fix introduced a problem (the refresh library button was not working correctly for epic games). This PR reverts the previous change and, instead, it adds a new measure: when running a `runner` command, if the command is currently running we return the same promise instead of running the command twice in parallel.

This will also prevent an issue we have during development because of react's double rendering of components in strict mode: if a component runs a command to (for example) get the game install info, it was running twice before, with this the command is running and the second call would pick the previous command's promise.

- the other change of this PR is it first checks if we are logged into Nile before trying to refresh its library

- it also fixes an issue where the page was reloaded when uninstalling a sideloaded app from the library, not it only redirects to the library if uninstalling from the game page

---

### DIFF
--- a/src/backend/storeManagers/legendary/library.ts
+++ b/src/backend/storeManagers/legendary/library.ts
@@ -143,6 +143,11 @@ export function refreshInstalled() {
   }
 }
 
+const defaultExecResult = {
+  stderr: '',
+  stdout: ''
+}
+
 /**
  * Get the game info of all games in the library
  *
@@ -150,12 +155,8 @@ export function refreshInstalled() {
  */
 export async function refresh(): Promise<ExecResult | null> {
   logInfo('Refreshing library...', LogPrefix.Legendary)
-  const isLoggedIn = LegendaryUser.isLoggedIn()
-  if (!isLoggedIn) {
-    return {
-      stderr: 'You must be logged into Epic Games to refresh this library!',
-      stdout: ''
-    }
+  if (!LegendaryUser.isLoggedIn()) {
+    return defaultExecResult
   }
 
   refreshLegendary()
@@ -173,10 +174,7 @@ export async function refresh(): Promise<ExecResult | null> {
     ['Game list updated, got', `${arr.length}`, 'games & DLCs'],
     LogPrefix.Legendary
   )
-  return {
-    stderr: '',
-    stdout: ''
-  }
+  return defaultExecResult
 }
 
 export function getListOfGames() {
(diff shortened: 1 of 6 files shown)


src/backend/storeManagers/legendary/library.ts
```diff
@@ -143,19 +143,20 @@ export function refreshInstalled() {
...
 export async function refresh(): Promise<ExecResult | null> {
   logInfo('Refreshing library...', LogPrefix.Legendary)
-  const isLoggedIn = LegendaryUser.isLoggedIn()
-  if (!isLoggedIn) {
-    return {
-      stderr: 'You must be logged into Epic Games to refresh this library!',
-      stdout: ''
-    }
+  if (!LegendaryUser.isLoggedIn()) {
+    return defaultExecResult
```
arielj, line 159: changing this for consistency with the other stores' `refresh` function

src/backend/storeManagers/nile/user.ts
```diff
@@ -115,4 +115,8 @@ export class NileUser {
 
     return user.extensions.customer_info
   }
+
+  public static isLoggedIn() {
+    return configStore.get_nodefault('userData') || false
```
arielj, line 120: please @imLinguin or @Gustash can you confirm if this is the correct way to check if we are logged in?

src/frontend/components/UI/UninstallModal/index.tsx
```diff
@@ -91,7 +89,6 @@ const UninstallModal: React.FC<UninstallModalProps> = function ({
       navigate('/')
     }
     storage.removeItem(appName)
-    refreshLibrary({ fullRefresh: true, checkForUpdates: false })
```
arielj, line 94: This was not needed. When uninstalling a game, the backend sends a `refreshLibrary` message to the frontend triggering this same call (we were running this twice on uninstall)

src/backend/storeManagers/nile/user.ts
Gustash: configStore is populated once when we migrate an existing nile config, if it exists, and whenever a user logs in/out, so this should be valid

flavioislima (approved):
Tested here and it is working fine, all other libraries are refreshing just fine as well, both on trigger and also when uninstalling or installing a game.

PR #2897, opened 2023-07-25
[FIX]: Multiple Amazon Games bugs

imLinguin:
- Updated nile to store install size in `installed.json`   
- Additionally I reproduced how nile creates folder names to avoid miss matches
- Changed how we determine if user is logged in, as username can be empty (added a fallback for login page to `Unknown` for that case)

### DIFF
--- a/src/backend/storeManagers/nile/library.ts
+++ b/src/backend/storeManagers/nile/library.ts
@@ -26,7 +26,7 @@ import {
 } from 'common/types/nile'
 import { existsSync, readFileSync, writeFileSync } from 'graceful-fs'
 import { installStore, libraryStore } from './electronStores'
-import { getNileBin } from 'backend/utils'
+import { getFileSize, getNileBin } from 'backend/utils'
 import { callRunner } from 'backend/launcher'
 import { dirname, join } from 'path'
 import { app } from 'electron'
@@ -47,6 +47,7 @@ export async function initNileLibraryManager() {
   refresh()
 }
 
+const ILLEGAL_CHARACTERS = ['?', ':', '"', '*', '<', '>', '|', '\\', '/']
 /**
  * Loads all the user's games into `library`
  */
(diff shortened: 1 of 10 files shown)


src/backend/storeManagers/nile/library.ts
```diff
@@ -66,6 +67,11 @@ function loadGamesInAccount() {
     } = productDetail
 
     const info = installedGames.get(product.id)
+    // Create save folder name like nile
+    const safeFolderName = ILLEGAL_CHARACTERS.reduce(
+      (acc: string, val: string) => acc.replaceAll(val, ''),
+      title ?? ''
+    )
```
Nocccer, line 74: Why not use `util.ts` function `removeSpecialcharacters` ?

Nocccer (changes requested):
LGTM. Left some suggestion to use `removeSpecialcharacters` function instead your own.

src/backend/storeManagers/nile/library.ts
imLinguin: This is what Nile removes from titles, We would need to change what characters are removed by Nile too (that's not an issue at all)

src/backend/storeManagers/nile/library.ts
```diff
@@ -75,11 +81,14 @@ function loadGamesInAccount() {
       install: info
         ? {
             install_path: info.path,
+            // For some time size was undefined in installed.json, that's why we
+            // need to keep this fallback to 0
+            install_size: getFileSize(info?.size ?? 0),
```
CommandMC, line 86: `info` is already known to not be `undefined`/`null` here
```suggestion
            install_size: getFileSize(info.size ?? 0),
```

src/backend/storeManagers/nile/library.ts
```diff
@@ -66,7 +66,8 @@ function loadGamesInAccount() {
     } = productDetail
 
     const info = installedGames.get(product.id)
-
+    // Create save folder name like nile
+    const safeFolderName = removeSpecialcharacters(title ?? '')
```
CommandMC, line 70: Is it really a good idea to use `''` as a fallback here? I can't imagine that creating a folder with that name would work

imLinguin: since `folder_name` for nile games is only used when deleting canceled installation `''` will be `false` therefore it won't cause any trouble

imLinguin: Also I just confirmed the case when title was undefined is resolved, so this is even less likely to happen

flavioislima (approved):

PR #2918, opened 2023-07-27
[Tech] Type-check Legendary commands

CommandMC:
This PR adds strict type checking to Legendary commands using [Zod](https://zod.dev/). Before a command can be ran, the caller of `runLegendaryCommand` has to verify that, for example, strings that should be non-empty are actually non-empty, numbers are integers & positive (if necessary), paths are valid paths, and so on

Details on how this is achieved / what the design decisions behind things were are in the commit messages

~~I've only tested this on Linux, I can test it on Windows in the coming days~~
Seems to be working fine on both Windows & Linux

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -63,7 +63,7 @@ import { gameManagerMap } from 'backend/storeManagers'
 import * as VDF from '@node-steam/vdf'
 import { readFileSync } from 'fs'
 import { LegendaryCommand } from './storeManagers/legendary/commands'
-import { Entries } from 'type-fest'
+import { commandToArgsArray } from './storeManagers/legendary/library'
 
 async function prepareLaunch(
   gameSettings: GameSettings,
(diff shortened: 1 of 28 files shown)


src/backend/utils/compatibility_layers.ts
```diff
@@ -372,18 +374,51 @@ export async function getGamingPortingToolkitWine(): Promise<
...
+  wineType: WineInstallation['type'],
+  wrapper: string,
+  asArray?: false
+): AllowedWineFlags
+export function getWineFlags(
+  wineBin: string,
+  wineType: WineInstallation['type'],
+  wrapper: string,
+  asArray: true
+): string[]
```
arielj, line 393: maybe have this be 2 functions? `getWineFlags` and `getWineFlagsArray`? I saw the `true` added when calling this function and I had to look for this code

I think it will make things easier to read

src/backend/launcher.ts
```diff
@@ -975,6 +977,63 @@ async function getWinePath({
   return stdout.trim()
 }
 
+/**
+ * Converts a LegendaryCommand to a parameter list passable to Legendary
```
arielj, line 981: can we move this function into some legendary specific file/folder?

src/backend/storeManagers/legendary/commands/base.ts
```diff
@@ -0,0 +1,37 @@
...
+export const LegendaryAppName = z
+  .string()
+  .refine((val) => hasGame(val), {
+    message: 'AppName was not found on account'
+  })
+  .brand('LegendaryAppName')
+export type LegendaryAppName = z.infer<typeof LegendaryAppName>
+
+export const LegendaryPlatform = z.enum(['Win32', 'Windows', 'Mac'] as const)
+export type LegendaryPlatform = z.infer<typeof LegendaryPlatform>
```
arielj, line 14: do we need this LegendaryPlatform type using zod? can't this be a normal typescript type?

Not sure how zod works so maybe this is needed to make this work, but we already have `export type LegendaryInstallPlatform = 'Windows' | 'Win32' | 'Mac'` in src/common/types/legendary.ts

src/backend/storeManagers/legendary/commands/base.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const NonEmptyString = z.string().min(1).brand('NonEmptyString')
+export type NonEmptyString = z.infer<typeof NonEmptyString>
+
+export const Path = z
+  .string()
+  .refine((val) => path.parse(val).root, 'Path is not valid')
+  .brand('Path')
```
arielj, line 22: just out of curiosity, I understand this check happens at runtime, right?

what should we expect to see in heroic if a path is invalid or a string is empty? some message in the logs?

arielj (commented):
this looks great, I love the idea of not having to pass an array of strings and have the options and validations for each legendary command :+1:

src/backend/storeManagers/legendary/commands/base.ts
CommandMC: Zod validates data at runtime, types don't exist anymore at that point
We *could* remove the definition from `common/types/legendary.ts` (& only import the one defined here), but I'm pretty sure we'd then have to switch to using `import type` everywhere (since Vite can't know that you *really* only need the type from this file, imports can have side effects after all)

src/backend/storeManagers/legendary/commands/base.ts
CommandMC: This depends on how the data is validated. If `SomeSchema.parse(...)` is used, an exception is thrown if the validation does not succeed. `.safeParse` on the other hand will always return an object of type `{ success: true, data: T } | { success: false, error: ZodError }`
I chose to use exclusively `.parse` for now to be a little more aggressive in finding potential issues & because there isn't a clear error propagation path in most functions, if we end up using more Zod in the future, we could validate the data at the source instead (potentially with better error handling there)

src/backend/storeManagers/legendary/commands/base.ts
arielj: I'm fine with keeping the 2 definitions, it's not an issue

src/backend/utils/compatibility_layers.ts
CommandMC: Yup, that definitely feels better, done in 292d0ecc84e15791741dcd8125b9230e138a76d3

src/backend/launcher.ts
CommandMC: I anticipate that I'm going to work on applying this same treatment to gogdl and Nile commands in the future, and then I'll need the function here, but for now it can live in Legendary's library
cefadad9e6e56d808cd8a8b75b995fa05f2e38e2

arielj (approved):
Looks good to me :+1:

Issue #2916, opened 2023-07-26
`EOS Overlay` won't install on Linux in `Flatpak` version `2.9.1` of the `Heroic Games Launcher`


MeteorKLINT:
### Describe the bug

Made a fresh installation of the `Flatpak` version `2.9.1` of `Heroic`. I tried to install the `EOS Overlay`. After I hit the button to install the EOS Overlay, I got the message `The EOS Overlay is being installed...`. The log tells me that the process is aborted close after, but heroic stays in its state forever, and I have to forcefully kill heroic to cancel the not working installation.

### Add logs

```text
(00:46:12) INFO:    [Legendary]:        Running command: XDG_CONFIG_HOME=/home/rieso/.var/app/com.heroicgameslauncher.hgl/config/heroic/legendaryConfig /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary --version
(00:46:12) INFO:    [Legendary]:        Legendary location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(00:46:12) INFO:    [Gog]:              GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(00:46:12) INFO:    [Connection]:       Connectivity: check-online
(00:46:12) INFO:    [Connection]:       Pinging external endpoints
(00:46:12) INFO:    [Connection]:       Connectivity: online
(00:46:12) INFO:    [Gog]:              Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/rieso/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json --version
(00:46:13) INFO:    [Nile]:             Running command: XDG_CONFIG_HOME=/home/rieso/.var/app/com.heroicgameslauncher.hgl/config/heroic/nile_config /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/nile --version
(00:46:13) INFO:    [Backend]:          

Heroic Version: 2.9.1 Boa Hancock
Legendary Version:  0.20.32 Dark Energy (hotfix #6)
GOGdl Version: 0.7.3
Nile Version: 1.0.0 Jonathan Joestar

Electron Version: 24.4.1
Chrome Version: 112.0.5615.204
NodeJS Version: 18.14.0

[...]

Issue #2905, opened 2023-07-26
`EOS Overlay` won't install on Windows 10 in version `2.9.1` of the `Heroic Games Launcher`.


Zliced13:
### Describe the bug

When I try to install **EOS Overlay** via the **Heroic Games Launcher**, I always get the following error as seen in this log.

```log
An error has occurred! Try to Logout and Login on your Epic account.
Do you want to install the overlay? [Y/n]:[cli] INFO: Preparing to start overlay install...[Core] WARNING: Unable to download manifest from "download.epicgames.com" (status: 403), trying next URL...[Core] WARNING: Unable to download manifest from "download2.epicgames.com" (status: 403), trying next URL...[Core] WARNING: Unable to download manifest from "download3.epicgames.com" (status: 403), trying next URL...[Core] WARNING: Unable to download manifest from "download4.epicgames.com" (status: 403), trying next URL...[cli] INFO: Install directory: C:\Users\USER\AppData\Roaming\heroic\tools\eos_overlay
[cli] INFO: Install size: 330.82 MiB
[cli] INFO: Download size: 141.52 MiBTraceback (most recent call last):
File "legendary\cli.py", line 3069, in <module>File "legendary\cli.py", line 3018, in main
File "legendary\cli.py", line 2268, in manage_eos_overlay
File "legendary\utils\cli.py", line 4, in get_boolean_choice
EOFError: EOF when reading a line
[16356] Failed to execute script 'cli' due to unhandled exception!
```

### Add logs

```text
(20:25:00) INFO:    [Legendary]:        Running command: XDG_CONFIG_HOME=C:\Users\USER\AppData\Roaming\heroic\legendaryConfig C:\Users\USER\scoop\apps\heroic-games-launcher\2.8.0\resources\app.asar.unpacked\build\bin\win32\legendary --version
(20:25:00) INFO:    [Legendary]:        Legendary location: C:\Users\USER\scoop\apps\heroic-games-launcher\2.8.0\resources\app.asar.unpacked\build\bin\win32\legendary
(20:25:00) INFO:    [Gog]:              GOGDL location: C:\Users\USER\scoop\apps\heroic-games-launcher\2.8.0\resources\app.asar.unpacked\build\bin\win32\gogdl
(20:25:00) INFO:    [Connection]:       Connectivity: check-online
(20:25:00) INFO:    [Connection]:       Pinging external endpoints
[...]

PR #2930, opened 2023-07-29
[Fix] Installation of EOS Overlay

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2916 and fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2905

I added some code to prevent running the same command twice in parallel (to prevent the epic logout issue), but I didn't know that `then` and `catch` on promises does NOT change the current promise but it actually returns a new object, so I was storing/caching the current promise without handling the catch.

Because the first step installing the overlay is to get the size and then abort the command, it was throwing an exception instead of handling it silently.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -781,7 +781,7 @@ async function callRunner(
     return currentPromise
   }
 
-  const promise = new Promise<ExecResult>((res, rej) => {
+  let promise = new Promise<ExecResult>((res, rej) => {
     const child = spawn(bin, commandParts, {
       cwd: runner.dir,
       env: { ...process.env, ...options?.env },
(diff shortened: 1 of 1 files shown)


CommandMC (approved):
Seems like the process still isn't *perfect* again, originally it also displayed the progress of the install in the sidebar, like regular game downloads do, but after looking at the code, this seems a lot more difficult to implement with the download manager & it *working* is of course the important part

arielj:
> Seems like the process still isn't _perfect_ again, originally it also displayed the progress of the install in the sidebar, like regular game downloads do, but after looking at the code, this seems a lot more difficult to implement with the download manager & it _working_ is of course the important part

looks like that has been broken for a long time, I tried with Heroic 2.7 and it was already not displaying the EOS installation in the sidebar

PR #2934, opened 2023-07-29
Update version in bug report template, and add workflow to automate this

arielj:
This PR does 2 things:
- Updates the `Latest version` in the bug template
- Adds a github workflow that runs when we publish a new release that will automatically create a PR doing this change

You can check it working here in this test repo: https://github.com/arielj/release-update-template/pull/6

---


flavioislima (approved):
Nice. Now we don't need to remember to do it on every release. Good stuff.

Issue #2891, opened 2023-07-24
Error when "Sync with Installed Epic Games"


Dominiquini:
### Describe the bug

When I selected the option "Sync with Installed Epic Games" in the settings, this error appears: "Invalid Path"

### Add logs

```text
(19:24:51) ERROR:   [Legendary]:        Error: Command failed: C:\Users\Rafael Dominiquini\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary egl-sync --enable-sync -y
'C:\Users\Rafael' is not recognized as an internal or external command,
operable program or batch file.

    at ChildProcess.exithandler (node:child_process:430:12)
    at ChildProcess.emit (node:events:513:28)
    at maybeClose (node:internal/child_process:1091:16)
    at ChildProcess._handle.onexit (node:internal/child_process:302:5)
```

### Steps to reproduce

1. Go to 'Settintgs'
2. Click on 'General'
3. Scroll down to 'Sync with Installed Epic Games'
4. Mark the checkbox
5. See the error and log

### Expected behavior

I don't know what this settings should do!

### Heroic Version

Latest Stable

### System Information

- OS: Windows 10 x64 (19045.3269)

PR #2964, opened 2023-08-06
Use runRunnerCommand to fix issue with spaces enabling egl-sync

arielj:
This PR should fix https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2891

We were still using `execAsync` to execute a legendary command instead of `runRunnerCommand` that handles spaces properly.

I also moved the logic of the function into the legendary/library.ts file, it makes more sense there.

I did test this in Linux sideloading the EGL app, but I don't have spaces in my username and I don't have a windows machine to test there either.

---

### DIFF
--- a/src/backend/storeManagers/legendary/library.ts
+++ b/src/backend/storeManagers/legendary/library.ts
@@ -2,7 +2,7 @@ import {
   createAbortController,
   deleteAbortController
 } from '../../utils/aborthandler/aborthandler'
-import { existsSync, readFileSync, readdirSync } from 'graceful-fs'
+import { existsSync, mkdirSync, readFileSync, readdirSync } from 'graceful-fs'
 
 import {
   GameInfo,
@@ -54,6 +54,7 @@ import { update } from './games'
 import axios from 'axios'
 import { app } from 'electron'
 import { copySync } from 'fs-extra'
+import { platform } from 'os'
 
 const allGames: Set<string> = new Set()
 let installedGames: Map<string, InstalledJsonMetadata> = new Map()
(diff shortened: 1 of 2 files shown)


flavioislima (approved):
LGTM

arielj:
@CommandMC I had to add a new LegendaryCommand after it was approved to resolve some conflicts, can you check this PR to make sure I did it the right way?

src/backend/storeManagers/legendary/library.ts
```diff
@@ -744,6 +745,57 @@ export async function getGameSdl(
 }
 
 /**
+ * Toggles the EGL synchronization on/off based on arguments
+ * @param path_or_action On Windows: "unlink" (turn off), "windows" (turn on). On linux/mac: "unlink" (turn off), any other string (prefix path)
+ * @returns string with stdout + stderr, or error message
+ */
+export async function toggleGamesSync(path_or_action: string) {
+  const isWindows = platform() === 'win32'
```
CommandMC, line 753: We already have a `isWindows` export in `backend/constants.ts`, might want to use that

src/backend/storeManagers/legendary/commands/egl_sync.ts
```diff
@@ -0,0 +1,16 @@
...
+  '--enable-sync'?: true
+  '--egl-wine-prefix'?: Path
+  '--disable-sync'?: true
+  '--egl-manifest-path'?: Path
+  '--one-shot'?: true
+  '--import-only'?: true
+  '--export-only'?: true
+  '--migrate'?: true
+  '--unlink'?: true
+}
```
CommandMC, line 14: This is a *super* minor thing, but for the other commands I've added parameters in the order they're displayed in Legendary's README/the output of `-h` (in this case, first `--egl-manifest-path`, then `--egl-wine-prefix`, and so on)

src/backend/storeManagers/legendary/library.ts
```diff
@@ -744,6 +745,57 @@ export async function getGameSdl(
...
+    subcommand: 'egl-sync',
+    '-y': true
+  }
+
+  if (path_or_action === 'unlink') {
+    command['--unlink'] = true
+  } else {
+    command['--enable-sync'] = true
+    if (!isWindows) {
+      command['--egl-wine-prefix'] = Path.parse(path_or_action)
```
CommandMC, line 774: I don't think we can be certain that this path is actually *valid* (it's user input after all; trusting the OS's file picker to do the validation for us sounds like a bad idea). Might want to do something like
```suggestion
      const pathParse = Path.safeParse(path_or_action)
      if (pathParse.success)
        command['--egl-wine-prefix'] = pathParse.data
      else
        return 'Error'
```

Doing this at the source would be the long-term goal of course, but that's far too great of a task to just push onto this PR :^)

src/backend/storeManagers/legendary/library.ts
arielj: changed

arielj:
thanks @CommandMC, I updated the PR

CommandMC (approved):

PR #2969, opened 2023-08-08
[Improv]: Add environment variables for use by wrappers

mtkennerly:
This adds two environment variables when launching games: `HEROIC_APP_NAME` and `HEROIC_APP_STORE`. Wrapper commands can use these to look up the app info in the Heroic config (e.g., to get the title or Wine prefix for save backup purposes, in my case).

I tested this on Windows 11 with GOG and Epic. I don't have any Amazon games, so I wasn't able to test that.

In order to test on my system, I made the `WrappersTable` display on Windows (which I'd like to submit in a separate PR after some testing), then configured a wrapper as a Python script. Main interesting part of the script:

```python
path = Path("C:/tmp/script.log")

with path.open("a") as f:
    f.write(f"{sys.argv}\n")
    for k, v in os.environ.items():
        if k.startswith("HEROIC_"):
            f.write(f"{k} = {v}\n")
```

I verified that the new environment variables were written out to the log file:

```
['C:/tmp/script.py', 'C:\\Users\\mtken\\Games\\Heroic\\Dead Cells\\deadcells.exe']
HEROIC_APP_NAME = 1237807960
HEROIC_APP_STORE = gog

['C:/tmp/script.py', 'C:\\Users\\mtken\\Games\\Heroic\\CaveStoryPlus\\CaveStory+.exe', '-AUTH_LOGIN=unused', '-AUTH_PASSWORD=<redacted>', '-AUTH_TYPE=exchangecode', '-epicapp=1dea8a6ddb544842a58e4b5c8675ff58', '-epicenv=Prod', '-EpicPortal', '-epicusername=<redacted>', '-epicuserid=<redacted>', '-epiclocale=en', '-epicsandboxid=78473822f724474d8e436f6bde735623']
HEROIC_APP_NAME = 1dea8a6ddb544842a58e4b5c8675ff58
HEROIC_APP_STORE = epic
```

---

I think these environment variables should be documented, but I'm not sure where that information belongs.

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -315,6 +315,7 @@ function setupWrapperEnvVars(wrapperEnv: WrapperEnv) {
   const ret: Record<string, string> = {}
 
   ret.HEROIC_APP_NAME = wrapperEnv.appName
+  ret.HEROIC_APP_RUNNER = wrapperEnv.appRunner
 
   switch (wrapperEnv.appRunner) {
     case 'gog':
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -45,6 +45,7 @@ import {
   GameInfo,
   Runner,
   EnviromentVariable,
+  WrapperEnv,
   WrapperVariable,
   ExecResult,
   GameSettings,
@@ -275,6 +276,7 @@ async function prepareWineLaunch(
 /**
  * Maps general settings to environment variables
  * @param gameSettings The GameSettings to get the environment variables for
+ * @param wrapperEnv The wrapper info to be added into the environment variables
  * @returns A big string of environment variables, structured key=value
  */
 function setupEnvVars(gameSettings: GameSettings) {
@@ -301,6 +303,20 @@ function setupEnvVars(gameSettings: GameSettings) {
   if (!process.env.LD_PRELOAD && !ret.LD_PRELOAD) {
     ret.LD_PRELOAD = ''
   }
+
+  return ret
+}
+
+/**
+ * Maps launcher info to environment variables for consumption by wrappers
+ * @param wrapperEnv The info to be added into the environment variables
+ * @returns Environment variables
+ */
+function setupWrapperEnvVars(wrapperEnv: WrapperEnv) {
+  const ret: Record<string, string> = {}
+
+  ret.HEROIC_APP_NAME = wrapperEnv.appName
+  ret.HEROIC_APP_STORE = wrapperEnv.appStore
 
   return ret
 }
@@ -982,6 +998,7 @@ export {
   launchCleanup,
   prepareWineLaunch,
   setupEnvVars,
+  setupWrapperEnvVars,
   setupWineEnvVars,
   setupWrappers,
   runWineCommand,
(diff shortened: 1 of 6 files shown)


imLinguin:
Looks good overall, could you also add it for sideload runner? It might not be useful for your use case, but it's good to have it consistent

mtkennerly:
@imLinguin Updated in 8db11ac7e0a7cd881350648a61fed31d2ee20fbf :)

I changed `HEROIC_APP_STORE` to `HEROIC_APP_SOURCE` since "store" doesn't make as much sense for sideloaded apps.

imLinguin (approved):

src/backend/storeManagers/gog/games.ts
```diff
@@ -464,9 +465,13 @@ export async function launch(
...
-  let commandEnv = isWindows
-    ? process.env
-    : { ...process.env, ...setupEnvVars(gameSettings) }
+  let commandEnv = {
+    ...process.env,
+    ...setupWrapperEnvVars({ appName, appRunner: 'gog' })
+  }
+  if (!isWindows) {
+    commandEnv = { ...commandEnv, ...setupEnvVars(gameSettings) }
+  }
```
CommandMC, line 474:
```suggestion
  let commandEnv = {
    ...process.env,
    ...setupWrapperEnvVars({ appName, appRunner: 'gog' }),
    ...(isWindows ? {} : setupEnvVars(gameSettings))
  }
```
This feels a little cleaner

src/backend/launcher.ts
```diff
@@ -305,6 +307,34 @@ function setupEnvVars(gameSettings: GameSettings) {
...
+    case 'legendary':
+      ret.HEROIC_APP_SOURCE = 'epic'
+      break
+    case 'nile':
+      ret.HEROIC_APP_SOURCE = 'amazon'
+      break
+    case 'sideload':
+      ret.HEROIC_APP_SOURCE = 'sideload'
+      break
+  }
```
Nocccer, line 333: Why not ?
```suggestion
 ret.HEROIC_APP_SOURCE = wrappernEnv.appRunner
```

Nocccer (changes requested):
Overall looks good.
I left some suggestion/question.

src/backend/launcher.ts
```diff
@@ -275,6 +276,7 @@ async function prepareWineLaunch(
 /**
  * Maps general settings to environment variables
  * @param gameSettings The GameSettings to get the environment variables for
+ * @param wrapperEnv The wrapper info to be added into the environment variables
```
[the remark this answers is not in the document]
mtkennerly: Ah, I originally tried adding the functionality here, but realized it worked better separate and forgot to reset the documentation. Fixed in 87ee893fa4cfdb712a78415abb808bc1d900dd48.

src/backend/storeManagers/gog/games.ts
mtkennerly: Good call. Updated in 5d1a6fe7520a240be97faff67fac600f14131d3a.

src/backend/launcher.ts
mtkennerly: I was thinking there could be a case in the future where it may help to distinguish the runner and the source, like if you could choose two different runners for the same store or if one runner supported multiple stores. Not sure if that's a valid/realistic concern given Heroic's design goals.

I added a `HEROIC_APP_RUNNER` env var in e78e680768ad5af6152af9070287a5cf6986722d. Just let me know if you'd prefer to have `HEROIC_APP_SOURCE` removed as well.

flavioislima (approved):
Looks good, thanks for that!

PR #2972, opened 2023-08-08
[FIX] Use wineserver -k when available

Etaash-mathamsetty:
Prevents wineboot -k from starting the prefix rather than stopping it when the prefix wasn't init in the first place.


flavioislima (approved):
Good stuff, noticed this issues while testing the latest release.

PR #2981, opened 2023-08-12
[Improv]: Confirm installing broken anticheat games

arielj:
This PR adds an extra confirmation dialog when trying to install a game that we have data about its anticheat support being broken/denied.

We keep getting users asking about Fortnite for example even with the message in both the game page and the install dialog.

This will add another barrier for users to understand these games will not work, and they have to click a button that says `I UNDERSTAND THIS WILL NOT WORK` to even install the game.

This has no impact on games where the anticheat works so it only adds one barrier to prevent extra support tickets.

You can see it working in this video:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/188464/b3ae8035-2bb7-493e-8d52-be649acc54ff

---

### DIFF
--- a/src/frontend/components/UI/Anticheat/index.tsx
+++ b/src/frontend/components/UI/Anticheat/index.tsx
@@ -1,5 +1,5 @@
-import React, { MouseEvent, useEffect, useState } from 'react'
-import { AntiCheatInfo, GameInfo } from 'common/types'
+import React, { MouseEvent } from 'react'
+import { AntiCheatInfo } from 'common/types'
 import { createNewWindow } from 'frontend/helpers'
 
 import { ReactComponent as InfoIcon } from 'frontend/assets/info_icon.svg'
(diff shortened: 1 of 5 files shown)


src/frontend/components/UI/Anticheat/index.tsx
```diff
@@ -10,30 +10,14 @@ import './index.scss'
...
-      gameInfo.title &&
-      gameInfo.namespace !== undefined
-    ) {
-      window.api
-        .getAnticheatInfo(gameInfo.namespace)
-        .then((anticheatInfo: AntiCheatInfo | null) => {
-          setAnticheatInfo(anticheatInfo)
-        })
-    }
-  }, [gameInfo])
```
arielj, line 35: moving this info fetch out and passing it as a propo, so we can see this info in the parent component

src/frontend/hooks/hasAnticheatInfo.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from 'react'
+import { AntiCheatInfo, GameInfo } from 'common/types'
+
+export const hasAnticheatInfo = (gameInfo: GameInfo) => {
```
arielj, line 4: custom hook so we can use it in different places without repeating the fetch logic

src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx
```diff
@@ -187,7 +190,37 @@ export default function DownloadDialog({
...
+        'install.anticheat-warning.message',
+        'The anticheat support is broken or denied. The game will not work. Do you want to install it anyway?'
+      ),
+      buttons: [
+        {
+          text: t(
+            'install.anticheat-warning.install',
+            'Yes (I understand it will not work)'
+          ),
+          onClick: async () => handleInstall(path, true)
```
arielj, line 206: if the user confirms the installation anyway, we call `handleInstall` with the `ignoreAnticheat` arg set to true

Nocccer (approved):
LGTM

Issue #2071, opened 2022-11-26
Disable Epic login if the OS is incompatible with Legendary


arielj:
### Problem description

It is really common on Discord to have users using old version of MacOS (High Sierra) and sometimes Windows 7 asking about issues during login.

The problem is that Legendary requires Windows 8.1+ and MacOS 12.0+ but we don't show that anywhere and we show no error when the login fails either. https://github.com/derrod/legendary#requirements

### Feature description

In the login screen, if we detect the OS is incompatible, the Epic login options should be disabled with a message explaining that we use Legendary and Legendary is not compatible with the OS version.

### Alternatives

No idea.

flavioislima:
Actually, we need to show a message that Heroic is not compatible with those systems, MS does not support windows 7 anymore so why should we?

So we can show a dialog if we detect one of those systems instead of disabling Epic and this dialog should not be able to close and after clicking OK Heroic would quit.

dylanmtaylor:
> Actually, we need to show a message that Heroic is not compatible with those systems, MS does not support windows 7 anymore so why should we?
> 
> So we can show a dialog if we detect one of those systems instead of disabling Epic and this dialog should not be able to close and after clicking OK Heroic would quit.

Windows 8.1 will reach end of support on January 10, 2023, so it might make sense to only support Windows 10+

PR #2983, opened 2023-08-13
[Mac] Disable the login buttons for MacOS older than version 12

arielj:
This PR disables the login buttons if we detect the OS is MacOS older than version 12 and shows a message to the user to clarify that.

Even with the readme and the information in the downloads page in the websites, users of MacOS 10/11 still try to run Heroic and ask for support when it doesn't work.

This PR makes it more explicit in the Log In page itself so the users can see the answer there without the need to ask in discord.

This is how it would look:
image

(I don't have that version of MacOS so I faked some values to generate the image, but I know the logic works since I also tested using version 14 as the limit (and I have MacOS 13) and the message was displayed)

This partially fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2071, but not completely since I didn't implement logic for old Windows systems.

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -344,7 +344,7 @@
         "epic": "Epic Games Login",
         "gog": "GOG Login",
         "message": "Login with your platform. You can login to more than one platform at the same time.",
-        "old-mac": "Your MacOS version is {{version}}. MacOS 12 or newer is required to log in the stores."
+        "old-mac": "Your macOS version is {{version}}. macOS 12 or newer is required to log in."
     },
     "message": {
         "sync": "Sync Complete",
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -343,7 +343,8 @@
         "amazon": "Amazon Login",
         "epic": "Epic Games Login",
         "gog": "GOG Login",
-        "message": "Login with your platform. You can login to more than one platform at the same time."
+        "message": "Login with your platform. You can login to more than one platform at the same time.",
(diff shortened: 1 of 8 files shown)


src/frontend/screens/Login/components/Runner/index.tsx
```diff
@@ -15,21 +15,15 @@ interface RunnerProps {
...
-  const [isLoggingOut, setIsLoggingOut] = React.useState(false)
-
-  let buttonText = ''
-  if (props.class === 'epic') {
-    buttonText = t('login.epic', 'Epic Games Login')
-  } else if (props.class === 'nile') {
-    buttonText = t('login.amazon', 'Amazon Login')
-  } else {
-    buttonText = t('login.gog', 'GOG Login')
-  }
```
arielj, line 32: this is extra in this PR, but I think this strings shouldn't be in this component but passed as props so I did a small refactor

public/locales/en/translation.json
```diff
@@ -343,7 +343,8 @@
         "amazon": "Amazon Login",
         "epic": "Epic Games Login",
         "gog": "GOG Login",
-        "message": "Login with your platform. You can login to more than one platform at the same time."
+        "message": "Login with your platform. You can login to more than one platform at the same time.",
+        "old-mac": "Your MacOS version is {{version}}. MacOS 12 or newer is required to log in the stores."
```
arielj, line 347: suggestions for this copy are welcomed

JKingweb: Small pedantic note: the proper capitalization is "macOS".

arielj: updated, thanks

Nocccer (approved):
LGTM

PR #3002, opened 2023-08-21
[Improv]: Add opt out of playtime tracking

### LINKED ISSUE #2906: opt out of playtime tracking
### Problem description

_No response_

### Feature description

I know playtime tracking is a heavily requested feature but I would really like a way to disable it. Either within Heroic or some kind of workaround.

Appreciate all the hard work. Thanks!

### Alternatives

_No response_

### Additional information

_No response_

imLinguin:
This PR adds two handlers for `suspend` and `resume` which now create separate sessions, avoiding situation where playtime was tracked when device was suspended (I have Steam Deck in mind here particularly)  
I also added toggle to disable playtime syncing

Closes #2906


src/frontend/screens/Settings/components/PlaytimeSync.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { useContext } from 'react'
+import { useTranslation } from 'react-i18next'
+import { ToggleSwitch } from 'frontend/components/UI'
+import useSetting from 'frontend/hooks/useSetting'
+import SettingsContext from '../SettingsContext'
+
+const PlaytimeSync = () => {
```
arielj, line 7: I understand we only support playtime sync for GOG games, right? Maybe that should be clarified with an info box component or something.

Also, I wonder if this setting would be better to be per game and not global (like if in the future you want to not sync one specific game). I can imagine users will want that and I think the change is not to much to adapt this PR to that? Just an idea though, not a blocker.

src/backend/main.ts
```diff
@@ -1049,24 +1078,41 @@ ipcMain.handle(
...
       sessionPlaytime + tsStore.get(`${appName}.totalPlayed`, 0)
     tsStore.set(`${appName}.totalPlayed`, Math.floor(totalPlaytime))
 
-    if (runner === 'gog') {
-      await updateGOGPlaytime(appName, startPlayingDate, finishedPlayingDate)
+    const { disablePlaytimeSync } = GlobalConfig.get().getSettings()
+    if (!disablePlaytimeSync) {
+      for (const session of playTimeSessions) {
+        if (runner === 'gog') {
+          await updateGOGPlaytime(appName, session.start, session.end)
```
arielj, line 1108: maybe we don't need the `await` here? just so syncing the playtime doesn't block other actions if for any reason the requests are slow or the update fails

src/backend/main.ts
```diff
@@ -1049,24 +1078,41 @@ ipcMain.handle(
...
-      (finishedPlayingDate.getTime() - startPlayingDate.getTime()) / 1000 / 60
+    currentSession.end = new Date()
+    playTimeSessions.push(currentSession)
+    tsStore.set(`${appName}.lastPlayed`, currentSession.end.toISOString())
+
+    const sessionPlaytime = playTimeSessions.reduce(
+      (acc, next) =>
+        acc + (next.end.getTime() - next.start.getTime()) / 1000 / 60,
+      0
+    )
```
arielj, line 1099: not a big deal, but I think it's better to do the division after you reduced the array

```suggestion
    const sessionPlaytime = playTimeSessions.reduce(
      (acc, next) =>
        acc + next.end.getTime() - next.start.getTime(),
      0
    ) / 60000
```

imLinguin:
From my testing, Steam Deck doesn't completely suspend, so the event is not fired.  
We'll have to look for alternatives on detecting it

src/backend/main.ts
imLinguin: I'm not sure how backend would behave when we post sessions in different order, I'd rather keep it in sync

arielj:
@imLinguin I understand this PR is doing 2 things: option to disable playtime tracking, and also trying to not try time when the steam deck is suspended

Maybe we can divide this in 2 so we can merge the new setting part

imLinguin:
Oh yes, let's do that

imLinguin:
I'll rollback the changes related to suspend on this PR, we'll consider possible solutions in the future

src/frontend/screens/Settings/sections/GeneralSettings/index.tsx
```diff
@@ -19,14 +19,15 @@ import {
...
+  WinePrefixesBasePath,
+  PlaytimeSync
 } from '../../components'
 
 export default function GeneralSettings() {
   const { t } = useTranslation()
 
   return (
-    <>
+    <div>
```
arielj, line 30: is this change needed for any reason?

I think I used a fragment on purpose back then, to not have an extra div

src/backend/storeManagers/gog/games.ts
```diff
@@ -989,7 +989,9 @@ export async function updateGOGPlaytime(
     return
   }
 
-  const response = await postPlaytimeSession({ ...data, appName })
+  const response = await postPlaytimeSession({ ...data, appName }).catch(
+    () => null
```
arielj, line 993: maybe we can log that something failed here? or re-try posting the playtime?

src/backend/main.ts
```diff
@@ -1065,8 +1070,16 @@ ipcMain.handle(
       sessionPlaytime + tsStore.get(`${appName}.totalPlayed`, 0)
     tsStore.set(`${appName}.totalPlayed`, Math.floor(totalPlaytime))
 
-    if (runner === 'gog') {
-      await updateGOGPlaytime(appName, startPlayingDate, finishedPlayingDate)
+    const { disablePlaytimeSync } = GlobalConfig.get().getSettings()
+    if (!disablePlaytimeSync) {
+      if (runner === 'gog') {
```
arielj, line 1075: maybe we can move this if before checking if the sync is disabled? because you disable play sync this will show a message for games of every store and can be confusing (like... I finish playing an Epic game and I have a message saying it's skipped, but if I enable it then it doesn't post the info because it's only for gog)

src/frontend/screens/Settings/sections/GeneralSettings/index.tsx
imLinguin: This makes it consistent with other screens and fixes some styling issues (don't remember what exactly)

imLinguin: Without div
image

With div
image

Other sections like Advanced Settings have divs too, so I made it consistent

src/backend/storeManagers/gog/games.ts
imLinguin: Below that, if the response fails for whatever reason we add it to retry queue, which will be ran on next Heroic startup. Isn't that pretty much that? What else do you suggest?

src/frontend/screens/Settings/sections/GeneralSettings/index.tsx
arielj: thanks for the explanation

src/backend/storeManagers/gog/games.ts
arielj: oh I see, what confused me was that the rest of the code was already in the file and this catch is new, so I imagined it was doing something extra, so this is to make response `null` when something fails so `!response` in the next line would be true?

I understand before it was throwing an exception instead of returning null?

imLinguin: > I understand before it was throwing an exception instead of returning null?

Technically `postPlaytimeSession` shouldn't throw any exceptions, however if we were ever to change how the function behaves we have this another layer of safety

arielj (approved):
looks good, would be good to get it in sync with main so CI runs correctly just in case

Jozxyqk6:
Forgive me for commenting on a closed topic, but this option seems to only disable playtime _sync_ but leave the _tracking_ enabled. 

Any option to also disable the actual tracking?

PR #3006, opened 2023-08-23
callRunner: Use PowerShell to run commands on Windows

### LINKED ISSUE #2046: Launching games on Windows isn't properly tracked.
### Describe the bug

When launching a game on Linux
It will go from `Play Now` to `Stop (Playing)`
And it will stay that way until you close the game.
But on windows it does this:
It will go from `Play Now` to `Stop (Playing)` back to `Play Now`

I don't know what the difference is, but on windows Heroic fails to track the game and this causes several issues.

- If you have cloud saves enabled, the sync won't occur when you close the game, potentially causing you to lose lots of progress in game.
- It fails to track time played which is a big bummer.
- You can't actually close the game from the launcher

CommandMC:
PowerShell's `Start-Process` cmdlet & its `-Wait` parameter are used to wait for a process tree to exit. This means we can now accurately track the "Playing" status of games on Windows.

TODO:
- [X] Since we now modify `commandParts`, hiding login tokens in `getRunnerCallWithoutCredentials` no longer works.  
  Adding a check like `if (runnerPath === 'powershell')` should  *work*, although I don't exactly like this solution

Closes #2046 

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1011,7 +1011,7 @@ async function callRunner(
       .join(',')
     commandParts = [
       'Start-Process',
-      `${fullRunnerPath}`,
+      `"\`"${fullRunnerPath}\`""`,
       '-Wait',
       '-ArgumentList',
       argsAsString,
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -946,6 +946,8 @@ interface RunnerProps {
 
 const commandsRunning = {}
 
+let shouldUsePowerShell: boolean | null = null
+
 function appNameFromCommandParts(commandParts: string[], runner: Runner) {
   let appNameIndex = -1
   let idx = -1
@@ -1004,8 +1006,12 @@ async function callRunner(
   let fullRunnerPath = join(runner.dir, bin)
 
   // On Windows: Use PowerShell's `Start-Process` to wait for the process and
-  // its children to exit
-  if (isWindows) {
+  // its children to exit, provided PowerShell is available
+  if (shouldUsePowerShell === null)
+    shouldUsePowerShell =
+      isWindows && !!(await searchForExecutableOnPath('powershell'))
+
+  if (shouldUsePowerShell) {
     const argsAsString = commandParts.map((part) => `"\`"${part}\`""`).join(',')
     commandParts = [
       'Start-Process',
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -26,6 +26,7 @@ import {
   flatPakHome,
   isLinux,
   isMac,
+  isWindows,
   isSteamDeckGameMode,
   runtimePath,
   userHome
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
LGTM

arielj:
can we merge this? I don't use windows so I have no idea, are we waiting for more tests on windows?

CommandMC:
For one, the TODO in the PR description still needs to be addressed.
Other than that, I think the backticks for the case that the argument includes a space aren't quite correct

Kajot-dev:
I've tested it on windows and can confirm it works.

As of the TODO: It would require quite bit of refactoring, but all in the same file, so I think it's doable. First we would have to extract a few functions:

- `commandParts` sanity check from `getRunnerCallWithoutCredentials` to a separate `sanitizeCommandParts`
- `commandParts` conversion to powershell's Start-Process to `convertToPowershell`
- command formatting (env, and runnerPath) from `getRunnerCallWithoutCredentials` to `formatCommand`

This way `getRunnerCallWithoutCredentials` would call `saniztizeCommandParts` and then `formatCommand` and would have its functionality *unchanged*. However, `callRunner` would no longer call `getRunnerCallWithoutCredentials`, but instead `saniztizeCommandParts` then if its windows `convertToPowershell` twice (once for sanitized and not-sanitized `commandParts`) and lastly `formatCommand` for logging purposes.

TLDR:

Extract: `sanitizeCommandParts`, `convertToPowershell`, `formatCommand`

Changes in functions:
- `getRunnerCallWithoutCredentials` calls `sanitizeCommandParts`, then `formatCommand` (functionality unchanged)
- `callRunner` calls `sanitizeCommandParts`, then if windows `convertToPowershell` x2 (sanitized and not-sanitized) and lastly `formatCommand`

This way sanitization happens before converting to Powershell and without changes to the functionality of `getRunnerCallWithoutCredentials`

flavioislima (changes requested):
Tested here and legendary is broken since my user name has spaces on it. 
all commands are failing.

src/backend/launcher.ts
```diff
@@ -862,13 +863,32 @@ async function callRunner(
...
-  const fullRunnerPath = join(runner.dir, runner.bin)
   const appName = commandParts[commandParts.findIndex(() => 'launch') + 1]
 
   // Necessary to get rid of possible undefined or null entries, else
   // TypeError is triggered
   commandParts = commandParts.filter(Boolean)
 
+  let bin = runner.bin
+  let fullRunnerPath = join(runner.dir, bin)
+
```
flavioislima, line 874: Since we change dir when executing the `spawn` command bellow, changing the `fullRunnerPath` to be only the `runner.bin` fixes the issue I am having with all legendary commands failing due to spaces in my user name.

CommandMC:
I sure wouldn't *complain* if this would end up in the next release

The issue you mentioned should be resolved now. Sorry it took so long, somehow missed the review notification

arielj:
@flavioislima can you confirm if the issue you had is resolved?

flavioislima (approved):
looks good now

PR #3016, opened 2023-08-29
[FIX] Unbreak updating games list if user is logged in to Amazon account, but has no games installed.

rdbrschf:
If no game is installed, `nile list-updates` will return an empty string. Unfortunately, this is not valid JSON and `JSON.parse()` later bails out with an error.

We're good to go if no game is installed either way, so break out early and don't try to parse an empty string.

---

### DIFF
--- a/src/backend/storeManagers/nile/library.ts
+++ b/src/backend/storeManagers/nile/library.ts
@@ -166,6 +166,14 @@ export async function listUpdateableGames(): Promise<string[]> {
   )
   deleteAbortController(abortID)
 
+  if (!output) {
+    /*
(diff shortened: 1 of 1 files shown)


CommandMC:
IMO the "better" fix here would be modifying Nile so it *does* return valid JSON (to be specific, an empty array) in this case

rdbrschf:
That would also work, sure. But making Heroic resilient to empty strings also doesn't sound like a bad idea. I'd do both, really.

flavioislima (approved):
LGTM

rdbrschf:
X-ref for nile: https://github.com/imLinguin/nile/pull/41

PR #3031, opened 2023-09-05
[FIX] Shutdown wine when game exits

Etaash-mathamsetty:


CommandMC:
Did this actually *do* something? When calling a runner, we wait for `stdout`/`stderr` to be closed. This only happens once all Wine processes quit. So I wouldn't expect this to ever help, say, a stuck launcher or something

Closing for now, feel free to re-open/re-create if the change is helpful

PR #3035, opened 2023-09-09
[FIX] Auto enabling anti cheat runtimes

Etaash-mathamsetty:
This has already been merged in the past, but got reverted in some refactoring

### DIFF
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -641,11 +641,14 @@ export async function install(
 
     if (anticheatInfo.anticheats.includes('Easy Anti-Cheat')) {
       gameConfig.setSetting('eacRuntime', true)
-      if (isFlatpak) gameConfig.setSetting('useGameMode', true)
+      if (isFlatpak) {
+        gameConfig.setSetting('useGameMode', true)
+      }
     }
 
-    if (anticheatInfo.anticheats.includes('BattlEye'))
+    if (anticheatInfo.anticheats.includes('BattlEye')) {
       gameConfig.setSetting('battleyeRuntime', true)
+    }
   }
 
   return { status: 'done' }
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -637,12 +637,15 @@ export async function install(
   const anticheatInfo = gameAnticheatInfo(getGameInfo(appName).namespace)
 
   if (anticheatInfo && isLinux) {
-    const gameSettings = await getSettings(appName)
+    const gameConfig = GameConfig.get(appName)
 
-    gameSettings.eacRuntime =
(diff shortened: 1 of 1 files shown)


src/backend/storeManagers/legendary/games.ts
```diff
@@ -637,12 +637,15 @@ export async function install(
   const anticheatInfo = gameAnticheatInfo(getGameInfo(appName).namespace)
 
   if (anticheatInfo && isLinux) {
-    const gameSettings = await getSettings(appName)
+    const gameConfig = GameConfig.get(appName)
+
+    if (anticheatInfo.anticheats.includes('Easy Anti-Cheat')) {
+      gameConfig.setSetting('eacRuntime', true)
+      if (isFlatpak) gameConfig.setSetting('useGameMode', true)
```
flavioislima, line 644: We had issues in the past with this approach, is safer to just use brackets like in other places.

src/backend/storeManagers/legendary/games.ts
```diff
@@ -637,12 +637,15 @@ export async function install(
...
+    }
 
-    gameSettings.eacRuntime =
-      anticheatInfo.anticheats.includes('Easy Anti-Cheat')
-    if (gameSettings.eacRuntime && isFlatpak) gameSettings.useGameMode = true
-    gameSettings.battlEyeRuntime = anticheatInfo.anticheats.includes('BattlEye')
+    if (anticheatInfo.anticheats.includes('BattlEye'))
```
flavioislima, line 647: same here.

flavioislima (approved):
LGTM

Etaash-mathamsetty:
now that I think about we should just have this on by default...

gabriele2000:
> now that I think about we should just have these runtimes on by default... they can't cause crashes/issues as far as I know

I have, on `~/.profile`, two `export` lines that points to the EAC and BE runtimes, are you sure this "automatic" feature won't create conflicts like it did a year ago?

imLinguin:
The variables will be shadowed (overwritten) when Heroic launches the game exe. With your configuration even if the option is disabled in Heroic, the anti cheats are enabled. 

I don't really understand why you do it like that, but I'm sure you have your reasons

gabriele2000:
> I don't really understand why you do it like that, but I'm sure you have your reasons

There were some games on Steam that required manual anticheat runtime arguments and I eventually got tired of this.
Also, it was before Heroic introduced the toggle for EAC and BE, if I'm not mistaken.

What's the advantage of not having the runtime always loaded?

Etaash-mathamsetty:
> > I don't really understand why you do it like that, but I'm sure you have your reasons
> 
> There were some games on Steam that required manual anticheat runtime arguments and I eventually got tired of this. Also, it was before Heroic introduced the toggle for EAC and BE, if I'm not mistaken.
> 
> What's the advantage of not having the runtime always loaded?

that is my point, there are no advantages, so we should have it loaded by default

PR #3056, opened 2023-09-17
[Refactor] Refactor abort controllers to simplify creation/deletion

arielj:
We have this pattern in a lot of places:

```
import {
  createAbortController,
  deleteAbortController
} from './utils/aborthandler/aborthandler'

...
...

const abortId = 'some-id'
runSomeCommand(
  ...
  createAbortController(abortId)
  {
    ...options
  }
)
deleteAbortController(abortId)
```

This PR is a refactor of that to only need the abortId when calling the functions and the abort controllers are created/deleted internally. It removes a lot of imports and explicit calls to `createAbortController` and `deleteAbortController`.

So now we don't need the imports and we can do

```
runSomeCommand(
  ...
  {
    abortId: 'some-id',
    ...options
  }
```

The abortId is also now optional, if we want to run something that we don't need to abort manually we can ignore that option.

We can still abort things when needed since the abort controllers are still being tracked and created with the same ID as before.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -65,6 +65,10 @@ import { readFileSync } from 'fs'
 import { LegendaryCommand } from './storeManagers/legendary/commands'
 import { commandToArgsArray } from './storeManagers/legendary/library'
 import { searchForExecutableOnPath } from './utils/os/path'
+import {
+  createAbortController,
+  deleteAbortController
+} from './utils/aborthandler/aborthandler'
 
 async function prepareLaunch(
   gameSettings: GameSettings,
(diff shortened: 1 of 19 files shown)


src/backend/launcher.ts
```diff
@@ -826,6 +829,9 @@ async function callRunner(
     return currentPromise
   }
 
+  const abortId = options?.abortId || appName || Math.random().toString()
```
arielj, line 832: if no abortId is passed, it tries to get the appName from the command, else it generates a random id

src/backend/wine/manager/ipc_handler.ts
```diff
@@ -19,12 +15,7 @@ ipcMain.handle('installWineVersion', async (e, release) => {
       progress
     })
   }
-  const result = await installWineVersion(
-    release,
-    onProgress,
-    createAbortController(release.version).signal
-  )
-  deleteAbortController(release.version)
+  const result = await installWineVersion(release, onProgress)
```
arielj, line 18: since the abortId is `release.version`, I'm just reading that inside `installWineVersion` so no need of a parameter, but it's created the same way as before

arielj:
I'm merging this, I tested a lot of things and if we get this merged we can have more developers testing before a release

I don't want to start getting conflicts

PR #3058, opened 2023-09-17
[UI] Fix wide card being extra tall

arielj:
This PR should fix these issues for good (hopefully)

We had a problem when using a controller:
image

And this was reported on Discord:
image

To reproduce the first one, just use a gamepad and it looks like that. I couldn't reproduce the second one manually but I could artificially (removing some classes). The fix I added should fix this in any case though since now the size will only be applied if the `justPlayed` class is actually present on that first element.

---

### DIFF
--- a/src/frontend/screens/Library/components/GameCard/index.css
+++ b/src/frontend/screens/Library/components/GameCard/index.css
@@ -13,12 +13,16 @@
   aspect-ratio: 173/275;
 }
 
+.gameCard.gamepad {
+  aspect-ratio: 3/4;
+}
+
 .gameCard.justPlayed {
   aspect-ratio: 275/205;
 }
 
-.gameCard.gamepad {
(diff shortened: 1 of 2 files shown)


src/frontend/screens/Library/components/GameCard/index.css
```diff
@@ -13,12 +13,16 @@
...
+}
+
 .gameCard.justPlayed {
   aspect-ratio: 275/205;
 }
 
-.gameCard.gamepad {
-  aspect-ratio: 3/4;
+.gameCard.gamepad.justPlayed {
+  aspect-ratio: 328/205;
```
arielj, line 25: we need a different ratio for the gamepad layout cause we remove the part with the play/settings/update buttons

flavioislima (approved):
LGTM

PR #3063, opened 2023-09-20
[UX] Don't use linux anticheat info on MacOS

arielj:
This PR adds the less inaccurate anticheat information for MacOS, currently we are showing linux support which means a lot of false positives.

I created https://github.com/Heroic-Games-Launcher/MacAnticheatData/blob/main/games.json with the content from AreWeAnticheatYet but all EAC/Battleye support changed to `broken` and all the others to `Unknown`.

With this change, if users report some game works on mac we can update the json file without the need of a new release.

The widget is the same, except I don't show the link to AWACY and I removed all the notes and updates to avoid people reporting issues to that repo for Mac games.

---


flavioislima (approved):
Good stuff!

PR #3066, opened 2023-09-22
[UI] Frameless window option (with theme support)

0xCmdrKeen:
First PR, requesting feedback.

## Summary

This adds a new setting "Use frameless window", which, if active during startup, will make Heroic create a frameless main window (i.e. no titlebar, window controls overlaid onto web content, see https://www.electronjs.org/docs/latest/tutorial/window-customization#create-frameless-windows).

## Features

- Uses native overlay controls on macOS and Windows and custom overlay controls on Linux
- Makes the sidebar a draggable area on frameless windows (meaning windows can be moved around by dragging on the sidebar instead of the titlebar)
- The height and color of the overlaid window controls can be controlled by the theme, see Theming Support below
- Built-in themes have been updated to respect the the presence of overlay controls when running in frameless mode

## Theming Support

The following variables can be set in a custom theme in order to change the appearance of the window controls:

- `--titlebar-height: <number>` will set the height of the window controls in pixels
- `--titlebar-color: <hex color>` will set the background color of the window controls (defaults to `--body-background`)
- `--titlebar-symbol-color: <hex color>` will set the foreground color of the window controls (defaults to `--accent`)

Also, the following variables are available to CSS authors for accommodating the presence of overlay controls:

- `--overlay-controls-height`: the actual height of the overlay controls, if present
- `--overlay-controls-width`: the actual width of the overlay controls, if present

Note that these variables are set regardless of whether frameless mode is active. If it IS enabled, an additional class `frameless` is added to the `<App>` element. 

## Screenshots

Windows 11, using Old School Heroic theme. Header automatically accommodates for missing space on the right.
Screenshot 2023-09-24 224155

Linux, using Old School Heroic theme and custom window controls.
Screenshot from 2023-09-24 23-14-06

MacOS, using Old School Heroic theme. No header adjustments necessary since window controls are on the left.
Screen Shot 2023-09-24 at 10 49 20 PM 

## To Do

A couple of things I'm looking to improve on right off the bat:

- [x] Remove duplication of functionality from `getThemeCSS` handler in `main.ts#1578-1588`
- [x] Cleaner CSS parsing that doesn't rely on regular expressions
- [x] If possible, allow references to other variables, since theme authors will likely want to use existing color definitions
- [x] Add support for built-in themes
- [x] Handle platform dependency (Linux doesn't support overlay controls)

## Checklist


flavioislima:
Tested this one and for me it does not loads the windows controls on Linux:
Screenshot_20230929_135236

flavioislima:
Tried the appimage as well and same result. I am on KDE 5.27.

0xCmdrKeen:
@flavioislima Indeed, can confirm, looks like the most recent commit accidentally broke this. I just pushed a fix for this, would you mind giving it another try?

0xCmdrKeen:
@let-the-world-forget-the-flower thanks for the praise
> for linux controls i'd suggest a box icon for the maximize instead, since that's a better looking/more common design choice
Yeah I'm not too happy about that icon either but I couldn't find anything more appropriate in the icon sets that already used by the app.

For instance, in Font Awesome, the maximize icon look like this: [solid](https://fontawesome.com/icons/window-maximize?f=classic&s=solid), [regular](https://fontawesome.com/icons/window-maximize?f=classic&s=regular)
Here's the restore icon: [solid](https://fontawesome.com/icons/window-restore?f=classic&s=solid), [regular](https://fontawesome.com/icons/window-restore?f=classic&s=regular)

Sure, they're not terribly awful, but also don't really look like native icons either.

Meanwhile, in Material UI, it's inexplicably [just a horizontal bar like the minimize icon, just placed at the top](https://mui.com/material-ui/material-icons/?query=maxi&selected=Maximize), and there is no restore icon at all.

If you can find any better icons (preferably in SVG format), I'll be happy to use those.

0xCmdrKeen:
> could something like `square-2-stack` from https://heroicons.com/ and another square/rectangle icon from a different open source icon library work? (e.g. bootstrap icons)

Hmm, that's not too bad, would be even better if it was flipped but that can be achieved. There ARE various squares in the existing Material Icons set, let me give it a try.

0xCmdrKeen:
> Would be good to also add some drag-able space at the top of the window on Linux

The whole sidebar is draggable space on a frameless window ;)

You can also double click on it to maximize/unmaximize the window.

arielj:
I'm trying this on MacOS and when I turn on the frameless option and restart, heroic fails to render the react app

It shows the loading spinner for a moment and then I get this error
Screenshot 2023-09-30 at 9 48 47 AM

src/frontend/index.tsx
```diff
@@ -134,6 +135,20 @@ window.setTheme = async (themeClass: string) => {
   }
 
   document.body.className = themeClass
+
+  if (navigator['windowControlsOverlay']?.visible) {
+    const titlebarOverlay = Object.fromEntries(
+      ['height', 'color', 'symbol-color']
+        .map((item) => [
+          camelCase(item),
```
arielj, line 143: I think we are adding a new library just to call this camelCase here and it's only for the `system-color` string

I'd say to just hardcode the value we want

I think it's overkill to add 5 dependencies for this one place (camelCase adds other `case`-related dependencies)

arielj: also, it looks like we already have another `camelcase` dependency (note this one is `camelcase` not `camelCase`) from what I see in the yarn.lock diff

src/frontend/index.scss
```diff
@@ -25,6 +25,13 @@ body {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   user-select: none;
+  /* Frameless window defaults */
+  --titlebar-color: var(--body-background);
+  --titlebar-symbol-color: var(--accent);
+  --overlay-controls-height: env(titlebar-area-height, 39px);
```
arielj, line 31: I wonder how this plays with system settings, like if a windows users changes their system to have bigger buttons I imagine this will not follow what the user wants, maybe we need to make this configurable in the Accessibility screen?

Maybe not for this PR but for a future improvement

src/backend/main.ts
```diff
@@ -513,6 +522,17 @@ ipcMain.handle('checkDiskSpace', async (event, folder) => {
   })
 })
 
+ipcMain.handle(
+  'isFrameless',
+  () =>
```
arielj, line 527: maybe, instead of exporting `getWindowProps` from mainWindow, we could export an `isFrameless` function that would to this logic, I think it will look cleaner and we don't export the window props variable

src/frontend/index.tsx
0xCmdrKeen: Agreed. Didn't notice the existing camelcase package, but it looks like that's only depended on by development packages so I'm not sure it can be relied on to be present in a production build. I'll just hardcode this instead.

src/backend/main.ts
0xCmdrKeen: Well in that case, I would almost prefer to augment the `BrowserWindow` class in our main window with a method `isFrameless()` so I can write `getMainWindow()?.isFrameless()` and do without any additional exports, but I'm not sure if it's worth all of the TypeScript incantations required to make that happen.

src/frontend/index.scss
0xCmdrKeen: Good question, I'll look into it.

0xCmdrKeen: Okay, so the custom WindowControls component actually plays along nicely and scales with the zoom factor set in the accessibility settings. For the native overlay controls, I could add some code to adjust the height according to the zoom factor in the handler for the `setZoomLevel` event (main.ts#250-254), but it would require having access to the existing window props because `BrowserWindow` does not have a method to read them out.

What are your thoughts on this?

0xCmdrKeen:
> I'm trying this on MacOS and when I turn on the frameless option and restart, heroic fails to render the react app

Not really sure what happened there, it works fine for me on MacOS. I pulled in the latest changes and still no issues. But please LMK if you figure out how to reproduce it.

src/frontend/index.scss
0xCmdrKeen: Okay, I gave this a try but it looks like the size of the native controls cannot be adjusted at runtime like the documentation suggests (at least not on Windows). However, it also occurred to me that if someone is using the zoom function in the app they are likely also already using the zoom function built into Windows that allow making everything, including window controls, bigger.

If this is a serious concern, we could make the native window controls optional. Basically, when frameless mode is enabled, we could show another checkbox for using native window controls if this is supported by the OS.

arielj: We can see how it is now and wait for feedback, don't want that to be a blocker since I don't even know what the right thing to do there haha.

arielj:
> > I'm trying this on MacOS and when I turn on the frameless option and restart, heroic fails to render the react app
> 
> Not really sure what happened there, it works fine for me on MacOS. I pulled in the latest changes and still no issues. But please LMK if you figure out how to reproduce it, will definitely fix it.

yeah no idea, I can't reproduce this anymore

I'm only worried that this can happen to users but I don't know how to handle a fallback solution from the code to catch this (since I can't reproduce it anymore I don't know what we would need to check)

Maybe what we can do is put a warning in the check to let users know this is experimental? so they turn this on knowing it could have some unexpected issue. Then, after some time, if we have no bugs reported because of this we can just remove the warning?

src/frontend/index.scss
0xCmdrKeen: Well, adding another toggleswitch for this really isn't that complicated, already did so last night. Can push if you want me to.

0xCmdrKeen:
>I'm only worried that this can happen to users but I don't know how to handle a fallback solution from the code to catch this (since I can't reproduce it anymore I don't know what we would need to check)

Yeah that would be annoying, I agree.

> Maybe what we can do is put a warning in the check to let users know this is experimental? so they turn this on knowing it could have some unexpected issue. Then, after some time, if we have no bugs reported because of this we can just remove the warning?

There already is a system in place for experimental features, right? I'm just not quite sure how it works. Do I just move the checkbox from the Settings screen to Advanced Settings and place it in the "Experimental Features" section? Is the "enable shiny feature" checkbox just a placeholder for now (it appears that way)? And is there some kind of process in place for a feature to "graduate" from experimental to normal once it's been tested enough in the wild? It looks like those settings are stored under a separate key in the configuration file, so it seems a user would lose their current preferences if a setting is moved around in the config.

Either way, I'm happy to put the frameless option there, would be a good opportunity to get rid of the "shiny feature" placeholder :)

arielj:
> > I'm only worried that this can happen to users but I don't know how to handle a fallback solution from the code to catch this (since I can't reproduce it anymore I don't know what we would need to check)
> 
> Yeah that would be annoying, I agree.
> 
> > Maybe what we can do is put a warning in the check to let users know this is experimental? so they turn this on knowing it could have some unexpected issue. Then, after some time, if we have no bugs reported because of this we can just remove the warning?
> 
> There already is a system in place for experimental features, right? I'm just not quite sure how it works. Do I just move the checkbox from the Settings screen to Advanced Settings and place it in the "Experimental Features" section? Is the "enable shiny feature" checkbox just a placeholder for now (it appears that way)? And is there some kind of process in place for a feature to "graduate" from experimental to normal once it's been tested enough in the wild? It looks like those settings are stored under a separate key in the configuration file, so it seems a user would lose their current preferences if a setting is moved around in the config.
> 
> Either way, I'm happy to put the frameless option there, would be a good opportunity to get rid of the "shiny feature" placeholder :)

We have the experimental features feature but I think this wouldn't work with that since this is already a toggleable option, we don't need to go with an experimental feature that would make it more complex. I think just adding a warning when somebody toggles the option on is enough, we do something like that for the `Use dedicated GPU` option when only 1 GPU is detected, you can use that for reference.

I imagine a confirmation message like "This feature is experimental, report any issue in GitHub", just to be extra safe.

arielj:
one more comment, if you update the PR with the latest main branch it should fix the fontawesome-svg-core errors in the github actions

0xCmdrKeen:
> We have the experimental features feature but I think this wouldn't work with that since this is already a toggleable option, we don't need to go with an experimental feature that would make it more complex. I think just adding a warning when somebody toggles the option on is enough, we do something like that for the `Use dedicated GPU` option when only 1 GPU is detected, you can use that for reference.
> 
> I imagine a confirmation message like "This feature is experimental, report any issue in GitHub", just to be extra safe.

Ok, can do. I was wondering whether it would make sense to put a dialog box here anyways to ask the user whether they want to relaunch the application right away after changing this setting. Could put a notice on that and remove it later.

src/frontend/components/UI/WindowControls/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <SvgButton
+        className="minimize"
+        title={t('window.minimize', 'Minimize window')}
+        onClick={handleMinimize}
+      >
+        <MinimizeIcon />
+      </SvgButton>
+      <SvgButton
+        className="maximize"
+        title={t('window.maximize', 'Maximize window')}
```
arielj, line 42: I think we should make this say `Restore window` when it's maximized like we do with the icon change

src/frontend/screens/Settings/components/UseFramelessWindow.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { ToggleSwitch } from 'frontend/components/UI'
+import useSetting from 'frontend/hooks/useSetting'
+
+const UseFramelessWindow = () => {
```
arielj, line 6: one thing I just realized is that we probably want to NOT show this option on the Steamdeck in gaming mode, since the minimize/maximize buttons will be broken

we have a `isSteamDeckGameMode` export in constants.ts in the backend, we would need to get this information to the frontend

arielj (approved):
Tested on linux with cinnamon and it works all good here :+1:

I left 2 comments

I'm only a bit concerned that the window drag is now in the sidebar and I don't know how discoverable that is for users

maybe the cursor can change when hovering the draggable areas of the sidebar and not just when actually dragging?? that way when you are using the app, you hover, it changes the icon and you have a visual hint that you can drag

src/frontend/screens/Settings/components/UseFramelessWindow.tsx
0xCmdrKeen: Hmm okay, I don't have a SteamDeck so no way to test this, but does it make sense to add a whole new IPC function `isSteamDeckGameMode` for this case (plus pulling that flag into the `GlobalState` so it can be accessed from `GeneralSettings` without requiring any further contortions to deal with asynchronously polling a flag)? 

Or would it be enough to use the `isFullscreen` flag for this purpose, which is already in the context, and whose value is determined by either `isSteamDeckGameMode` or `isCLIFullscreen` being true (see `main.ts:658`)? Does it make sense to show this option when CLI fullscreen mode is enabled? And should we also make sure not to enable frameless mode in `main_window.ts` if either of these are true?

arielj:
One I idea I want to implement is to inject some static information already in the preload script when we create the main window.

if you add something like this in src/backend/preload.ts:

```ts
contextBridge.exposeInMainWorld(
  'isSteamDeckGameMode',
  process.env.XDG_CURRENT_DESKTOP === 'gamescope'
)
```

then you can use `window.isSteamDeckGameMode` in the frontend anywhere without any async call

these system information will never change so it's safe to just inject like that and way simpler

I want to include more things there in the future but I think it's fine to start with that now that it's needed and can be refactored later

0xCmdrKeen:
> One I idea I want to implement is to inject some static information already in the preload script when we create the main window.
> 
> if you add something like this in src/backend/preload.ts:
> 
> ```ts
> contextBridge.exposeInMainWorld(
>   'isSteamDeckGameMode',
>   process.env.XDG_CURRENT_DESKTOP === 'gamescope'
> )
> ```
> 
> then you can use `window.isSteamDeckGameMode` in the frontend anywhere without any async call
> 
> these system information will never change so it's safe to just inject like that and way simpler

Ah that's very interesting, I could have made use of that for the `isFrameless` flag as well instead of stringing that through the IPC layer and into the global context. I tried another approach using `webContents.executeJavaScript()`, but that ended up getting lost if the webContents were ever reloaded. I just tried this approach and it seems to stick.

Would it make sense to add another file to hold any other constants that might need to be made available on the frontend in the future? And should I use the same approach for `isFrameless` instead of the current one (since that can't change during runtime either)?

arielj:
> Would it make sense to add another file to hold any other constants that might need to be made available on the frontend in the future? And should I use the same approach for `isFrameless` instead of the current one (since that can't change during runtime either)?

That's what I want to do eventually so I think it's fine, anything that is constant that we need to expose to the frontend we could add that there and avoid a lot of boilerplate.

And I was thinking that even for non-constant values, we could always inject the initial state of the app with that to avoid the first IPC calls to get the initial states ready even before react loads.

0xCmdrKeen:
> > Would it make sense to add another file to hold any other constants that might need to be made available on the frontend in the future? And should I use the same approach for `isFrameless` instead of the current one (since that can't change during runtime either)?
> 
> That's what I want to do eventually so I think it's fine, anything that is constant that we need to expose to the frontend we could add that there and avoid a lot of boilerplate.
> 
> And I was thinking that even for non-constant values, we could always inject the initial state of the app with that to avoid the first IPC calls to get the initial states ready even before react loads.

Yeah, that would be neat, the problem is what to name that file? There already is a `constants.ts` in same directory, and I'm not sure it's safe to just expose all of those to the frontend. If they go on `window`, we need to make sure not to overwrite anything important that's already there. 

We could put another `constants.ts` in the `api` directory, but then they would all end up in the `window.api` namespace, which is already getting quite crowded (and might contain functions by the same name). Also wouldn't be super great to break pattern and load `api/constants.ts` directly into `window` in `preload.ts`. 

Not quite sure what to do here. I'm quite tempted to just put that snippet into `preload.ts` directly (and perhaps add `isFrameless` there as well) until we have a better idea in the future.

arielj:
> Not quite sure what to do here. I'm quite tempted to just put that snippet into `preload.ts` directly (and perhaps add `isFrameless` there as well) until we have a better idea in the future.

I'll start with this, just add:

```ts
contextBridge.exposeInMainWorld(
  'isSteamDeckGameMode',
  process.env.XDG_CURRENT_DESKTOP === 'gamescope'
)
```

to preload.ts, we can refactor that into other files if we need to in the future

flavioislima:
@0xCmdrKeen I believe we can merge this one before the next release, just need to fix the conflicts now.

Etaash-mathamsetty:
there's an issue with electron 27 where the buttons just scroll up with the content, would be nice if that got fixed, I think it's the only issue

0xCmdrKeen:
@Etaash-mathamsetty I can't seem to be able to reproduce this, however, after pulling the merge commit I am noticing that some screens are no longer using the full window height (see screenshot).

Screen Shot 2023-10-30 at 3 49 07 AM

0xCmdrKeen:
@Etaash-mathamsetty @arielj okay, this should take care of it.

Kajot-dev:
With this feature enabled "Exit to System Tray" stops working on Linux (though I've not tested it on other platforms).
While closing the window with native controls Heroic stays minimized in the system tray (as expected).
While closing the frameless window with custom window controls Heroic terminates even though "Exit to System Tray" is enabled.

Etaash-mathamsetty:
> With this feature enabled "Exit to System Tray" stops working on Linux (though I've not tested it on other platforms).
> While closing the window with native controls Heroic stays minimized in the system tray (as expected).
> While closing the frameless window with custom window controls Heroic terminates even though "Exit to System Tray" is enabled.
> 
> EDIT: But using Alt+F4 works correctly every time

Yep I can reproduce this, nice catch

0xCmdrKeen:
> While closing the frameless window with custom window controls Heroic terminates even though "Exit to System Tray" is enabled.

Hmm, I was using the same `handleQuit'` helper that the "Quit" button in the sidebar uses. Looks like that actually bypasses the `close` handler that would minimize to the tray instead. I added another IPC function that calls the `close` handler instead, that should fix it.

flavioislima:
I wonder, since this is an experimental feature, should it be under the `Experimental Features` Setting? 
image

flavioislima:
Another thing I found, not a huge deal since it is still experimental and optional. But it is not possible to resize the sidebar anymore. Also, the only way to move the window is by clicking don't the sidebar as well.

But that's ok, we can improve it later :)

0xCmdrKeen:
> I wonder, since this is an experimental feature, should it be under the `Experimental Features` Setting?

Yes, I also suggested this [here](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3066#issuecomment-1742312977), but @arielj was against it.

As far as not being able to resize the sidebar goes, that's a good catch. Might be a result of making it a draggable area, so I'm not entirely sure what if anything can be done about this, but honestly, I didn't even know the sidebar was resizable for the longest time, and I never really missed that feature, because the initial size seemed entirely appropriate to me.

arielj:
it's just that adding it as a experimental feature is more to enable disable components, feels weird to enable the feature to show the toggle and then have to also enable the toggle?

arielj:
also, the only reason I asked for that is because I had one issue that rendered heroic unusable, but I was not able to reproduce it in any way after that, so it was more to cover ourselves

0xCmdrKeen:
@arielj I'm not trying to blame you, just pointing out a part of the discussion that @flavioislima might have missed. 

Since it's just a simple toggle, it COULD go into the experimental features section, but I'm fine with either solution. Just cool to finally see this merged.

PR #3076, opened 2023-09-25
[FIX]: `searchForExecutableOnPath` hanging or false failing

Nocccer:
The function `searchForExecutableOnPath` had two major bugs:

- It resolve with null on stderr, even `which` reported the correct path. This happens for example inside steam, where calling `which` throws this Error `ERROR: ld.so: object '/home/niklas/.steam/debian-installation/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.` before but reports the correct path afterwards.
- It does not hang indefinitely if neither `stdout`, `stderr` or `error` is called

The solution is to wait till spawn calls `close` and resolve with null. If `stdout` is called we resolve earlier with the given path.

---

### DIFF
--- a/src/backend/utils/os/path/index.ts
+++ b/src/backend/utils/os/path/index.ts
@@ -11,13 +11,20 @@ async function searchForExecutableOnPath(
   executable: string
 ): Promise<string | null> {
   return new Promise<string | null>((resolve) => {
+    // no need to check stderr or error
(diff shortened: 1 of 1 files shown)


imLinguin (approved):
Looks good

PR #3089, opened 2023-09-27
[Feature]: Add gamescope support

Nocccer:
This PR adds support for gamescope configuration per game.
- Upscaling can be toggled on/off and configured
- Limiter can be toggled on/off and configured

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/61798668/fb6653b3-ccf4-46e9-ad80-653a134d4fae

---


flavioislima:
@Nocccer tested here with Flatpak and its working mostly fine. A few small improvements:
Some help texts are not really helpful, we could follow something similar to what is on Gamescope docs.
The scaling method should be select with FSR, NVidia Integer Scaling (if I am not wrong its open source and works on all GPUs but its worse than FSR in most cases I guess), and Integer.
Using Interger and FSR won't work I guess since Interger will keep the fullscreen window but setting the game window itself to the internal game resolution. For instance:

Game resolution 1440p upscaled to 4k, the game window does not use the whole screen when using interger:
Screenshot_20230929_132337

flatpak/templates/com.heroicgameslauncher.hgl.metainfo.xml.template
flavioislima, line 1: I think we dont need to update this file anymore, in fact, we can remove the prepareFlatpak script in total since I added the other github action that clone the flatpak repo and opens a PR there with all the updates, including the changelog as well.

flavioislima:
One bug I just found:
Games fail to launch if Gamescope is Disabled. 

image

Nocccer:
> Another thing, one good improvement would be also to check if gamescope is available and disable the `Summary` component and show a message on the side saying that Gamescope is not available.
> 
> Also, should we also add the Stretch scaling option? I wonder if people would use it to play 4:3 games from GOG or Amazon for instance and stretch it to 16:9.

Ok so i added the message if gamescope can not be found. I also wrapped upscale and limiter behind a toggle.
The design sucks. I am not that good at CSS and before i add tons of new components @arielj or you find better solutions.

Nocccer:
> One bug I just found: Games fail to launch if Gamescope is Disabled.
> 
> image

Should be fixed now.

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -0,0 +1,334 @@
...
+                />
+              }
+              onChange={(event: ChangeEvent<HTMLInputElement>) => {
+                setGamescope({
+                  ...gamescope,
+                  gameWidth:
+                    setResolution(event.currentTarget.value) ??
+                    gamescope.gameWidth
+                })
+              }}
```
arielj, line 153: Just a suggestion, we use this code in 4 places, maybe we could move this inside `setResolution` and do `setResolution('gameWidth', event.currentTarget.value)` for example

and do this `setGamescope....` code there instead

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -0,0 +1,334 @@
...
+            onChange={(event: ChangeEvent<HTMLSelectElement>) =>
+              setGamescope({
+                ...gamescope,
+                windowType: event.currentTarget.value
+              })
+            }
+            value={gamescope.windowType}
+          >
+            {['fullscreen', 'borderless'].map((opt, i) => (
+              <option key={i}>{opt}</option>
```
arielj, line 250:
```suggestion
              <option key={i} value={opt}>
                {t(`gamescope.windowType.${opt}`, opt)}
              </option>
```

We need to add this also somewhere before this return for the `yarn i18next` command:

```
// t('gamescope.windowType.fullscreen', 'Fullscreen')
// t('gamescope.windowType.borderless', 'Borderless')
```

We can do something similar for the upscaling method select, to say `AMD FidelityFX™ Super Resolution` instead of `fsr` for example?

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -0,0 +1,334 @@
...
+        <ToggleSwitch
+          htmlId="gamescopeUpscaleToggle"
+          value={gamescope.enableUpscaling || false}
+          handleChange={() =>
+            setGamescope({
+              ...gamescope,
+              enableUpscaling: !gamescope.enableUpscaling
+            })
+          }
+          title={t('setting.gamescope.enableUpscaling', 'Enables Upscaling')}
```
arielj, line 96:
```suggestion
          title={t('setting.gamescope.enableUpscaling', 'Enable Upscaling')}
```

arielj (commented):
If we add a class to the div in line 84 like `gamescopeSettings` we can fix the `(i)` alignment issue with something like:

```
.gamescopeSettings .selectFieldWrapper {
  grid-template-areas:
    'label info'
    'select select';
  grid-template-columns: max-content 1fr;
}

.gamescopeSettings .textInputFieldWrapper {
  grid-template-areas:
    'label info'
    'input input';
  grid-template-columns: max-content 1fr;
}
```

And we can also replace the divs with margin 10 used as separators with:

```
.gamescopeSettings .row {
  display: flex;
  gap: 10px;
}
```
And changing the `toggleRow` class with `.row`. So we don't use `toggleRow` for things components that are not toggles

src/backend/launcher.ts
```diff
@@ -121,6 +124,92 @@ async function prepareLaunch(
...
+    !isSteamDeckGameMode
+  ) {
+    const gameScopeBin = await searchForExecutableOnPath('gamescope')
+    if (!gameScopeBin) {
+      return {
+        success: false,
+        failureReason:
+          'Gamescope is enabled, but `gamescope` executable could not be found on $PATH'
+      }
+    }
```
arielj, line 139: I wonder if we should puts something in the logs and ignore gamescope instead of failing?

because if a user uninstalls gamescope and a game has it configured or it's configured in the defaults, there's no way for the user to disable gamescope form Heroic because we show `We could not found gamescope on the PATH. Install it or add it to the PATH.` and hide the gamescope settings.

Or maybe we can always show the gamescope settings but show a warnings at the beginning of that section saying the executable was not found, so users can disable it.

flavioislima: Heroic should check for gamescope at launch as well, so it can ignore the gamescope settings if was not found.
This would avoid breaking some games in this example.

But I also vote for the disabled setting instead of removing it, so we can be more clear.

Etaash-mathamsetty: yeah disabling the feature if no gamescope found is the best way in my opinion

flavioislima:
Tested here and the feature itself works pretty well (FPS limit won't work for me for some reason but I can see that the command is correct).

I believe some of the improvements like the style can be done on another PR, I will also need to move it to a tab anyway once I merge my tabs settings PR. So I think solving some of the other comments here should be fine to merge.

src/backend/launcher.ts
arielj: just to clarify, currently the feature is partially disabled when the binary is not found (it shows a message and no actual settings)

that is what would cause this issue: if you had that available, enabled it, and then uninstall gamescope, the settings are disabled but the game would fail to run and we have no way of disabling it since the settings are gone

so we have to either:
- not fail if it's enabled but not available
- not hide the settings if it's enabled, even if not available to allow disabling it (we could hide the settings if disabled and not available to prevent enabling it)
- do both things to be extra safe

flavioislima:
I think we can merge the PR like this and then improve the Style and message on another PR.
Sounds good @arielj @Nocccer

flavioislima:
One thing we need is a new component that is an input with the exclamation icon inside it, on the left or right, to show the help text. Outside like this is not fine. we can improve that before the next release though.

flatpak/templates/com.heroicgameslauncher.hgl.metainfo.xml.template
flavioislima: True, so we will need both then.

flavioislima (approved):
LGTM!
Good stuff!

public/locales/en/translation.json
```diff
@@ -533,6 +550,11 @@
         "FsrSharpnessStrenght": "FSR Sharpness Strength",
         "fsync": "Enable Fsync",
         "gamemode": "Use GameMode (Feral Game Mode needs to be installed)",
+        "gamescope": {
+            "enableLimiter": "Enable FPS Limiter",
+            "enableUpscaling": "Enables Upscaling",
+            "missingMsg": "We could not found gamescope on the PATH. Install it or add it to the PATH."
```
flavioislima, line 556:
```suggestion
            "missingMsg": "Heroic could not find gamescope on your system. Install it or add it to the PATH."
```

I think it is easier to understand this way.

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -0,0 +1,334 @@
...
+      />
+    )
+  }
+
+  if (!isInstalled) {
+    return (
+      <div style={{ color: 'red' }}>
+        {t(
+          'setting.gamescope.missingMsg',
+          'We could not found gamescope on the PATH. Install it or add it to the PATH.'
```
flavioislima, line 77:
```suggestion
          'Heroic could not find gamescope on your system. Install it or add it to the PATH.'
```

arielj:
> One thing we need is a new component that is an input with the exclamation icon inside it, on the left or right, to show the help text. Outside like this is not fine. we can improve that before the next release though.

with my suggested change, the `(i)` icon is next to the label like:

```
Game width (i)
[________________________]
```

I can push those changes, I already tested them

flavioislima:
> > One thing we need is a new component that is an input with the exclamation icon inside it, on the left or right, to show the help text. Outside like this is not fine. we can improve that before the next release though.
> 
> with my suggested change, the `(i)` icon is next to the label like:
> 
> ```
> Game width (i)
> [________________________]
> ```
> 
> I can push those changes, I already tested them

Fine by me, that works as well

flavioislima:
Merging this one as well and we can improve it afterwards

PR #3103, opened 2023-10-01
[FIX] Enable anti cheat runtimes by default

Etaash-mathamsetty:
This is probably the best way to do it, rather than just reading the anticheat data. Proton does the same thing for most titles I think


src/backend/config.ts
```diff
@@ -323,7 +324,9 @@ class GlobalConfigV0 extends GlobalConfig {
       winePrefix: isWindows ? '' : defaultWinePrefix,
       wineVersion: defaultWine,
       enableEsync: true,
-      enableFsync: true
+      enableFsync: true,
+      eacRuntime: isLinux,
```
Nocccer, line 328: Probably for native linux games this should be false aswell.

Nocccer (commented):
As long as it does not use more free space on each game we could enable it always. How big are this runtimes and are they copied over to the prefix?

Etaash-mathamsetty:
> As long as it does not use more free space on each game we could enable it always. How big are this runtimes and are they copied over to the prefix?

It doesn't take any additional storage

src/backend/config.ts
Etaash-mathamsetty: Anti cheat runtime settings are only applied for non native games, so it doesn't matter.

Etaash-mathamsetty:
updated to automatically download the runtimes as well

src/backend/config.ts
```diff
@@ -323,7 +324,9 @@ class GlobalConfigV0 extends GlobalConfig {
       winePrefix: isWindows ? '' : defaultWinePrefix,
       wineVersion: defaultWine,
       enableEsync: true,
-      enableFsync: true
+      enableFsync: isLinux,
```
Etaash-mathamsetty, line 327: fsync uses a linux specific API and can only should be enabled by default on linux

src/backend/launcher.ts
```diff
@@ -269,6 +270,14 @@ async function prepareWineLaunch(
     }
   }
 
+  if (gameSettings.eacRuntime && !isInstalled('eac_runtime')) {
+    download('eac_runtime')
+  }
+
+  if (gameSettings.battlEyeRuntime && !isInstalled('battleye_runtime')) {
+    download('battleye_runtime')
```
arielj, line 278: should we `await` these calls? so we don't launch the game before they are downloaded

also, should we handle the case when something fails to download? like no internet connection for example, since I understand that setting to true when it's not available is a problem

flavioislima: agree on that as well.

Etaash-mathamsetty: if you are offline then the anti cheat not running doesn't matter in most cases right?

CommandMC (changes requested):
You'll also have to auto-enable GameMode if we're on Flatpak

src/backend/config.ts
CommandMC: This may be nitpicky, but I don't think this should be in this PR

Etaash-mathamsetty:
> You'll also have to auto-enable GameMode if we're on Flatpak

gamemode is always installed on the flaptak right?

CommandMC:
Yes, the Flatpak always installs it. See [com.heroicgameslauncher.hgl.yml#L111](https://github.com/flathub/com.heroicgameslauncher.hgl/blob/b9d84ca5d03d35d1d0c5602b180ee0a9f9b46adc/com.heroicgameslauncher.hgl.yml#L111)

kohend (approved):
Looks good to me

flavioislima (approved):

PR #3106, opened 2023-10-03
[Fix] Fix "Run Game Offline" not working on windows and mac

arielj:
In the prepareLaunch function we have an early return if not linux and it was only returning the discord presence flag but not the offline flag, so even with the setting the launch commands were not created with this argument. It only worked on Linux since it didn't do that early return.

---


flavioislima (approved):
Good catch

PR #3110, opened 2023-10-04
[Refactor] Introduce Recoil to start migrating to a more efficient state management

arielj:
We have discussed this issue many times: the global state controls a lot of different global values and updating any of those triggers unnecessary re-renders throughout the app because it uses the Context API.

We talked many times about migrating to a more efficient state manager to be able to re-render only the components affected for a state changing instead.

I checked many options and Recoil https://recoiljs.org seems to be simple and it feels really close to `useState`, it's really easy to migrate to this tool without a lot of changes and we can do it gradually.

In this PR I'm migrating the epic library and username state to Recoil while keeping everything else in the GlobalState. With this approach we can slowly move different states at a time to Recoil in small targeted PRs until we move everything.

Any new global state we want to add we can now do it efficiently using Recoil (no need to actually update GlobalStateFunctional, we just need to add a new atom and use it with either useRecoilState or useRecoilValue where it's needed).

One limitation of Recoil is that it requires functional components and the GlobalState component is a class component, so, temporarily, we can use the new GlobalStateFunctional wrapper component to use recoil state and pass the state as props to GlobalState. Eventually we will be able to remove GlobalState completely because it will be migrated to a function component (which is also something we talked about doing many times).

This PR is really small cause it's the first, I plan to do some PRs moving more than one state at a time to not need dozens of PRs to move everything, but I plan to group them in a way that we could target the review/QA to specific places.

I imagine we'll still use the Context Api for some things, GlobalState doesn't only share state but also functions, and the SettingsContext and GameContext are already scoped to specific parts of the app, but for global state I'd like to move everything to be either a Recoil state OR a window object property (some things could be injected directly into the app with the preload.ts script since some things never change and don't need to be reactive).

---

### DIFF
--- a/src/frontend/components/UI/SearchBar/index.tsx
+++ b/src/frontend/components/UI/SearchBar/index.tsx
@@ -13,6 +13,8 @@ import './index.scss'
 import { faXmark } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { GameInfo } from '../../../../common/types'
+import { epicState } from 'frontend/state/epic_state'
+import { useRecoilValue } from 'recoil'
 
 function fixFilter(text: string) {
   const regex = new RegExp(/([?\\|*|+|(|)|[|]|])+/, 'g')
@@ -26,7 +28,8 @@ const RUNNER_TO_STORE = {
 }
 
 export default React.memo(function SearchBar() {
-  const { handleSearch, filterText, epic, gog, sideloadedLibrary, amazon } =
+  const epic = useRecoilValue(epicState)
+  const { handleSearch, filterText, gog, sideloadedLibrary, amazon } =
     useContext(ContextProvider)
   const { t } = useTranslation()
   const navigate = useNavigate()
--- a/src/frontend/screens/Login/components/SIDLogin/index.tsx
+++ b/src/frontend/screens/Login/components/SIDLogin/index.tsx
@@ -16,7 +16,7 @@ interface Props {
 export default function SIDLogin({ backdropClick }: Props) {
   const epicLoginUrl = 'https://legendary.gl/epiclogin'
 
-  const { epic } = useContext(ContextProvider)
+  const { epicLogin } = useContext(ContextProvider)
   const { t } = useTranslation('login')
   const [input, setInput] = useState('')
   const [status, setStatus] = useState({
@@ -38,7 +38,7 @@ export default function SIDLogin({ backdropClick }: Props) {
       loading: true,
       error: false
     })
-    await epic.login(sid).then(async (res) => {
+    await epicLogin(sid).then(async (res) => {
       console.log(res)
       if (res === 'done') {
         await window.api.getUserInfo()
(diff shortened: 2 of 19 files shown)


src/frontend/components/UI/SearchBar/index.tsx
```diff
@@ -13,6 +13,8 @@ import './index.scss'
 import { faXmark } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { GameInfo } from '../../../../common/types'
+import { epicState } from 'frontend/state/epic_state'
```
arielj, line 16: we can refactor these imports when more are added with an index file

src/frontend/components/UI/SearchBar/index.tsx
```diff
@@ -13,6 +13,8 @@ import './index.scss'
...
+import { useRecoilValue } from 'recoil'
```
arielj, line 17: we use useRecoilValue to read a state

src/frontend/screens/Login/components/SIDLogin/index.tsx
```diff
@@ -16,7 +16,7 @@ interface Props {
 export default function SIDLogin({ backdropClick }: Props) {
   const epicLoginUrl = 'https://legendary.gl/epiclogin'
 
-  const { epic } = useContext(ContextProvider)
+  const { epicLogin } = useContext(ContextProvider)
```
arielj, line 19: the GlobalState object still exposes the epic login and logout functions since that's not state

src/frontend/state/GlobalState.tsx
```diff
@@ -56,14 +58,12 @@ interface Props {
   children: React.ReactNode
   i18n: i18n
   t: T
+  epic: EpicState
+  setEpic: SetterOrUpdater<EpicState>
```
arielj, line 62: we need to add these props temporarily until we migrate everything that needs getters and setters inside GlobalState

src/frontend/state/GlobalStateFunctional.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { epicState } from './epic_state'
+import { useRecoilState } from 'recoil'
+import GlobalState from './GlobalState'
+
+interface Props {
+  children: JSX.Element
+}
+
+export function GlobalStateFunctional({ children }: Props) {
```
arielj, line 10: wrapper around GlobalState to be able to use Recoil

src/frontend/state/epic_state.ts
arielj, line 1: this is called `epic_state` now but we can group multiple states in a single file, I'll probably replace this with `stores_state.ts` handling epic, gog, amazon, sideloaded games, all in a single file since they are all similar

flavioislima:
That's interesting but Recoil still does not convince me as a stage management when compared with Mobx or Redux. It still has same flaws as the React Context related with re-renders, at least from some articles I've read some time ago.

Brett is currently refactoring the GlobalState to seu Mobx and looks promising, I am willing to implement it here and it is pretty simple to use and especially to test a State using Mobx. Have you tried it before?

arielj:
I'll close this PR for now, but I think we should still make a decision at some point

moving into a different state manager that can handle a global state more efficiently allows us to do a few things that we can't do now (for example, if you download a wine and you move out of the wine manager screen while it's being unzipped, if you go back to the wine manager screen that status is lost because the state is kept in the component and not globally, there are many places in the app that have this problem and a global state can solve it, but we can't put everything in the global context today)

PR #3128, opened 2023-10-11
[FIX-Linux] Snaps Issues Part 1

flavioislima:
Fixes:

- [x] Legendary access permissions.
- [x] Get rid of `curl` dependency (#3129) 
- [x] Get rid of `tar` dependency (#3181)  since Snaps does not extract tar.xz files by default.
- [x] Text not showing correctly on native dialogs (showing squares instead). Need to run on the terminal: `rm ~/.cache/fontconfig/* ~/snap/heroic/common/.cache/fontconfig/*`

Current Discussions:
https://forum.snapcraft.io/t/help-fixing-heroic-games-launcher-snap/37248/5
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/discussions/3127

### For Part 2

- [ ] Access to Steam and Lutris folders to look for Proton and Wine.
- [ ] Gamemode
- [ ] Mangohud
- [ ] Gamescope

---


flavioislima:
> Hey, these changes probably are not needed. And as said before in the forum, please don't use electron-builder. It'll be worse to maintain later. Please.

It is actually worse for us to maintain a separate thing, if we can do it using electron-builder would be easier since we don't need to change our workflows.

Because otherwise we will need a specific workflow just for snaps and find a way to automate the update and release of it on every release as well. 

So unless it is impossible to solve with it then we would try a different way.
I think to fix the other issues now I just need to make the layouts work, right? so I can symlink the binaries on the system to snap?

soumyaDghosh:
```
name: heroic
base: core22 
version: '2.9.2' 
summary: Single-line elevator pitch for your amazing snap 
description: |
  This is my-snap's description. You have a paragraph or two to tell the
  most important story about your snap. Keep it under 100 words though,
  we live in tweetspace and your description wants to look good in the snap
  store.

grade: devel
confinement: devmode 
compression: lzo

parts:
  heroic:
    plugin: dump
    source: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/releases/download/v$SNAPCRAFT_PROJECT_VERSION/heroic_$SNAPCRAFT_PROJECT_VERSION_$CRAFT_TARGET_ARCH.deb
  
  deps:
    plugin: nil
    stage-packages:
      - curl
    prime:
      - usr/bin/curl

plugs:
  dot-local-share:
    interface: personal-files
    write:
      - $HOME/.local/share/Steam
      - $HOME/.local/share/lutris
      - 
apps:
  heroic:
    command: opt/Heroic/heroic
    desktop: usr/share/applications/heroic.desktop
    environment:
      TMPDIR: $XDG_RUNTIME_DIR
    extensions:
      - gnome
```

Bringing up my latest manifest here again to compare. Heroic is a game launcher, a very sensitive app, that needs to be fast and efficient. Now, here with this manifest, `curl` is working prefectly. I can download wine and proton. Now here, check that I only allowed the `curl` binary to stage and no other file, this way I can keep the thing in check. Now, regarding the workflow, what I am thinking to build is

1. New Tag made, Deb package made
2. After that the github workflow will chaneg the Line 3, `version: 2.9.2` to the latest tag
3. The snap package will be made by the snapcraft builder, released to snapcraft directly from the github workflow

Another alternative way, not using github workflow to build the snap

After 2:
This repository will be connected with snapcraft website
It'll auto build from the latest changed manifest, when the manifest changes, if not, just click the `Trigger New Build` button and it'll build and release to the edge channel.

flavioislima:
@soumyaDghosh yes I think electron-builder might now allow those options and we will need to go with our own snapcraft file then.

The way that you described could work, but then we trigger the workflow on `published` like we are doing for the flatpak and the AUR which are totally automated now as well. The snap should be something similar to what we do on the AUR, maybe more simpler since we don't need the even check the md5sum of the package for instance.

About maintaining the Snap, I am fine with it, no problem. But even that might not be needed if we can automate it.

soumyaDghosh:
Forgot to mention, Electron Builder uses `core20` where as `core22` is the current ongoing base and soon `core24` will be there based on Ubuntu 24.04 LTS. Latest bases are necessary for latest software drivers etc. Also, core22 brings significant performance boost for Wayland.

flavioislima:
> Forgot to mention, Electron Builder uses `core20` where as `core22` is the current ongoing base and soon `core24` will be there based on Ubuntu 24.04 LTS. Latest bases are necessary for latest software drivers etc. Also, core22 brings significant performance boost for Wayland.

Updated electron-builder to the latest version and using core22 now works fine. 
Testing here your snapcraft.yml as well.

soumyaDghosh:
My snap manifest isn't complete yet. I most probably hae one with most of the things working. I will share it soon

ashuntu:
Is there a reason you are not staging `curl`? I.e. `stagePackages: ["default", "curl"]`

flavioislima:
> @flavioislima Is there a reason you are not staging `curl`? I.e. `stagePackages: ["default", "curl"]`

I tried but it didn't work. I will try again but on Ubuntu.

flavioislima:
> question about the curl access, do we need curl? what I mean is... we have node, it should be able to do anything we are using curl for without curl
> 
> maybe it's easier to remove the curl dependency than to fix the access to curl?

For sure. We can use EasyDL lib to download all packages, would be a medium size pr to refactor everything but it's possible.

But we still need to figure it out the rest of the plug-ins like mangohud, etc.

flavioislima:
I will modify this PR so it will generate a snap to download so we can test it.

src/backend/storeManagers/legendary/library.ts
```diff
@@ -666,7 +666,11 @@ export async function runRunnerCommand(
   if (!options.env) {
     options.env = {}
   }
-  options.env.XDG_CONFIG_HOME = dirname(legendaryConfigPath)
+
+  // if not on a SNAP environment, set the XDG_CONFIG_HOME to the same location as the config file
+  if (!process.env.SNAP) {
+    options.env.XDG_CONFIG_HOME = dirname(legendaryConfigPath)
```
Etaash-mathamsetty, line 672: does chaning XDG_CONFIG_HOME under snap cause issues? if it does does it still cause issues if a different variable is used?

imLinguin:
> For sure. We can use EasyDL lib to download all packages, would be a medium size pr to refactor everything but it's possible.


Don't we have axios that's capable of that already?

soumyaDghosh:
> @flavioislima Is there a reason you are not staging `curl`? I.e. `stagePackages: ["default", "curl"]`

All these uncertainties are the reason, I am still on the side of a custom manifest.

src/backend/storeManagers/legendary/library.ts
flavioislima: Yes, it breaks the SNAPs because then legendary was trying to write its configs to a place that it is not possible so it was giving permission issues.

flavioislima:
> > For sure. We can use EasyDL lib to download all packages, would be a medium size pr to refactor everything but it's possible.
> 
> Don't we have axios that's capable of that already?

Yes but Axios was not intended to download files, it is possible but it is not stable. EasyDL is a pretty good library that does that and supports progress, split files into chunks and continuing to download in case something fails, different from axios that does not have any of these features.

flavioislima:
> > @flavioislima Is there a reason you are not staging `curl`? I.e. `stagePackages: ["default", "curl"]`
> 
> All these uncertainties are the reason, I am still on the side of a custom manifest.

Tried here on Ubuntu and on Arch (with multipass). The curl package is successfully installed but Heroic still cannot access it.
When I try adding gamescope, mangohud or gamemode to the list of stage packages, Snapcraft fails saying that the packages are not available.
So I am starting to think that the only way would be for Heroic to use Classic confinement but it does not meet the requirements for the Snap Store team to approve it.

This leads me to two choices:

1. Refactor the download part of Heroic that uses Curl so it is not required anymore and show a message to the users saying that some features might not work or are not available on Snaps and recommend it to use a different package if they need that (for Mangohud, Gamemode, and Gamescope), those are not really a HUGE deal to most people so I believe that is fine (perhaps Mangohud only).
2. Drop the Snap support for good, which I think it is not the best decision since it is the default install method on Ubuntu and Derivates.

For 1, I still need approval from the Snap Store to be able to access those home folders we need. I just tried to publish the package on Edge channel and they denied it because of these new accesses.

ashuntu:
I still would be on the side of a custom snapcraft yaml, as I think you'd have trouble getting GameMode and MangoHUD to work only by adding their packages via electron-builder. Maybe in the meantime, a message to users about functionality wouldn't be a bad idea, though.

We include GameMode and MangoHUD as separate parts in the Steam Snap if you're looking for inspiration there (https://github.com/canonical/steam-snap/blob/main/snap/snapcraft.yaml#L250-L310). Steam is a finicky case since we force the client to run in 32-bit, so functionality isn't very consistent. I'd be curious to see how well something similar would work for Heroic.

flavioislima:
> I still would be on the side of a custom snapcraft yaml, as I think you'd have trouble getting GameMode and MangoHUD to work only by adding their packages via electron-builder. Maybe in the meantime, a message to users about functionality wouldn't be a bad idea, though.
> 
> We include GameMode and MangoHUD as separate parts in the Steam Snap if you're looking for inspiration there (https://github.com/canonical/steam-snap/blob/main/snap/snapcraft.yaml#L250-L310). Steam is a finicky case since we force the client to run in 32-bit, so functionality isn't very consistent. I'd be curious to see how well something similar would work for Heroic.

Ok, so for now I will remove the need of curl from the code and fix these other issues and in the future we try to bring these other tools to the snap package.

flavioislima:
I am removing the `curl` dependency here on this PR:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3129

soumyaDghosh:
> I still would be on the side of a custom snapcraft yaml, as I think you'd have trouble getting GameMode and MangoHUD to work only by adding their packages via electron-builder. Maybe in the meantime, a message to users about functionality wouldn't be a bad idea, though.
> 
> We include GameMode and MangoHUD as separate parts in the Steam Snap if you're looking for inspiration there (https://github.com/canonical/steam-snap/blob/main/snap/snapcraft.yaml#L250-L310). Steam is a finicky case since we force the client to run in 32-bit, so functionality isn't very consistent. I'd be curious to see how well something similar would work for Heroic.

@flavioislima So, should I share a PR on the snap workflow?

flavioislima:
> > I still would be on the side of a custom snapcraft yaml, as I think you'd have trouble getting GameMode and MangoHUD to work only by adding their packages via electron-builder. Maybe in the meantime, a message to users about functionality wouldn't be a bad idea, though.
> > We include GameMode and MangoHUD as separate parts in the Steam Snap if you're looking for inspiration there (https://github.com/canonical/steam-snap/blob/main/snap/snapcraft.yaml#L250-L310). Steam is a finicky case since we force the client to run in 32-bit, so functionality isn't very consistent. I'd be curious to see how well something similar would work for Heroic.
> 
> @flavioislima So, should I share a PR on the snap workflow?

sure

flavioislima:
> > > > I still would be on the side of a custom snapcraft yaml, as I think you'd have trouble getting GameMode and MangoHUD to work only by adding their packages via electron-builder. Maybe in the meantime, a message to users about functionality wouldn't be a bad idea, though.
> > > > We include GameMode and MangoHUD as separate parts in the Steam Snap if you're looking for inspiration there (https://github.com/canonical/steam-snap/blob/main/snap/snapcraft.yaml#L250-L310). Steam is a finicky case since we force the client to run in 32-bit, so functionality isn't very consistent. I'd be curious to see how well something similar would work for Heroic.
> > > 
> > > 
> > > @flavioislima So, should I share a PR on the snap workflow?
> > 
> > 
> > sure
> 
> Problem, I want to test this pr, but where can I get the deb file from?

Sorry, missed your reply. I will change this PR so it will generate a .deb during build

flavioislima:
Added message with warning of missing features:
image

soumyaDghosh:
@flavioislima sorry to inform you that heroic being a big project needs a lot of attention and problem is my sem exam is coming soon. So, I'm halting on this thing now. I think if you don't mind, we can restart again after December!

flavioislima:
> @flavioislima sorry to inform you that heroic being a big project needs a lot of attention and problem is my sem exam is coming soon. So, I'm halting on this thing now. I think if you don't mind, we can restart again after December!

No problem. This is also getting more complicated. The app cannot even extract a tar.xz file since the core app is so limited on Snap. So I also need to change the extraction methods to use a npm package instead of native `tar` command.

soumyaDghosh:
> > @flavioislima sorry to inform you that heroic being a big project needs a lot of attention and problem is my sem exam is coming soon. So, I'm halting on this thing now. I think if you don't mind, we can restart again after December!
> 
> No problem. This is also getting more complicated. The app cannot even extract a tar.xz file since the core app is so limited on Snap. So I also need to change the extraction methods to use a npm package instead of native `tar` command.

Those are possible, and can be done. What I suggest now is to hold any code changes just for the sake of snap. It'll later make a hell for the package.

flavioislima:
> 
> Those are possible, and can be done. What I suggest now is to hold any code changes just for the sake of snap. It'll later make a hell for the package.

Yeah, but at the same time having the extraction, download, etc. Done from Heroic itself will be better since we won't rely on external dependencies and packages. So in the end these are necessary refactors.

flavioislima:
Merging this one. Should at least fix part of the issues. 
Will continue with more fixes afterwards. Perhaps only by using a snapcraft.yml build file we can have a 100% fully working app.

PR #3129, opened 2023-10-12
[Tech] Refactor Download Methods

flavioislima:
Replace all methods that use `curl` with the `easydl` lib.
EasyDL uses only native nodejs methods so has no dependencies. Split all files into chunks, supports resuming interrupted downloads, and has built-in progress report.

With this, we remove one package dependency from the app and also fix the wine manager on Snaps.

Also removed the downloaded and average speed from the wine manager since it was too bloated with info and breaking the line. So I kept only percentage and ETA.

---


flavioislima:
fixing tests...

flavioislima:
I tried to update the tests for downloading and installing but they won't work since the download is handled by an external lib and it does not support `file://` for URL.

Tried to solve it in different ways but realized that we would just adjust the code for the tests to pass. 
Perhaps an e2e tests makes more sense for this page instead of jest.

Issue #3126, opened 2023-10-10
Remote debugging is enabled by default on release builds


C-512L:
### Describe the bug

Heroic right now listens for remote debugging by default. I noticed this while looking at #3079, specifically at [this line](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/1020e7e2d6fb5379613f8d96b3d61036c7cb0d00/src/backend/main.ts#L155C1-L155C63). I think this would be better disabled on release builds since the average user would not need this and opens the possibility of exploiting it maliciously. If someone wants to debug, it should be explicitly passed to the launch args.

### Add logs

```text
Not applicable
```

### Steps to reproduce

1. Open the launcher
2. Check  with `ss -tlpn` if heroic is listening on remote debugging

### Expected behavior

The launcher not enabling remote debugging without passing manually a command line flag.

### Heroic Version

Latest Stable (Flatpak)

### System Information

Not applicable

PR #3130, opened 2023-10-13
Remove remote debugger access

arielj:
This PR removes the configuration that enabled the chrome remote debugger access.

It was adding during the stores refactor.

We don't really need it for development since we already turn on the dev tools inside heroic, and it's insecure to have that enabled in production releases since other apps can access information without the user knowing. Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3126

---


Etaash-mathamsetty (approved):
lgtm

PR #3132, opened 2023-10-13
Delete old logs, keep only the last 2

arielj:
This PR changes the logic to delete old log files. Instead of deleting things older than 1 month ago, we delete everything except the last 2 launches of the app.

Older logs haven't really been useful when debugging problems, there's no need to keep them.

I did include 2 more changes:
- I moved the date to the beginning of the filename, so it's easier to see the logs for the different stores from the same date
- I removed the milliseconds from the datetime, we don't need that precision and it just makes the filename harder to read

---


flavioislima (approved):
LGTM

PR #3133, opened 2023-10-13
[DX] Require Node 16 or greater for development

arielj:
We need at least Node 16 because we use the `replaceAll` function for strings.

Adding this will prevent wasting the time I wasted trying to figure out why a call to `replaceAll` was failing for me haha.

---


flavioislima (approved):
LGTM

flavioislima:
Actually, 18 would be ideal since it is the current LTS.

But since 16 its not EOL, so might be fine.

PR #3135, opened 2023-10-14
[UX] Help component

arielj:
This PR adds a component to have Help for users built in into the app.

We can use this to add information about the components in the interface, details about how to use specific screens or features, anything we want. The idea is to allow people to find answers on their own without leaving the app and also adds an opportunity to translate all that information.

I added it as an experimental feature for now, I only added a few basic strings.

**Important:**
- ~it needs Electron 26 to be able to use the popover API, this is needed to be able to show an element ABOVE dialogs like the install or setting dialogs, since those are in the top layer and we need this new API to put elements in there~
- ~because of this I understand it wouldn't work on some machines so this is not to merge until we figure out how to upgrade Electron safely~

This is a super basic example of how it works since I added only a few things there:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/188464/507f79c5-810e-41c6-9f66-3403f9b0fed9

We can put anything for the element's content using the `hasHelp` hook.

---


src/backend/main.ts
```diff
@@ -491,7 +491,7 @@ ipcMain.on('unlock', () => {
     unlinkSync(join(gamesConfigPath, 'lock'))
     if (powerId) {
       logInfo('Stopping Power Saver Blocker', LogPrefix.Backend)
-      return powerSaveBlocker.stop(powerId)
+      powerSaveBlocker.stop(powerId)
```
arielj, line 494: this is needed because of the electron upgrade

Etaash-mathamsetty:
electron 27 works properly, 26 has known bugs. I think 27 will be released soon, so it should be safer to upgrade to it. there's still the issue with the gamepad navigation on electron 27.

arielj:
> electron 27 works properly, 26 has known bugs. 27 has been released, so it should be safer to upgrade to it. there's still the issue with the gamepad navigation on electron 27.

I see, looks like it has issues focusing elements inside an element with scroll when using the spatial navigation

the gamepad works fine, the issue is the spatial navigation (same issue happens moving with the keyboard)

the controls work fine in screens that are not scrollable like the login or the game page

do you know if there's a reported issue for that problem?

arielj:
I found a workaround for the problem, I made a PR to upgrade to Electron 27 and once that's done I can update this one https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3136

flavioislima:
> I found a workaround for the problem, I made a PR to upgrade to Electron 27 and once that's done I can update this one #3136

do you want to merge this PR before the next release?

arielj:
> > I found a workaround for the problem, I made a PR to upgrade to Electron 27 and once that's done I can update this one #3136
> 
> do you want to merge this PR before the next release?

nono, it's too barebones, it's gonna be useless for everyone the way it is now, not even useful as an experimental feature to release yet

I think it can be merged after the next release once I fix the conflicts and that issue that I needed electron 27 to fix to have that for us to test more internally and to start adding help content

.eslintrc
```diff
@@ -35,7 +35,15 @@
...
     "react/no-unknown-property": [
       "error",
-      { "ignore": ["partition", "allowpopups", "useragent"] }
+      {
+        "ignore": [
+          "partition",
+          "allowpopups",
+          "useragent",
+          "popover",
+          "popovertarget"
```
arielj, line 44: typescript doesn't understand these properties, they are too new

flavioislima (approved):
lgtm

PR #3147, opened 2023-10-21
[UX] Show runners and installed games logs in Settings > Logs

arielj:
This PR changes the general Logs screen to allow users to see not only the general log but also the logs for legendary/gogdl/nile and the logs for each of the currently installed apps/games.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/188464/bba8429b-8edd-404a-b8a1-b23a4b055d6c

I removed the option to the the old general log on purpose since I don't think is ever used by users (when we ask for logs we want to see the most recent one). It also frees up some space for all the new options.

The motivations to do this are:
- it makes it easier for use when we ask users for the specific runner logs
- hopefully now users looking for games logs will see them in the sidebar and click there, since it happens too often that we ask for games logs and users share general logs
- this adds more visibility in general to the end user, easier to discover information

If anybody needs the previous logs they can still open the logs folder and look for them, but I'd consider that a more advanced usage.

---


src/frontend/screens/Settings/sections/LogSettings/index.tsx
```diff
@@ -115,37 +121,58 @@ export default function LogSettings() {
...
-        </span>
-      )}
-      {refreshing ? (
-        <span className="log-box">
-          <UpdateComponent inline />
+          {[
+            ['Heroic', 'heroic'],
+            ['Epic/Legendary', 'legendary'],
+            ['GOG', 'gogdl'],
+            ['Amazon/Nile', 'nile']
```
arielj, line 130: I didn't add translations for these names on purpose, I've seen them translated and translated incorrectly when they shouldn't be translated actually

flavioislima (approved):
Codewise looks good.
Some improvements would be to make the item that is selected with a different color because right now it changes the size but it is not that visible which one is selected.

Also, this one I am not sure if it is related with this PR, but the GOG logs are showing the user token instead of 'redacted'.
Maybe would be good to fix it before next release.

arielj:
@flavioislima I updated the code adding a `logSanitizer` option to callRunner so we can customize the sanitization for each command individually

I added a function to cleanup the gog and amazon data, but I don't have an amazon account so I can't test it properly (it's based on the types of the output)

would be great to get someone with an amazon account checking if the log is still exposing info

@imLinguin can you check that maybe?

flavioislima:
Testing this one again, I can still see my Access Token and refresh tokens on the logs for GOG.
Nile it's not showing them anymore though.

This is the GOG command that is exposing the codes:
`gogdl --auth-config-path`

flavioislima:
Probably this line:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/876da593cd5eabab0e60f24fa33446db65e28cc0/src/backend/storeManagers/gog/library.ts#L1026

I wonder if we really need to add this information to the logs or even log this command since it just checks if the user is logged in or not.

arielj:
> Testing this one again, I can still see my Access Token and refresh tokens on the logs for GOG. Nile it's not showing them anymore though.
> 
> This is the GOG command that is exposing the codes: `gogdl --auth-config-path`

can you check what's the complete command? `--auth-config-path` is used for all commands, check at the end of the string like `gogdl --auth-config-path "/Users/arielj/Library/Application Support/heroic/gog_store/auth.json" auth`

this is what I see when I open heroic in GOG's logs

Screenshot 2023-10-27 at 2 20 47 PM

arielj:
> Testing this one again, I can still see my Access Token and refresh tokens on the logs for GOG. Nile it's not showing them anymore though.
> 
> This is the GOG command that is exposing the codes: `gogdl --auth-config-path`

can you check what's the complete command? `--auth-config-path` is used for all commands, check at the end of the string like `gogdl --auth-config-path "/Users/arielj/Library/Application Support/heroic/gog_store/auth.json" auth`

this is what I see when I open heroic in GOG's logs

Screenshot 2023-10-27 at 2 20 47 PM

arielj:
I fixed some sizing issue and added underline to the current log (I didn't want to go too wild with the indication of the current log to keep it simple)

flavioislima:
Ok I think its fine now. We can merge it 👍🏽

PR #3150, opened 2023-10-21
Disable gamepad console log in devtools

arielj:
This was turned on at some point for debugging something and it's not really useful now, it just adds noise in the dev tools console when using a gamepad.

---

### DIFF
--- a/src/frontend/helpers/gamepad.ts
+++ b/src/frontend/helpers/gamepad.ts
@@ -307,7 +307,7 @@ export const initGamepad = () => {
       const controller = gamepads[index]
       if (!controller) return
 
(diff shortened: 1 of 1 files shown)


flavioislima (approved):

flavioislima:
another idea would be to have a setting on advanced settings for that.
Since it is a frontend-only debugging, it does not need to go to the backend and write the config.json file.
So it can be disabled by default and if someone needs to debug a new controller it just turns it on.

arielj:
Maybe we can have that advanced feature if we need to debug things again, it have not happened in a looooooong time, most users have standard controllers anyway.

PR #3151, opened 2023-10-22
[Refactor] Move the library status/filters out of the GlobalContext into a LibraryContext

arielj:
I started working on implementing some redesign for the library and noticed the current state management of the library is spread all over the place:
- most state is in GlobalState
- some state is in Library
- some state is in the children of the library

The state of the library doesn't need to be global, moving that out saves extra re-renderings, like re-rendering the sidebar when the search changes. Also, decoupling the different components will make it easier for me to implement some of the design changes for the filtering. It also makes the code more consistent, ActionIcons was updating the local storage of the browser to store the values of some filters but the other filters were updated by GlobalState.

Now all these:
- search text
- platform filter
- store filter
- sorting
- layout
- visibility filter

are handled by the Library component and shared with a context to also avoid prop drilling

One more thing I noticed and fixed: the platform filter and the toggle to show/hide non-available games were not getting stored

this also makes the global state simpler for whenever we move out of the global context into a state store

---

### DIFF
--- a/src/frontend/components/UI/ActionIcons/index.tsx
+++ b/src/frontend/components/UI/ActionIcons/index.tsx
@@ -14,7 +14,7 @@ import {
 } from '@fortawesome/free-regular-svg-icons'
 
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import React, { useCallback, useContext } from 'react'
+import React, { useContext } from 'react'
 import { useTranslation } from 'react-i18next'
 import ContextProvider from 'frontend/state/ContextProvider'
 import FormControl from '../FormControl'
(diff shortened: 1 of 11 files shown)


src/frontend/components/UI/ActionIcons/index.tsx
```diff
@@ -14,55 +14,38 @@ import {
...
-
-  const toggleShowFavourites = useCallback(() => {
-    setShowFavourites(!showFavourites)
-  }, [showFavourites])
+    setShowNonAvailable,
+    sortDescending,
+    setSortDescending,
+    sortInstalled,
+    setSortInstalled
+  } = useContext(LibraryContext)
```
arielj, line 48: some state came from the global context, some from props, now it's all from the library context

src/frontend/state/GlobalState.tsx
```diff
@@ -673,11 +637,8 @@ class GlobalState extends PureComponent<Props> {
 
   async componentDidMount() {
     const { t } = this.props
-    const { epic, gameUpdates = [], libraryStatus, category } = this.state
-    const oldCategory: string = category
-    if (oldCategory === 'epic') {
-      this.handleCategory('all')
-    }
```
arielj, line 680: more than a year ago we migrated the `epic` filter to be called `legendary`, by now any user should already have the correct store filter saved in the local storage so we can remove this migration workaround

flavioislima:
Could not find any other issue.

flavioislima (changes requested):
Only issue is with the favorites.

arielj:
> Testing this one here now. One issue I found is that the favorites are not being updated until you restart the app. First I added a few games to it and then clicking on the favorites button was showing nothing. Then After going to the General settings and changing the first lane to 'Favorites' made it work. But then after adding more favorites won't work until restart. image

thanks for testing, just pushed a fix for that

flavioislima (approved):
Works perfectly now. I noticed a bit of performance improvement as well when changing filters, layout, etc.
Maybe we won't need mobx or recoil in the short term anymore? 👀 
something to think about.

PR #3170, opened 2023-10-29
Electron 27: more fixes (gallery lazy load)

arielj:
After using the latest main with Electron 27 I noticed the library felt slow and it was because the lazy loading of the elements below the fold broke with all the changes related to which element is the scroll element.

This PR fixes that to render it fast again.

I also removed a leftover console.log.

I'll keep testing to see if I find any other issue.

---

### DIFF
--- a/src/frontend/App.css
+++ b/src/frontend/App.css
@@ -16,11 +16,8 @@
   /* Measurements */
 }
 
-body,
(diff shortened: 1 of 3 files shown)


flavioislima (approved):
lgtm

PR #3174, opened 2023-10-29
[Fix] Prevent useEffect infinite loops

arielj:
This PR fixes a problem I noticed in some components in the Advanced screen that include the state variable as a dependency for useEffects that update that state. This can lead to infinite loops and it's not correct to do it.

This can be seen when running a clean install of Heroic and going to Settings > Advanced, there's an infinite loop in the dev tools console.

This is solved once the experimental features are modified, but we shouldn't have this loop to begin with.

The useEffect that updates the Copy to clipboard status is safe because it returns early if false so it doesn't call a state setter.

---

### DIFF
--- a/src/frontend/screens/Settings/components/ExperimentalFeatures.tsx
+++ b/src/frontend/screens/Settings/components/ExperimentalFeatures.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect } from 'react'
+import React, { useContext } from 'react'
 import { useTranslation } from 'react-i18next'
 import useSetting from 'frontend/hooks/useSetting'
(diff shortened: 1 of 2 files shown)


CommandMC (approved):
Almost like *someone* didn't know how React worked back when they wrote this code (to clarify, this "someone" was me)

Thanks!

flavioislima:
> Almost like _someone_ didn't know how React worked back when they wrote this code (to clarify, this "someone" was me)
> 
> Thanks!

no worries about that, I also did this mistake a few times 👍🏽

flavioislima (approved):
Good catch!

PR #3184, opened 2023-11-01
[UX] Warn about the use of %command% and env variables in game arguments

arielj:
It's really common for users trying to make a game work to read instructions for Steam and try to apply them in Heroic.

It happened many times that users would set environment variables and the `%command%` placeholder in the `Game Arguments` of Heroic and that usually creates more problems.

This PR adds some error messages when we detect env variables or the `%command%` placeholder so users can see it's incorrect.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/188464/2c1ddb3c-aa02-40fc-8797-555dc45775d4

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -433,6 +433,10 @@
             "info": "Set environment variables to append to the command."
         },
         "gameargs": {
+            "error": {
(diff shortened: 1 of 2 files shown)


flavioislima (approved):
Nice catch

PR #3185, opened 2023-11-01
mock xhmikosr decompress and decompress-targz

BrettCleary:
@flavioislima if you don't want to test any functionality with the new decompress packages, this is one way to mock them out so the existing tests can run

otherwise we need to explore one of the suggested solutions by jest
```
    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
```

or refactor to use vitest so we can share the vite build pipeline with the test build pipeline

---


flavioislima:
Thanks, man. I tried mocking the package but not using those exports. Let's hope this fixes it 🤞🏽

PR #3192, opened 2023-11-04
[Tech] Move to pnpm

CommandMC:
This involved
- specifying our Electron dependency properly (we were missing the "github:" protocol)
- adding some missing '@types' packages (and in turn fixing some type issues that weren't caught before)
- fixing imports from 'react-router' (it's 'react-router-dom')
- installing pnpm in Workflow runs
- "Find and replace"ing quite a few "yarn"s with "pnpm"s

---


flavioislima:
Perhaps lets keep using npm or yarn on AppVeyor since it is just one of the builds and all the rest is working fine.

CommandMC:
Using something other than pnpm would still leave us without a lock file, but I guess that's better than it not working at all

CommandMC:
Silly meme

PR #3198, opened 2023-11-04
[Tech] Deduplicate our Workflows

CommandMC:
- The "install-deps" action is used to have one place to declare (common) dependencies for building
- The build-base workflow is used for a generic "build Heroic for all platforms and upload the results" step (since we do that multiple times as well)
- Job strategies were used wherever possible to not repeat (nearly) the same steps for 3 OSs

@flavioislima I'm assuming that we'd need to configure the repo so it no longer requires a "build" check for PRs (the equivalent file is `build-prs.yml` now), otherwise GH will keep showing this "expected" build task that'll never actually start (since the file no longer exists)
I shuffled the files around a bit since it felt unclear which file was doing what before

---


.github/actions/install-deps/action.yml
flavioislima, line 1: I only vote for us to use NPM here instead of yarn since it gave me several issues during the latest release and did not work at all making me waste more than 2h debugging. While changing to npm make it work on first try.
Also, about the worrying of the lock file, the npm uses yarn.lock file if it is available since v7:
```This command installs a package and any packages that it depends on. If the package has a package-lock, or an npm shrinkwrap file, or a yarn lock file, the installation of dependencies will be driven by that```

flavioislima (changes requested):
Just change the install deps to use npm and we are fine 👍🏽

.github/actions/install-deps/action.yml
CommandMC: This PR's intent was *just* to deduplicate our files, not to change them in any way

Moving to npm (or any other package manager) requires changes not just in this files, and is best left to its own PR (#3192)

CommandMC: Oh hang on, I see what you mean now, we were using npm for some of these files while using yarn for others (I had this patch laying around for a bit, so I didn't notice this change)

Should be fixed now

flavioislima: perfect. Thanks!

flavioislima (approved):
Pretty good stuff!

Issue #3200, opened 2023-11-05
When opening a store, if the saved url is external to the store we can't go back to the store


arielj:
### Describe the bug

In 2.10.0 I implemented a feature to store the last visited url for each store.

If the user navigated to a url that is external to the current store, it's saved and then there's no way to go back to the store without clearing the local storage manually of finding a way to navigate back to the store following links.

### Add logs

```text
Nothing relevant
```

### Steps to reproduce

1. Open a store
2. Click links around until you are in an external url
3. go to the library
4. click the same store again
5. We are in a page that is not the actual store

### Expected behavior

We should either not store the url, or we should ignore the stored url if the url does not belong to the store.

### Heroic Version

Latest Stable

### System Information

Not relevant

Liametekudasai:
same bug,  can't go back to the original store

jp438-2:
All three of my store links (Epic, GOG and Prime Gaming) all lead to the Epic Game store now (and the GOG one leads to the EGS Turnip Boy Commits Tax Evasion page). I have no idea how it happened and I can't do anything about it. This really needs to be fixed!

nenadandric:
I'm facing the same issue. Quick workround is to go to library page, select game from GOG store, then go to dot menu and select store page

kn-f:
Same issue here on Heroic 2.10.0.
Even rebooting my PC or clearing Heroic cache is not working, now I'm stuck to a page in Prime gaming no matter what store I choose.
In 2.9 this wasn't happening

arielj:
This will be fixed in the next release, until now the workarounds are:

Option 1)
Ctrl/Cmd+shit+I to open devtools
look for the application tab in devtools
search for local storage in the sidebar
delete the entries that start with `last-url-...`

Option 2)
Open the `store page` of a game from the 3 dots menu in the game's page

Zykino:
Thanks for the workaround. Option 2 did not work for me (it opened the GoG shop, but in the general shop bar, not fixing the "shop gog" item.

The option 1 did work… but I managed to open the dev tools only once, and I’m not sure how/why. Do I need to select the shop, or the application? I tried both but did not have a consistent setup.

Can I stress a point in my issue that: I don’t care which is the last url visited for each shop? I like to have heroic remember if I was on EGS or GoG, but I don’t really want to have it remember if I was visiting game A, B or C (at least, not after a reboot of the application).

[...]

PR #3208, opened 2023-11-05
[Fix] Prevent storing external urls for stores

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3200

We now store the last url visited when navigating the stores, but we can store external urls and then there's no way to go back to the store.

With this PR we now only store urls that are allowed instead of storing everything. And in case a user already has an invalid url stored, we now also check to only re-open in the url if it's one of the allowed.

---

### DIFF
--- a/src/frontend/components/UI/Sidebar/components/SidebarLinks/index.tsx
+++ b/src/frontend/components/UI/Sidebar/components/SidebarLinks/index.tsx
@@ -93,7 +93,7 @@ export default function SidebarLinks() {
   }
 
   // if we have a stored last-url, default to the `/last-url` route
-  const lastStore = localStorage.getItem('last-store')
+  const lastStore = sessionStorage.getItem('last-store')
   if (lastStore) {
     defaultStore = lastStore
   }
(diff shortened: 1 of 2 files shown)


Etaash-mathamsetty (approved):

CommandMC:
Storing the URLs inside `sessionStorage` instead might be safer. They don't have to persist between restarts, right?

arielj:
> Storing the URLs inside `sessionStorage` instead might be safer. They don't have to persist between restarts, right?

I meant this to be stored between restarts, I didn't think about making it not persistent. Not sure what users will expect though.

I guess we can wait for feedback. In the issue I was fixing with that feature both tings were mentioned: storing between restarts as not necessary and also as someone mentioned to support that after restarts, so I don't know haha.

flavioislima:
I vote for it to NOT be persistent across restarts. 
It could be safer in case something goes wrong during the navigation.

imLinguin:
I agree, there is no need for that to be persistent across restarts. I think it's even nicer for UX

arielj:
changed it to use sessionStorage

flavioislima (approved):

PR #3210, opened 2023-11-05
[FIX] Wine Stuck sometimes at adding registry key

K900:
Otherwise, newer Wine versions will prompt to overwrite it and loop there forever.

---

### DIFF
--- a/src/backend/storeManagers/legendary/setup.ts
+++ b/src/backend/storeManagers/legendary/setup.ts
@@ -20,7 +20,12 @@ export const legendarySetup = async (appName: string) => {
 
   // Fixes games like Fallout New Vegas and Dishonored: Death of the Outsider
   await runWineCommandOnGame(appName, {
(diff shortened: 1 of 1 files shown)


arielj:
can you explain the problem better? how can this be reproduced?

K900:
I'm not entirely sure _how_ it can be reproduced - it happens consistently on some of my Wine prefixes, but basically, the `reg add` command will just get stuck. `strace`ing it shows it gets stuck on an "overwrite? [yes/no]" prompt, which can be bypassed with `/f`.

Etaash-mathamsetty:
> I'm not entirely sure _how_ it can be reproduced - it happens consistently on some of my existing Wine prefixes, but not new ones. Basically, the `reg add` command will just get stuck. `strace`ing it shows it gets stuck on an "overwrite? [yes/no]" prompt, which can be bypassed with `/f`.

that means that the registry key is already there, which is a strange, but I don't see any problem with adding a /f here

Etaash-mathamsetty (approved):

K900:
Not really sure why the CI failed here, but seems unrelated?

PR #3213, opened 2023-11-05
[Tech] Simplify our E2E tests

CommandMC:
Our E2E test configuration, while being very robust, was a bit overkill for what we're doing with it. Instead of running the tests in a Docker container to isolate them, we can simply set Heroic's configuration directory to a temporary one. In addition, the "packaged" and "dev" tests only differ slightly (one having Heroic's code in the `build` directory, while the other has it zipped up in a single file somewhere in `dist`), but this difference is not significant enough to warrant two separate test runs IMO

I hope that by making it easier to run these tests, we can start to use them properly in the future

---


arielj (approved):
looks good to me

PR #3218, opened 2023-11-06
[Tech] Move to the `electron-vite` build tool

CommandMC:
[electron-vite](https://electron-vite.org/) is a slightly newer option for building Electron applications with Vite. Other than having some saner defaults and being generally more popular in the ecosystem, it also allows us to finally upgrade to Vite 4, which comes with noticeable speed improvements

I've done *some* testing of this on Linux and Windows, but due to architectural issues (the great CommonJS vs ESM split), literally any module we import might break with this. This is sadly unavoidable if we want to upgrade to a newer Vite version. If a module like this crops up in the future, we have to add it to `electron.vite.config.ts`'s `dependenciesToNotExternalize` array
**Due to this issue, more testing than normal will be necessary for this PR**

---


arielj (commented):
I tested some things like login, install, play, uninstall, install a wine, filtering, changing settings

all on linux seems to work fine

I didn't test building the app for release though

maybe we can merge this into main so we all use this more during development? not sure what others think

imLinguin (approved):
works as expected

PR #2748, opened 2023-05-30
Prettier - Move exclamation mark to outside the single quotes of flatpak-build

ram1660:
This is my first pull request.

In Windows when you try to run `yarn prettier` or `yarn prettier-fix` the commands won't work because the `!` is being inside the single quotes of the "flatpak-build" and there maybe be difference between Windows and Linux where Windows treats the '!flatpak-build' as a folder while we are trying to ignore the folder. 
By moving the exclamation mark to the outside of the single quote you will be able to run prettier and ignore the "flatpak-build" folder.
I have tested this in Windows and WSL Ubuntu and the and it works on both of them.

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [x] Tested the feature and it's working on a current and clean install.
- [x] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #3219, opened 2023-11-06
[Tech] Truly ignore `flatpak-build`

CommandMC:
This is an alternative PR to #2748
I didn't like the fact that we had to explicitly add the folder when invoking Prettier in the first place. It's already in our `.prettierignore`, so it should be... ignored. Newer versions of Prettier do this correctly, so this was solved by a simple update (and subsequently fixing all the new formatting rules)
In the process, I noticed that ESLint also had problems with this folder. Adding it to `.eslintignore` was all *it* needed

---

- [X] N/A - Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 sign/**
+flatpak-build/**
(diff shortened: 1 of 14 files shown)


flavioislima (approved):
lgtm

PR #3223, opened 2023-11-08
[Fix/Tech] Fixup Snap publishing in `build-base` workflow

CommandMC:
Of course, the *one thing* I can't test locally breaks
The workflow tried to access `snapcraftEdgeId`, but the secret in question is named `snapcraftIdEdge`

---

### DIFF
--- a/.github/workflows/build-base.yml
+++ b/.github/workflows/build-base.yml
@@ -36,7 +36,7 @@ jobs:
           sudo apt-get install --no-install-recommends -y snapcraft
           yarn dist:linux Snap --publish=always
         env:
-          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.snapcraftEdgeId }}
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.snapcraftIdEdge }}
         if: runner.os == 'Linux' && inputs.publish-snap
       - name: Build AppImage version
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
np, S!$% happens 🤷🏽

.github/workflows/build-base.yml
```diff
@@ -36,7 +37,7 @@ jobs:
           sudo apt-get install --no-install-recommends -y snapcraft
           yarn dist:linux Snap --publish=always
         env:
-          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.snapcraftEdgeId }}
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.snapcraftIdEdge }}
```
flavioislima, line 40: Actually:
```suggestion
          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_ID_EDGE }}
```

CommandMC: The secret name here depends on the name given in line 12. Shared workflows like this don't have access to the actual repo's secrets (the reason behind this is that something like `actions/checkout@v3` is also a shared workflow, and you don't want that having access to your own repo's secrets)

flavioislima: Ah ok, I get it.

PR #3227, opened 2023-11-09
[FIX-macOS] Wine Downloader fails to extract files 

### LINKED ISSUE #3202: Crashing while Downloading Wine in MacOs
I tried to download wine from wine manager , when the download completes Launcher getting crash 
Downloaded through Brew
OS : MacOs
version : Sonoma

### Add logs

```text
(10:53:02) INFO:    [WineDownloader]:   Removed wine version Wine-Crossover-latest succesfully.
(10:53:04) INFO:    [WineDownloader]:   Start installation of wine version Wine-Crossover-latest
(10:53:08) INFO:    [Backend]:          Downloaded: 11.41 MB / 175.5 MB  @4.56 MB/s (6.50%)
(10:53:10) INFO:    [Backend]:          Downloaded: 28.66 MB / 175.5 MB  @8.92 MB/s (16.33%)
(10:53:13) INFO:    [Backend]:          Downloaded: 53.9 MB / 175.5 MB  @8.48 MB/s (30.71%)
(10:53:15) INFO:    [Backend]:          Downloaded: 68.49 MB / 175.5 MB  @9.28 MB/s (39.03%)
(10:53:17) INFO:    [Backend]:          Downloaded: 85.39 MB / 175.5 MB  @8.77 MB/s (48.66%)
(10:53:19) INFO:    [Backend]:          Downloaded: 108.02 MB / 175.5 MB  @8.61 MB/s (61.55%)
(10:53:20) INFO:    [Backend]:          Downloaded: 118.87 MB / 175.5 MB  @8.96 MB/s (67.73%)
(10:53:21) INFO:    [Backend]:          Downloaded: 127.8 MB / 175.5 MB  @8.5 MB/s (72.82%)
(10:53:24) INFO:    [Backend]:          Downloaded: 148.97 MB / 175.5 MB  @8.46 MB/s (84.89%)
(10:53:25) INFO:    [Backend]:          Downloaded: 157.73 MB / 175.5 MB  @7.37 MB/s (89.88%)
(10:53:27) INFO:    [Backend]:          Downloaded: 172.59 MB / 175.5 MB  @8.85 MB/s (98.35%)
(10:53:28) INFO:    [DownloadManager]:  Downloader: Finished downloading https://github.com/Gcenx/winecx/releases/download/crossover-wine-23.0.1/wine-crossover-23.0.1-osx64.tar.xz
(10:53:28) INFO:    [DownloadManager]:  Downloader: Finished downloading https://github.com/Gcenx/winecx/releases/download/crossover-wine-23.0.1/wine-crossover-23.0.1-osx64.tar.xz
(10:53:28) WARNING: [WineDownloader]:   No checksum provided. Download of Wine-Crossover-latest could be invalid!
```

### Steps to reproduce

1. Just Download wine from Wine Manager.
2. Wait until downloads 100%.
3. That's it launcher gets crashed.

### Expected behavior

A way to install wine in Mac Os

### Screenshots

_No response_

### Heroic Version

Latest Stable

### System Information

- OS : MacOs
- Version : Sonoma
- Model : Macbook Pro M1 16" 

### Additional information

_No response_

flavioislima:
This fixes #3202 

But it also get back to using `tar` on Linux except for Snaps. so if you do not have a mac to test, just test the behavior on linux, it is the same.

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -34,7 +34,8 @@ import {
   GITHUB_API,
   isMac,
   configStore,
-  isLinux
+  isLinux,
+  isSnap
 } from './constants'
 import {
   logError,
@@ -1348,10 +1349,57 @@ interface ExtractOptions {
 }
 
 async function extractFiles({ path, destination, strip = 0 }: ExtractOptions) {
-  return decompress(path, destination, {
-    plugins: [decompressTargz(), decompressTarxz()],
-    strip
-  })
+  if (!isSnap && (path.endsWith('.tar.xz') || path.endsWith('.tar.gz'))) {
+    try {
+      await extractNative(path, destination, strip)
+    } catch (error) {
+      logError(`Error: ${error}`, LogPrefix.Backend)
+    }
+  } else {
+    try {
+      await extractDecompress(path, destination, strip)
+    } catch (error) {
+      logError(`Error: ${error}`, LogPrefix.Backend)
+    }
+  }
+}
+
+async function extractNative(path: string, destination: string, strip: number) {
+  logInfo(
+    `Extracting ${path} to ${destination} using native tar`,
+    LogPrefix.Backend
+  )
+  const { code, stderr } = await spawnAsync('tar', [
+    '-xf',
+    path,
+    '-C',
+    destination,
+    `--strip-components=${strip}`
+  ])
+  if (code !== 0) {
+    logError(`Extracting Error: ${stderr}`, LogPrefix.Backend)
+    return { status: 'error', error: stderr }
+  }
+  return { status: 'done', installPath: destination }
+}
+
+async function extractDecompress(
+  path: string,
+  destination: string,
+  strip: number
+) {
+  logInfo(
+    `Extracting ${path} to ${destination} using decompress`,
+    LogPrefix.Backend
+  )
+  try {
+    await decompress(path, destination, {
+      plugins: [decompressTargz(), decompressTarxz()],
+      strip
+    })
+  } catch (error) {
+    logError(`Error: ${error}`, LogPrefix.Backend)
+  }
 }
 
 export {
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -1353,13 +1353,13 @@ async function extractFiles({ path, destination, strip = 0 }: ExtractOptions) {
     try {
       await extractNative(path, destination, strip)
     } catch (error) {
-      logError(`Error: ${error}`, LogPrefix.Backend)
+      logError(['Error:', error], LogPrefix.Backend)
     }
   } else {
     try {
       await extractDecompress(path, destination, strip)
     } catch (error) {
-      logError(`Error: ${error}`, LogPrefix.Backend)
+      logError(['Error:', error], LogPrefix.Backend)
     }
   }
 }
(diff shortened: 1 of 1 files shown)


src/backend/utils.ts
```diff
@@ -1348,10 +1349,57 @@ interface ExtractOptions {
...
 async function extractFiles({ path, destination, strip = 0 }: ExtractOptions) {
-  return decompress(path, destination, {
-    plugins: [decompressTargz(), decompressTarxz()],
-    strip
-  })
+  if (!isSnap && (path.endsWith('.tar.xz') || path.endsWith('.tar.gz'))) {
+    try {
+      await extractNative(path, destination, strip)
+    } catch (error) {
+      logError(`Error: ${error}`, LogPrefix.Backend)
```
CommandMC, line 1356:
```suggestion
      logError(['Error:', error], LogPrefix.Backend)
```
Passing the error directly to the logging function makes it print more information if available

src/backend/utils.ts
```diff
@@ -1348,10 +1349,57 @@ interface ExtractOptions {
...
+    }
+  } else {
+    try {
+      await extractDecompress(path, destination, strip)
+    } catch (error) {
+      logError(`Error: ${error}`, LogPrefix.Backend)
```
CommandMC, line 1362: Same thing here ...

src/backend/utils.ts
```diff
@@ -1348,10 +1349,57 @@ interface ExtractOptions {
...
+    `Extracting ${path} to ${destination} using decompress`,
+    LogPrefix.Backend
+  )
+  try {
+    await decompress(path, destination, {
+      plugins: [decompressTargz(), decompressTarxz()],
+      strip
+    })
+  } catch (error) {
+    logError(`Error: ${error}`, LogPrefix.Backend)
```
CommandMC, line 1401: ... and here

src/backend/utils.ts
```diff
@@ -1348,10 +1349,57 @@ interface ExtractOptions {
 }
 
 async function extractFiles({ path, destination, strip = 0 }: ExtractOptions) {
-  return decompress(path, destination, {
-    plugins: [decompressTargz(), decompressTarxz()],
-    strip
-  })
+  if (!isSnap && (path.endsWith('.tar.xz') || path.endsWith('.tar.gz'))) {
```
CommandMC, line 1352: I'm not sure if the second condition makes sense here. `decompress` will also only work with these two file types (since we manually pass the two plugins for them to it)
So something like
```ts
if (!path.endsWith('.tar.xz') && !path.endsWith('.tar.gz')) {
  logError(['Unsupported archive:', path], LogPrefix.Backend)
  return
}

if (!isSnap)
// ...
```
might be more accurate

CommandMC (commented):
In general, this seems to work fine (tested on Linux)

arielj (approved):
tested on an intel mac and it works fine (it worked fine before too though, the issue was only for M1/2/3 chips)

src/backend/utils.ts
flavioislima: Yes, it fails if we pass a different archive. I just think this way is safer, but might be too much or not needed like you said.

CommandMC: I don't think this is safer right now, actually. I doubt those plugins are designed to work with archives not intended for them (and them failing might lead to a crash instead of just an error message, as seen here), so bailing out early should be the safer option

Regardless, as far as I know we *currently* don't feed other archive formats to this function, so *for now* this doesn't really matter

Issue #2466, opened 2023-02-18
GamePage image is missing logo when available


arielj:
### Describe the bug

For games that have a separate asset for the logo, we are not showing the logo in the game page:

![image](https://user-images.githubusercontent.com/188464/219874647-d4551845-283b-4432-bbdc-6e3033fbd724.png)

![image](https://user-images.githubusercontent.com/188464/219874804-611ac6df-935c-490b-bb93-274d617fb60d.png)

### Add logs

```text
Nothing relevant
```

### Steps to reproduce

1. Open the game page for any game that has a logo asset (rayman legends, alan wake's, batman arkham, faster than light, celeste, and many more)

### Expected behavior

The logo should be placed over the image like it's done in the library

### Heroic Version

Latest Stable

### System Information

Not relevant

thiyagarajan17:
please assign the issue.. would love to work on it!

flavioislima:
> 
> please assign the issue.. would love to work on it!

No need to assign. Just let us know that you want to work on that. And also if you are stuck go to our discord for help. 👍

PR #3245, opened 2023-11-19
Fix Issue #2466 GamePage image is missing logo when available

hashkar123:
Fix Issue #2466 GamePage image is missing logo when available

Add logo support in GamePage and GamePicture.
In GamePage component, add logo variable and pass it to GamePicture component. In GamePicture component, add logo parameter; add logo element (same as the CachedImage logo element in GameCard)

There was no need to add CSS, since there was  already CSS for logo in GamePicture  that positions it at the bottom center.

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -231,6 +231,7 @@ export default React.memo(function GamePage(): JSX.Element | null {
       art_square,
       art_cover,
       art_background,
+      art_logo: logo = undefined,
       install: { platform: installPlatform },
       is_installed
     } = gameInfo
@@ -325,7 +326,11 @@ export default React.memo(function GamePage(): JSX.Element | null {
             {/* OLD DESIGN */}
             {!experimentalFeatures.enableNewDesign && (
               <>
-                <GamePicture art_square={art_square} store={runner} />
+                <GamePicture
+                  art_square={art_square}
+                  logo={logo}
+                  store={runner}
+                />
                 <NavLink
                   className="backButton"
                   to={backRoute}
@@ -394,7 +399,11 @@ export default React.memo(function GamePage(): JSX.Element | null {
                   <DotsMenu gameInfo={gameInfo} handleUpdate={handleUpdate} />
                   {!isBrowserGame && <SettingsButton gameInfo={gameInfo} />}
                   <div className="mainInfo">
-                    <GamePicture art_square={art_cover} store={runner} />
+                    <GamePicture
+                      art_square={art_cover}
+                      logo={logo}
+                      store={runner}
+                    />
                     <div className="store-icon">
                       <StoreLogos runner={runner} />
                     </div>
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -231,7 +231,7 @@ export default React.memo(function GamePage(): JSX.Element | null {
       art_square,
       art_cover,
       art_background,
-      art_logo: logo = undefined,
+      art_logo = undefined,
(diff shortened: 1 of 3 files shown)


src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -231,6 +231,7 @@ export default React.memo(function GamePage(): JSX.Element | null {
       art_square,
       art_cover,
       art_background,
+      art_logo: logo = undefined,
```
arielj, line 234: maybe we can just use `art_logo` everywhere? it makes it easier to understand that it's the same thing that comes from the gameInfo and we already use `art_...` for the other images, makes it more consistent

arielj: also, I think there's no need to do the `= undefined` since it will be undefined if it is undefined anyway

src/frontend/screens/Game/GamePicture/index.tsx
```diff
@@ -6,10 +6,17 @@ import fallbackImage from 'frontend/assets/heroic_card.jpg'
 
 interface Props extends React.ImgHTMLAttributes<HTMLImageElement> {
   art_square: string
+  logo?: string | undefined
```
arielj, line 9: do we need the `| undefined`? I think because it's optional with the `?` it already includes undefined as a possibility

src/frontend/screens/Game/GamePicture/index.tsx
```diff
@@ -6,10 +6,17 @@ import fallbackImage from 'frontend/assets/heroic_card.jpg'
...
   store: string
 }
 
-function GamePicture({ art_square, store, className, ...props }: Props) {
+function GamePicture({
+  art_square,
+  logo = undefined,
```
arielj, line 15: same here as the other places, I'd just call this `art_logo` and without the default `undefined`, makes naming more consistent with the other places and it's undefined if not defined already

src/frontend/screens/Game/GamePicture/index.tsx
```diff
@@ -27,13 +34,22 @@ function GamePicture({ art_square, store, className, ...props }: Props) {
...
   return (
     <div className="gamePicture">
-      <CachedImage
-        alt="cover-art"
-        className={`gameImg ${className}`}
-        src={src}
-        fallback={fallback}
-        {...props}
-      />
+      {
```
arielj, line 37: are these `{ ... }` needed?

hashkar123:
Thanks for the review!
I'll get these fixed tomorrow or the day after it.

I'll make sure to check everything :D

arielj:
I'm testing this and it works fine, but I think the logo is too big compared to the other places we do a similar thing:

In the library it looks like this:
image

In the old design it looks like this:
image

In the new design it looks like this:
image


I think the old CSS that was there is not used by anything else. I played a bit with it and with:

```css
/* in src/frontend/screens/Game/GamePicture/index.css */
.gamePicture > .gameLogo {
  position: absolute;
  bottom: 50%;
  transform: translateY(50%);
  width: 40%;
  min-width: 140px;
}
```

it looks like this:
image

image

which looks more consistent with the library

hashkar123:
I also think the logo looked weird in the game page. It would be less weird if it was more consistent with the Library's logo.

I could do that.
Do you want me to make the logo look the same as in the library? It just wasn't clear if you wanted me to change that or not.

arielj:
> I also think the logo looked weird in the game page. It would be less weird if it was more consistent with the Library's logo.
> 
> I could do that. Do you want me to make the logo look the same as in the library? It just wasn't clear if you wanted me to change that or not.

yes please do that change, I think it looks better to make it more consistent

arielj (approved):
Looks great now, thanks!

PR #3253, opened 2023-11-21
[Fix] Workflows: Disable publishing when building the AppImage

CommandMC:
Electron-Builder tries to be smart and auto-enables publishing if if's on the main branch of a repo (at least it seems that way, it doesn't do this in the PR workflow; can't find anything about this in the docs), which then promptly fails.
Explicitly turn it off

I can't test this locally, let's hope 3rd time's the charm

---


flavioislima (approved):
should work now

PR #3254, opened 2023-11-22
[UI] Fix height of install and settings dialogs

arielj:
After one of my last PRs, the install dialog is too tall even when not enough content. This PR fixes that by making the install modal use the height on its content while also letting the settings be bigger to use more height in small screens.

Before:
image

After:
image

---

### DIFF
--- a/src/frontend/screens/Settings/components/SettingsModal/index.scss
+++ b/src/frontend/screens/Settings/components/SettingsModal/index.scss
@@ -10,3 +10,17 @@
     width: 100%;
   }
 }
+
+.InstallModal__dialog:has(.settingsDialogContent) {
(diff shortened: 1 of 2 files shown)


src/frontend/screens/Settings/components/SettingsModal/index.scss
```diff
@@ -10,3 +10,17 @@
     width: 100%;
   }
 }
+
```
arielj, line 13: the issue was caused by some CSS to make the settings dialog use more height in small screens, I removed some css in the other file and implemented that here we some media queries.

flavioislima (approved):
TEsted here and looks good. 👍🏽

PR #3266, opened 2023-11-25
[UI/UX] Filters dropdown in Library. Show only installed

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1881 and adds a filter requested in discord (show only installed games).

Now all filtering options are grouped in a dropdown that shows up when clicking/hovering the `Filters` button.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/188464/e0782feb-199f-4fc5-ace0-a3d73075b9ff

This makes the UI more consistent anda easier to maintain:
- it's easier to add more filters without dealing with sizes and positioning
- it's easier to translate elements without dealing with sizes
- we can use more text and less icons with hover tooltips (it's more accessible)
- all filters are in the same place

I also included 2 more fixes closely related:
- I added a `Show Installed Only` filter that was requested many times
- The `Show Favourites Only` filter was not composable (turning it on was ignoring all other filters)

It should honor the older filtering configurations (so if a user was filtering only GOG, only GOG will be enabled in the new filters)

Note two changes:
- because the store filter is now composable instead of exclusive, there was some code in the library header to say `Epic`/`GOG`/etc in the library title instead of `All Games` but that can't be done now cause we can toggle more than 1 store at a time
- the library refresh icon was triggering a background refresh ONLY for when the GOG filter was selected, that doesn't make much sense now so I made it refresh in the background for all store (we could make it refresh in the foreground for all stores if desired instead)

---


public/locales/en/translation.json
```diff
@@ -247,14 +245,11 @@
     "GOG": "GOG",
     "gog-store": "GOG Store",
     "header": {
-        "hide_non_available_games": "Hide non-available games",
-        "ignore_hidden": "Ignore Hidden",
-        "platform": "Platform",
-        "show_all_games": "Show all games",
```
arielj, line 253: these are not needed anymore

src/frontend/components/UI/ActionIcons/index.tsx
```diff
@@ -4,61 +4,32 @@ import {
...
-    ? t('header.ignore_hidden', 'Ignore Hidden')
-    : t('header.show_hidden', 'Show Hidden')
-
-  const showFavouritesTitle = showFavourites
-    ? t('header.show_all_games', 'Show all games')
-    : t('header.show_favourites_only', 'Show Favourites only')
-
-  const showNonAvailableTitle = showNonAvailable
-    ? t('header.hide_non_available_games', 'Hide non-available games')
-    : t('header.show_available_games', 'Show non-Available games')
```
arielj, line 60: moving all the filters from here to LibraryFilters

src/frontend/components/UI/Header/index.css
```diff
@@ -18,15 +18,6 @@
   );
 }
 
-@media screen and (max-width: 1000px) {
-  .Header {
-    grid-template-columns: 1fr min-content;
-    grid-template-areas:
-      'search search'
-      'filters filters';
```
arielj, line 26: we don't need to drop into a second row now that the `filters` element is smaller

src/frontend/screens/Library/components/LibraryHeader/index.tsx
```diff
@@ -17,7 +15,7 @@ export default React.memo(function LibraryHeader({
   handleAddGameButtonClick
 }: Props) {
   const { t } = useTranslation()
-  const { category, showFavourites } = useContext(LibraryContext)
+  const { showFavourites } = useContext(LibraryContext)
```
arielj, line 18: we don't have a single category anymore (I didn't like the word `category` anyway haha), so the code that depended on this was removed

src/frontend/screens/Library/constants.ts
arielj, line 1: we can't change the library title based on the category cause now they are composable, so removing this

flavioislima:
That is looking pretty great. I will do some testing during the week :D

flavioislima:
Testing here. Overall looks good.
Some good additions would be to:
- Have a ALL Button that will select everything
- Do not allow to unselect all stores and all platforms if possible (I imagine some people might come to us asking why their library is gone).
- Also, I am not sure if that is expected but selecting all stores by deselecting all platforms makes it show native linux games + Browser games.

If those are too much we can improve it in another PR since most of the feature is working fine :)

arielj:
I implemented that if you don't select and platform it behaves like if all platforms are selected. Same with the stores, selecting none is like selecting all. To avoid that problem of users deselecting everything and then complaining haha.
I didn't add an `All` cause that felt different to the rest of the options (each check is composable, we would have to handle the `All` in a different way, I wanted to avoid the extra code.

the third element seems to be a bug, I'll check that later today

arielj:
Just pushed a fix with the missing windows games with nothing selected.

For the other two:
- we can allow users to unselect everything, it will just show all games so I think it's safe, if somebody expect that to not show any game I can jump in and ask "but why?"
- for the `All` button, I'd wait to see if it's something needed, it will make the UI inconsistent with a button or a check box that works different than the rest (and we would need 2 of those one for platforms and one for stores)

flavioislima:
> Just pushed a fix with the missing windows games with nothing selected.
> 
> For the other two:
> 
>     * we can allow users to unselect everything, it will just show all games so I think it's safe, if somebody expect that to not show any game I can jump in and ask "but why?"
> 
>     * for the `All` button, I'd wait to see if it's something needed, it will make the UI inconsistent with a button or a check box that works different than the rest (and we would need 2 of those one for platforms and one for stores)

Sounds good, lets gather some feedback.

i believe this one is ready to be merged since it is not breaking anything, so it can be on the next release.

flavioislima (approved):
LGTM

PR #3267, opened 2023-11-25
[FIX] Disable DXVK and VKD3D if vulkan is not supported

Etaash-mathamsetty:
Should help out those ancient systems that don't support vulkan at all

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -253,10 +253,12 @@
     "GOG": "GOG",
     "gog-store": "GOG Store",
     "header": {
-        "all_categories": "All Categories",
+        "categories": "Categories",
         "filters": "Filters",
+        "no_categories": "No custom categories. Add categories using each game menu.",
         "only": "only",
         "reset": "Reset",
+        "select_all": "Select All",
         "show_available_games": "Show non-Available games",
         "show_favourites_only": "Show Favourites only",
         "show_hidden": "Show Hidden",
--- a/src/frontend/components/UI/CategoryFilter/index.css
+++ b/src/frontend/components/UI/CategoryFilter/index.css
@@ -1,9 +1,0 @@
-#custom-category-selector {
-  width: 324px;
-}
-
-@media screen and (max-width: 1000px) {
-  #custom-category-selector {
-    width: auto;
-  }
-}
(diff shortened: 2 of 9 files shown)


src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -139,6 +140,11 @@ export default function GamesSettings() {
       }
       getIsNative()
     }
+    const getIsVulkanSupported = async () => {
+      const isSupported = await window.api.isVulkanSupported()
+      setIsVulkanSupported(isSupported)
+    }
+    getIsVulkanSupported()
```
arielj, line 147: since the vulkan support cannot change at runtime, we could expose that information at boot time in the preload script like we do here https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/preload.ts#L7

it makes the handling of this information way easier without having to deal with ipc messages and async calls, just read from windows

src/backend/tools/index.ts
```diff
@@ -750,6 +750,22 @@ function getVkd3dUrl(): string {
   return 'https://api.github.com/repos/Heroic-Games-Launcher/vkd3d-proton/releases/latest'
 }
 
+export function isVulkanSupported(): boolean {
+  if (!isLinux) {
+    return isMac
+  }
```
arielj, line 756: I think it's more clear to do

```ts
if (isMac) return true
```

src/backend/tools/index.ts
```diff
@@ -750,6 +750,22 @@ function getVkd3dUrl(): string {
...
+  if (!any_gpu_supports_version([1, 1, 0])) {
+    logWarning(
+      'No GPU with Vulkan 1.1 support found, Vulkan will not work',
+      LogPrefix.DXVKInstaller
+    )
+    return false
+  }
+
+  return true
+}
```
arielj, line 767: maybe this can be calculated once and put in a variable and then we just return that variable? so we don't print that warning every single time this method is called for the cases this is not supported

src/backend/preload.ts
```diff
@@ -2,12 +2,14 @@
...
 import { contextBridge } from 'electron'
 import api from './api'
+import { isVulkanSupported } from './utils/graphics/vulkan'
 
 contextBridge.exposeInMainWorld('api', api)
 contextBridge.exposeInMainWorld(
   'isSteamDeckGameMode',
   process.env.XDG_CURRENT_DESKTOP === 'gamescope'
 )
+contextBridge.exposeInMainWorld('isVulkanSupported', isVulkanSupported())
```
Etaash-mathamsetty, line 12: this doesn't work because the constants are not initialized yet, I have no clue how to fix it

PR #3075, opened 2023-09-25
[FIX] mark paths starting with XDG_RUNTIME_DIR as invalid

Etaash-mathamsetty:
All it needs is a popup message box to inform the user and some testing :)

Use the following Checklist if you have changed something on the Backend or Frontend:

- [x] Tested the feature and it's working on a current and clean install.
- [x] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #3271, opened 2023-11-27
[FIX] Mark wineprefixes starting with XDG_RUNTIME_DIR as invalid

Etaash-mathamsetty:
Prevents more user error

### DIFF
--- a/src/backend/api/misc.ts
+++ b/src/backend/api/misc.ts
@@ -103,6 +103,8 @@ export const getFonts = async (reload: boolean) =>
   ipcRenderer.invoke('getFonts', reload)
 export const checkDiskSpace = async (installPath: string) =>
   ipcRenderer.invoke('checkDiskSpace', installPath)
+export const checkWinePrefix = async (prefix: string) =>
+  ipcRenderer.invoke('checkWinePrefix', prefix)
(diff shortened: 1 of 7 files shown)


arielj:
@Etaash-mathamsetty you have to explain your PRs better, what is the issue this is fixing? why is this fix implemented like this? it's hard to review your PRs without any context of the problem or the reasons of the fix

Etaash-mathamsetty:
context: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3075

arielj:
maybe instead of adding specific directories as invalid we can validate the path the pick and check if it's writable instead? then we can cover any path and not the ones we hardcode

Etaash-mathamsetty:
> maybe instead of adding specific directories as invalid we can validate the path the users pick and check if it's writable instead? then we can cover any path and not only the ones we hardcode checks

I think those /run/user paths are writable, but I'll confirm this after trying it out

flavioislima:
WE already have a method on backend to validate if the path is writable or not so just use it and that should fix the issue @Etaash-mathamsetty

CommandMC:
I don't think this is too helpful for Wine prefixes. Users rarely modify that path themselves

Closing primarily due to inactivity. Feel free to re-open/re-create if you're interested in continuing this

PR #3275, opened 2023-11-29
[Mac] Allow setting the prefixes folder on Mac too

arielj:
This option is present only on Linux, but we should also allow users to set this on Mac for the "wines" that work with prefixes like Wine-Staging.

Currently the prefixes dir is the default `~/Games/Heroic/Prefixes/default` and can't be changed.

---


flavioislima (commented):
Looks good, but what is being shown if the selected wine is Crossover?

arielj:
> Looks good, but what is being shown if the selected wine is Crossover?

This setting is shown in the general settings, it's independent of what's selected as the wine for a given game. (like in linux it's independent to games being win or native)

Issue #3026, opened 2023-09-03
Heroic not running gptk app under a wineprefix, only the default wineprefix (~/.wine)


blackxfiied:
### Describe the bug

Even when you specify a wine prefix in heroic, only the default wine prefix is used when launching games with game porting toolkit set as the wine version.

### Add logs

```text
(15:03:22) INFO:    [Legendary]:        Running command: XDG_CONFIG_HOME="/Users/blackxfiied/Library/Application Support/heroic/legendaryConfig" /opt/homebrew/bin/legendary --version
(15:03:22) INFO:    [Legendary]:        Legendary location: /opt/homebrew/bin/legendary
(15:03:22) INFO:    [Gog]:              GOGDL location: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/gogdl
(15:03:22) INFO:    [Connection]:       Connectivity: check-online
(15:03:22) INFO:    [Connection]:       Pinging external endpoints
(15:03:22) INFO:    [Backend]:          DRM module staus {
  "oimompecagnajdejgnnjijobebaeigek": {
    "name": "Widevine Content Decryption Module",
    "status": "new",
    "version": "4.10.2557.0"
  }
}
(15:03:22) WARNING: [Backend]:          Protocol already registered.
(15:03:22) INFO:    [Gog]:              Running command: /Applications/Heroic.app/Contents/Resources/app.asar.unpacked/build/bin/darwin/gogdl --auth-config-path "/Users/blackxfiied/Library/Application Support/heroic/gog_store/auth.json" --version
(15:03:23) INFO:    [Backend]:          Loading Screen Ready
(15:03:23) INFO:    [Frontend]:         Refreshing undefined Library
(15:03:23) WARNING: [Backend]:          refresh not implemented on Sideload Library Manager
(15:03:23) INFO:    [Legendary]:        Refreshing library...
(15:03:23) INFO:    [Legendary]:        Refreshing Epic Games...
(15:03:23) INFO:    [Legendary]:        Game list updated, got 11 games & DLCs
(15:03:23) INFO:    [Backend]:          Frontend Ready
(15:03:23) WARNING: [Backend]:          listUpdateableGames not implemented on Sideload Library Manager
(15:03:23) INFO:    [Backend]:          Checking for new Heroic Updates
(15:03:23) INFO:    [Connection]:       Connectivity: online
[...]

PR #3277, opened 2023-11-29
[Fix] WINEPREFIX ignored when using GPTK

Mrredstone5230:
Just a quick fix to bug #3026. Without this fix, WINEPREFIX isn't set in the environment variables in the launch options, and therefore GPTK uses the user's default wine prefix (in my case ~/.wine). I doubt this fix is crazy perfect, but it's simple and it works!

Tested using Heroic latest on macOS 14.0 with Satisfactory Early Access. (The game is now using the right prefix's save files!)

----

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -493,6 +493,8 @@ function setupWineEnvVars(
       break
     case 'crossover':
       ret.CX_BOTTLE = wineCrossoverBottle
+    case 'toolkit':
+      ret.WINEPREFIX = winePrefix
   }
   if (gameSettings.showFps) {
     isMac ? (ret.MTL_HUD_ENABLED = '1') : (ret.DXVK_HUD = 'fps')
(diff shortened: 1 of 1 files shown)


Mrredstone5230:
Well, I'm confused as to why it's so angry about syntax: the 2 lines right over my commit are quite almost the same...

Mrredstone5230:
I think this PR should be ready for review, but I don't have the permissions to do so.

arielj:
one thing I'm worried about this is that if someone has a game already running using the wrong `.wine` prefix, this will break that because it will set the correct prefix (maybe losing saves for example)

I'm not sure what's the best way to handle that (we should keep the wrong prefix for current installations and have this fix for new installations, but the launch command doesn't know about that)

Mrredstone5230:
This would be dirty, but maybe at the next update, on first opening of the launcher, it could force the wine prefix of all GPTK games to the default wine prefix... 🤷‍♂️

flavioislima:
> This would be dirty, but maybe at the next update, on first opening of the launcher, it could force the wine prefix of all GPTK games to the default wine prefix... 🤷‍♂️

That is not ideal as well since some games might already be working fine on the `.wine` prefix and also some people might lose their saves. Not lose, they will be there, but 99% of people will think that Heroic has erased them.

So, for now, lets keep as it is and if people come to discord we know how to proper explain the issue to them.

PR #3300, opened 2023-12-07
[improv]: electron side changes to window behavior

imLinguin:
window now has a background which should prevent flashing of white,   
window is shown after its contents are loaded - removed our custom loadingScreenReady event

additionally I replaced deprecated `protocol.registerStringProtocol` in favor of `protocol.handle`

### DIFF
--- a/src/backend/api/misc.ts
+++ b/src/backend/api/misc.ts
@@ -67,7 +67,6 @@ export const openWebviewPage = (url: string) =>
 export const setZoomFactor = (zoom: string) =>
   ipcRenderer.send('setZoomFactor', zoom)
 export const frontendReady = () => ipcRenderer.send('frontendReady')
-export const loadingScreenReady = () => ipcRenderer.send('loadingScreenReady')
 export const lock = () => ipcRenderer.send('lock')
 export const unlock = () => ipcRenderer.send('unlock')
 export const login = async (sid: string) => ipcRenderer.invoke('login', sid)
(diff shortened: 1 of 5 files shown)


arielj:
I think we need to find a different solution because the problem seems to be that the `show: false` option we use when we initialize the window is not really working for some reason, because the our code is saying "start hidden, and only show the window once the frontend tells you the loading screen of heroic is displayed" but then the window is displayed way earlier than that.

This pr is kinda working around that issue but not fully solving it, that hardcoded color is not great (cause now the page wont blink, but it will start black and then change to another color).

I played around and seems like the issue is this call to `maximize()` https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/main_window.ts#L89

We should wait until the application is ready and the window is displayed to decide if it should be maximized instead. Because I also noticed the blinking doesn't seem to happen if the app is not maximized when closed.

arielj:
related to the removal of the loadingScreenReady event, maybe it can be replaced with https://www.electronjs.org/docs/latest/api/browser-window#event-ready-to-show instead of `did-finish-load`?

the different between the custom event we have and the electron events is that ours should wait until the loading screen is ready, and the electron one doesn't not about the loading screen and it can show the app before that (without a blink, but also without the loading screen animation)

imLinguin:
Indeed, `maximize` causes the window to `show`. Will fix that...

> I'm not sure about that change though

I think it should be fine, the loading indicator is the first thing that displays on the page after being "loaded".

arielj (approved):
Looks good to me!

PR #3303, opened 2023-12-08
[UI/Fix] Categories improvements: make them composable, update style, fix combination with filters

arielj:
This PR does a few different things related to the new Categories feature:

- Updated the style to match the new style used in Filters (this depends on that other PR)
- Made the categories composable, so we can show more than 1 category at a time
- Added the `only` action like we have in the filters, so we can easily filter by a single category
- Replaced the `All categories` "category" with a more clear `Select all` button now that they are toggles
- Fixed an issue when filters like `Show hidden` or `Show Non-Available` were ignored if showing a category
- Save the selected categories when changed and restore on boot (this was partially implemented before but was not working)

One thing that I did NOT implement intentionally is a way to add categories directly from the dropdown, we have to refactor how the categories creation is handled because it currently depends on adding the category for a given game. So to not make this PR bigger, for now I added a message that lets the user know they can add categories using the game menus instead of showing `all / uncategorized`.

---


flavioislima (approved):
amazing stuff, looks a lot better now.
I think the next step (on another Pr though) would be to add a button to add new categories from the categories dropdown, then it will be perfect.

Issue #2030, opened 2022-11-19
Adding to Steam can fail due to error checking that isn't required


Zarklord:
### Describe the bug

Trying to add games to steam throws this error in the log
`(12:20:30) ERROR:   [Shortcuts]:        Can't remove "Goat Simulator 3" from Steam user "<redacted>". "C:\Program Files (x86)\Steam\userdata\<redacted>\config\shortcuts.vdf" is corrupted!`

`One of the game entries is missing the AppName parameter!`

the thing is, my shortcuts.vdf isn't corrupt, it just contained an entry like this:
```
	{
		"appid" "-1493849556"
		"appName" "Dolphin Emulator"
		"exe" "\"E:\\NintendoEmulation\\SteamDolphin\\Dolphin.exe\""
		"StartDir" "E:\\NintendoEmulation\\SteamDolphin\\"
		"icon" "E:\\NintendoEmulation\\Tools\\Dolphin icon.png"
		"ShortcutPath" ""
		"LaunchOptions" ""
		"IsHidden" "0"
		"AllowDesktopConfig" "1"
		"AllowOverlay" "1"
		"OpenVR" "0"
		"Devkit" "0"
		"DevkitGameID" ""
		"DevkitOverrideAppID" "0"
		"LastPlayTime" "1666485127"
		"FlatpakAppID" ""
		"tags"
		{
			"0" "favorite"
			"1" "Emulators"
		}
	}
```
Notice how it has `appName`, but not `AppName` or `appname`, this is due to the error checking not properly checking case insensitivity.
This shortcut does work on steam, so the error checking is at fault here.

IMO, its not really on this program to check for valid VDF entries, it should just add the entry, and possibly warn the user that the shortcuts file is potentially corrupt, and so they might need to go delete the file.

### Add logs

(12:20:31) INFO:    [Shortcuts]:        Prepare Steam images for Goat Simulator 3
(12:20:32) INFO:    [Shortcuts]:        Prepare Steam images for Goat Simulator 3
(12:20:33) INFO:    [Shortcuts]:        Prepare Steam images for Goat Simulator 3
(12:20:33) INFO:    [Shortcuts]:        Prepare Steam images for Goat Simulator 3
(12:20:34) INFO:    [Shortcuts]:        Prepare Steam images for Goat Simulator 3
(12:20:35) INFO:    [Shortcuts]:        Prepare Steam images for Goat Simulator 3
[...]

PR #3320, opened 2023-12-15
[Fix] Steam Shortcuts / Make checkIfAlreadyAdded case-insensitive

### LINKED ISSUE #3124: Heroic may fail to detect entries it added to Steam
### Describe the bug

An entry added to Steam by Heroic can fail to be detected by Heroic if a user adds it to a collection or any other minor edit that rewrites its shortcuts.vdf entry.

### Add logs

```text
< no relevant logs available >
```

### Steps to reproduce

1. Open Heroic
2. Select game
3. Add to Steam
4. Restart Steam
5. Add game to collection
6. Heroic (possibly) no longer registers the game as added to Steam

### Expected behavior

Games that have had minor edits such as adding it to a collection should still be detected as added by Heroic

### Screenshots

_No response_

### Heroic Version

Latest Stable

### System Information

- OS [e. g. "Ubuntu"]: Windows

### Additional information

This seems like it is somewhat related to #2357 

Since Steam is apparently inconsistent in the usage of "appname", "appName", "AppName", etc. when Heroic adds an entry to Steam it will use one form but Steam may use another. Steam appears to be entirely case-insensitive in this regard so it doesn't affect it, but when it updates an entry to add it to a collection it will rewrite it to whatever form it uses and Heroic will no longer detect it.
I checked this by changing the form back and Heroic detects the entry again.

A possible fix would be to make Heroic case-insensitive as well.

One minor complication I could see with that fix is since it appears Heroic detects based entirely on game title, if a user happened to have another instance of an entry with that title (possibly a different version), then Heroic will incorrectly detect it as its own entry and could possibly delete it.

HazardousBackup:
This change should make checkIfAlreadyAdded insensitive to the case of 'AppName' so if it changes it'll still be recognized and able to be removed and not double added.

Fixes #3124 

Change is based on [this StackOverflow post](https://stackoverflow.com/a/47538066)

Tested but not on a completely clean install

---

### DIFF
--- a/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
+++ b/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
@@ -184,7 +184,13 @@ function checkIfShortcutObjectIsValid(
  */
 function checkIfAlreadyAdded(object: Partial<ShortcutObject>, title: string) {
   const shortcuts = object.shortcuts ?? []
-  return shortcuts.findIndex((entry) => entry.AppName === title)
+  return shortcuts.findIndex(
+    (entry) =>
+      entry[
+        Object.keys(entry).find((key) => key.toLowerCase() === 'appname') ??
+          'AppName'
+      ] === title
+  )
 }
 
 /**
--- a/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
+++ b/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
@@ -425,7 +425,12 @@ async function removeNonSteamGame(props: {
     const shortcutObj = content.shortcuts.at(index)!
 
     const exe = shortcutObj.Exe
-    const appName = shortcutObj.AppName
+    const appName =
+      shortcutObj[
+        Object.keys(shortcutObj).find(
+          (key) => key.toLowerCase() === 'appname'
+        ) ?? 'AppName'
+      ]
 
     // remove
     content.shortcuts.splice(index, 1)
--- a/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
+++ b/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
@@ -145,6 +145,17 @@ function hasParameterCaseInsensitive(object: ShortcutEntry, key: string) {
   return Object.keys(object).some((k) => k.toLowerCase() === keyAsLowercase)
 }
 
+/** Return AppName property case insensitive
+ *  @param {Object} object
+ *  @returns Title of Shortcut Entry
+ */
+function getAppName(object: ShortcutEntry) {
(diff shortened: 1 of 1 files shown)


flavioislima:
   - - -
   - - -

HazardousBackup:

arielj:
I'm curious, is heroic adding the information with the wrong case for the key? or is steam transforming them in some way?

because maybe we should make heroic use whatever case steam expects instead? do you have an example of this data with keys that are not `AppName`? maybe we should add the games to steam with `appname` as the key if that makes them all consistent for example

HazardousBackup:
> I'm curious, is heroic adding the information with the wrong case for the key? or is steam transforming them in some way?
> 
> because maybe we should make heroic use whatever case steam expects instead? do you have an example of this data with keys that are not `AppName`? maybe we should add the games to steam with `appname` as the key if that makes them all consistent for example

It's a bit more than that based on #2030

Apparently Steam isn't 100% consistent on the use of case and can vary over time or OSes, but that doesn't matter to Steam because it doesn't appear to care about the case at all (you can use randomized case with no issues). However, Steam does like to convert entrees to whatever case is standard for it at the moment.

So Heroic can't really rely on just changing `AppName` to `appname` because it seems like it could change to `appName` or something else without warning later on.

HazardousBackup:
I just realized that a change needs to be made to `removeNonSteamGame` as well. It uses the AppName value of the shortcut, which since it might be a different case could end up being undefined and so the AppID used for removing the artwork won't be correct.

HazardousBackup:
I've committed the change to removeNonSteamGame and tested it on my install.
Before:
Add to Steam -> images added
Change Collection
Remove from Steam -> images stay

After:
Add to Steam -> images added
Change Collection
Remove from Steam -> images removed (as expected)

It's basically the same as the first change and I don't see any other points where the shortcuts file is being accessed that way, so it should be complete.

arielj (commented):
maybe this is a bit nitpicky, but since we already have something similar in this function

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts#L143

maybe we move getting the value to a function too for consistency:

```
function getAppName(object) {
  const foundKey = Object.keys(object).find((key) => key.toLowerCase() === 'appname') ?? 'AppName'
  return object[foundKey]
}
```

then you can do `return getAppName(shortcuts)` and `const appName = getAppName(surtcubOjb)`

HazardousBackup:
@arielj Nitpicky or not, it's not a big deal and it helps with consistency. Had I noticed that other function then I likely would've implemented it the same way.

Bonus is that it helps keep the other functions cleaner.

arielj (approved):
Looks good to me :+1:

Issue #3324, opened 2023-12-16
[Bug] Heroic UI does not respect UI scale after a resize


allora:
### Describe the bug

When using a UI scale that is not 100%, the UI will revert to 100% scale after resizing the window.

### Add logs

```text
(21:48:46) INFO:    [Legendary]:        Legendary location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(21:48:46) INFO:    [Gog]:              GOGDL location: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(21:48:46) INFO:    [Connection]:       Connectivity: check-online
(21:48:46) INFO:    [Connection]:       Pinging external endpoints
(21:48:46) INFO:    [Nile]:             Refreshing library...
(21:48:46) INFO:    [Nile]:             Refreshing Amazon Games...
(21:48:46) INFO:    [Nile]:             Running command: NILE_CONFIG_PATH=/home/allie/.config/heroic/nile_config /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/nile library sync
(21:48:46) INFO:    [Nile]:             Game list updated, got 14 games
(21:48:46) INFO:    [Backend]:          DRM module staus {
  "oimompecagnajdejgnnjijobebaeigek": {
    "name": "Widevine Content Decryption Module",
    "status": "new",
    "version": "4.10.2710.0"
  }
}
(21:48:46) WARNING: [Backend]:          Protocol already registered.
(21:48:46) INFO:    [Connection]:       Connectivity: online
(21:48:46) INFO:    [Gog]:              Checking if login is valid
(21:48:46) INFO:    [Gog]:              Running command: /opt/Heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/allie/.config/heroic/gog_store/auth.json auth
(21:48:46) INFO:    [Backend]:          AreWeAntiCheatYet data downloaded
(21:48:46) INFO:    [Backend]:          Loading Screen Ready
(21:48:46) INFO:    [Nile]:             Saved user data to config file
(21:48:46) INFO:    [Frontend]:         Refreshing undefined Library
(21:48:46) WARNING: [Backend]:          refresh not implemented on Sideload Library Manager
[...]

Issue #2156, opened 2022-12-07
Zoom resets when switching from FULLSCREEN to WINDOWED and vice versa (Windows 11)


RaikoWasTaken:
### Describe the bug

Zoom, edited through accessibility settings, resets when switching from FULLSCREEN to WINDOWED and vice versa. I need to edit it again for it to go back to normal.
I'm using windows 11
This does not happen when I'm still in the accessibility settings page 

### Add logs

```text
(17:44:51) INFO:    [Connection]:       Connectivity: check-online
(17:44:51) INFO:    [Connection]:       Pinging external endpoints
(17:44:51) INFO:    [Legendary]:        Running command: C:\Users\Raiko\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary --version
(17:44:51) INFO:    [Legendary]:        Legendary location: C:\Users\Raiko\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary
(17:44:51) INFO:    [Gog]:              GOGDL location: C:\Users\Raiko\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\gogdl
(17:44:51) WARNING: [Backend]:          Protocol already registered.
(17:44:51) INFO:    [Connection]:       Connectivity: online
(17:44:52) INFO:    [Gog]:              Running command: C:\Users\Raiko\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\gogdl --version
(17:44:53) INFO:    [Backend]:          VCRuntime is installed
(17:44:53) INFO:    [Backend]:          Checking for new Heroic Updates
(17:44:53) INFO:    [Backend]:          Checking for current version changelog
(17:44:53) INFO:    [Frontend]:         Refreshing Library
(17:44:53) INFO:    [Legendary]:        Refreshing library...
(17:44:53) INFO:    [Legendary]:        Refreshing Epic Games...
(17:44:53) INFO:    [Legendary]:        Game list updated, got 5 games & DLCs
(17:44:53) INFO:    [Backend]:          Frontend Ready
(17:44:53) INFO:    [ProtocolHandler]:  received 'heroic://launch/0a2d9f6403244d12969e11da6713137b'
(17:44:53) ERROR:   [Gog]:              Could not get game info for 0a2d9f6403244d12969e11da6713137b, returning empty object. Something is probably gonna go wrong soon
[...]

PR #3325, opened 2023-12-16
[UI/Fix] Set zoom limits along with zoom level instead of zoom factor

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3324 and fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2156

Looks like in some cases, the zoom factor set with the accessibility settings was getting reset when the windows size changed.

I could not reproduce it during development but I was able to reproduce it using the Flatpak container (both on my linux desktop and on the steam deck). Set a zoom of 80%, move to another screen, and resize the window.

This PR uses the zoomLevel instead of zoomFactor and also sets the visual level limit to 1 and that seems to work correctly (tested both during development and with the generated flatpak).

---


src/backend/main.ts
```diff
@@ -276,9 +276,9 @@ const processZoomForScreen = (zoomFactor: number) => {
   const screenSize = screen.getPrimaryDisplay().workAreaSize.width
   if (screenSize < 1200) {
     const extraDPIZoomIn = screenSize / 1200
-    return zoomFactor * extraDPIZoomIn
+    return (zoomFactor * extraDPIZoomIn - 1) / 0.2
   } else {
-    return zoomFactor
+    return (zoomFactor - 1) / 0.2
```
arielj, line 281: instead of setting a zoom factor (like 1.2, 1.4, 0.8, etc), `setZoomLevel` expects an integer with increments of `0.2` zoom

so a zoom of 80% is 0.8 zoom factor so it's `-1` for zoom level
120% is 1.2 zoom factor so it's 1 zoom level

flavioislima (approved):
lgm

PR #3330, opened 2023-12-18
[FEAT] Show Genres and Release date on Game Page

flavioislima:

---


src/frontend/screens/Game/GamePage/components/ReleaseDate.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    return (
+      windowsReleaseDate ||
+      t('label.unknownReleaseDate', 'Unknown Release Date')
+    )
+  }
+
+  return (
+    <div className="releaseDate">
+      {t('label.releaseDate', 'Release Date: ')}: {getReleaseDate()}
```
arielj, line 52:
```suggestion
      {t('label.releaseDate', 'Release Date')}: {getReleaseDate()}
```

image

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -355,7 +357,13 @@ export default React.memo(function GamePage(): JSX.Element | null {
                     <DotsMenu gameInfo={gameInfo} handleUpdate={handleUpdate} />
                   </div>
                   <div className="infoWrapper">
+                    <Genres genres={wikiInfo?.pcgamingwiki?.genres} />
```
arielj, line 360:
```suggestion
                    <Genres genres={wikiInfo?.pcgamingwiki?.genres || []} />
```

just a suggestion so we don't have to consider `undefined` for this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3330/files#diff-b822c7f041f98de8200abf8822131fecf49d7e6fa1456370c34a738d604bc76eR8, makes it a bit simpler

src/frontend/screens/Game/GamePage/components/ReleaseDate.tsx
```diff
@@ -0,0 +1,57 @@
...
+type ReleaseDateProps = {
+  date: string[] | undefined
+  isLinuxNative: boolean
+  isMacNative: boolean
+}
+
+const ReleaseDate: React.FC<ReleaseDateProps> = ({
+  date,
+  isLinuxNative,
+  isMacNative
```
arielj, line 14: we should be able to get these 2 and `isLinux` and `isMac` all from the GameContext so we don't need to pass them as props and we don't need to check `platform === ...`

```ts
const { is } = useContext(GameContext)
```

and then

```ts
      if (
        (platformName === 'Linux' && is.linuxNative && is.linux) ||
        (platformName === 'OS X' && is.macNative && is.mac)
      ) {
        return releaseDate
      }
```

arielj:
there seems to be some edge cases when parsing the release dates:
image
image

arielj:
something that maybe could be in a future release would be to parse the date and then use internationalization, because now it's displayed as is in english

Etaash-mathamsetty:
Instead of printing the month as February I think we should use the number instead, then no translations are needed

arielj:
> Instead of printing the month as February I think we should use the number instead, then no translations are needed

even if showing just number, we should still use localization, different languages have different conventions to show dates with just numbers (some countries use DD/MM/YYY,Y others YYYY-MM-DD, others MM/DD/YYYY)

src/frontend/screens/Game/GamePage/components/ReleaseDate.tsx
flavioislima: I tottally forgot about this new context you added on your refactor. Nice!

src/frontend/screens/Game/GamePage/index.tsx
flavioislima: true, will fix

flavioislima:
Yes, using Intl would be nice but we need to check if we can convert the current received string from PCGW to a real date.

flavioislima:
@arielj fixed all issues you pointed out. Thanks for testing it.

public/locales/en/translation.json
```diff
@@ -346,7 +346,8 @@
         }
     },
     "label": {
-        "loading": "Loading"
+        "loading": "Loading",
+        "releaseDate": "Release Date: "
```
arielj, line 350:
```suggestion
        "releaseDate": "Release Date"
```

src/frontend/screens/Game/GamePage/components/ReleaseDate.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const dateObj = new Date(date)
+
+  // Check if dateObj is a valid date
+  if (isNaN(dateObj.getTime())) {
+    return null
+  }
+
+  return dateObj.toLocaleDateString(undefined, options)
```
arielj, line 30: we should probably use the user's selected language here instead of undefined, now it uses the system's language instead of heroic's language

image

flavioislima: I was thinking about that and I think the system language is ideal in this case.
Because imagine if the person, like me, is from Brazil, latin America in general. And then uses Heroic in English. We will see the date in US standards and that is not ideal IMO. 
Also because the language is different from metrics.

imagine if we decide to show measurements in inches and foot instead of meters and centimeters just because we want heroic translations to be in English 😓

flavioislima: Also, it is the same logic we are applying on the time container to show the information of last played, etc.

arielj: yeah, honestly I don't know what users expect, I'm fine with this

arielj (approved):
Looks good :+1:

PR #3331, opened 2023-12-18
[Fix] Resolve 7/8 dependabot alerts

CommandMC:
The version of `steam-shortcut-editor` we were using required an old version of lodash. In terms of functionality, nothing changed between 3.1.1 and 3.1.3, so we can just update it.

The last alert is about `axios`, which we can't easily update right now (we'd need #3218 for that).

---


arielj (approved):
looks good to me

PR #3335, opened 2023-12-20
[Experimental] Automatic installation of known winetricks fixes

arielj:
This PR adds an experimental feature to install known winetricks fixes for games when the Prefix is created. This is waaaay less complete than https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2176 but it was really simple to implement and we can have this working while the other more complete solution is implemented.

For my library, this PR makes more than 30 games work out of the box.

I added a bunch of known fixes from my tests in https://github.com/arielj/test-appname-winetricks (temporarily while we test things, then I plan to move it into the heroic org). We can update things over time adding more known winetricks fixes (I have more fixes to add but it's late already).

The way this works is:
- when a game is installed, we try to download the corresponding `appName-store.json` file from the repository that holds that info in parallel and store it
- when a prefix is updated, if the experimental feature is toggled on and one of those files is stored matching appName and store, we install the winetricks packages from the json one by one

You can see the list of games I added fixes for in https://github.com/arielj/test-appname-winetricks?tab=readme-ov-file#fixes-in-the-repo, you should be able to install and run any of those out of the box now (you can try with and without the experimental feature and clearing the prefix to compare)

For games that are already installed, you can download the corresponding file from the repo and put it in `<heroic-config-path>/fixes/<appName>-<store>.json`.

---


Etaash-mathamsetty:
I think legendary supports the game's setup scripts, I think some games can be fixed if we run those rather than creating a long list of winetricks fixes (fall guys for example would be fixed)

arielj:
> I think legendary supports the game's setup scripts, I think some games can be fixed if we run those rather than creating a long list of winetricks fixes (fall guys for example would be fixed) This PR is still useful, but maybe we can clump the functionalities together into one PR?

I'd rather keep 2 different functionalities in 2 different PRs, they are not really connected to need them together, that would just delay this PR that works independently and it's already implemented and functional

Etaash-mathamsetty:
maybe we can add a eos overlay thingy as well into the list for winetricks, for games that need it installed

arielj:
I insist, other features can be added in another PR, I don't see the point of blocking this on something extra that I have to investigate how to do with new logic that is unrelated to what's implemented here

there's a more complex solution that aims to solve multiple cases of workarounds that are not winetricks related in commandmc's PR

src/frontend/screens/Settings/components/ExperimentalFeatures.tsx
```diff
@@ -4,13 +4,13 @@ import useSetting from 'frontend/hooks/useSetting'
...
 
-const FEATURES = ['enableNewDesign']
+const FEATURES = ['enableNewDesign', 'automaticWinetricksFixes']
 
 const ExperimentalFeatures = () => {
   const { t } = useTranslation()
   const [experimentalFeatures, setExperimentalFeatures] = useSetting(
     'experimentalFeatures',
-    { enableNewDesign: false }
+    { enableNewDesign: false, automaticWinetricksFixes: false }
```
flavioislima, line 13: Will this be shown on Windows?

flavioislima (approved):
Looks good but I could not find the code that avoids rendering the option on Windows so would be good to double-check.

src/frontend/screens/Settings/components/ExperimentalFeatures.tsx
arielj: yeah, this is displayed on windows

it doesn't really do anything on windows, but I'll update it to not show the option to avoid confusing the user

arielj:
I'll do 3 things later today:
- fix conflicts
- hide the experimental flag on windows
- transfer the repo from my account to Heroic

PR #3354, opened 2023-12-27
[UX] Improvements when using a controller to navigate the interface

arielj:
This PR improves the UX of heroic with a controller with a few tweaks:

- Pressing the `B` button (in the steam deck layout, I think it's `O` if using a PS controller) now closes any dialog, not only the install dialog, and it doesn't go back in history if a dialog is open

This was really annoying, now we can press `B` to close the Settings dialog, Categories dialog, or the Uninstall dialog for example

- After closing a dialog, the element that was focused before opening it gains focus again

this was REALLY annoying, it feels way better now, for example when opening the settings or logs for a card, after closing the dialog we are still focused on the same card instead of nowhere

- When moving the focus in the library, now cards are scrolled into view so they are completely visible

this fixes an issue that when you were at the top or bottom of the screen, you had to do a few more moves to actually get the screen to scroll before the focus changes to the next element

---

### DIFF
--- a/src/frontend/components/UI/Dialog/components/Dialog.tsx
+++ b/src/frontend/components/UI/Dialog/components/Dialog.tsx
@@ -5,7 +5,8 @@ import React, {
   SyntheticEvent,
   useCallback,
   useEffect,
-  useRef
+  useRef,
+  useState
 } from 'react'
 
 interface DialogProps {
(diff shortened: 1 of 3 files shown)


flavioislima (approved):
looks good

PR #3355, opened 2023-12-27
[Tech] Refactor how we write to game logs

arielj:
We currently use explicit calls to `writeFileSync` and `appendFileSync` when we want to write game logs. This PR extracts that into more descriptive logger helper methods (`initGameLog` and `appendGameLog`) to simplify the usage so we can abstract the actual file name calculation and writing of the content.

I'm doing this because I want to add more logging and make it all async in the future, so I want to first simplify the API.

Nothing is changed in the functionality, logs should work just the same as before.

---

### DIFF
--- a/src/backend/logger/logfile.ts
+++ b/src/backend/logger/logfile.ts
@@ -6,10 +6,15 @@ import {
   appendFileSync
 } from 'graceful-fs'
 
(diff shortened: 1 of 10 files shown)


flavioislima (approved):
LGTM

PR #3356, opened 2023-12-27
Install EOS overlay by default

Etaash-mathamsetty:
Enable the EOS overlay by default to fix some games from complaining about the overlay dlls being missing. 

~~Will be testing shortly~~


Etaash-mathamsetty:
im not sure why this constant legendaryConfigPath is not initialized, and im not sure how to fix it either

arielj:
> im not sure why this constant legendaryConfigPath is not initialized, and im not sure how to fix it either. As far as I can tell it should be initialized already

the `constants` module is mocked for tests, you probably need to define that config in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/__mocks__/constants.ts for the tests to work

Etaash-mathamsetty:
> > im not sure why this constant legendaryConfigPath is not initialized, and im not sure how to fix it either. As far as I can tell it should be initialized already
> 
> the `constants` module is mocked for tests, you probably need to define that config in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/__mocks__/constants.ts for the tests to work

it's not just the tests, I can't get regular old heroic to start either

CommandMC:
You are introducing circular dependencies with this (basically, Vite can't figure out a way to correctly lay out the code).

Take this small example:
```ts
// a.ts
import { c } from './c'
export const a = 'a' + c

// b.ts
import { a } from './a'
export const b = 'b'

// c.ts
import { b } from './b'
export const c = 'c'
```
The `c` imported in `a.ts` will be undefined, since there's no way to arrange the files properly (in reality, Vite might notice that you're not using `b` in `c.ts` and thus drop the import, but you get the idea)

You now have two options:
1. Move the constants defined in `eos_overlay.ts` out of the global scope and into a function definition
  ```ts
  const getEosOverlayConstants = () => ({
    currentVersionPath: ...,
    installedVersionPath: ...,
    // ...
  })
  ```
2. Move the constants from `eos_overlay.ts` into `constants.ts`  
  I wouldn't recommend this, as they are really only relevant for the EOS overlay functions
3. Untangle our dependency spaghetti
  This would involve removing as many imports as possible from `constants.ts`, possibly moving functions into other files, and generally reworking our "we have one file for all constants" model. Suffice it to say, I wouldn't recommend this either

Etaash-mathamsetty:
> You are introducing circular dependencies with this (basically, Vite can't figure out a way to correctly lay out the code).
> 
> Take this small example:
> 
> ```ts
> // a.ts
> import { c } from './c'
> export const a = 'a' + c
> 
> // b.ts
> import { a } from './a'
> export const b = 'b'
> 
> // c.ts
> import { b } from './b'
> export const c = 'c'
> ```
> 
> The `c` imported in `a.ts` will be undefined, since there's no way to arrange the files properly (in reality, Vite might notice that you're not using `b` in `c.ts` and thus drop the import, but you get the idea)
> 
> You now have two options (well, technically 3, but the third one really isn't practical):
> 
>     1. Move the constants defined in `eos_overlay.ts` out of the global scope and into a function definition
> 
> 
> ```ts
> const getEosOverlayConstants = () => ({
>   currentVersionPath: ...,
>   installedVersionPath: ...,
>   // ...
> })
> ```
> 
>     2. Move the constants from `eos_overlay.ts` into `constants.ts`
>        I wouldn't recommend this, as they are really only relevant for the EOS overlay functions
> 
>     3. Untangle our dependency spaghetti
>        This would involve removing as many imports as possible from `constants.ts`, possibly moving functions into other files, and generally reworking our "we have one file for all constants" model. Suffice it to say, I wouldn't recommend this either, but it is the direction we have to go if we want to avoid this spontaneous combustion in the future

thanks for the advice :)

Etaash-mathamsetty:
ok now after my last bug fix to make sure it ony enables if it's not enabled already it breaks legendary and the apps can't get the account id

Etaash-mathamsetty:
I guess ill just revert that change for now

src/backend/storeManagers/legendary/eos_overlay/eos_overlay.ts
```diff
@@ -14,11 +14,19 @@ import { Path } from '../commands/base'
...
+function getConstants(): {
+  currentVersionPath: string
+  installedVersionPath: string
+  defaultInstallPath: string
+} {
+  const currentVersionPath = join(legendaryConfigPath, 'overlay_version.json')
+  const installedVersionPath = join(legendaryConfigPath, 'overlay_install.json')
+  const defaultInstallPath = join(toolsPath, 'eos_overlay')
+  return { currentVersionPath, installedVersionPath, defaultInstallPath }
+}
```
arielj, line 28: I think you can simplify this by converting each of the constants into separate functions, I did it like that in this PR since I also had to import `isEnabled` https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3394/files#diff-efa75d3c5004afcce158bc8224b64a644dd5f88c4a9d0f7a03e1f6f6253963c7R17

Etaash-mathamsetty: ill wait for the other one to merge and then ill just rebase it on top of yours

arielj (approved):
Looks good to me, and works as expected from my tests.

PR #3357, opened 2023-12-27
Update legendary to v0.20.34

Etaash-mathamsetty:
Simple legendary update so that we are no longer using some -git binaries and are instead using the binaries from a release. 


public/bin/darwin/legendary
arielj, line 1: I see the permissions change from 755 to 644, don't we need the execute permissions in these files?

Etaash-mathamsetty: Definitely not needed for windows, but idk about Mac os, I'll change all the perms to what they were

flavioislima (approved):

flavioislima:
Better merge this one right away so we can get any issues when testing the other PRs.

PR #2817, opened 2023-06-22
[Feat/Mac] Add Whisky Support

IsaacMarovitz:
This PR adds support for Whisky.

**TODO:**
- [ ] Add Whisky bottle detection
- [ ] Test test tset

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #3358, opened 2023-12-28
[FEAT-macOS] Add Whisky Support

flavioislima:
- Finds and show Whisky Wine and set it as type `toolkit` type.
- Fix DXVK always installing on macOS due to `autoVkd3dInstall` always being true.

---


flavioislima:
I am using almost the same code from the old PR by @IsaacMarovitz on his old PR #2817. So all credits to him.

src/backend/launcher.ts
```diff
@@ -352,10 +352,10 @@ async function prepareWineLaunch(
     if (gameSettings.autoInstallDxvk) {
       await DXVK.installRemove(gameSettings, 'dxvk', 'backup')
     }
-    if (gameSettings.autoInstallDxvkNvapi) {
+    if (isLinux && gameSettings.autoInstallDxvkNvapi) {
       await DXVK.installRemove(gameSettings, 'dxvk-nvapi', 'backup')
     }
-    if (gameSettings.autoInstallVkd3d) {
+    if (isLinux && gameSettings.autoInstallVkd3d) {
```
CommandMC, line 358: IMO, instead of gating these behind `isLinux`, we should make sure the relevant settings values are never `true` on macOS in the first place

src/backend/tools/index.ts
```diff
@@ -191,6 +191,10 @@ export const DXVK = {
       return true
     }
 
+    if (isMac && tool !== 'dxvk') {
+      return true
+    }
```
CommandMC, line 196: Somewhat of the same issue here: Instead of just returning true for macOS, change callers of this so they don't try to do this when on macOS

src/backend/launcher.ts
Etaash-mathamsetty: that's already been done, see the change where the defaults have been changed

flavioislima: Thats been done like Etaash said they still come true on macOS and I could not find exactly why. on the app settings it is true only if it is linux but on mac somehow still sets to true the vkd3d. So with the current logic, if vkd3d is true it installs the dxvk as well and it breaks the GPTK.

src/backend/tools/index.ts
flavioislima: the thing here is that we don't handle this logic on frontend, that is the one that calls this method.
We only ask for dxvk and then the installer figure it out what tool to install.

src/backend/launcher.ts
CommandMC: Ah I see, it *was* `true` by default once, so now that it's set to that value, it stays on that (even if we change the default)

I guess this is fine then

src/backend/tools/index.ts
CommandMC: So what I gather from that is that the Frontend should only ever request the `dxvk` tool... So why do you need this check then?

Anyways, assuming you've tested this for all Wine types available on macOS, I don't see a huge issue here.
We should go through and actually figure out what's wrong here, but that can happen someday in the future

src/backend/config.ts
```diff
@@ -137,12 +138,14 @@ abstract class GlobalConfig {
...
     const gamingPortingToolkitWine = await getGamingPortingToolkitWine()
+    const getWhysky = await getWhisky()
 
     return new Set([
       ...gamingPortingToolkitWine,
       ...crossover,
       ...wineOnMac,
-      ...wineskinWine
+      ...wineskinWine,
+      ...getWhysky
```
Etaash-mathamsetty, line 148: small spelling error here

Etaash-mathamsetty (commented):
it looks fine, but I can't test on a mac os device unfortunately

src/backend/config.ts
flavioislima: fixed

flavioislima:
@Etaash-mathamsetty if you guys can't test just check the code but leave a review.

CommandMC (approved):

PR #3364, opened 2023-12-30
Install legendary prerequisites

Etaash-mathamsetty:
Installs the prerequisites that legendary gives in the json when installing a game, we can rip out the dedicated ubisoft stuff with this change. 


Etaash-mathamsetty:
the only issue is to move all the setup functions so that they only happen post install or on first run, otherwise the setup functions will constantly run each launch

src/backend/main.ts
```diff
@@ -1099,12 +1093,6 @@ ipcMain.handle(
       return false
     })
 
-    // Stop display sleep blocker
-    if (powerDisplayId !== null) {
-      logInfo('Stopping Display Power Saver Blocker', LogPrefix.Backend)
-      powerSaveBlocker.stop(powerDisplayId)
-    }
```
arielj, line 1106: is this intentional? looks unrelated to installing pre-requisites

src/backend/storeManagers/legendary/setup.ts
```diff
@@ -30,107 +27,27 @@ export const legendarySetup = async (appName: string) => {
...
-    const oneWeekAgo = new Date()
-    oneWeekAgo.setDate(oneWeekAgo.getDate() - 7)
-    if (cachedAt > oneWeekAgo) {
-      return true
+  if (
+    gameInfo.install?.platform === 'Windows' ||
+    gameInfo.install?.platform === 'Win32' ||
+    gameInfo.install?.platform === 'windows'
+  ) {
+    const info = await getInstallInfo(appName, gameInfo.install.platform)
```
arielj, line 35: this call to getInstallInfo can throw an exception, in other places in the code we do a try/catch just in case

src/backend/storeManagers/legendary/setup.ts
```diff
@@ -30,107 +27,27 @@ export const legendarySetup = async (appName: string) => {
...
+    if (
+      info.manifest.prerequisites &&
```
arielj, line 37: just curious, I see we have this typed as an object, I wonder if there's no game that has this as an array of prerequisites maybe?

the ones I tested they have either 1 prerequisite or null, but I'm curious if we should handle that possibility

arielj (commented):
Looks good! I tested with a few games with and without prerequisites (including ubisoft) and all works fine

I just left some comments and one question

arielj:
Testing with RollerCoaster Tycoon 3 I get this error:
image

maybe we need to skip the args for the command when it's an empty string?

```
			"prerequisites": {
				"ids": [
					"DirectXJun2010Redist"
				],
				"name": "DirectX June 2010 Redistributable",
				"path": "directx_Jun2010_redist/DXSETUP.exe",
				"args": ""
			},
```

src/backend/storeManagers/legendary/setup.ts
```diff
@@ -30,107 +27,27 @@ export const legendarySetup = async (appName: string) => {
...
+      info.manifest.prerequisites.path.length > 0
+    ) {
+      await runWineCommandOnGame(appName, {
+        commandParts: [
+          join(
+            gameInfo.install.install_path ?? '',
+            info.manifest.prerequisites.path
+          ),
+          info.manifest.prerequisites.args
+        ],
```
imLinguin, line 47: args is a string right? you should shlex.split them

src/backend/storeManagers/legendary/setup.ts
Etaash-mathamsetty: where? I don't see a try catch anywhere

Etaash-mathamsetty: there's a try catch within the function itself it looks like

Etaash-mathamsetty:
> Testing with RollerCoaster Tycoon 3 I get this error: image
> 
> maybe we need to skip the args for the command when it's an empty string?
> 
> ```
> 			"prerequisites": {
> 				"ids": [
> 					"DirectXJun2010Redist"
> 				],
> 				"name": "DirectX June 2010 Redistributable",
> 				"path": "directx_Jun2010_redist/DXSETUP.exe",
> 				"args": ""
> 			},
> ```

i think the shlex.split should fix this, could you retest?

src/backend/storeManagers/legendary/setup.ts
arielj: The function has a try/catch inside, but when it catches an error it throws a new one https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/a460b9721910a7a55ed59ccb15aae9992dad15ec/src/backend/storeManagers/legendary/library.ts#L245

I found one getInstallInfo using the try/catch here https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/a460b9721910a7a55ed59ccb15aae9992dad15ec/src/frontend/screens/Library/components/InstallModal/DownloadDialog/index.tsx#L267 but now I see it's a different `getInstallInfo` function, not that one


I do think we should handle that possible throw though (I think the other places that are not handling it are something that we should fix eventually too)

Etaash-mathamsetty: makes sense

arielj:
> i think the shlex.split should fix this, could you retest?

yep! the shlex.split fixes that for rollercoaster tycoon

arielj (approved):
Looks good to me!

I only tested on linux, I understand this has no impact on windows.

Also, the only game that really needed the pre-requisites for me was XCOM 2 and that now works out of the box with this change.

For the other games I own it looks like the pre-requisites are not really needed for the games to work (I didn't need the pre-requisites for any other game), but I think it's good to install them anyway, doesn't seem to create any issue anyway.

src/backend/storeManagers/legendary/setup.ts
```diff
@@ -30,107 +29,31 @@ export const legendarySetup = async (appName: string) => {
...
-const downloadIfNotCached = async (cachePath: string, url: string) => {
-  if (existsSync(cachePath)) {
-    // if the cached file exist but it was cached more than a week ago, download a new one
-    // an outdated installer can make the installation fail according to some report
-    const cachedAt = statSync(cachePath).mtime
-    const oneWeekAgo = new Date()
-    oneWeekAgo.setDate(oneWeekAgo.getDate() - 7)
-    if (cachedAt > oneWeekAgo) {
-      return true
+  if (
```
flavioislima, line 32: Maybe we could just:
```
const winPlatforms = ['windows', 'Windows', 'Win32']
if (winPlatforms.includes(gameInfo.install?.platform)) {
```

flavioislima (approved):
left a small comment, looks good otherwise.

Issue #3341, opened 2023-12-21
Uninstalling Gamescope can restrict access to games through GUI


Alzarath:
### Describe the bug

After enabling Gamescope on a game, if you uninstall Gamescope, the menu gives the information "We could not found gamescope on the PATH. Install it or add it to the PATH." and hides Gamescope options. However, the options are still there and the game will attempt to play with gamescope, but it can't because it's not installed.

### Add logs

```text
N/A
```

### Steps to reproduce

1. Enable any Gamescope options with Gamescope installed
2. Uninstall Gamescope
3. Attempt to launch the game

### Expected behavior

When Gamescope is uninstalled, either the game will ignore the Gamescope options entirely, or the options will be visible for the user to make corrections.

### Heroic Version

Latest Stable

### System Information

Arch Linux x86_64

### Additional information

Workarounds involve, of course, re-installing Gamescope, or deleting the gamescope block from the game's config file.

arielj:
For reference if anyone wants to work on fixing this, I reported this possibility during the development of the feature, pointing to the problematic code https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3089/files#r1349736177

PR #3367, opened 2023-12-30
[Fix] Ignore gamescope if enabled and not found instead of failing

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3341

Instead of failing with an error, ignore the `gamescope` setup if enabled but not found.

This can happen if gamescope was configured for a game and then uninstalled.

With this change it will just show a warning in the logs and still run the game as if the setting was disabled.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -136,83 +136,81 @@ async function prepareLaunch(
   ) {
     const gameScopeBin = await searchForExecutableOnPath('gamescope')
     if (!gameScopeBin) {
-      return {
-        success: false,
-        failureReason:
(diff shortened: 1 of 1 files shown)


Etaash-mathamsetty (approved):
this diff is hard to read but if I read it right it looks fine

arielj:
> this diff is hard to read but if I read it right it looks fine

yeah, github is trying to be too smart, I just changed the return with a `logWarning` and moved the rest into an `else` (just added indentation to most of the lines there), github is going crazy but there's no other change on that part of the code apart of the extra indentation

PR #3371, opened 2024-01-01
[UX] Add offline support filter

arielj:
This PR adds a new filter to list only games flagged as supporting offline mode.

This can be really handy when deciding what to install for the steam deck for example knowing it would work on the go with no internet.

(I understand that some games are mislabeled, but we can't do anything about that)

---


flavioislima (approved):
Tested here and looks good 👍🏽

Issue #3328, opened 2023-12-17
Cancelling the installation of a Wine version does not update the status of the item


arielj:
### Describe the bug

When cancelling the installation of a Wine version, the download stops but the status stays showing the `installing` state (progres, eta, cancel button).

### Add logs

```text
Nothing relevant.
```

### Steps to reproduce

1. Start installing a Wine
2. Wait a moment so the progress starts
3. Cancel the installation
4. Status is stuck until the items are re-rendered

### Expected behavior

The status of the canceled item should change to the `not installed` status

### Heroic Version

Latest Stable

### System Information

Not relevant

PR #3374, opened 2024-01-02
[Fix/UI] Fix status after aborting wine installation. Make width of progress more stable

arielj:
This PR fixes 2 problems related to the download progress of wines:

- After canceling the installation of a wine, the progress was stuck as `downloading` instead of reverting back to idle state. Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3328

- The width of the progress was not constant so it kept jumping around as wider/narrower digits changed. Now I added a new line and it's centered so it doesn't move that much:

image

---

### DIFF
--- a/src/frontend/screens/WineManager/components/WineItem/index.tsx
+++ b/src/frontend/screens/WineManager/components/WineItem/index.tsx
@@ -94,6 +94,11 @@ const WineItem = ({
           default:
             break
         }
+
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
Looks good.
I think this whole page needs a redesign and maybe we already have one on the Figma file.

PR #3384, opened 2024-01-04
Write game logs async. Write messages in bulks.

arielj:
This PR is a small refactor on how we write game logs to make the write async instead of inline before launching games.

This has a few benefits:
- games start launching faster since we don't wait for logs to be written
- if any issue happens while writing to logs, games start anyway

Also, with this change, appended messages are queued and written to disk at 1 second intervals instead of writing them immediately to disk, reducing the amount of writes we do.

Also, with this refactor, we can now include more logs in a single place (the initLog method), unblocking many issues related to adding more information in the logs.

---


src/backend/logger/logger.ts
```diff
@@ -355,10 +355,89 @@ export function logFileLocation(appName: string) {
...
+
+  constructor(appName: string) {
+    this.appName = appName
+    this.initialized = false
+    this.filePath = lastPlayLogFileLocation(appName)
+  }
+
+  logMessage(message: string) {
+    // initialize the value if undefined
+    if (this.queue === undefined) this.queue = []
```
CommandMC, line 375: The commend doesn't really add anything here

```suggestion
    this.queue ??= []
```

src/backend/logger/logger.ts
```diff
@@ -355,10 +355,89 @@ export function logFileLocation(appName: string) {
...
+  async initLog() {
+    try {
+      const info = await getSystemInfo()
+      const systemInfo = await formatSystemInfo(info)
+
+      // init log file and then append message if any
+      writeFile(
+        this.filePath,
+        'System Info:\n' + `${systemInfo}\n` + '\n',
+        () => {
```
CommandMC, line 397: This will swallow up errors
```suggestion
        (error) => {
          // Handle `error` somewhere, not sure if `throw`ing it would work since we're in a callback
```

I'd also prefer the use of the `fsPromises.writeFile` function here (since you're already in an async method)

src/backend/logger/logger.ts
```diff
@@ -355,10 +355,89 @@ export function logFileLocation(appName: string) {
...
+  appendMessages() {
+    const messagesToWrite = this.queue
+
+    // clear pending message if any
+    this.queue = []
+
+    // clear timeout if any
+    delete this.timeoutId
+
+    if (!messagesToWrite || !messagesToWrite.length) return
```
CommandMC, line 416:
```suggestion
    if (!messagesToWrite?.length) return
```

src/backend/logger/logger.ts
```diff
@@ -355,10 +355,89 @@ export function logFileLocation(appName: string) {
...
+
+    // if we have messages, write them and check again in 1 second
+    // we start the timeout before writing so we don't wait until
+    // the disk write
+    this.timeoutId = setTimeout(() => this.appendMessages(), 1000)
+
+    try {
+      appendFileSync(this.filePath, messagesToWrite.join(''))
```
CommandMC, line 424: Shouldn't the join separator here be `\n`?

src/backend/logger/logger.ts
```diff
@@ -355,10 +355,89 @@ export function logFileLocation(appName: string) {
   return join(gamesConfigPath, `${appName}.log`)
 }
 
+const logsWriters: { [appName: string]: LogWriter } = {}
```
CommandMC, line 358:
```suggestion
const logsWriters: Record<string, LogWriter> = {}
```

src/backend/logger/logger.ts
```diff
@@ -355,10 +355,89 @@ export function logFileLocation(appName: string) {
...
+      appendFileSync(this.filePath, messagesToWrite.join(''))
+    } catch (error) {
+      // ignore failures if messages could not be written
+    }
+  }
+}
+
 export function appendGameLog(appName: string, message: string) {
-  appendFileSync(lastPlayLogFileLocation(appName), message)
+  if (!logsWriters[appName]) logsWriters[appName] = new LogWriter(appName)
```
CommandMC, line 432:
```suggestion
  logsWriters[appName] ??= new LogWriter(appName)
```

src/backend/logger/logger.ts
```diff
@@ -355,10 +355,89 @@ export function logFileLocation(appName: string) {
...
+  logsWriters[appName].logMessage(message)
+}
+
+export function initGameLog(appName: string) {
+  if (!logsWriters[appName]) logsWriters[appName] = new LogWriter(appName)
```
CommandMC, line 437:
```suggestion
  logsWriters[appName] ??= new LogWriter(appName)
```

src/backend/logger/logger.ts
arielj: that was my first guess too, but it adds extra new lines in the middle, looks like the logs we put already include the new lines at the end

src/backend/logger/logger.ts
flavioislima: I have to say that this new JS feature makes the code a lot more unreadable IMO.
is up to you guys to use it or not, I will try not to.

for me doing a `IF` is always easier to understand than using this `??=` 😅 😂

CommandMC: I'd argue that a lot of "JS-isms" (specifically around null-checking "shortcuts" and functional programming) are hard to read when first confronted with them.
Nullish coalescing assignment *is* ever-so-slightly faster, but feel free to use whatever you wish to use

arielj: at least for me I'm used to this syntax from other languages, I didn't know it was a thing in javascript

if it helps I can say you get used to it very quickly and then you miss that when it doesn't exist

src/backend/logger/logger.ts
```diff
@@ -355,10 +356,94 @@ export function logFileLocation(appName: string) {
...
+        this.appendMessages()
+      } catch (error) {
+        logError(
+          [`Failed to initialize log ${this.filePath}:`, error],
+          LogPrefix.Backend
+        )
+      }
+    } catch (error) {
+      logError(['Failed to fetch system information', error], LogPrefix.Backend)
+    }
```
CommandMC, line 410: The try-catch blocks here are now a little hard to follow

```suggestion
    let systemInfo
    try {
      systemInfo = await getSystemInfo().then(formatSystemInfo)
    } catch (error) {
      logError(['Failed to fetch system information', error], LogPrefix.Backend)
      return
    }

    // init log file and then append message if any
    try {
      await writeFile(
        this.filePath,
        'System Info:\n' + `${systemInfo}\n` + '\n'
      )

      this.initialized = true
      this.appendMessages()
    } catch (error) {
      logError(
        [`Failed to initialize log ${this.filePath}:`, error],
        LogPrefix.Backend
      )
    }
```

arielj: I'm refactoring that in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3394/files#diff-3b36a7de185377bbb618ac5e7bc20b0780c6ea78b5d5ac53c7b9a7469448a988L388, what do you think?

(also using `writeFileSync`, realized there's no need to do `await writeFile` if we can just use the sync function?)

CommandMC: That's much more readable IMO, thanks

*As far as I know* `writeFileSync` blocks execution until the write operation is finished, meanwhile Promise-based APIs are "transparent" to the JS engine (so it knows that we're waiting on something and can in the meantime handle other things, like callbacks or IPC events)

flavioislima (approved):
Looks god

src/backend/logger/logger.ts
flavioislima: yeah, makes sense. Maybe its the age that is affecting me and making me stand for the 'old ways' 😂

Issue #3383, opened 2024-01-04
Offline Mode Not Appearing for Some Epic Games


Ponjen:
### Describe the bug

Games that support Offline Mode or are DRM free don't have an Offline Mode toggle in Heroic. Three games as examples are Soulstice, DNF Duel, and 20 Minutes Till Dawn. Heroic says these games require an online connection, but this is wrong. DNF Duel and Soulstice work 100% offline on Epic's launcher and supports Offline Mode. 20 Minutes Till Dawn is DRM free. However, because Heroic doesn't offer the Offline Mode toggle for these games, an arbitrary online connection is forced in order to boot them.

### Add logs

```text
System Info:

CPU: 8x AMD Custom APU 0405

Memory: 12.36 GB (used: 7.32 GB)

GPUs:

  GPU 0:

    Name: Advanced Micro Devices, Inc. [AMD/ATI] VanGogh [AMD Custom GPU 0405]

    IDs: D=163f V=1002 SD=0123 SV=1002

    Driver: amdgpu

OS: SteamOS 3.5.7 holo (linux)

The current system is a Steam Deck (model: LCD)

We are running inside a Flatpak container

Software Versions:

  Heroic: 2.11.0 Kumachi

  Legendary: 0.20.33 Undue Alarm

  gogdl: 0.7.3

  Nile: 1.0.0 Jonathan Joestar

Game Settings: {

	"autoInstallDxvk": true,

	"autoInstallVkd3d": true,

	"preferSystemLibs": false,

	"enableEsync": true,

	"enableFsync": true,

	"nvidiaPrime": false,

	"enviromentOptions": [],

	"wrapperOptions": [],

	"showFps": false,

	"useGameMode": false,

	"language": "",

	"wineVersion": {

		"bin": "/home/deck/.local/share/Steam/compatibilitytools.d/GE-Proton8-25/proton",

		"name": "Proton - GE-Proton8-25",

		"type": "proton"

	},

	"winePrefix": "/home/deck/Games/Heroic/Prefixes/Baldurs Gate 3",

	"wineCrossoverBottle": ""

}

Game launched at: Wed Jan 03 2024 19:41:45 GMT-0500 (Eastern Standard Time)

Launch aborted: Offline mode not supported
```

### Steps to reproduce

1. Disable internet connection
2. Launch a DRM free EGS game
3. See Error "offline mode is not supported"

### Expected behavior

[...]

PR #3385, opened 2024-01-04
[UX] Allow setting offline mode even if game does not have the metadata

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3383

Some games seem to not have the `CanRunOffline` metadata but still work fine in offline mode. This PR allows users to toggle that on so they can still try to run the game.

When the option is turned on and the game does not allow offline mode, this warning is displayed:

image

This will also solve a problem that users sometimes can't find the option and they ask over Discord. Now the option will be always there. In some cases it may not work but we have the warning to let users know it's not heroic's fault but the game not supporting offline mode.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -86,10 +86,9 @@ async function prepareLaunch(
 
   // Check if the game needs an internet connection
   if (!gameInfo.canRunOffline && offlineMode) {
-    return {
-      success: false,
-      failureReason: 'Offline mode not supported'
-    }
+    logWarning(
+      'Offline Mode is on but the game does not allow offline mode explicitly.'
+    )
   }
 
   // Update Discord RPC if enabled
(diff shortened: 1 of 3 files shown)


src/backend/launcher.ts
```diff
@@ -86,10 +86,9 @@ async function prepareLaunch(
 
   // Check if the game needs an internet connection
   if (!gameInfo.canRunOffline && offlineMode) {
-    return {
-      success: false,
-      failureReason: 'Offline mode not supported'
-    }
+    logWarning(
+      'Offline Mode is on but the game does not allow offline mode explicitly.'
+    )
```
arielj, line 91: we don't stop the launching of the game, but we still log this information cause it's important

src/frontend/screens/Settings/components/OfflineMode.tsx
```diff
@@ -26,17 +28,28 @@ const OfflineMode = () => {
 
   const [offlineMode, setOfflineMode] = useSetting('offlineMode', false)
 
-  if (isDefault || !canRunOffline) {
+  if (isDefault) {
```
flavioislima, line 31: This feature is for Epic games only, right? because GOG will work fine offline since they don't have DRM. Same for Amazon games. So we could hide it for GOG or disable it and show a information about it.

arielj: I changed this to only show the option for Epic games.

flavioislima (approved):
nice stuff

PR #3389, opened 2024-01-05
[Fix] Hide hidden games in Favourites and Recent Games sections

Sasikuttan2163:
Hidden games were shown in the Favourites section of the Library before even if "Show Hidden" was unticked. Now they are not shown when hidden. If all favourite games are hidden, the favourites section is not shown at all.

Before with Show Hidden off
image

Before with Show Hidden on
image

After with Show Hidden off
image

After with Show Hidden on
image

---


src/frontend/screens/Library/index.tsx
```diff
@@ -306,13 +306,24 @@ export default React.memo(function Library(): JSX.Element {
...
+  const favouriteGamesList = useMemo(() => {
+    return showHidden
+      ? favouriteGames.list
+      : favouriteGames.list.filter(
+          (game) =>
+            !hiddenGames.list
+              .map((hidden: HiddenGame) => hidden?.appName)
+              .includes(game.appName)
+        )
+  }, [favouriteGames, showHidden, hiddenGames])
```
arielj, line 318: I think this would be more readable if we do:

```ts
const favouriteGamesList = useMemo(() => {
  if (showHidden) {
    return favouritesGames.list
  }

  return ... thefileredlist
}
```

that ternary operator is kinda complex


also, with that change you can move the `hiddenGames.list.map` thing outside of the filter callback, now it's being calculated for every element:

```ts
const favouriteGamesList = useMemo(() => {
  if (showHidden) {
    return favouritesGames.list
  }

  const hiddenAppNames = hiddenGames.list.map((hidden: HiddenGame) => hidden?.appName)
                                                                      // do we need this "?" here?, can this be undefined?

  return favouriteGames.list.filter((game) => !hiddenAppNames.includes(game.appName))
}
```

Sasikuttan2163: Oh yeah that makes sense, I'll patch it up.

Sasikuttan2163:
Now, games set to hide are hidden in Recent Games section too.

Before with Show Hidden off
image

After with Show Hidden off
image


Before with Show Hidden on
image

After with Show Hidden on
image

Known bug:
The first game in the Recent Games list does not show a masked thumbnail.
image

Sasikuttan2163:
Fixed issue of first game in Recents page not having a greyed out image if it is hidden and Show Hidden is enabled. 
image

Now I think it should be ready.

flavioislima (approved):
Looks good. Thanks for fixing that!

PR #3393, opened 2024-01-05
Fixes related to "Installing Prerequisites" status

arielj:
This fixes 2 issues:

- we were showing the `Installing Prerequisites` for GOG games but we don't really do that yet
- we were setting the `Installing Prerequisites` status when running an exe on the prefix but never resetting that state

I also moved the `Installing Prerequisites` status for the Amazon setup to be displayed after the prefix is verified instead of before, to happen right before the post install step.

---


Etaash-mathamsetty (approved):
Looks good

PR #3398, opened 2024-01-06
Remove fixes file when a game is uninstalled

arielj:
In https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3335 we are adding the feature to install known winetricks fixes. That downloads the `fixes` file from the https://github.com/Heroic-Games-Launcher/known-fixes repo during the installation of a game.

The problem is that we have no way or removing the file once the winetricks fix is not needed anymore, so if a game is downloaded today and a fix is not needed anymore in 2 months, if I install the game it will still run the winetricks commands.

This PR changes that so when a game is uninstalled, if a fixes file exists it gets deleted. That way, if we install the game in a few months and the fix is not needed, it won't be in the `known-fixes` repo and we don't download anything.

I did a small refactor cause I needed that map from runner to store and it was already defined in 3 other places

---

### DIFF
--- a/src/backend/downloadmanager/utils.ts
+++ b/src/backend/downloadmanager/utils.ts
@@ -13,6 +13,7 @@ import { fixesPath } from 'backend/constants'
 import path from 'path'
 import { existsSync, mkdirSync } from 'graceful-fs'
 import { platform } from 'os'
+import { storeMap } from 'common/utils'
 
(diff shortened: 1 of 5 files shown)


CommandMC (approved):
Ideally the "doing the refactor" and "introducing the change" parts of this PR would be in two different commits...
but we'll squash it down when merging anyways, so that's not a huge issue

PR #3401, opened 2024-01-06
[UX/Linux] Open main window when clicking the Tray icon

arielj:
We were using the `double-click` event but that only works on Mac and Windows, not on Linux https://www.electronjs.org/docs/latest/api/tray#event-double-click-macos-windows

This PR changes the code to use `click` instead which works on all platforms.

Other electron apps like discord or slack work like this too.

---


Etaash-mathamsetty (approved):
lgtm let me test

Etaash-mathamsetty:
doesn't seem to work that well, prolly cuz click also opens context menu

flavioislima (approved):
looks good

PR #3402, opened 2024-01-06
[UX] Add Accessibility option to disable closing dialogs when clicking outside

arielj:
This PR adds an Accessibility option to let users decide if they want the dialogs to be closed or not when clicking outside.

Clicking outside of dialogs currently (not in main but in the stable release) closes dialogs without any confirmation by the user. This is problematic because clicking outside by mistake can mean information is lost.

Current main branch disables the click outside to close the dialog, but some users might want that to work (it's a common thing to do).

This PR gives the users the option to pick which behavior they want.

One thing I noticed is that disabling the click outside also disables closing dialogs with the `Esc` key, so I added a handled for that too.

---


flavioislima (approved):
lgtm

PR #3394, opened 2024-01-06
[Logs] Add more info in game logs

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3187, and fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3186 and fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3104

This PR adds more information in the logs to make them more useful when users share them over Discord.

New info:
- Name of the game/app
- Runner
- If it's native or not
- Install dir
- If the game is running in Desktop or Game mode (only on a Steamdeck)
- If the EOS overlay is enabled or not (for epic games only)
- If there are winetricks packages installed (for non-native games)
- Also added a `===== End of game logs =====` line at the end so we know users didn't cut log when sharing

I also included a few extra changes that I needed:
- I need the gameInfo in the log functions, so I had to update some code to handle game info instead of just app name
- Nile setup function was running even with undefined GameInfo
- I had to do a small refactor of the eos_overlay import because it was going in an import error
- Moved the log of the setting inside the initLog function to do that also async

The log now looks like this:

```
Launching "A Short Hike" (legendary)
Native? No
Installed in: /home/ariel/Games/Heroic/AShortHike

System Info:
CPU: 16x AMD Ryzen 7 3700X 8-Core Processor
Memory: 16.63 GB (used: 8.66 GB)
GPUs:
  GPU 0:
    Name: Advanced Micro Devices, Inc. [AMD/ATI] RX 5700 XT RAW II
    IDs: D=731f V=1002 SD=5701 SV=1682
    Driver: amdgpu
OS: Linux Mint 21.2 (Victoria) (linux)

The current system is not a Steam Deck
We are not running inside a Flatpak container

Software Versions:
  Heroic: 2.11.0 Kumachi
  Legendary: 0.20.34 Direct Intervention
  gogdl: 0.7.3
  Nile: 1.0.0 Jonathan Joestar

Game Settings: {
	"autoInstallDxvk": true,
	"autoInstallVkd3d": true,
	"preferSystemLibs": false,
	"enableEsync": true,
	"enableFsync": true,
	"nvidiaPrime": false,
	"enviromentOptions": [],
	"wrapperOptions": [],
	"showFps": false,
	"showMangohud": false,
	"useGameMode": false,
	"language": "",
	"wineVersion": {
[...]

PR #3404, opened 2024-01-07
Revert logging if the EOS overlay is enabled

arielj:
In a previous PR I added a line to log if the EOS overlay is enabled or not: this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3394/files#diff-3b36a7de185377bbb618ac5e7bc20b0780c6ea78b5d5ac53c7b9a7469448a988R429-R437

This works fine when the prefix is already created but produces an error when the prefix is not ready yet. We have to log this information after making sure the prefix is initialized.

I'll revert this change for now since this other PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3356 will enable the EOS overlay by default anyway

We can include this in the log in the future (I just don't want this to be a blocker or a problem for a release).

---


Etaash-mathamsetty (approved):
lgtm, no need to log it if it's installed already

PR #3405, opened 2024-01-07
Revert printing winetricks packages in logs

arielj:
Similar to https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3404

With the new information in the logs, one of the lines was the list of winetricks packages installed. The problem is that, when the prefix is not yet created, this call to get the winetricks packages initializes the prefix at the wrong point of the process and all the setup code is skipped.

I'll revert this for now, in the future I'll make change some of this to add the logs only after the prefix is ready.

---


Etaash-mathamsetty (approved):
lgtm

PR #3410, opened 2024-01-07
Only try to do EOS operations on a path if there's actually a Wineprefix there

CommandMC:
Previously, trying to enable/disable/check the status of the EOS overlay for a game with an invalid prefix directory would lead to an error message.
We're now handling this correctly (not enabling/disabling it, always reporting that it's disabled)

Technically, this works by adding:
- a new type, `ValidWinePrefix`. It's an extension of the `Path` type, meaning it already makes sure the string inside it is a valid path, and it now also verifies that there's a `user.reg` file inside the directory it's pointing to
- a new function inside the EOS handler, `getWinePrefixFolder`. It deduplicates some of the logic used in the `enable`/`disable`/`isEnabled`, and now also makes sure the configured prefix path is contains a valid prefix

Error message before:
image

No error message now:
image

---

### DIFF
--- a/src/backend/storeManagers/legendary/commands/base.ts
+++ b/src/backend/storeManagers/legendary/commands/base.ts
@@ -1,6 +1,7 @@
 import { z } from 'zod'
 import path from 'path'
 import { hasGame } from '../library'
+import { existsSync } from 'graceful-fs'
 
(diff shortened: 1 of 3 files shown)


flavioislima (approved):
Nice fix!

PR #3420, opened 2024-01-08
General repo cleanup

CommandMC:
Some (early) spring cleaning now that the release is out of the way

These are a couple small changes mostly just focusing on readability. Functionality is not changed in any way


arielj (approved):
this looks good to me :+1:

PR #3428, opened 2024-01-11
[Tech/Demo] Migrate Wine Manager status state to Zustand

CommandMC:
This is a proof-of-concept of using Zustand (https://github.com/pmndrs/zustand) for one of our many states-to-be-managed

I don't know how much I have to explain here, the code looks very straight-forward *to me*, but it's probably confusing in some places if you're not used to the syntax. Please ask questions! With that said, a general explanation:

First off, we've had somewhat of a code split here before (with some state being updated by the Frontend and some by the Backend). This *would* work just fine now as well, but I thought I might as well move everything over to the Backend while I'm at it. Thus, the `installWineVersion` and `removeWineVersion` IPC callbacks no longer return anything, instead settings state directly themselves

On the Frontend side, we first create a store to hold our state (in our case, the easiest structure is a Record, mapping wine version names to their state), and then give the Backend an option to update this state using the (already present but now somewhat modified) `progressOfWineManager` Frontend message.
The WineItem components then listen to their respective version's state updates (using `useShallow` to really only re-render when the specific version changes). Everything else is then handled by Zustand and React

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,8 @@
     "steam-shortcut-editor": "3.1.3",
     "tslib": "2.5.0",
     "xvfb-maybe": "^0.2.1",
-    "zod": "3.22.3"
+    "zod": "3.22.3",
+    "zustand": "^4.4.7"
   },
   "devDependencies": {
     "@electron/notarize": "^2.3.0",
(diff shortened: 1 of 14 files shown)


arielj (commented):
this looks pretty good to me, this already solves the problem I had with the frontend losing the wine installation progress status temporarily while components are mounted

I like that the migration can be done really granular

arielj (approved):

Issue #2663, opened 2023-04-29
Add an explanation of what "Clear Heroic Cache" and "Reset Heroic" do


arielj:
### Problem description

We have the 2 buttons in the settings, but there's no indication of they will do. If you click the "Clear Heroic Cache" button a message saying "Cache cleared: Heroic cache cleared." appears but it's still no indication of that it's clearing.

The "Reset Heroic" button only tells you what it will clear after you click it, there should be a way to know what it will do without a user having to click on it.

### Feature description

Next to the "Clear Heroic Cache" and "Reset Heroic" buttons there should be elements that a user can interact with to see what each of those buttons will do, explaining for example what are the things we cache, similar to the modal that appears when clicking the Reset Heroic button but without the user having to click the actual actions.

PR #3431, opened 2024-01-12
[UX] Add info about what Reset Heroic and Clear Heroic Cache do

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2663

image
image
image

So users can see what's going to happen before clicking the buttons.

---


CaptainMorgan12:
Is there a way to perform these actions in the terminal?

CommandMC:
@CaptainMorgan12 This is not an issue tracker. If you have a question about Heroic, please ask it either on our Discord or on the "Issues" page

CaptainMorgan12:
Ok thanks will post it there.

src/frontend/screens/Settings/components/ClearCache.tsx
```diff
@@ -0,0 +1,81 @@
...
+        </ul>
+        {t('settings.advanced.clearCache.help6', 'This will NOT delete:')}
+        <ul>
+          <li>{t('settings.advanced.clearCache.help7', 'Store login')}</li>
+          <li>{t('settings.advanced.clearCache.help8', 'Installed games')}</li>
+          <li>{t('settings.advanced.clearCache.help9', 'Games settings')}</li>
+          <li>
+            {t('settings.advanced.clearCache.help10', 'Heroic configuration')}
+          </li>
+        </ul>
```
CommandMC, line 62: I'd be great if it would be possible to make this whole list freely translatable (I worry about languages not conforming to this format), but I don't think i18next offers something for this

CommandMC (approved):

src/frontend/screens/Settings/components/ClearCache.tsx
arielj: yeah, I don't really know how to mix a list with translations as a single translation (I could add `{{}}` placeholder to replace with the html tags, but it would be a mess as a single string)

CommandMC: I don't think that'd work actually (it would just display the tags as text, e.g. `<ul><li>foo</li>...`). You'd have to use `dangerouslySetInnerHTML`, which is, well, dangerous

Issue #3433, opened 2024-01-12
getSteamDeckInfo does not detect Steam Deck with APU 0932


m3e-g:
### Describe the bug

Some of the Steam Deck OLED devices have different APU names and thus fail the current detection.

```
System Info:
CPU: 8x AMD Custom APU 0932
Memory: 15.53 GB (used: 2.28 GB)
GPUs:
  GPU 0:
    Name: Advanced Micro Devices, Inc. [AMD/ATI] undefined
    IDs: D=1435 V=1002 SD=0123 SV=1002
    Driver: amdgpu
OS: SteamOS 3.5.7 holo (linux)

The current system is not a Steam Deck
```

### Add logs

```text
Launching "Marvel's Guardians of the Galaxy" (legendary)
Native? No
Installed in: /run/media/mmcblk0p1/Heroic/MarvelGOTG

System Info:
CPU: 8x AMD Custom APU 0932
Memory: 15.53 GB (used: 2.28 GB)
GPUs:
  GPU 0:
    Name: Advanced Micro Devices, Inc. [AMD/ATI] undefined
    IDs: D=1435 V=1002 SD=0123 SV=1002
    Driver: amdgpu
OS: SteamOS 3.5.7 holo (linux)

The current system is not a Steam Deck
We are running inside a Flatpak container

Software Versions:
  Heroic: 2.12.0 Vegapunk
  Legendary: 0.20.34 Direct Intervention
  gogdl: 0.7.3
  Nile: 1.0.0 Jonathan Joestar

Game Settings: {
	"autoInstallDxvk": true,
	"autoInstallVkd3d": true,
	"preferSystemLibs": false,
	"enableEsync": true,
	"enableFsync": true,
	"launcherArgs": "-nolauncher",
	"nvidiaPrime": false,
	"enviromentOptions": [],
	"wrapperOptions": [],
	"savesPath": "",
	"showFps": false,
	"targetExe": "",
	"useGameMode": false,
	"language": "",
	"wineVersion": {
		"bin": "/home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/Wine-GE-latest/bin/wine",
		"name": "Wine - Wine-GE-latest",
		"type": "wine",
		"lib": "/home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/Wine-GE-latest/lib64",
[...]

PR #3434, opened 2024-01-12
Added the second AMD APU code value to Steam Deck detection

m3e-g:
This should fix the https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3433

I've added the new APU name to the first check. If neither is true, isDeck is set to false.

### DIFF
--- a/src/backend/utils/systeminfo/steamDeck.ts
+++ b/src/backend/utils/systeminfo/steamDeck.ts
@@ -7,7 +7,11 @@ type SteamDeckInfo =
   | { isDeck: false; model?: undefined }
 
 function getSteamDeckInfo(cpus: CpuInfo[], gpus: GPUInfo[]): SteamDeckInfo {
-  if (cpus[0]?.model !== 'AMD Custom APU 0405') return { isDeck: false }
+  if (
+    cpus[0]?.model !== 'AMD Custom APU 0405' &&
+    cpus[0]?.model !== 'AMD Custom APU 0932'
+  )
+    return { isDeck: false }
 
   const primaryGpu = gpus.at(0)
   if (!primaryGpu || primaryGpu.vendorId !== '1002') return { isDeck: false }


flavioislima:
   - - -
   - - -

m3e-g:

src/backend/utils/systeminfo/steamDeck.ts
```diff
@@ -7,7 +7,11 @@ type SteamDeckInfo =
   | { isDeck: false; model?: undefined }
 
 function getSteamDeckInfo(cpus: CpuInfo[], gpus: GPUInfo[]): SteamDeckInfo {
-  if (cpus[0]?.model !== 'AMD Custom APU 0405') return { isDeck: false }
+  if (
+    cpus[0]?.model !== 'AMD Custom APU 0405' &&
+    cpus[0]?.model !== 'AMD Custom APU 0932'
+  )
+    return { isDeck: false }
```
arielj, line 14: I'm checking on my steamdeck and it looks like we have a `HOSTNAME` env variable with the value `steamdeck`, maybe we can just use that info for this instead? I imagine it's more future proof

what do you think @CommandMC ?

We have other variables like:
```
HOME=/home/deck
LOGNAME=deck
```

and many env variables including `/home/deck` or `steamdeck`

m3e-g: `HOSTNAME` is changable from the steam ui itself and I haven't checked how multi users are handled on the deck (may have different `HOME`?). 

Problem I see with the environment variables is that anyone running steamos image on their machine would have the same set, and hardware ID is harder to spoof. (If someone did it, then he had his purpose)

CommandMC: > I haven't checked how multi users are handled on the deck

As far as I can tell, they aren't, you're always the `deck` user account

Regardless, I agree, we should only look at things definitively proving we're on the Steam Deck hardware, not just SteamOS or something that's very similar to what the Deck might look like

CommandMC (approved):

PR #3440, opened 2024-01-15
[Ref/Fix] Improving checkDiskSpace

CommandMC:
Our `checkDiskSpace` IPC function had a couple issues in edge-case scenarios, which could lead to Heroic freezing. Those are now resolved

To reproduce the issue (in the current Heroic release)
1. Use Heroic on Windows
2. Specify a nonexistent path in the install dialog

It'd seem that an issue like this was also discovered on Linux by a user on our Discord ([here](<https://discord.com/channels/812703221789097985/1196122444172378212/1196122444172378212>)). I can't reproduce this myself, but it might have the same root cause as the Windows issue

---


src/backend/utils/filesystem/windows.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!disk.FreeSpace || !disk.Size) continue
+
+    if (path.startsWith(disk.Caption))
+      return { freeSpace: disk.FreeSpace, totalSpace: disk.Size }
+  }
+
+  return { freeSpace: 0, totalSpace: 0 }
+}
+
+async function isWritable_windows(path: Path): Promise<boolean> {
```
flavioislima, line 59: Since I am not a powershell expert I asked copilot to check some improvements here and one of the answers makes sense:

**Error handling for the PowerShell command**: The function currently assumes that the PowerShell command will always succeed. It might be a good idea to add some error handling for the case where the command fails. For example, you could check if `stdout` is empty or if it contains an error message, and return `false` in these cases.


What do you think?

Probably same for the other command about `getDiskInfo`

CommandMC: Error checking is already happening below:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/49fc1bd83fa7738365cf8110d81be4a0ff78a62c/src/backend/utils/filesystem/windows.ts#L71-L76

The command is supposed to output an array of AccessControlEntries in JSON format, so we try to parse the output as JSON & then verify the structure of what we parsed with Zod. If anything goes wrong with PowerShell, either parsing as JSON will fail (for example with empty output / a PS error message), or Zod will catch type mismatches (in case PowerShell ever does a breaking API change, which I don't see happening)

This same pattern (run PS and make it output JSON, parse output with JSON and Zod) is already used in other places we run PowerShell, for example here:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/9b0982c85fc74d9bb976c3a7a6fe43d805b0dce0/src/backend/utils/systeminfo/osInfo/windows.ts#L10-L28

flavioislima: Makes sense, I think Copilot only checked the PS command itself but missed the try/catch around the JSON parse that can really fail if the PS command doesn't work.

flavioislima (approved):
lgtm

PR #3446, opened 2024-01-16
[UX] Prefer using release dates from each runner

imLinguin:
Now we'll use release dates obtained from each runners API. We were already making a requests to obtain a store URL for example, so I just re-used that.

I skipped that for Amazon for now...  
We have each game's release for each entry in nile's `library.json`, we'd probably need to include that data in GameInfo object.


flavioislima (approved):
LGTM

Issue #3450, opened 2024-01-18
Epic library is not refreshed correctly at startup


arielj:
### Describe the bug

After buying a game, opening Heroic shows the library refreshing but the newly bought game does not show up. An extra refresh is needed (either manually refreshing or re-opening Heroic).

The game is listed by Legendary, so the issue is Heroic not updating the internal copy of the library.

I "bought" the Prince of Persia demo ant this is the log when I open Heroic:

### Add logs

```text
[17/1/2024, 21:56:10] LEGENDARY_CONFIG_PATH=/home/ariel/.config/heroic/legendaryConfig/legendary /home/ariel/dev/oss/HeroicGamesLauncher/public/bin/linux/legendary list --third-party
[cli] INFO: Logging in...
[Core] INFO: Logging in...
[cli] INFO: Getting game list... (this may take a while)
[17/1/2024, 21:56:11] LEGENDARY_CONFIG_PATH=/home/ariel/.config/heroic/legendaryConfig/legendary /home/ariel/dev/oss/HeroicGamesLauncher/public/bin/linux/legendary --version
legendary version "0.20.34", codename "Direct Intervention"
[Core] INFO: Fetching metadata for 4 app(s).

Available games:
...
... all my games
...
 * Prince of Persia: The Lost Crown Demo (App name: b32890f4b0b0454aae1d876b00f2410a | Version: 1.3)
  - This game can be activated directly on your Ubisoft account and does not require legendary to install/run. Use "legendary activate --uplay" and follow the instructions.
...
...
```

### Steps to reproduce

1. With Heroic closed, buy/add a free game in the Epic store
2. Open Heroic
3. Wait until the loading animation ends
4. Search for the game > it's not in the library
5. Click the Refresh library button
6. Search for the game > it is in the library

### Expected behavior

The initial load should updated the library once it finished the background refresh.

### Heroic Version

Latest Stable

### System Information

Not relevant

PR #3452, opened 2024-01-19
[Fix] Wait for legendary refresh to finish before storing array

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3450

We were not waiting for the promise to finish, so the library was not updated with the new results from legendary, we were writing the old library to the store instead, requiring a second reload to get things updated.

---


Etaash-mathamsetty (approved):
lgtm

PR #3456, opened 2024-01-21
[UX] Add some copy in the Wine manager explaining what are the different Wines in linux

arielj:
This PR addressed a common problem that users try `Wine-GE-Proton-LoL` without knowing it will most likely fail if not used for League of Legends.

This can also help address common questions:
- which wine variant is recommended (based on Glorious Eggroll's reddit posts)
- why we allow downloading proton-ge if wine-ge is recommended
- hopefully users understand the part about Proton-GE-Proton providing useless logs too

image
image
image

One extra note: since I had to add some style and the scss file simple enough, I moved to be a plain css file since we can now use native css nesting. We can eventually stop using sass.

---


src/frontend/screens/WineManager/index.tsx
```diff
@@ -110,6 +114,43 @@ export default function WineManager(): JSX.Element | null {
     }
   }, [repository])
 
+  let wineVersionExplanation = <></>
```
CommandMC, line 117: The `useMemo` hook should be used for values like this, e.g.
```ts
const wineVersionExplanation = useMemo(() => { /* switch case in here, returning the appropriate value */ }, [repository.type])
```

That'll prevent the value from being re-calculated on every re-render

src/frontend/screens/WineManager/index.css
CommandMC, line 1: Ideally use `git mv` when renaming a file (even when making changes to it). That'll keep its history alive (right now, it looks like this whole file's contents were added by this PR, which isn't the case) and also makes it easier to review what's actually relevant

arielj: I just tried `git mv` and the thing is that this PR does update the whole file (I added the `.wineManager` class wrapper to nest elements inside, so all lines have an updated indentation)

src/frontend/screens/WineManager/index.css
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  .infoBox {
+    width: 80%;
+    margin: 0 auto;
+    & svg {
+      margin-inline-end: 0.5rem;
+    }
+  }
```
arielj, line 54: this `.infoBox` is the only new thing in this file, the rest is just updated indentation (I added a `.wineManager` wrapper selector)

src/frontend/screens/WineManager/index.css
CommandMC: Hmm, usually GitHub picks up on files renamed with `git mv` (it should show a "Renamed from ..." note somewhere), and telling it to then not show whitespace changes should only show the new parts of the file

But oh well, this isn't a huge issue

flavioislima (approved):
lgtm

PR #3461, opened 2024-01-21
[Feature] Allow Steam AppId override

HamsterExAstris:
Some games have multiple Steam AppIds, e.g. when separate language editions are sold separately. Even if PCGamingWiki lists the alternative(s), there's no way for a user to select which one to use for images when adding a game to their Steam library.

This PR allows the AppId value to be overridden on a title-by-title basis, in case the detected one is wrong or not the user's preferred mapping.

example:
https://www.pcgamingwiki.com/wiki/Sabbat_of_the_Witch
https://steamdb.info/app/888790/ - English, but no library_hero.jpg or logo.png
https://steamdb.info/app/2458530/ - Chinese, with all the images that HGL uses

I couldn't find examples of tests for similar functionality - if there's something here I _should_ be adding tests for, please let me know!

---


HamsterExAstris:
I misunderstood the docs so I only ran the E2E tests locally, not CI. Working on fixing...

arielj:
There's something I think might be confusing with how this works for the end user, because most users will have the option to automatically add games to steam and in order to set this setting you first have to install the game, find out the image is wrong, find the right steam id, remove the shortcut > configure the id > re add the shortcut

I think it would be better/simpler for the user to have an option to update the shortcut entering some image url instead of fixing it indirectly (by setting the steamid and re-adding it)

I imagine an entry in the 3 dots menu to allow users to edit the shortcut that opens a dialog showing the current image, current text and then you set a new URL or text and save it.

what do you think? I know it's really different but I feel it will be more user friendly

arielj:
Closing since there's been no comments or updates for more than a year

PR #3471, opened 2024-01-25
[Rework] The config system

CommandMC:
Our old settings system was a solid implementation of *a* system to serve this purpose, but over the years, it became clear that *we* need some features to work ever-so-slightly differently. "I'll quickly implement those", I thought... This is the result

- Config values can now be global-only, game-only, or shared (before, they were only either global-only or shared). For some keys (e.g. a game's target EXE), it just doesn't make sense to have a global value
- Default values are now defined in one place
  No more duplicating across Backend, Frontend hook, and Frontend state
- All config keys can be reset to their default values with one click (and now properly handle default state)
  In my opinion, settings should fundamentally have a special state for "Not set by the user". Take the boolean value for DXVK as an example, the user should be able to either explicitly enable/disable it, or leave it to the default (and with that us) to decide the best option for them. Storing this "Not Set" state separately allows us to later change the underlying value (for example turning an option on by default that was off before) with no further action needed by the user and retroactively applying for all their games
  Adding to that, a common problem in support requests is that the user changed something they don't know about, and we then have to instruct them that "This is how X works, *this* is the default value, set it like *this*". A dedicated button for this makes the process much easier
- Proper type checking, everywhere
  Due to, admittedly, our misuse of the old system, there was some confusion between what a settings key *should* be and what it ends up being at runtime. Everything is now type-checked with Zod; what you think a key should be *is* now what a key is
- One Frontend interface for config keys
  The Frontend had... 5? different ways of looking at settings (specifically the global ones). Now we only have one, shared everywhere

For now, this PR mainly needs testing. In the process of this, I had to migrate the old config format to a more... defined new one. I've tried to do this as carefully as possible, but it's possible I've missed some special value of a key somewhere, which will then lead to the new system not accepting the migrated config. As such, please **back up your `config.json` and `GamesConfig` folders** before running this PR.

Some TODOs:
- [X] Thorough testing on all platforms
- [X] `TextInputField` needs to handle the "reset to default" properties (I wasn't able to position the icon correctly)
- [ ] Generally, the UI on this is still a little rough
- [X] Documentation on how all this works internally

---


CommandMC:
I've been dogfooding this basically ever since opening the PR, everything seems to be working fine on both Linux and Windows. Migrating the config to the new format works, setting/resetting keys works, and the default values are respected properly

There's sadly no way to shrink this PR down, migrating just *some* of the config keys to an entirely new system while keeping other keys on the old one doesn't make sense, and that's where the biggest part of the changes comes from

CommandMC:
Going to close this for now as I'll probably never rebase it. Planning on picking some of these changes up in separate, smaller PRs to hopefully make it more review-friendly

PR #3476, opened 2024-01-28
Support additional gamescope options

Etaash-mathamsetty:
Funny story behind this PR: I was trying to play some random Styx game from GOG that had broken mouse capture on kwin wayland, I opened it through gamescope, it works but on the wrong monitor this time, then I wrote the code for this PR really quickly to change the displays, and guess what ... gamescope doesn't support changing the displays. I will have to make a PR there as well

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -214,6 +214,10 @@ async function prepareLaunch(
         }
       }
 
+      gameScopeCommand.push(
+        ...shlex.split(gameSettings.gamescope.additionalOptions)
+      )
+
       // Note: needs to be the last option
       gameScopeCommand.push('--')
     }
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -215,7 +215,7 @@ async function prepareLaunch(
       }
 
       gameScopeCommand.push(
-        ...shlex.split(gameSettings.gamescope.additionalOptions)
+        ...shlex.split(gameSettings.gamescope.additionalOptions ?? '')
       )
 
       // Note: needs to be the last option
--- a/src/frontend/screens/Settings/components/Gamescope.tsx
+++ b/src/frontend/screens/Settings/components/Gamescope.tsx
@@ -25,7 +25,8 @@ const Gamescope = () => {
     upscaleWidth: '',
     upscaleMethod: 'fsr',
     fpsLimiter: '',
-    fpsLimiterNoFocus: ''
+    fpsLimiterNoFocus: '',
+    additionalOptions: ''
   })
   const [fetching, setFetching] = useState(true)
   const [isInstalled, setIsInstalled] = useState(false)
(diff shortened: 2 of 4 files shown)


Etaash-mathamsetty:
this generates a lot of log spam, I think there should be a way to fix this though

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -352,6 +353,29 @@ const Gamescope = () => {
...
+          <FontAwesomeIcon
+            className="helpIcon"
+            icon={faCircleInfo}
+            title={t(
+              'help.gamescope.additionalOptions',
+              'Additional commandline flags to pass into gamescope.'
+            )}
+          />
+        }
+        onChange={(event: ChangeEvent<HTMLInputElement>) =>
```
CommandMC, line 372: You might want to use `onBlur` instead of `onChange` here to only update the config when the user is *done* editing (clicks somewhere else)

src/backend/launcher.ts
```diff
@@ -211,6 +211,10 @@ async function prepareLaunch(
         }
       }
 
+      gameScopeCommand.push(
+        ...shlex.split(gameSettings.gamescope.additionalOptions)
+      )
```
CommandMC, line 216: You have to guard this, `additionalOptions` might be undefined (the types around the config system are a little unsafe)
```suggestion
      if (gameSettings.gamescope.additionalOptions)
        gameScopeCommand.push(
          ...shlex.split(gameSettings.gamescope.additionalOptions)
        )
```
I don't think this lints, don't merge as-is

imLinguin (approved):
LGTM

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -352,6 +353,29 @@ const Gamescope = () => {
...
+          <FontAwesomeIcon
+            className="helpIcon"
+            icon={faCircleInfo}
+            title={t(
+              'help.gamescope.additionalOptions',
+              'Additional commandline flags to pass into gamescope.'
+            )}
+          />
+        }
+        onBlur={(event: ChangeEvent<HTMLInputElement>) =>
```
Etaash-mathamsetty, line 372: onBlur doesn't work for some reason

flavioislima:
I believe that would be good to also add the HDR option.
And when enabled, add the DXVK_HDR=1 to the env variables. 
this is how it works on my machine.

Etaash-mathamsetty:
i think on change is the best option since otherwise you need to have a onChange and an onBlur

imLinguin:
`onBlur` didn't work because you were relying on the value from `useSetting` it didn't change because you couldn't input anything in the first place due to lack of meaningful `onChange`. To implement this properly you need to add separate state for holding input value, then update `useSetting` value `onBlur`

imLinguin:
That would look like this
```patch
   const [fetching, setFetching] = useState(true)
   const [isInstalled, setIsInstalled] = useState(false)
 
+  const [additionalOptions, setAdditionalOptions] = useState(gamescope.additionalOptions)
+
   useEffect(() => {
     setFetching(true)
     window.api
@@ -358,7 +360,7 @@ const Gamescope = () => {
         label={t('options.gamescope.additionalOptions', 'Additional Options')}
         htmlId="additionalOptions"
         placeholder=""
-        value={gamescope.additionalOptions}
+        value={additionalOptions}
         afterInput={
           <FontAwesomeIcon
             className="helpIcon"
@@ -369,7 +371,8 @@ const Gamescope = () => {
             )}
           />
         }
-        onChange={(event: ChangeEvent<HTMLInputElement>) =>
+        onChange={(event: ChangeEvent<HTMLInputElement>) => setAdditionalOptions(event.currentTarget.value)}
+        onBlur={(event: ChangeEvent<HTMLInputElement>) =>
           setGamescope({
             ...gamescope,
             additionalOptions: event.currentTarget.value

```

I imagine while we are at it we could do the same thing for other input fields in this component

Etaash-mathamsetty:
> `onBlur` didn't work because you were relying on the value from `useSetting` it didn't change because you couldn't input anything in the first place due to lack of meaningful `onChange`. To implement this properly you need to add separate state for holding input value, then update `useSetting` value `onBlur`

yeah I was trying to fix this, but I just didn't connect it with useState lol
I was using a regular old variable

flavioislima (approved):
lgtm

PR #3477, opened 2024-01-28
Cleanup `platform` checks. Remove old code.

arielj:
This PR does 3 cleanup/refactor tasks:

- remove some code that is not really needed anymore (I'll add comments in the diff)
- use `isWindows/Linux/Mac` constants when possible instead of `process.platform` or `platform()`
- expose the platform name in the preload file so we don't need the extra IPC calls to get that info

---

### DIFF
--- a/e2e/api.spec.ts
+++ b/e2e/api.spec.ts
@@ -40,10 +40,3 @@ electronTest('gets heroic, legendary, and gog versions', async (app) => {
     expect(compareVersions(gogdlVersion, '0.7.1')).toBeGreaterThanOrEqual(0)
   })
 })
-
-electronTest('test ipcMainInvokeHandler', async (app) => {
-  const page = await app.firstWindow()
-  const platform = await page.evaluate(async () => window.api.getPlatform())
-  console.log('Platform: ', platform)
-  expect(platform).toEqual(process.platform)
-})
--- a/src/backend/config.ts
+++ b/src/backend/config.ts
@@ -275,13 +275,6 @@ class GlobalConfigV0 extends GlobalConfig {
       winePrefix
     } as AppSettings
 
-    // TODO: Remove this after a couple of stable releases
-    // Get settings only from config-store
-    const currentConfigStore = configStore.get_nodefault('settings')
-    if (!currentConfigStore?.defaultInstallPath) {
-      configStore.set('settings', settings)
-    }
-
     return settings
   }
 
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -24,7 +24,7 @@ import {
 } from 'electron'
 import 'backend/updater'
 import { autoUpdater } from 'electron-updater'
-import { cpus, platform } from 'os'
+import { cpus } from 'os'
 import {
   access,
   constants,
@@ -94,7 +94,9 @@ import {
   createNecessaryFolders,
   fixAsarPath,
   isSnap,
-  fixesPath
+  fixesPath,
+  isWindows,
+  isMac
 } from './constants'
 import { handleProtocol } from './protocol'
 import {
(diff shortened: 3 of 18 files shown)


e2e/api.spec.ts
```diff
@@ -40,10 +40,3 @@ electronTest('gets heroic, legendary, and gog versions', async (app) => {
     expect(compareVersions(gogdlVersion, '0.7.1')).toBeGreaterThanOrEqual(0)
   })
 })
-
-electronTest('test ipcMainInvokeHandler', async (app) => {
-  const page = await app.firstWindow()
-  const platform = await page.evaluate(async () => window.api.getPlatform())
-  console.log('Platform: ', platform)
-  expect(platform).toEqual(process.platform)
-})
```
arielj, line 49: I removed this test because I removed the `getPlatform` function, that info is now exposed in the preload.ts file so we don't need an extra call to the backend

src/backend/main.ts
```diff
@@ -831,12 +827,12 @@ ipcMain.handle('getAlternativeWine', async () =>
...
 
-ipcMain.handle('readConfig', async (event, config_class) => {
-  if (config_class === 'library') {
+ipcMain.handle('readConfig', async (event, configClass) => {
+  if (configClass === 'library') {
     await libraryManagerMap['legendary'].refresh()
     return LegendaryLibraryManager.getListOfGames()
   }
-  const userInfo = await LegendaryUser.getUserInfo()
+  const userInfo = LegendaryUser.getUserInfo()
```
arielj, line 835: this await was not needed, the linter was flagging it

src/backend/main.ts
```diff
@@ -831,12 +827,12 @@ ipcMain.handle('getAlternativeWine', async () =>
   GlobalConfig.get().getAlternativeWine()
 )
 
-ipcMain.handle('readConfig', async (event, config_class) => {
-  if (config_class === 'library') {
+ipcMain.handle('readConfig', async (event, configClass) => {
+  if (configClass === 'library') {
```
arielj, line 831: just to follow conventions

src/backend/storeManagers/gog/library.ts
```diff
@@ -696,19 +696,6 @@ export async function getInstallInfo(
     }
   }
   installInfoStore.set(installInfoStoreKey, info)
-  if (!info) {
```
arielj, line 699: a few lines above we do `const info = { ..... }`, I couldn't find any way this could be false since it's defined right before this line

src/backend/storeManagers/gog/user.ts
```diff
@@ -115,26 +115,6 @@ export class GOGUser {
     return JSON.parse(stdout)
   }
 
-  /**
-   * Migrates existing authorization config to one supported by gogdl
```
arielj, line 119: I checked with linguin, we don't need this anymore, it was here temporarily to migrate credentials over a few releases

src/backend/main.ts
```diff
@@ -234,9 +235,7 @@ async function initializeWindow(): Promise<BrowserWindow> {
     handleExit()
   })
 
-  if (isWindows) {
-    detectVCRedist(mainWindow)
-  }
+  detectVCRedist(mainWindow)
```
arielj, line 238: `detectVCRedist` already checks if `isWindows` inside the function

src/backend/config.ts
```diff
@@ -275,13 +275,6 @@ class GlobalConfigV0 extends GlobalConfig {
       winePrefix
     } as AppSettings
 
-    // TODO: Remove this after a couple of stable releases
-    // Get settings only from config-store
-    const currentConfigStore = configStore.get_nodefault('settings')
-    if (!currentConfigStore?.defaultInstallPath) {
-      configStore.set('settings', settings)
-    }
-
```
arielj, line 284: this is not needed anymore

flavioislima (approved):
LGTM

PR #3482, opened 2024-01-31
[FIX]: set initial state properly for password modal

imLinguin:
Since we are loading password from backend asynchronously we need to add useEffect that will react to those changes. It's a small but annoying bug


arielj (approved):
looks good to me

PR #3484, opened 2024-01-31
[Fix] Caching in `downloadFile`

CommandMC:
We were looking at the birthtime of the file, which doesn't actually get updated when re-downloading (as such, nothing was ever cached).

A *very quick* local test suggests that this is working, but I don't have the time to properly test this on multiple systems/OSs right now.

In case this does work as-is everywhere, I've also created the [pci_ids_stable_branch](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/tree/pci_ids_stable_branch) branch from our latest release and the two commits, ideally we'd release this ASAP as a 2.12.1.

---


Etaash-mathamsetty:
I understand the fix, but why is this so important?

Etaash-mathamsetty (approved):
lgtm

PR #3485, opened 2024-02-01
[Test/e2e] Add a mechanism to stub binaries

arielj:
This PR is an initial implementation to be able to stub binaries during e2e tests.

The current `e2e/api.spec.ts` tests are not really doing an e2e test: those tests are testing that the `window.api` functions return a version for each binary but are not really testing the boundaries of the app from the user point of view. They are more like an API test.

With this PR we can test navigating Heroic as a user would, going to Settings > Advanced where the versions are displayed, and we can now stub the binaries to test different outcomes. The current tests only stub a single stdout value, but the subbing supports setting a response Promise if we want to do something more complex to mimic a download progress for example?

I also added a few `if (process.env.CI === 'e2e')` guards in commands that were performing external requests, we shouldn't do that (and eventually we should also add a method to stub those to be able to test different scenarios).

---

### DIFF
--- a/e2e/api.spec.ts
+++ b/e2e/api.spec.ts
@@ -4,14 +4,11 @@ import { electronTest } from './helpers'
 
 declare const window: { api: typeof import('../src/backend/api').default }
 
-electronTest('renders the first page', async (app) => {
-  const page = await app.firstWindow()
+electronTest('renders the first page', async (app, page) => {
   await expect(page).toHaveTitle('Heroic Games Launcher')
 })
 
-electronTest('gets heroic, legendary, and gog versions', async (app) => {
-  const page = await app.firstWindow()
-
+electronTest('gets heroic, legendary, and gog versions', async (app, page) => {
   await test.step('get heroic version', async () => {
     const heroicVersion = await page.evaluate(async () =>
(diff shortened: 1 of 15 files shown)


e2e/api.spec.ts
```diff
@@ -4,14 +4,11 @@ import { electronTest } from './helpers'
 
 declare const window: { api: typeof import('../src/backend/api').default }
 
-electronTest('renders the first page', async (app) => {
-  const page = await app.firstWindow()
+electronTest('renders the first page', async (app, page) => {
```
arielj, line 7: I'm passing `app` and `page` from the helper to avoid calling `const page ....` for every test.

e2e/settings.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+          }
+        } else {
+          return {
+            stdout: '',
+            stderr: ''
+          }
+        }
+      }
+    )
+  })
```
arielj, line 24: I tried to create a helper method to abstract this evaluate/emit code, but I keep running into this issue `Error: electronApplication.evaluate: ReferenceError: .... is not defined` because of how the evaluation context changes

I can add more details if needed

PR #3505, opened 2024-02-02
[PROTON]: use run verb for setup commands

imLinguin:
We were using `waitforexitandrun` for setup commands, which is slow as it first ensures that wineserver exited properly.
`run` is made for exactly the purpose of setup. With that we can re-use wineserver process and in turn make it faster. 

I also made changes to legendary's setup, as it was using function that could recursively call itself in some conditions. This change is here to prevent that from happening in the future when we make changes to when game setup commands are executed.

Proton-GE relies on the verb to determine if it should apply protonfixes, which will be useful for us with introduction of ULWGL.

There is one command I'm not yet sure about - `shutdownWine` function. I'll take care of it in ULWGL PR though

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -338,7 +338,7 @@ async function prepareWineLaunch(
       sendFrontendMessage('gameStatusUpdate', {
         appName,
         runner: 'gog',
-        status: 'playing'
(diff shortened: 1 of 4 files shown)


flavioislima (approved):
lgtm

Issue #3289, opened 2023-12-04
Games don't launch a second time


Gaemfreak:
### Describe the bug

The first time I open Heroic and launch a game it works fine. The second time I try to relaunch the game during that first instance of Heroic being open, say I closed the game and came back later to play again, the game will not launch unless I restart Heroic.

This happens every time. When I open Heroic and go to my library, I get a "Cannot get game info" error. Refresh library does nothing. Restarting Heroic fixes it once until it breaks when trying to launch the game a second time.

### Add logs

```text
Heroic: (20:18:39) INFO:    [Backend]:          Heroic is already running, quitting this instance

Fortnite: 

System Info:
CPU: 16x AMD Ryzen 7 3700X 8-Core Processor             
Memory: 34.29 GB (used: 11.01 GB)
GPUs:
  GPU 0:
    Name: NVIDIA Corporation GA102 [GeForce RTX 3080]
    IDs: D=2206 V=10de SD=3885 SV=3842
    Driver: 546.01
OS: Microsoft Windows 11 Pro 10.0.22621 (win32)

The current system is not a Steam Deck
We are not running inside a Flatpak container

Software Versions:
  Heroic: 2.11.0 Kumachi
  Legendary: 0.20.33 Undue Alarm
  gogdl: 0.7.3
  Nile: 1.0.0 Jonathan Joestar

Game Settings: {
	"autoInstallDxvk": true,
	"autoInstallVkd3d": true,
	"preferSystemLibs": false,
	"enableEsync": true,
	"enableFsync": true,
	"nvidiaPrime": false,
	"enviromentOptions": [],
	"wrapperOptions": [],
	"showFps": false,
	"useGameMode": false,
	"language": ""
}

Game launched at: Sun Dec 03 2023 20:24:38 GMT-0800 (Pacific Standard Time)

Launch Command: HEROIC_APP_NAME=Fortnite HEROIC_APP_RUNNER=legendary HEROIC_APP_SOURCE=epic C:\Users\UserPC\AppData\Local\Programs\heroic\resources\app.asar.unpacked\build\bin\win32\legendary launch Fortnite --language en

Game Log:
An exception occurred when launching the game:
Error: EPERM: operation not permitted, open 'C:\Users\UserPC\AppData\Roaming\heroic\logs\2023-12-04T02_16_16-legendary.log'
    at Object.openSync (node:fs:603:3)
[...]

PR #3507, opened 2024-02-03
[Fix/Refactor] Make all logs async

arielj:
This PR is a continuation of https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3384.

In that PR I moved the `lastPlay` logs to be async, and this PR moves the rest of the logs to behave the same.

This has 2 main benefits:
- we write logs in batch every 1 second when there are enqueued message instead of writing once for every message (avoid writes when they happen too quickly which is an expensive operation)
- if, for any reason, writing to a log fails, the main action that is logging something is not impacted and can continue normally ( should fix https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3289 for example)

I included one fix: there was an error in the logs when stopping a sideloaded app

---


src/backend/logger/__tests__/logfile.test.ts
```diff
@@ -25,29 +25,33 @@ describeSkipOnWindows('logger/logfile.ts', () => {
...
-      [
-        expect.stringContaining(`Open invalid/${year}-`),
-        expect.objectContaining(
-          Error(`ENOENT: no such file or directory, open 'invalid/${year}-`)
-        )
-      ],
-      { prefix: 'Backend', skipLogToFile: true }
-    )
-  })
+  // FIXME: this test for some reason is correct in CI but `app.getPath` returning 'invalid'
```
flavioislima, line 28: Maybe run it if the ENV for CI is true.
Something like:
`if (process.env.GITHUB_ACTIONS) {`

flavioislima (approved):
LGTM

src/backend/logger/__tests__/logfile.test.ts
arielj: one thing I'm seeing now is that this is testing a case that will never happen

because from the getPath method, when it returns a string it's a valid path and when it's invalid it throws an error (that we are not catching), it should never return an invalid path https://www.electronjs.org/docs/latest/api/app#appgetpathname

we seem to be handling an invalid path incorrectly

I'd rather remove this test to avoid a false sense of correctness, we should actually do a try/catch in `createNewLogFileAndClearOldOnes` and return if the path is invalid

Issue #3249, opened 2023-11-21
Stuck at `getting download/install size` for Epic Games titles


a-cristi:
### Describe the bug

Trying to install games from Epic Games Store hangs indefinitely with a loading icon indicating that `Getting download size...` and `Getting install size...` are still in progress. A similar thing can be seen on a game's page, with `Download Size: Getting download size...` and `Install Size:Getting install size...`.

I waited around 10 minutes with no visible progress. Logging out and back in seems to not fix this issues. Games that are already installed seem to work fine.

### Add logs

```text
(09:50:34) INFO:    [Legendary]:        Legendary location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary

(09:50:34) INFO:    [Gog]:              GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl

(09:50:34) INFO:    [Connection]:       Connectivity: check-online

(09:50:34) INFO:    [Connection]:       Pinging external endpoints

(09:50:34) INFO:    [Backend]:          DRM module staus {

  "oimompecagnajdejgnnjijobebaeigek": {

    "name": "Widevine Content Decryption Module",

    "status": "new",

    "version": "4.10.2710.0"

  }

}

(09:50:34) WARNING: [Backend]:          Failed to register protocol with OS.

(09:50:34) INFO:    [Connection]:       Connectivity: online

(09:50:34) INFO:    [Gog]:              Checking if login is valid

(09:50:34) INFO:    [Gog]:              Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json auth

(09:50:35) INFO:    [Backend]:          AreWeAntiCheatYet data downloaded

(09:50:35) INFO:    [Backend]:          Loading Screen Ready

(09:50:35) INFO:    [Frontend]:         Refreshing undefined Library

(09:50:35) WARNING: [Backend]:          refresh not implemented on Sideload Library Manager

(09:50:35) INFO:    [Gog]:              Running command: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json auth

(09:50:35) INFO:    [Legendary]:        Refreshing library...

[...]

PR #3508, opened 2024-02-03
[Fix] Ensure manifest in install info

arielj:
When getting the install info for legendary games, it sometimes happens that we have invalid data cached that doesn't have `manifest` info.

In the past that lead to the InstallDialog to be stuck fetching the install info forever, with the latest changes for the gogdl_refactor that produces a javascript exception killing the react app: `Error: Cannot use 'in' operator to search for 'perLangSize' in null]`

This PR addresses that with 3 changes:
- if the cached information for legendary does NOT have a `manifest` property, it doesn't use the cache and fetches a new one
- if the result of fetching the info does not have a `manifest` property, it's retried up to 3 times
- if the result of fetching the info does not have a `manifest` property after the retries, it throws an error instead of storing the wrong data

I had to do some changes in the parameters to be able to support different extra params for the different stores.

I understand this fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3249

---

### DIFF
--- a/src/backend/downloadmanager/downloadqueue.ts
+++ b/src/backend/downloadmanager/downloadqueue.ts
@@ -126,12 +126,10 @@ async function addToQueue(element: DMQueueElement) {
   } else {
     const installInfo = await libraryManagerMap[
       element.params.runner
-    ].getInstallInfo(
-      element.params.appName,
-      element.params.platformToInstall,
-      element.params.branch,
-      element.params.build
-    )
+    ].getInstallInfo(element.params.appName, element.params.platformToInstall, {
+      branch: element.params.branch,
+      build: element.params.build
+    })
 
     element.params.size = installInfo?.manifest?.download_size
       ? getFileSize(installInfo?.manifest?.download_size)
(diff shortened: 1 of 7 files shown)


src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -126,12 +126,10 @@ async function addToQueue(element: DMQueueElement) {
...
-    ].getInstallInfo(
-      element.params.appName,
-      element.params.platformToInstall,
-      element.params.branch,
-      element.params.build
-    )
+    ].getInstallInfo(element.params.appName, element.params.platformToInstall, {
+      branch: element.params.branch,
+      build: element.params.build
+    })
```
arielj, line 132: `branch` and `build` are not in an optional `options` 3rd argument for gog

src/backend/storeManagers/sideload/library.ts
```diff
@@ -73,7 +68,7 @@ export function addNewApp({
 
 /* eslint-disable @typescript-eslint/no-unused-vars */
 
-export function installState(appName: string, state: boolean) {
+export function installState() {
```
arielj, line 71: we don't need the arguments in this functions because we don't use them

removing to avoid importing types we don't need

flavioislima (approved):
lgtm

Issue #3511, opened 2024-02-03
Heroic will automatically install the -LoL version if it is the latest released Wine-GE version.


TheRsKing:
### Describe the bug

Heroic will automatically install the -LoL version if it is the latest released Wine-GE version:

### Add logs

```text
(14:20:39) INFO:    [WineDownloader]:   wine versions updated
(14:20:39) INFO:    [DXVKInstaller]:    dxvk updated!
(14:20:39) INFO:    [Backend]:          Starting the Download Queue
(14:20:40) INFO:    [Backend]:          Downloaded: 21.63 MB / 213.45 MB  @17.43 MB/s (10.13%)
(14:20:41) INFO:    [Backend]:          Downloaded: 81.86 MB / 213.45 MB  @44.73 MB/s (38.35%)
(14:20:42) INFO:    [Backend]:          Downloaded: 145.96 MB / 213.45 MB  @66.1 MB/s (68.38%)
(14:20:43) INFO:    [Backend]:          Downloaded: 198.68 MB / 213.45 MB  @40.74 MB/s (93.08%)
(14:20:44) INFO:    [Backend]:          Downloaded: 212.64 MB / 213.45 MB  @10.84 MB/s (99.62%)
(14:20:45) INFO:    [DownloadManager]:  Downloader: Finished downloading https://github.com/GloriousEggroll/wine-ge-custom/releases/download/GE-Proton8-27-LoL/wine-lutris-GE-Proton8-27-LoL-x86_64.tar.xz
(14:20:45) INFO:    [Backend]:          Extracting /home/username/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/wine-lutris-GE-Proton8-27-LoL-x86_64.tar.xz to /home/username/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/wine/Wine-GE-Proton8-27-LoL using native tar
(14:20:54) INFO:    [WineDownloader]:   Finished installation of wine version Wine-GE-Proton8-27-LoL
(14:20:54) DEBUG:   [Backend]:          Unable to load Steam Libraries, libraryfolders.vdf not found
(14:20:54) INFO:    [Backend]:          Changing wine version to Wine - Wine-GE-Proton8-27-LoL
[...]

PR #3512, opened 2024-02-03
[Fix] Filter out -LoL build when installing wine at boot

arielj:
When Heroic is opened for the first time, if no wine is found it auto installs the latest available Wine-GE.

This was not filtering out the `LoL` builds.

You can test this behavior if you have other wines installed by going to main.ts and commenting out the `if` at:

```ts
  setTimeout(async () => {
    // Will download Wine if none was found
    const availableWine = await GlobalConfig.get().getAlternativeWine()
    DXVK.getLatest()
    Winetricks.download()
    if (!availableWine.length) {
      downloadDefaultWine()
    }
  }, 2500)
```

like:

```ts
  setTimeout(async () => {
    // Will download Wine if none was found
    DXVK.getLatest()
    Winetricks.download()
    downloadDefaultWine()
  }, 2500)
```

Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3511

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -858,7 +858,10 @@ export async function downloadDefaultWine() {
   // use Wine-GE type if on Linux and Wine-Crossover if on Mac
   const release = availableWine.filter((version) => {
     if (isLinux) {
-      return version.version.includes('Wine-GE-Proton')
+      return (
+        version.version.includes('Wine-GE-Proton') &&
+        !version.version.endsWith('-LoL')
+      )
     } else if (isMac) {
       return version.version.includes('Wine-Crossover')
     }


imLinguin (approved):
tested, works as expected

Nocccer:
Wierd i already added a filter for lol. If you fetch all available wine-ge the first one is always the latest none lol or sc build

arielj:
> Wierd. It means my filtering does not work here:
> 
> https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/f320fded71708119d67ab94acc7eef6866605e9c/src/backend/wine/manager/downloader/utilities.ts#L58-L60
> 
> Actually i would adapt this isntead adding a filter on top!
> 
> Update: Ok it still works. Wonder why it still downloads LoL and not the first wine-ge in the list of available wine-ge? image

this is unrelated to installing `wine-ge-latest`, it's a different code path

wine-ge-latest is indeed the last one that is not -LoL

PR #3515, opened 2024-02-04
[UX] Show error if using Wine-LoL for not League of Legends

arielj:
Because of this bug we had https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3512, a lot of users end up with Wine-GE-Proton-LoL installed and configured by default. Some users also do that by mistake.

This PR adds an error message to prevent users from using Wine-GE-LoL for games that are not League of Legends from Epic.

image

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -76,6 +76,10 @@
                 "message": "Something went wrong with the update, please check the logs or try again later!",
                 "title": "Update Error"
             },
+            "wine-lol-not-lol": {
+                "message": "Wine-GE-Proton-LoL is being used for a game that is not \"League of Legends\". Please install a Wine-GE-Proton version that does not end in -LoL using the Wine Manager section and configure it in the game's settings.",
+                "title": "Incompatible Wine Version"
+            },
             "winetricks": {
                 "message": "Winetricks returned the following error during execution:{{newLine}}{{error}}",
                 "title": "Winetricks error"
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -78,6 +78,8 @@ import { download, isInstalled } from './wine/runtimes/runtimes'
 import { storeMap } from 'common/utils'
 import { runWineCommandOnGame } from './storeManagers/legendary/games'
 
+const leagueOfLegendsAppId = '64b0c77d07f644e6a2326a1fd7ab9926'
+
 async function prepareLaunch(
   gameSettings: GameSettings,
   gameInfo: GameInfo,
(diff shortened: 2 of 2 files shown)


Etaash-mathamsetty:
this looks fine, but i think it's too extreme of a change tbh

arielj:
> this looks fine, but i think it's too extreme of a change tbh

why do you think it's extreme? it just prevents running games with the wrong wine and should avoid a lot of support requests

Etaash-mathamsetty:
> > this looks fine, but i think it's too extreme of a change tbh
> 
> why do you think it's extreme? it just prevents running games with the wrong wine and should avoid a lot of support requests

the support requests usually aren't caused by the LoL wine itself, but some other misconfiguration

imLinguin:
That doesn't cover sideload, LoL is a standalone game in general.

arielj:
> > > this looks fine, but i think it's too extreme of a change tbh
> > 
> > 
> > why do you think it's extreme? it just prevents running games with the wrong wine and should avoid a lot of support requests
> 
> the support requests usually aren't caused by the LoL wine itself, but some other misconfiguration. That's the main reason. The LoL patches may also fix a game in the future, and we shouldn't bar people who are simply testing

I can show a warning and let the user continue instead of just preventing it

I did try with a really simple game like 20 minutes til dawn and it doesn't even open with wine-ge-lol

arielj:
> That doesn't cover sideload, LoL is a standalone game in general.

I guess I can check the executable name? but I don't know how sideloading LoL works, do you add the exe file directly? what's the file name? or do you add the riot launcher? and in that case, is there something else that tells the launcher that you are trying to run LoL and not another game?

I'll change this to show a dialog where users can decide to launch the game or not instead of just blocking the launching

flavioislima (approved):
I am fine with the warning, but this might not be needed as soon as LOL stops working on linux so might be just wasted code.
But is up to you guys. Code looks good.
I think skipping sideloaded games is fine as well.

arielj:
@Etaash-mathamsetty I updated this to show a dialog with the option to continue anyway

@imLinguin I also included code to detect if the game is sideloaded (checking that the executable must be inside a `Riot Games/League of Legends` folder), I tested sideloading it and it works as expected (does not show the warning for LoL using -LoL and shows the warning for other games using -LoL)

arielj:
The main reason for this was because we installed the `-LoL` build by default for a lot of users because of the bug. But now that the latest version is not a `-LoL` and no new `-LoL` versions will happen I guess we can ignore this and just handle any support request in discord.

I'm going to close this one. The code will be here anyway if at any point we need this again (who knows, maybe LoL decides to remove the anticheat eventually? no idea)

Issue #3104, opened 2023-10-01
List components installed with Winetricks in logs


arielj:
### Problem description

When users ask for support, we usually tell them to do different steps like installing `vcrun20XX` or `d3dcompiler_47` or other winetricks stuff.

We have no way of knowing if they actually did the task.

### Feature description

Working on this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3102 I added a new function to get the list of things Winetricks installed in a prefix.

Would be useful for support to include that information in the logs.

### Alternatives

No idea, I guess not adding this information in the logs is the alternative.

### Additional information

Running winetricks is kinda slow in the computer I'm now so we should have this happen async.

Also, we have to make sure we don't try to do this for games that don't use wine, or calling the function will create/update a prefix.

PR #3516, opened 2024-02-04
[Logs] Log installed Winetricks packages when launching game

arielj:
Before the previous release I had to rever this change because calling winetricks before the wineprefix was ready was creating insues.

This PR fixes that by only logging Winetricks packages if the wine prefix was created successfully.

I added the `afterInit` async callback so it can be printed at the right position of the logs and still be async.

Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3104

---

### DIFF
--- a/src/backend/logger/logger.ts
+++ b/src/backend/logger/logger.ts
@@ -384,7 +384,7 @@ class LogWriter {
     if (this.initialized && !this.timeoutId) this.appendMessages()
   }
 
-  async initLog(afterInit: Promise<string> | null) {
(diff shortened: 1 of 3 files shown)


flavioislima (approved):
looks good

arielj:
I'm closing this, I did some more testing and it's till not working as expected.

I mixed the function that checks if the wine config is a valid wine with the function that sets up the prefix which happens later.

I'll work on this at another time, I'll close this for now.

PR #3518, opened 2024-02-04
[DX] Optimize ESLint includes

arielj:
The current `tsconfing.eslint.json` config is overriding the `include` property telling eslint to include all files and not only the files we care (I understand it's trying to lint too much).

Without this change I can't push any code, the `eslint` process starts consuming RAM like crazy until Node crashed with a memory limit error after it gets to 8GB of ram.

With these changes it has no issues and it goes from like a minute of loading then crash to processing all the files in 11 seconds without crashing.

I can see with the `--debug` flag that it's still parsing all the files from the project that we care about (all ts/tsx/js files inside src and e2e) so this seems to be safe. Any feedback is appreciated, I'm not an expert on these configurations.

---


CommandMC:
Hmm, I have a hard time reproducing this issue. I *did* get it once, but now can't trigger it again; deleted the cache of course

My original intent with the change (including all .js files in ESLint) was to lint "random" files like `downloadCount.js`. Now I'm however unsure if we even *want* that (not every one-off script needs perfect code, after all). Maybe removing the `"**/*.js"` from the original include array is the easier solution.

arielj:
> Hmm, I have a hard time reproducing this issue. I _did_ get it once, but now can't trigger it again; deleted the cache of course

For me it happens 100% of the times, with and without cache (it even freezes my whole system temporarily) :/

I've been changing that config locally to be able to push other branches

> My original intent with the change (including all .js files in ESLint) was to lint "random" files like `downloadCount.js`. Now I'm however unsure if we even _want_ that (not every one-off script needs perfect code, after all). Maybe removing the `"**/*.js"` from the original include array is the easier solution.

I guess we can add files manually if we want a file to be linted instead? to prevent this issue and still have linting (sometimes we might forget to add some file, but someone will notice eventually)

Etaash-mathamsetty:
I want to try this PR could you pls rebase?

arielj:
> I want to try this PR could you pls rebase?

I'll do it later, but you only need this change https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3518/files#diff-5053c0bda6e3ba2fcfa5bbe118ecf5c60fa9f099b78b3e6deb66e5679424f515R3

arielj:
I removed the updates of the eslint stuff, some packages were already outdated but updating them makes them incompatible with other eslint packages and I didn't want to waste time figuring that out since it's not relevant for this fix.

The main thing is the "includes" value to limit the number of files that are processed

CommandMC (approved):

PR #3519, opened 2024-02-04
[FIX] Power saver blocker disabling after library updates

Etaash-mathamsetty:
Title, I think this is reason why the power saver blocker doesn't work

### DIFF
--- a/src/backend/api/misc.ts
+++ b/src/backend/api/misc.ts
@@ -67,7 +67,7 @@ export const openWebviewPage = (url: string) =>
 export const setZoomFactor = (zoom: string) =>
   ipcRenderer.send('setZoomFactor', zoom)
 export const frontendReady = () => ipcRenderer.send('frontendReady')
-export const lock = () => ipcRenderer.send('lock')
+export const lock = (playing: boolean) => ipcRenderer.send('lock', playing)
 export const unlock = () => ipcRenderer.send('unlock')
 export const login = async (sid: string) => ipcRenderer.invoke('login', sid)
 export const logoutLegendary = async () => ipcRenderer.invoke('logoutLegendary')
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -553,7 +553,10 @@ ipcMain.on('lock', (e, playing: boolean) => {
     powerId = powerSaveBlocker.start('prevent-app-suspension')
   }
 
-  if (playing && (!displaySleepId || !powerSaveBlocker.isStarted(displaySleepId))) {
+  if (
+    playing &&
+    (!displaySleepId || !powerSaveBlocker.isStarted(displaySleepId))
+  ) {
     logInfo('Preventing display to sleep', LogPrefix.Backend)
     displaySleepId = powerSaveBlocker.start('prevent-display-sleep')
   }
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -28,9 +28,7 @@ import { cpus } from 'os'
 import {
   existsSync,
   rmSync,
-  unlinkSync,
   watch,
-  writeFileSync,
   readdirSync,
   readFileSync
 } from 'graceful-fs'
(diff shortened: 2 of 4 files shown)


arielj:
I'm curious how is the lock even working when playing a game, because we are playing with no other operation then `pendingOps` would be false and it would never lock the thing in the first place

I'm always bad with filtering and negative logic but I think we should remove the `game.status !== 'playing'` from the pendingOps filter maybe instead?

then `playing` would be considered as a pending op and it will not only not trigger the unlock but also trigger the lock

Etaash-mathamsetty:
> I'm curious how is the lock even working when playing a game, because if we are playing with no other operation then `pendingOps` would be false and it would never lock the thing in the first place
> 
> I'm always bad with filtering and negative logic but I think we should remove the `game.status !== 'playing'` from the pendingOps filter maybe instead?
> 
> then `playing` would be considered as a pending op and it will not only not trigger the unlock but also trigger the lock

yeah me as well, i think it almost instantly unlocks it again
your logic makes sense to me, I will modify this PR

flavioislima:
I think we broke the logic at some point a long time ago.

I believe that we used to have two logics:
For playing, we were blocking the screen to turn off and the machine to sleep as well.

For downloading, etc. We were blocking only the machine to sleep.

Looking at the code now in `main.ts`, I can see that we are only blocking the machine to suspend. So even if you are playing the screen will turn off and should not be like this.

The Electron docs says:

```
Note: prevent-display-sleep has higher precedence over prevent-app-suspension. Only the highest precedence type takes effect. In other words, prevent-display-sleep always takes precedence over prevent-app-suspension.

For example, an API calling A requests for prevent-app-suspension, and another calling B requests for prevent-display-sleep. prevent-display-sleep will be used until B stops its request. After that, prevent-app-suspension is used.
```

so your logic looks good @Etaash-mathamsetty , but a proper fix would need to include the display-sleep as well. 

So in the past we used to pass an argument to the 'lock' command. Could be either `play` or `download`, and then we treated the priority on the backend with the correct command.

At some point we removed it and kept only one, which won't work properly.

aalosb:
Any updates here? I NEEEEEEEED this. Ever since I installed heroic my windows machine will not put the display to sleep, nor will it go into suspended mode. 

Since I don't care to compile a build of heroic myself, I'm going back to the epic launcher (yuck) until this can get merged in.

arielj:
> Since I don't care to compile a build of heroic myself, I'm going back to the epic launcher (yuck) until this can get merged in.

you can download the portable binary from https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/actions/runs/8804826042?pr=3519 no need to compile heroic yourself

aalosb:
> (actually it would be good to have someone test this to check if it solves the issue, since it seems to be system and DE dependent

On Win 11 pro.

Looks like this actually sorta works.

I can launch the game and it acquires the screen lock 

```C:\Windows\System32>powercfg /requests
DISPLAY:
[PROCESS] \Device\HarddiskVolume3\Users\aalosb\AppData\Local\Temp\2fVm3vUXQMipsCIrZVhtZQBckLh\Heroic.exe
Electron

EXECUTION:
None.
```

and then I can exit the game, and the request is released and no long shows

```
C:\Windows\System32>powercfg /requests
DISPLAY:
None.

C:\Windows\System32>powercfg /requests
DISPLAY:
None.

EXECUTION:
[PROCESS] \Device\HarddiskVolume3\Users\aalosb\AppData\Local\Temp\2fVm3vUXQMipsCIrZVhtZQBckLh\Heroic.exe
Electron
```

Whereas using the current installer version 2.14.1.321 it will acquire the display power request, but never release it after closing out the game.



However this branch still does acquire an EXECUTION lock, so it still prevents the PC from going to sleep while heroic is running, but at least it now lets my screens turn off, so slight improvement.

Etaash-mathamsetty:
i don't see any bug in my code but it's possible that there's a bug, will let someone else look into it

flavioislima (approved):
Tested here and everything seems to be working fine

Issue #2001, opened 2022-11-06
Current game being installed sometimes gets reordered in the library


arielj:
### Describe the bug

At some point we added this feature that, when ordering but installation status, the order would be:
- installed games
- game being installed
- non-installed games

Currently, when installing a game, sometimes this custom ordering works and sometimes it doesn't.

### Add logs

Nothing relevant.

### Steps to reproduce

1. Start installing a game
2. It is NOT listed right after the currently installed games
3. Move around in heroic doing different things (maybe having more elements in the queue? something triggers a reorder)
4. Now the game being installed is right after the currently installed games

### Expected behavior

The ordering should be consistent, the currently installing game should be listed right after the installed games when ordering by install status.

### System Information

- OS [e. g. "Ubuntu"]: Linux Mint
- Heroic Version [e. g. 2.1.1]: 2.5.0-beta2 (but I don't think it started in this beta)

PR #3523, opened 2024-02-05
[UX] Update order in library when an install starts

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2001

We have code to sort games by: installed > installing > not installed, but when an installation starts it's currently not being sorted unless a re-render is triggered.

This PR fixes that.

To test this you can start installing a game and the card doesn't move to the top, then change to another screen, go back to the library and it's now sorted.

---

### DIFF
--- a/src/frontend/screens/Library/index.tsx
+++ b/src/frontend/screens/Library/index.tsx
@@ -17,13 +17,7 @@ import Fuse from 'fuse.js'
 import ContextProvider from 'frontend/state/ContextProvider'
 
 import GamesList from './components/GamesList'
-import {
-  FavouriteGame,
-  GameInfo,
-  GameStatus,
-  HiddenGame,
-  Runner
-} from 'common/types'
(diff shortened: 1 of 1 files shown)


src/frontend/screens/Library/index.tsx
```diff
@@ -242,15 +236,13 @@ export default React.memo(function Library(): JSX.Element {
...
-
-    setInstalling(newInstalling)
-  }, [libraryStatus])
+  const installing = useMemo(
+    () =>
+      libraryStatus
+        .filter((st) => st.status === 'installing')
+        .map((st) => st.appName),
+    [libraryStatus]
+  )
```
arielj, line 245: replaced useState with useMemo to make it simpler

also removed the type since it's already inferred anyway, removes one import

flavioislima (approved):
Tested here and works and looks good 👍🏽

PR #3527, opened 2024-02-05
[Fix] Only download Winetricks every 7 days

CommandMC:
Before, we'd re-download Winetricks on every Heroic startup. This isn't a good idea for many reasons

`utils/inet`'s `downloadFile` function already has the option to cache if the file it's pointing to is younger than X seconds, so we can just use it.
Ideally we'd consolidate the two `downloadFile`s into one, but they have a very different featureset & function, so that'd be a larger refactor

---

### DIFF
--- a/src/backend/tools/index.ts
+++ b/src/backend/tools/index.ts
@@ -50,6 +50,10 @@ import { lt as semverLt } from 'semver'
 import { createAbortController } from '../utils/aborthandler/aborthandler'
 import { gameManagerMap } from '../storeManagers'
 import { sendFrontendMessage } from '../main_window'
+import {
+  DAYS,
+  downloadFile as downloadFileInet
+} from '../utils/inet/downloader'
 
 interface Tool {
   name: string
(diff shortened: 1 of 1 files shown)


arielj:
I'd make it even longer than 24hs, I'd pick like 1 week minimum.

I don't think we need to be on the edge of winetricks. Most of the use cases we have are installing d3dcompiler/vcrun/etc

and if a user really needs to get it updated we can instruct them to clear the cache

arielj:
Like... if I understand this right, between the last 2 releases of winetricks it took like 10 months
image

CommandMC:
We're not looking at Winetricks *releases*, we're looking at the master version (precisely because releases are so far apart)

Bumping up the delay does sound sensible though. Are you sure clearing the cache also clears this file?

arielj:
> We're not looking at Winetricks _releases_, we're looking at the master version (precisely because releases are so far apart)

I see, at least from a quick look at the commits it looks like once a month is enough? I see the last commit is from 2 weeks ago for example

> Are you sure clearing the cache also clears this file?

Oh now I'm not sure, I was expecting the clear cache to clear this cache too haha... well... we can instruct the user to delete the file manually haha

Etaash-mathamsetty:
> I'd make it even longer than 24hs, I'd pick 1 week minimum, or even 1 month.
> 
> I don't think we need to be on the edge of winetricks. Most of the use cases we have are installing d3dcompiler/vcrun/etc
> 
> and if a user really needs to get it updated we can instruct them to clear the cache

agreed, 1 week seems good

CommandMC:
> I was expecting the clear cache to clear this cache too

Heroic as a whole doesn't really know about this file being cached, so that's currently not the case. Could of course make that a thing if this ends up being an issue, but Winetricks has also made some changes over the years so slightly-outdated versions should no longer be an issue

arielj (approved):
looks good to me

flavioislima (commented):
lgtm

Issue #3483, opened 2024-01-31
Excessive downloads of pci.ids


gollux:
### Describe the bug

I am the maintainer of the PCI ID database and I am getting an excessive number of HTTP requests (more than 30000 per day) for the `pci.ids` file from clients with User-Agent set to `HeroicGameLauncher`. This is becoming unsustainable, so I would like to kindly request your cooperation with reducing the load.

At the very minimum, you should download the compressed version of the file and cache it locally.

However, it seems that all you need is identifying the GPU of the machine where your software is running. For this purpose, the PCI ID database can be queried via DNS for a specific ID only (see `lib/names.c` in the [pciutils](https://github.com/pciutils/pciutils) project).

### Add logs

```text
Does not apply.
```

### Steps to reproduce

Does not apply.

### Expected behavior

Downloading files responsibly.

### Heroic Version

Latest Stable

### System Information

Linux.

CommandMC:
Hello. First off, I'd like to apologize for the mess here. Our intent *was* to cache the response, however it seems the expiry time conditions were not checked properly.

> However, it seems that all you need is identifying the GPU of the machine where your software is running. For this purpose, the PCI ID database can be queried via DNS for a specific ID only (see lib/names.c in the [pciutils](https://github.com/pciutils/pciutils) project).

I'm not *the best* at reading C, but I don't think NodeJS, this project's language, offers the same low-level DNS functions? I'll look into this once I have some more time though.  
We've discussed some alternatives, such as only re-downloading the file if the user's hardware changes, but that'd require more complex changes.

For now, I've created [this PR](<https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3484>) to mitigate the issue (properly caching the file, as mentioned above, and bumping up our cache timeout to 30 days to hopefully lessen the load further).

gollux:
Thanks for the quick reaction! A hot fix of caching issues is very welcome.

Let us think about what to do next. I am no NodeJS expert, but I think it should be possible to query DNS for TXT records using the [`dns.resolveTxt`](https://www.geeksforgeeks.org/node-js-dns-resolvetxt-method/) method.

Another possibility is to bundle a snapshot `pci.ids` with your releases and download a newer version only if the user explicitly asks for it, or perhaps if you fail to identify the device.

(This is based on my assumption that you are using the `pci.ids` to identify the GPU. Correct me if I am wrong, please.)

[...]

PR #3528, opened 2024-02-05
[Ref] Use DNS text queries to resolve PCI IDs

CommandMC:
As discussed in #3483

To test this:
- Open the current Heroic release and navigate to the System Information tab. Note down the name of your GPU.
- Open this version, navigate to the same tab. The same name as before should be displayed.

I've tested this on both real and virtual machines, and on both Windows and Linux. macOS does not need testing, as detecting GPUs in the first place isn't done there.

Internally, we now keep two `CacheStore`s for (1) vendor names and (2) device names. The vendor name cache is simply indexed by their ID, device names are indexed with either `<DeviceSubID>_<VendorSubId>_<DeviceId>_<VendorId>`, if there are sub-IDs, or just `<DeviceId>_<VendorId>` if there aren't. This key format is handy, since replacing the `_` with `.` nets us the DNS query prefix.

Both caches are valid for 7 days. I've toned this down from 30, but the bandwidth used should still be much less than before (since we're just requesting individual IDs).

CC @gollux

This requires #3527, as `downloadFile` isn't used anymore (so Unimported will fail)

---

### DIFF
--- a/src/backend/api/misc.ts
+++ b/src/backend/api/misc.ts
@@ -9,8 +9,8 @@ import {
 } from 'common/types'
 import { NileRegisterData } from 'common/types/nile'
 
-export const clearCache = (showDialog?: boolean) =>
-  ipcRenderer.send('clearCache', showDialog)
+export const clearCache = (showDialog?: boolean, fromVersionChange?: boolean) =>
+  ipcRenderer.send('clearCache', showDialog, fromVersionChange)
 export const resetHeroic = () => ipcRenderer.send('resetHeroic')
 
 export const openWeblate = () => ipcRenderer.send('openWeblate')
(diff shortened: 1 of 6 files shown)


arielj:
I'm not sure I understand the reason to clear the cache after 7 days. If the information that's cached already matches my GPU, why would we need to discard it and do a new request?

it's because the name of my device can change in that endpoint?

CommandMC:
Correct, device names can change (they don't do often, but they might)

7 days might indeed still be too frequent for *most* use cases, but my layman's guess would be that one of these queries per user every 7 days still is *not much at all* in terms of data transferred

gollux:
I am pretty fine with a couple of requests per 7 days per user.

flavioislima (approved):
lgtm

PR #3535, opened 2024-02-08
[Fix] Pass `--skip-sdl` when updating/repairing Epic games

CommandMC:
If we're updating/repairing a game with SDL that was either synced from EGS or imported (or is missing SDL tags for some other reason), the command will get stuck waiting for the user to choose these tags.
Passing `--skip-sdl` resolves this issue. It'll always set the "default" tags, which might not be the user's choice, but it's better than getting stuck like this.

Testing this yourself is probably not that easy. Could of course download a game with SDL, delete the tags from Legendary's config, and then try a repair in Heroic, but that's rather involved.
I've done precisely this, and this change indeed resolves the issue.

---

### DIFF
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -501,7 +501,8 @@ export async function update(
   const command: LegendaryCommand = {
     subcommand: 'update',
     appName: LegendaryAppName.parse(appName),
-    '-y': true
+    '-y': true,
+    '--skip-sdl': true
   }
   if (maxWorkers) command['--max-workers'] = PositiveInteger.parse(maxWorkers)
   if (downloadNoHttps) command['--no-https'] = true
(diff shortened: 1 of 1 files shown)


arielj:
Would we need any conditional to not pass this flag if the game is installed through Heroic? since you mention the flag is needed when it's synced or imported

code looks good to me, but that case was not clear for me

CommandMC:
Legendary already handles those cases for us. The logic works like this (feel free to check out [the code](https://github.com/derrod/legendary/blob/7fefdc497329276fc8c41d20e72600e0112352d1/legendary/cli.py#L932-L941) yourself):
- If the game doesn't support SDL, skip
  - If the game already has SDL tags configured, skip
    - If `--skip-sdl` is passed, select all required tags and save them
    - If it isn't, prompt the user for desired tags

arielj:
oh I see, the flag means `skip the sdl prompt`, not `skip installing sdl`

arielj (approved):

PR #3536, opened 2024-02-08
[UX] Add messages when there are no games to display in the library

arielj:
This PR adds 2 messages when we have no games to show:

- If the user has no games at all in the library, we show a message so they can either log in or sideload games
- If the user has at least one game in the library but nothing is displayed, we show a message to let the user know the filters are filtering all games

image
image

I'm using this `Trans` element that supports rendering components and HTML inside it, so the `log in` and `add game` elements are interactive in the message. At the same time it provides safety against translations including unwanted HTML since it can only show the HTML we allow (check the translation strings have `<1>` and `<3>`, i18next handle that internally, so translators have to use them for elements to render. Then i18next replaces those tags with the ones from the children.

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -222,6 +222,10 @@
         },
         "title": "Downloads"
     },
+    "emptyLibrary": {
+        "noGames": "Your library is empty. You can <1>log in</1> using a store or click <3></3> to add one manually.",
+        "noResults": "The current filters produced no results."
+    },
     "epic": {
         "offline-notification-body": "Online services may not work fully as Epic Games servers are offline!",
         "offline-notification-title": "offline"
--- a/src/frontend/screens/Library/components/EmptyLibrary/index.css
+++ b/src/frontend/screens/Library/components/EmptyLibrary/index.css
@@ -1,0 +1,14 @@
+.noResultsMessage {
+  font-size: 2rem;
+  max-width: 500px;
+  margin: auto;
+  padding-bottom: 20%;
+  .sideloadGameButton {
+    margin: 0;
+    display: inline;
+    vertical-align: text-bottom;
+  }
+  & a {
+    text-decoration: underline;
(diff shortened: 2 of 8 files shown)


src/frontend/screens/Library/components/LibraryHeader/index.tsx
```diff
@@ -38,12 +35,7 @@ export default React.memo(function LibraryHeader({
             ? t('favourites', 'Favourites')
             : t('title.allGames', 'All Games')}
           <span className="numberOfgames">{numberOfGames}</span>
-          <button
-            className="sideloadGameButton"
-            onClick={handleAddGameButtonClick}
-          >
-            {t('add_game', 'Add Game')}
-          </button>
+          <AddGameButton />
```
arielj, line 38: extracted this into a component to use it in the empty library message

src/frontend/screens/Library/components/EmptyLibrary/index.css
```diff
@@ -0,0 +1,14 @@
+.noResultsMessage {
+  font-size: 2rem;
+  max-width: 500px;
+  margin: auto;
+  padding-bottom: 20%;
```
arielj, line 5: this message looks odd vertically centered, this padding moves it a bit up

public/locales/en/translation.json
```diff
@@ -222,6 +222,10 @@
         },
         "title": "Downloads"
     },
+    "emptyLibrary": {
+        "noGames": "Your library is empty. You can <1>log in</1> using a store or click <3></3> to add one manually.",
```
arielj, line 226: these `<1>` and `<3>` are intentional, the `Trans` component replaces them with the tags from the children

flavioislima (approved):
This is a pretty good UX improvement that we have been missing for a long time. Should be on the next release and will help a lot of new users.

PR #3540, opened 2024-02-08
[Ref/Feat] Clean up our routing and make it faster

CommandMC:
Our Routes were a bit of an organically grown mess, with different approaches to do the same things, and generally using outdated React Router features. This PR cleans those up, and improves Heroic startup time in the process.

I did remove one route when "converting" from Route elements to objects: `/last-url`, I couldn't find any components navigating to it, so it seems like this was always unreachable?
Other than that, I think we should look into making our WebView component less reliant on specific `location.pathname` values, instead just *telling it* what it's supposed to do (I've opted to choose params over props there, but I don't have any preference really, just seemed more convenient). Didn't want to touch too many things before a release though.

---

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -86,16 +86,8 @@ const router = createHashRouter([
         Component: Login
       },
       {
-        path: 'epicstore',
-        element: <WebView store="epic" />
-      },
-      {
-        path: 'gogstore',
-        element: <WebView store="gog" />
-      },
-      {
-        path: 'amazonstore',
-        element: <WebView store="amazon" />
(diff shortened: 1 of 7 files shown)


arielj:
> I did remove one route when "converting" from Route elements to objects: `/last-url`, I couldn't find any components navigating to it, so it seems like this was always unreachable?

this is probably a leftover from when I implemented the feature to store the last url for the stores that then needed some tweaks

I think it's not used anymore. It's easy to check though, if the feature to ropen the stores in the last url works as expected it's all good

arielj (approved):
looks good to me

PR #3541, opened 2024-02-09
Tech: gogdl update 

imLinguin:
Updated gogdl and added a check if gog redist directory exists - in order to make sure ISI is always installed

### DIFF
--- a/src/backend/downloadmanager/downloadqueue.ts
+++ b/src/backend/downloadmanager/downloadqueue.ts
@@ -143,7 +143,7 @@ async function addToQueue(element: DMQueueElement) {
       installInfo &&
       'dependencies' in installInfo.manifest
     ) {
-      const newDependencies = installInfo.manifest.dependencies
+      const newDependencies = installInfo.manifest.dependencies || []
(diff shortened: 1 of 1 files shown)


Etaash-mathamsetty (approved):
lgtm

src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -142,7 +144,7 @@ async function addToQueue(element: DMQueueElement) {
       'dependencies' in installInfo.manifest
     ) {
       const newDependencies = installInfo.manifest.dependencies
-      if (newDependencies?.length) {
+      if (newDependencies?.length || !existsSync(gogRedistPath)) {
```
Etaash-mathamsetty, line 147: Looking over this again shouldn't this be an and?

imLinguin: we are about to create that path, it's a check to make sure it's not removed and we always have ISI downloaded

imLinguin: I might need to add a fallback to newDependencies in case it's undefined

arielj (approved):

PR #3544, opened 2024-02-10
[Refactor] Sidebar cleanup (and a bit more that's related)

arielj:
This PR is a refactor to cleanup the code related to the Sidebar component.

I does a few things:
- Extracted a new `SidebarItem` component, so we don't need to use `<button>` and `NavLink` directly in SidebarLinks (this helps remove a lot of duplication around creating those items)
- Replaced the .scss files with .css, we are not really using any sass feature
- Refactored the css files to use native css nesting and moved the styles to the corresponding components' css (we had a lot of css in the main `sidebar.css` file that is easier to find in each component's css, we also had css for `Sidebar__item` in multiple css files)
- I added new CSS properties specific to the sidebar that can help when theming, they fallback to what we already had so nothing changes, but new themes can use better variables that don't cause unexpected side effect
- Because of this cleanup, I had to remove the `state: { fromGameCard: ...}` that was not really used anymore since the settings for a game were moved to the SettingsModal component. This led into some changes needed also in the `useSettingsContext` context and the `Settings` screen to remove the code that was using that

**How to test/qa this**

The sidebar should behave the same way as before (I didn't really change any CSS, I just moved things to different places and added nesting, and the new custom properties have fallbacks to the previous values):
- Resizing
- Hiding content during resize
- sizes, colors, etc depending on the theme
- etc

---


src/frontend/components/UI/Sidebar/components/CurrentDownload/index.css
arielj, line 1: Most of this file is the previous `index.scss` file renamed and adding the `.Sidebar {` nesting, and some new code extracted from the `Sidebar/index.css` file that fits better here

src/frontend/components/UI/Sidebar/components/CurrentDownload/index.tsx
```diff
@@ -52,37 +52,35 @@ export default React.memo(function CurrentDownload({ appName, runner }: Props) {
...
-      </Link>
-    </>
+          <Box sx={{ minWidth: 35 }}>
+            <Typography variant="body2">{`${Math.round(
+              progress.percent || 0
+            )}%`}</Typography>
+          </Box>
+        </Box>
+      </div>
+    </Link>
```
arielj, line 84: This is just me removing the extra `<>` fragment that was not needed, the diff is just the indentation

src/frontend/components/UI/Sidebar/components/HeroicVersion/index.css
arielj, line 1: The code here is just CSS extracted from `Sidebar/index.css`, and refactored to use nesting

src/frontend/components/UI/Sidebar/components/SidebarItem/index.css
```diff
@@ -0,0 +1,111 @@
...
+    --sidebar-item-inactive-text-color: var(
+      --navbar-inactive,
+      var(--navbar-accent)
+    );
+    --sidebar-hover-text-color: var(--text-hover);
+    --sidebar-item-active-text-color: var(
+      --navbar-active,
+      var(--accent-overlay, var(--accent))
+    );
+    --sidebar-active-item-background: var(--navbar-active-background);
```
arielj, line 12: this file is mostly extraction of `Sidebar__item` styles that were in other places, but I also added these new custom properties to help themming

src/frontend/components/UI/Sidebar/index.css
```diff
@@ -0,0 +1,140 @@
...
+  .SidebarLinks {
+    grid-area: links;
+  }
+
+  .heroicVersion {
+    grid-area: version;
+    margin-block: 0 var(--sidebar-vertical-padding);
+    margin-inline: var(--sidebar-horizontal-padding) var(--space-3xs);
+    -webkit-app-region: no-drag;
+  }
```
arielj, line 40: I'm still keeping these types of properties here in the parent and not in the component's css because these styles should be controlled by the parent since they are styles "outside" the component

src/frontend/components/UI/Winetricks/index.tsx
```diff
@@ -21,7 +21,7 @@ export default function Winetricks({ onClose }: Props) {
     setLoading(true)
     try {
       const components = await window.api.winetricksListInstalled(
-        runner,
+        runner!,
```
arielj, line 24: This is not needed because `runner` can be undefined in the Settings Context, but cannot be undefined in this case (we would never show the winetricks option anywhere without a game, so runner would be defined

src/frontend/screens/Settings/index.tsx
```diff
@@ -127,12 +104,7 @@ function Settings() {
       <SettingsContext.Provider value={contextValues}>
         <div className={`Settings ${type}`}>
           <div role="list" className="settingsWrapper">
-            <NavLink
-              to={returnPath}
-              role="link"
-              className="backButton"
-              state={{ gameInfo: gameInfo }}
-            >
+            <NavLink to="/" role="link" className="backButton">
```
arielj, line 107: Since we moved the settings for specific games into a modal, this was always `/` anyway, we don't need all the code to detect to go back to a game page because that doesn't happen anymore

src/frontend/components/UI/Winetricks/index.tsx
CommandMC: It'd be more robust if you'd pass the runner via this component's props (and then only take a `Runner`, not a `Runner | undefined`). You'd then not need these assertions

Assertions like this are dangerous, since if we ever change the conditions for showing this component (to no longer look at the context's runner), this will silently break

src/frontend/screens/Settings/index.tsx
CommandMC: This should go to `/library` for now (until #3540 is merged). Navigating to `/` just navigates to `/library` with an extra re-render

src/frontend/components/UI/Winetricks/index.tsx
arielj: I'll updated the `if (....)` that we have above in this component to not render anything if we don't have a runner (we already render `<></>` when `isDefault` which implies `runner` is undefined, but making that explicit will get rid of these `!`)

src/frontend/components/UI/Sidebar/components/SidebarLinks/index.tsx
```diff
@@ -101,288 +95,159 @@ export default function SidebarLinks() {
...
-              >
-                <span>
-                  {t(
-                    'settings.navbar.games_settings_defaults',
-                    'Game Defaults'
-                  )}
-                </span>
-              </NavLink>
+              <SidebarItem
+                url="/settings/app/default/general"
```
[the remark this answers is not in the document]
arielj: yep, wrong link, I'll actually do a bit more refactoring, we don't really need the `app/default` part of these urls anymore

src/backend/main.ts
```diff
@@ -1676,7 +1676,7 @@ ipcMain.on('processShortcut', async (e, combination: string) => {
       break
     // hotkey to open the settings on frontend
     case 'ctrl+k':
-      sendFrontendMessage('openScreen', '/settings/app/default/general')
+      sendFrontendMessage('openScreen', '/settings/general')
```
arielj, line 1679: we don't need these `app/default` routes anymore, the Settings component always renders Heroic's settings, game's settings are not handled by this url

src/frontend/screens/Settings/components/Tools/index.tsx
```diff
@@ -19,12 +19,12 @@ export default function Tools() {
...
   const isWindows = platform === 'win32'
 
-  if (isDefault || isWindows) {
+  if (isDefault || isWindows || !runner) {
     return <></>
   }
 
   type Tool = 'winecfg' | string
-  async function callTools(tool: Tool, exe?: string) {
+  const callTools = async (tool: Tool, exe?: string) => {
```
arielj, line 27: I had to change these to arrow functions because the typescript complains about `runner` being possible undefined inside `function()` (even though this happens after the `if`)

looks like, because `this` is different in the 2 cases, runner can be undefined for named functions but it's correct with arrow functions

src/frontend/screens/Settings/index.tsx
```diff
@@ -68,41 +63,24 @@ function Settings() {
...
   const contextValues: SettingsContextType | null = useSettingsContext({
-    appName,
-    gameInfo,
-    runner
+    appName
   })
 
   // render `loading` while we fetch the settings
-  if (!title || !contextValues) {
+  if (!currentConfig || !contextValues) {
```
arielj, line 78: we were checking `title` here, but title doesn't really change, it's better to check for `currentConfig` since it's what the comment is saying originally and that's the real intention

flavioislima (approved):
LGTM

PR #3547, opened 2024-02-10
[UX] Simplify install dialog hiding wine settings by default. Warn about shared wine directory

arielj:
It's really common to see support requests of users that clicked the `Use Default Wine Settings` without understanding what it is for.

This PR aims to solve that with 2 things:
- the wine settings are hidden behind a checkbox (most users don't really need to touch those fields ever, the install dialog is simpler)
- then, if they check `Use Default Wine Settings` a warning is displayed to tell them to not use this option if they don't know what they are doing

Also, now when checking the `Use Default Wine Settings` option, the wine and prefix field are not hidden so users can see what that check box is doing (the fields are disabled so they don't change them).

image

image

image

---

### DIFF
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -200,7 +200,7 @@
     "setting": {
         "show-wine-settings": "Show Wine settings",
         "use-default-wine-settings": "Use Default Wine Settings",
-        "warn-use-default-wine-settings": "Only use this option if you know what you are doing.<br/>Sharing the same prefix for multiple games can create problems<br/>if their dependencies are incompatible.",
+        "warn-use-default-wine-settings": "Only use this option if you know what you are doing.<br/>Sharing the same prefix for multiple games can create problems if their dependencies are incompatible.",
         "winecrossoverbottle": "CrossOver Bottle"
     },
     "sideload": {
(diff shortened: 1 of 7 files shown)


imLinguin:
I think having a button instead of the checkbox makes more sense here. We are hiding that element when it's checked, so it doesn't need to be a checkbox

arielj:
> I think having a button instead of the checkbox makes more sense here. We are hiding that element when it's checked, so it doesn't need to be a checkbox

right, I originally added it as a checkbox cause I was not hiding it, but then there were 2 checkboxes together and it looked weird

arielj:
looks kinda weird with a button, I'm trying all the styles for buttons we have and it looks off, not sure exactly why

image

flavioislima:
Can you make this button smaller or maybe use it like a link?
I think it is too big and calls too much attention on the dialog. Especially when installing games with a lot of DLCs, languages, etc.

Another idea would be to use a collapsible element, which may make more sense.

arielj:
> Can you make this button smaller or maybe use it like a link? I think it is too big and calls too much attention on the dialog. Especially when installing games with a lot of DLCs, languages, etc.
> 
> Another idea would be to use a collapsible element, which may make more sense.

I tried making it smaller and it also looks odd (I didn't push any of that, the code in this PR is still the checkbox).

I can try with a details/summary element but I don't know if we have a style for that anywhere (I used it for the help component but it just the basic browser style, doesn't look great)

arielj:
what about this with a details/summary element?

image
image

I think it looks better

flavioislima (approved):

PR #3550, opened 2024-02-11
[FIX] Apply top padding only to external link warning dialog

arielj:
There was a regression with this change https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/commit/bda0871bb1e2d324091154549f2842be864cfe14#diff-d9b11fef9fc96a7cf2707aa13652ab1309131d88dcdba70f00cd97fcccc25830R16 causing all dialogs to have extra top padding.

Causing this visual glitch:
image

This PR fixes that by applying the padding only to the external link warning dialog.

I also changed the hardcoded `20px` to the css property that's used at the bottom so it's more theme-friendly.

(we should re-write this component in the future, it's using margin properties as padding)

---

### DIFF
--- a/src/frontend/components/UI/Dialog/index.css
+++ b/src/frontend/components/UI/Dialog/index.css
@@ -13,7 +13,6 @@
   display: flex;
   flex-direction: column;
   padding: 0;
-  padding-top: 20px;
   overflow: auto;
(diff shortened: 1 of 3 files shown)


imLinguin (approved):

flavioislima (approved):
lgtm

PR #3562, opened 2024-02-12
[Fix] Sideloaded games on Windows

CommandMC:
With my changes to callRunner (to use PowerShell on Windows), sideloaded games actually broke on Windows (for two different reasons!)
This fixes both of those problems (details as always in commit messages)

To test:
Well, try to run sideloaded games on Windows. Won't work on 2.13.0, will work with this PR.

---


flavioislima (approved):
LGTM

Issue #2507, opened 2023-02-28
Best way to run something before and after e.g. changing keyboard layout


alexheretic:
### Problem description

I often want to set my keyboard layout to qwerty for games, rather than my default of colemak.

In steam I can do this with launch settings `setxkbmap gb; %command%; setxkbmap gb colemak`.

For some games I've used a similar way to set mouse sensitivity before & after the game launch.

I can't see an equivalent way to do so in heroic. Is there a good way to achieve this?

### Feature description

Configure commands to run before & after the game.

Anuskuss:
Until this is done properly, you can use this as your wrapper script:

```bash
#!/bin/bash

setxkbmap gb
"$@"
setxkbmap colemak
```

PR #3565, opened 2024-02-13
[Feature] Configure scripts to run before and after a game is launched

arielj:
This PR closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2507. This can be used to change keyboard layout, screen resolution, or trigger any script the user wants to.

This PR adds 2 configuration options for games to set a script that would run before the game is launched, and one that would run after the game closes:

image

The code will wait for the script to exit before continuing with the launching of the game, so if something should run in the background in parallel to the game it must be handled by the script itself. For example, in bash:

```
#!/bin/bash

echo "this runs before the game" # this is sync
bash ./another-script.bash & # note the `&` at the end, this will run in parallel and finish when it needs to
echo "this also runs before the game" # this is sync too
```

With a file like this, Heroic will process the first and last line and proceed to launch the game. It will also process the second line but it won't wait for it to finish, so make sure the synchronous code exits and doesn't leave Heroic waiting forever.

Note that scripts are run in the context of the install directory of the game (working directory is set to the install_path of the game).

I only tested this on linux, so any feedback is welcomed, specially on windows.

I imagine the must will have to be accessible in the Flatpak's permissions if using Heroic flatpak.

---


src/backend/main.ts
```diff
@@ -1168,6 +1176,75 @@ ipcMain.handle(
...
+        ...execOptions
+      })
+
+      child.stdout.on('data', (data) => {
+        logInfo(data.toString(), LogPrefix.Backend)
+      })
+
+      child.stderr.on('data', (data) => {
+        logInfo(data.toString(), LogPrefix.Backend)
+      })
```
CommandMC, line 1210: Shouldn't these go into the game log?

src/backend/main.ts
```diff
@@ -1168,6 +1176,75 @@ ipcMain.handle(
...
+        reject(err.message)
+      })
+    })
+  } else {
+    return Promise.resolve(true)
+  }
+}
+
+function runAfterLaunchScript(gameInfo: GameInfo, gameSettings: GameSettings) {
+  if (gameSettings.afterLaunchScriptPath) {
```
CommandMC, line 1231: Same early return note here
```ts
  if (!gameSettings.afterLaunchScriptPath)
    return
  // Remaining code
```

src/backend/main.ts
```diff
@@ -1168,6 +1176,75 @@ ipcMain.handle(
...
+    logInfo(
+      [
+        'Running script after',
+        gameInfo.title,
+        `(${gameSettings.afterLaunchScriptPath})`
+      ],
+      LogPrefix.Backend
+    )
+    const output = execSync(gameSettings.afterLaunchScriptPath, {
```
CommandMC, line 1240: Using `execSync` here isn't *that* great since it blocks the Backend thread. Ideally use `exec` with a callback like
```ts
(error, stdout) => {
  if (error) logError(error, LogPrefix.Backend)
  logInfo(stdout, LogPrefix.Backend)
}

File: src/backend/main.ts [hunk 4]
```diff
@@ -1168,6 +1176,75 @@ ipcMain.handle(
...
+      child.on('exit', () => {
+        resolve(true)
+      })
+
+      child.on('error', (err: Error) => {
+        logError(
+          ['Error running before script: ', err.message],
+          LogPrefix.Backend
+        )
+        if (err.stack) logError(err.stack, LogPrefix.Backend)
```
CommandMC, line 1221: You can just pass the `err` directly to `logError`, it already outputs either its `stack` or `message` depending on which is present (and the `stack` includes the `message`)

src/common/types.ts
```diff
@@ -193,6 +193,8 @@ export interface GameSettings {
   wrapperOptions: WrapperVariable[]
   savesPath: string
   gogSaves?: GOGCloudSavesLocation[]
+  beforeLaunchScriptPath: string
+  afterLaunchScriptPath: string
```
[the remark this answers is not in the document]
arielj: where would it be the right place to add a default? I'm doing `const [scriptPath, setScriptPath] = useSetting('afterLaunchScriptPath', '')` but I imagine you mean somewhere else (I know the settings is kinda messy and you did a lot of refactoring for that other PR)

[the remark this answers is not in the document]
CommandMC: Well, a `useSetting` in the Frontend will also only affect the Frontend (if the Backend is "first" in looking at the settings object, for it the property will just be undefined)

You'd have to add these properties to the object returned by `GlobalConfigV0`'s `getFactoryDefaults` method here:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/2dc65ec96e7787133f03a58a60b03a4dc055630c/src/backend/config.ts#L318-L319
then destructure them here for the GameConfig:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/2dc65ec96e7787133f03a58a60b03a4dc055630c/src/backend/game_config.ts#L233-L234
and finally re-pack them right after:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/2dc65ec96e7787133f03a58a60b03a4dc055630c/src/backend/game_config.ts#L261-L262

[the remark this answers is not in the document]
arielj: thanks! done

src/backend/launcher.ts
```diff
@@ -1281,6 +1283,89 @@ async function getWinePath({
...
+  gameInfo: GameInfo,
+  gameSettings: GameSettings
+) {
+  if (!gameSettings.beforeLaunchScriptPath) {
+    return true
+  }
+
+  appendGamePlayLog(
+    gameInfo,
+    `Running script before ${gameInfo.title} (${gameSettings.afterLaunchScriptPath})\n`
```
CommandMC, line 1296:
```suggestion
    `Running script before ${gameInfo.title} (${gameSettings.beforeLaunchScriptPath})\n`
```

src/backend/launcher.ts
```diff
@@ -1281,6 +1283,89 @@ async function getWinePath({
...
+ *   use the after script to kill any running process if that's the case
+ */
+async function runScriptForGame(
+  gameInfo: GameInfo,
+  scriptPath: string
+): Promise<boolean | string> {
+  return new Promise((resolve, reject) => {
+    const child = spawn(scriptPath, {
+      cwd: gameInfo.install.install_path,
+      ...execOptions
```
CommandMC, line 1346: We don't want a shell here, this breaks if the path has spaces.

The one other usage of `execOptions`, running `which` for Winetricks, also shouldn't use a shell, so we should probably just remove this completely

CommandMC (changes requested):

src/backend/launcher.ts
arielj: thanks, I removed this and applied the other suggestion

CommandMC (approved):

flavioislima:
About the Settings itself, do you think would be better to do like this:
- Select Alt EXE
- Game Arguments
- New section called Scripts (similar to the wrapper command subtitle
- Select script to run before
- Select script to run after

I think would be more intuitive then having script before, exe, script after and then game arguments. Even though they are related, they do not serve the same purpose:
image

arielj:
@flavioislima just pushed a change re-ordering them

Issue #3415, opened 2024-01-07
Allow to rename category


Birdie0:
### Problem description

Category can't be renamed at the moment and if in case I would like to improve title I'll have to create another one, duplicate each title to it and after remove old category.

### Feature description

Allow to rename category with right click or on category window.

### Alternatives

Renaming categories in `config.json` file.

Issue #3406, opened 2024-01-07
Add the ability to add new categories from the "Categories" dropdown


arielj:
### Problem description

Currently, users can see the `Categories` element at the top right side of the library but when there are no categories we show a message that they can add categories from the game's menu.

This is not great because the user is already there in the dropdown and now they have to find which menu we are talking about.

### Feature description

In the Categories dropdown, we have an action to add new categories (could be a button that opens something, or just a field with a submit button, anything that looks good)

### Alternatives

Not sure, current behavior is an alternative but it's not great UX

### Additional information

If we add something like a modal to manage categories (to add/delete, maybe edit? them), we should probably then remove the option to remove categories when editing a game's categories to avoid users deleting them by mistake (they may think the trash can icon only remove the game from the category when it actually deletes the category altogether).

PR #3573, opened 2024-02-15
[Feature] Add option to add/edit/delete categories from categories dropdown

arielj:
Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3415
Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3406

This PR adds a new dialog to add/edit/delete categories without having to right click a game card.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/188464/93538965-5b6a-4f1f-9d4e-ed38e45a0c44

Some notes:
- I ordered the categories alphabetically for now, at some point we can let users reorder them but currently it the order changes when editing which is not good
- The video was recorded before the last changes so some issue that can be seen in the videos are fixed

---


flavioislima:
Codewise and feature-wise looks good.
But this design I think doesn't work:
image


I think both buttons should be at the same place: bottom.
try something like:
Select ALL
Manage Categories.

I think that would look better.

arielj:
> Codewise and feature-wise looks good. But this design I think doesn't work: image
> 
> I think both buttons should be at the same place: bottom. try something like: Select ALL Manage Categories.
> 
> I think that would look better.

just pushed this change

flavioislima:
Tested here and looks good, will merge to add to the next release.

One minor improvement for the future though: When clicking on Select All it also selects the 'Uncategorized' this has the same effect as selecting none.

flavioislima (approved):
LGTM

PR #3575, opened 2024-02-15
[FIX] Adtraction fallback

imLinguin:
When user has blocked the adtraction we should fallback to load gog url directly  
I also changed types to `Electron.WebviewTag` in places where webview was used, to avoid types issues in the future.

### DIFF
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -11,7 +11,7 @@ import { useNavigate, useLocation, useParams } from 'react-router'
 import { UpdateComponent } from 'frontend/components/UI'
 import WebviewControls from 'frontend/components/UI/WebviewControls'
 import ContextProvider from 'frontend/state/ContextProvider'
-import { Runner, WebviewType } from 'common/types'
+import { Runner } from 'common/types'
 import './index.css'
 import LoginWarning from '../Login/components/LoginWarning'
 import { NileLoginData } from 'common/types/nile'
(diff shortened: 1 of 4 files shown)


src/frontend/screens/WebView/index.tsx
```diff
@@ -217,8 +217,24 @@ export default function WebView({ store }: Props) {
         }
       }
 
-      webview.addEventListener('dom-ready', loadstop)
+      const onerror = (error: Electron.DidFailLoadEvent) => {
+        const { errorCode, validatedURL } = error
+        // Connection refused
+        if (
+          errorCode === -102 &&
```
arielj, line 224: maybe we can ignore this check? in case it fails to load adtraction for any reason, not just connection refused

imLinguin: huh, maybe you are right. As long as the url matches we should be good to go

src/frontend/screens/WebView/index.tsx
```diff
@@ -217,8 +217,19 @@ export default function WebView({ store }: Props) {
...
 
-      webview.addEventListener('dom-ready', loadstop)
+      const onerror = ({ validatedURL }: Electron.DidFailLoadEvent) => {
+        // Connection refused
+        if (validatedURL && validatedURL.match(/track\.adtraction\.com/)) {
+          const parsedUrl = new URL(validatedURL)
+          const redirectUrl = parsedUrl.searchParams.get('url')
+          if (redirectUrl) {
+            webview.loadURL(redirectUrl)
+          }
```
arielj, line 227: should we add something like:

```
else {
  webview.loadURL("gog.com url here")
}
```

I understand there's a chance that `redirectUrl` can be empty/undefined and it's probably better to fallback to something that works instead of staying in the the adtraction page that failed

(I think in practice that shouldn't happen? but since the if is there I think we should handle the else for completeness)

arielj (approved):
looks good to me!

flavioislima (approved):
Nice!

PR #3591, opened 2024-02-24
[Fix] Print sideload game install dir in logs

arielj:
This PR fixes an issue with sideloaded games printing `Installed in: undefined` instead of the real app folder.

---

### DIFF
--- a/src/backend/logger/logger.ts
+++ b/src/backend/logger/logger.ts
@@ -450,11 +450,17 @@ class GameLogWriter extends LogWriter {
     // init log file and then append message if any
     try {
       // log game title and install directory
+
+      const installPath =
+        this.gameInfo.runner === 'sideload'
+          ? this.gameInfo.folder_name
+          : this.gameInfo.install.install_path
+
       await writeFile(
         this.filePath,
         `Launching "${this.gameInfo.title}" (${runner})\n` +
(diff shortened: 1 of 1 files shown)


Etaash-mathamsetty (approved):
lgtm

CommandMC (changes requested):
I would be more in favor of making sideloaded games' GameInfo consistent with the others (reporting the install path in `install.install_path`, instead of having to look at a different property to get that info)

flavioislima (approved):
lgtm

arielj:
> I would be more in favor of making sideloaded games' GameInfo consistent with the others (reporting the install path in `install.install_path`, instead of having to look at a different property to get that info)

I agree with that, but I'd rather have that in another PR since it's a more complicated change (it requires fixing the game info for games that are already installed and updating other parts of the code to use `install_path` instead). This PR is just meant to fix the info in the logs, refactoring the sideload games sounds way more complicated just to get this thing working.

CommandMC:
It is definitely more work, and I get why you'd not want to do it (in this PR, and in general). I just worry that "patch-work" fixes like this will make the code harder to read & work with. I mean, the fact that this mistake was made in the first place already proves this.

I've said "Oh well, fine for now, we'll fix it up later" too many times, but I guess this'll be number <too many times + 1>

PR #3593, opened 2024-02-25
[Log] Print installed winetricks packages in logs

arielj:
Ok, third time is the charm. I tried to implement this 2 times before and both times it was broken, but now it's fine.

This fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3104

---


src/backend/logger/logger.ts
```diff
@@ -361,43 +362,76 @@ const logsWriters: Record<string, LogWriter> = {}
...
+
+          // support a promise returning a string or an array of strings
+          if (typeof result === 'string') {
+            result = [result]
+          }
+          messagesToWrite = messagesToWrite.concat(result)
+        } catch (error) {
+          logError(error, LogPrefix.Backend)
+        }
+      }
```
CommandMC, line 431: Assuming we keep this code (see my other comment), this can still be simplified:
```suggestion
      try {
        const result = await item
        messagesToWrite = Array.isArray(result) ? result : [result]
      } catch (error) {
        logError(error, LogPrefix.Backend)
      }
```

src/backend/logger/logger.ts
```diff
@@ -504,6 +538,28 @@ export async function initGamePlayLog(gameInfo: GameInfo) {
...
+              ? installedPackages.join(', ')
+              : 'none'
+
+            resolve(`Winetricks packages: ${packagesString}\n\n`)
+          })
+          .catch((error) => {
+            reject(error)
+          })
+      })
+    )
```
CommandMC, line 559: This feels strange to me. Why not await the promise in here and then just log the packages, avoiding all the "You can put a Promise<string | string[]> into the queue" stuff
```suggestion
  const installedPackages = await Winetricks.listInstalled(gameInfo.runner, gameInfo.app_name)
  const packagesString = installedPackages
    ? installedPackages.join(', ')
    : 'none'
  logWriter.logMessage(`Winetricks packages: ${packagesString}\n\n`)
```

arielj: The problem of waiting for the promise here is that it takes some time to wait and during that time other messages are appended to the log (the game starts and sends output), so this ends up mixed with the game output instead of appended to the logs before the game launches

with this approach, the promise gets enqueued before the game starts, so the winetricks info is consistently logged before the game output cause it's before those in the messages queue

CommandMC: I see, hadn't considered that. This is quite clever!

src/backend/logger/logger.ts
arielj: I pushed this refactor

I do have to keep the contact part cause I'm processing multiple elements of the queue (and each element can be either a string or an array of strings)

CommandMC: Oh right, of course

CommandMC (approved):

PR #3613, opened 2024-03-03
[UX] WIP: Improve gamepage

Kajot-dev:
This PR aims to improve UI/UX of the redesigned gamepage view

## Main fixes:
- Moved "report issues" link a bit so both panels will always be the same height
- Moved `TabPanel` into separate component as it was duplicated and gamepage used the same logic for the 3rd time
- Instead of buttons, used tabs with icons
  - Two new translations will be needed for tab headers
- Dots menu and game settings are on the right side back again
- Fixed `HowLongToBeat` not displaying properly
  - It will scale better with translations which have longer text (no more ellipses, but still not perfect)
  - It now uses MUI's Accordion
- Added initial styles for uninstall button (logic not implemented yet)
- Centered play button (see rationale)

## A little rationale for centered button
- Button was not resizable which caused visual glitches on some translations
- However I didn't want to go back to the old design where the button was stretched

### Main info tab

**Before:**
Screenshot_20240303_233535
**After:**
Screenshot_20240303_233339

### Extra info tab

**Before:**
Screenshot_20240303_233459
**After:**
Screenshot_20240303_233401

### Sys requirements tab

**Before:**
Screenshot_20240303_233523
**After:**
Screenshot_20240303_233435

## TODO:

- [ ] Use more `--space-*` CSS variables for more consistency
- [ ] Implement uninstall button
- [ ] Make header (icons, back button) and footer (report issues) constant height so scrolling can be done by each panel independently

---


flavioislima:
   - - -
   - - -
   **Kajot-dev** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You can retrigger this bot by commenting **recheck** in this Pull Request</sub>

Kajot-dev:

Kajot-dev:
- Top buttons have gained a shadow and are now sticky
- Anticheat warning does not have unnecessary margin and is scaled correctly now
image
- Removed min-height and increased padding to compensate in installinfo/extrainfo rows. Since height is now bounded by the padding, opening HLTB does not have this "jumping" effect
- Implemented Uninstall button

Redesigned gamepage should now be feature complete!

src/frontend/screens/Game/GamePage/components/HLTBInline.tsx
```diff
@@ -0,0 +1,48 @@
...
+  if (!howlongtobeat) {
+    return null
+  }
+
+  return (
+    <div className="hltbWrapper">
+      <Accordion expanded={isExpanded} onChange={handleChange}>
+        <AccordionSummary
+          expandIcon={<ExpandMore />}
+          aria-control="hltb-content"
```
arielj, line 33: I get an invalid prop warning in the console with this

image

src/frontend/screens/Game/GamePage/components/HLTBInline.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  const howlongtobeat = wikiInfo.howlongtobeat
+
+  if (!howlongtobeat) {
+    return null
+  }
+
+  return (
+    <div className="hltbWrapper">
```
arielj, line 29: I'm not sure which component is problematic, but when I move using a controller (or keyboard), the item for the HLTB info doesn't have the outline (the background is a biiiit darker)

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/188464/4c207812-e47e-450a-a23b-3b5edbd0db5e

src/frontend/components/UI/WikiGameInfo/components/HowLongToBeat/index.scss
```diff
@@ -1,5 +1,7 @@
 .howLongToBeat {
```
arielj, line 1: some change here broke the style in the old design
image

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -461,16 +480,42 @@
     }
   }
 
+  .topRowWrapper {
+    display: flex;
+    flex-direction: row;
+    gap: min(var(--space-lg), 50px);
+    position: sticky;
```
arielj, line 487: I think we should aim to not have to scroll this screen at all (so this being sticky shouldn't be needed). Scrolling is kinda annoying on the Steam Deck and when we have scroll in the whole screen it means the `play` button can start outside of the viewport when it makes sense for this button to always be visible since it's kinda the main action we want.

I think at some point I had that implemented (the other components would shrink if needed) but I can't see it working that way in the main branch either so I don't know what happened there.

If something like system requirements or game description are long we can have scroll on those elements that are not that critical for example.

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -461,16 +480,42 @@
...
+      }
+    }
+
+    .opts-left {
+      justify-content: flex-start;
+    }
+
+    .opts-right {
+      justify-content: flex-end;
+    }
```
arielj, line 507: these utility classes doesn't seem to be needed, the parent has `justify-content: space-between;` anyway

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -461,16 +480,42 @@
     }
   }
 
+  .topRowWrapper {
+    display: flex;
+    flex-direction: row;
+    gap: min(var(--space-lg), 50px);
```
arielj, line 486: this gap is not really needed, we have only 2 children and they are separated on the 2 sides of the element

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -461,16 +480,42 @@
     }
   }
 
+  .topRowWrapper {
+    display: flex;
+    flex-direction: row;
```
arielj, line 485: this is also not needed, row is the default anyway

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -461,16 +480,42 @@
...
+    gap: min(var(--space-lg), 50px);
+    position: sticky;
+    top: var(--space-lg);
```
arielj, line 488: this is not doing anything, the element is `position: static` so top has no effect

Kajot-dev: It has the position `sticky`, not `static`, but since we don't want the whole screen to scroll (like you mentioned earlier), I will remove `sticky` and `top` entirely

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -416,17 +434,22 @@ export default React.memo(function GamePage(): JSX.Element | null {
...
+                      title={t2('webview.controls.back', 'Go Back')}
+                    >
+                      <ArrowBackIosNew />
+                    </NavLink>
+                  </div>
+                  <div className="opts opts-right">
+                    {!isBrowserGame && <SettingsButton gameInfo={gameInfo} />}
+                    <DotsMenu gameInfo={gameInfo} handleUpdate={handleUpdate} />
+                  </div>
+                </div>
```
arielj, line 451: a bit nit picky, but we can avoid using the `opts` wrappers if we have the 3 elements inside `topRowWrapper` and then you set `.topRowWrapper .backButton { margin-inline-end: auto }`

(you'd need to move the gap from `.opts` to `.topRowWrapper`)

src/frontend/screens/Game/GamePage/index.scss
Kajot-dev: Main info element was *partially* implemented like this, so description would become scrollable, but `max-height` was too high and whole screen scrolled anyway. I will try to fix this.

src/frontend/screens/Game/GamePage/components/HLTBInline.tsx
Kajot-dev: Good catch, it's probably MUI stuff :/

src/frontend/screens/Game/GamePage/components/HLTBInline.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { useContext, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import GameContext from '../../GameContext'
+import { Speed, ExpandMore } from '@mui/icons-material'
+import { Accordion, AccordionSummary, AccordionDetails } from '@mui/material'
+import HowLongToBeat from 'frontend/components/UI/WikiGameInfo/components/HowLongToBeat'
+
+const HLTBInline = () => {
```
arielj, line 8: maybe we can include the different logic from this inside the current `HLTB` component? and depending on the `experimentalFeatures.enableNewDesign` value? I did that already in the Scores component to make it inline

The reason is that at some point we are going to remove the old design and we'll have to remember to remove one of these HLTB components. If we just add some conditional code in the current HLTB component it's going to be easier to clean in the future.

src/frontend/screens/Game/GamePage/components/MainButton.tsx
```diff
@@ -128,6 +142,13 @@ const MainButton = ({ gameInfo, handlePlay, handleInstall }: Props) => {
 
   const is_installed = gameInfo.is_installed
 
+  const extraClassName = other.className
+
+  if (extraClassName) {
+    // do not override internal classNames
+    delete other.className
+  }
```
arielj, line 150: I understand this `className` prop is only used to add the `mainBtn` class, right? I'd rather just hardcode the `mainBtn` class inside this component since we are not going to use that for anything else

it will be added in the old design too, but it will do nothing there if there's no CSS anyway

src/frontend/screens/Game/GamePage/index.scss
arielj: yeah, sorry, I was playing around and I remove the sticky incorrectly so it was just static for me when I checked this :see_no_evil:

src/frontend/screens/Game/GamePage/index.scss
Kajot-dev: Fixed

src/frontend/screens/Game/GamePage/index.scss
Kajot-dev: Fixed

src/frontend/screens/Game/GamePage/index.scss
Kajot-dev: Fixed

src/frontend/components/UI/WikiGameInfo/components/HowLongToBeat/index.scss
Kajot-dev: Fixed

src/frontend/screens/Game/GamePage/components/HLTBInline.tsx
Kajot-dev: Fixed, it was kinda tricky, because of MUI

src/frontend/screens/Game/GamePage/components/HLTBInline.tsx
Kajot-dev: Good point! Only downside is that `isExpanded` will be present regardless of the design since react hook must be at component top level.

src/frontend/screens/Game/GamePage/index.tsx
Kajot-dev: Fixed

src/frontend/screens/Game/GamePage/components/MainButton.tsx
Kajot-dev: Fixed

src/frontend/screens/Game/GamePage/components/HLTBInline.tsx
arielj: yeah, not a big deal though, it's just a state that won't be used in the old design, but I wouldn't worry about that

Kajot-dev: Fixed

src/frontend/screens/Game/GamePage/index.scss
Kajot-dev: Should be fixed now. Now gamepage should almost always adapt to full screen and not be scrollable. One side note that under 600px gamepage *will* become scrollable since scrollable game description is not sufficient (for preventing overflow) anymore.

Kajot-dev:
@arielj All issues should be addressed now. I see that one check has failed, but it was during installation of `node-gyp` so it's not related to this PR

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -428,12 +428,43 @@
 }
 
 .App:not(.oldDesign) .gameConfigContainer {
-  display: flex;
-  padding: 2vh 2vw;
-  height: 100%;
-  gap: min(3vw, 50px);
+  display: grid;
+  position: relative;
```
imLinguin, line 432: Now that we have a `position: relative;` here, some images are slightly offset to the right, because background image is `width: 100vw;`. I'd suggest changing this to 100% in order to keep it centered

(disabled filters for better visibility in the example)

100vw
image

100%
image



Whole image:
image

Kajot-dev: I will change it to `100%` for both width and height since when viewport is <= 600px, then 100vh <= 100% and the image does not cover the whole background

Kajot-dev: Fixed!

arielj (changes requested):
I was testing this, I found only one issue, now that the 3 dots button is back on the side of the screen, the menu open in the wrong direction for RTL languages:

image

The old design handles the direction this opens for reference (if I remember right it's some css based on the RTL class).

then there's one more issue related to RTL languages, the `tabs` MUI component seems to not support it properly and the arrow are not the right icons and one of the icons never shows up, but I don't know if this is fixable if it's an issue with MUI itself

other than those 2 issues it looks great!

Kajot-dev:
> I was testing this, I found these 2 issues:
> 
> now that the 3 dots button is back on the side of the screen, the menu open in the wrong direction for RTL languages:
> 
> image
> 
> The old design handles the direction this opens for reference (if I remember right it's some css based on the RTL class).
> 
> then there's one more issue related to RTL languages, the `tabs` MUI component seems to not support it properly and the arrows are not the right icons and one of the icons never shows up, but I don't know if this is fixable if it's an issue with MUI itself
> 
> other than those 2 issues it looks great!

O I didn't even know, that Heroic supports RTL! As of MUI tabs this issue affects all MUI tabs in heroic so I would guess that it's a global MUI issue (I found this: https://github.com/mui/material-ui/issues/30207 ; it's closed but not resolved really :)). I will think of some wrapper for the `Tabs` component which would correct this

Kajot-dev:
> I was testing this, I found these 2 issues:
> 
> now that the 3 dots button is back on the side of the screen, the menu open in the wrong direction for RTL languages:
> 
> image
> 
> The old design handles the direction this opens for reference (if I remember right it's some css based on the RTL class).
> 
> then there's one more issue related to RTL languages, the `tabs` MUI component seems to not support it properly and the arrows are not the right icons and one of the icons never shows up, but I don't know if this is fixable if it's an issue with MUI itself
> 
> other than those 2 issues it looks great!

Fixed dots menu with a simple CSS tweak.
Fixed all MUI tabs across Heroic (We need to use mui Theme with specified direction as mui has internal logic which depends on that). So what happened was that CSS was RTL but the MUI logic was expecting LTR.

Kajot-dev:
Build on macos-12 failed while detaching a drive, not related to the PR

Kajot-dev:
Fixed some more issues (and ofc all of these that you mentioned):
- Nav buttons now respect color scheme
- Background is scaled up by 10% to mitigate ugly edges on blurred image (this was not visible at first, since image was scaled to 100vw and now is scaled to 100%)
- Instead of image filter `brightness(30%)` it now uses `opacity` to blend with the background. This makes background darker on dark themes (text is light) and brighter on light themes (text is dark usually)
- Also styled progressbar using `LinearProgress` (since both sidebar and downloads tab are using it) - this also affects old design slightly
- Icons on the main button are now larger and icon is always on the left side (this affects old design slightly)

Now I think everything is working as it should

arielj (approved):
hey @Kajot-dev can you update the PR to have the latest code from main? just to be sure it's all good (there's no conflicts but shouldn't hurt to do that and I can merge this)

Kajot-dev:
@arielj Pulled changes from main. Failed e2e, I think its unrelated (fails on testing a webview and I cannot reproduce the error), but you can rerun a Github job just in case

arielj:
thanks!

PR #3616, opened 2024-03-05
[WORKAROUND]: run physxlegacy setup without user input

imLinguin:
As we noticed in https://github.com/Heroic-Games-Launcher/known-fixes/pull/7  
the PHYSXLEGACY setup doesn't have a silent switch, or at least it doesn't respect it.

I came up with this simple solution, it should make the experience more hands free.

### DIFF
--- a/src/backend/storeManagers/gog/setup.ts
+++ b/src/backend/storeManagers/gog/setup.ts
@@ -350,6 +350,14 @@ async function setup(
 
       const commandParts = [exePath, ...exeArguments]
 
+      // HACKS zone
+
+      // Force hands-free setup for PHYSXLEGACY
+      if (dep === 'PHYSXLEGACY') {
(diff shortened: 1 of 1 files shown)


arielj (approved):
Works as expected. Tested with Empire Earth 3 with and without this change.

With this change physx legacy is installed silently.

flavioislima (approved):
LGTM!

Not sure if this could be called a hack, you are just adapting the command to what the dependency needs.

If was the case that we would need to handle several more dependencies with different commands then we could use a mapping object for it.

For me this is just a normal fix and I would remove the Hack commend before merging. 
But I will leave it up to you 👍🏽

imLinguin:
Maybe a workaround is better suited for this

imLinguin:
Maybe... but here we are branching out from the standard way even more 😄

Issue #3315, opened 2023-12-13
[macOS]Considering adding msync support for mac users


Coulin9:
### Feature description

I noticed that Wine-9.0-rc1 has added msync support,so is it possible to open msync in the game settings like opening esync?
Msync can greatly improve the game experience.

Coulin9:
> > I noticed that Wine-9.0-rc1 has added msync support
> 
> That’s inaccurate, I’ve only applied msync patchset to wine-staging (since 8.18), the patchset is _not_ applied to wine-devel packages.

Yeah,my mistake.I mean the latest staging packages.
I can only turn on msync by setting environment variables. I think adding a switch is more user-friendly than configuring environment variables.
The switch can be shown when the user selects the wine-staging packages.This can also remind users that this is a package which support msync.

theaspirational:
how to turn on msync by setting environment variables? I am using whisky bottle prefix and msync is checked in whisky settings. but i don't know if it works when i run gta v through heroic. Thx

Coulin9:
> how to turn on msync by setting environment variables? I am using whisky bottle prefix and msync is checked in whisky settings. but i don't know if it works when i run gta v through heroic. Thx

Maybe you can check the game boot log to see if msync is enabled

LeaveNhA:
I wanna get some log with WINEDEBUG="+msync" but, where is the log going I don't get it.

flavioislima:
Per https://github.com/marzent/wine-msync we would just need to copy/paste the ESYNC setting that we have today and use the `WINEMSYNC=1` environmental variable when it's enabled.

flavioislima:
> > Per https://github.com/marzent/wine-msync we would just need to copy/paste the ESYNC setting that we have today and use the `WINEMSYNC=1` environmental variable when it's enabled.
> 
> Actually need to set `WINEEYSNC=1` & `WINEMSYNC=1` due to a D3DMetal bug

Perfect, did not know that. Thanks for the info!

flavioislima:
Closing this one since it was implemented by #3630

PR #3630, opened 2024-03-15
Mac msync

Mariaboni:
Hey, guys! I implemented Msync (here it is the issue https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3315) but I cannot test it, since I don't have a Mac. I was able to check that the Enable Msync option was working fine on the Game Settings
screenshotMsync

---


src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -42,6 +42,7 @@ import SyncSaves from '../SyncSaves'
 import FooterInfo from '../FooterInfo'
 import { Tabs, Tab } from '@mui/material'
 import { GameInfo } from 'common/types'
+import EnableMsync from '../../components/EnableMsync'
```
flavioislima, line 45: This import could be inside the above one that is already importing other components from the same folder

src/frontend/screens/Settings/components/EnableMsync.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const { platform } = useContext(ContextProvider)
+    const { isMacNative } = useContext(SettingsContext)
+    const isMac = platform === 'darwin'
+    const [enableMsync, setEnableMsync] = useSetting('enableMsync', false)
+
+    if (!isMac || isMacNative){
+        return <></>
+    }
+
+    return(
```
flavioislima, line 22: Prettier will probably complain about these, you need to fix the styling by running
`yarn prettier-fix`

flavioislima (commented):
Overall looks good, but left a few comments.
I can test it on macOS.

src/common/types.ts
```diff
@@ -171,6 +171,7 @@ export interface GameSettings {
   enableDXVKFpsLimit: boolean
   enableEsync: boolean
   enableFSR: boolean
+  enableMsync: boolean
```
CommandMC, line 174: Whenever you're adding a new key to the settings, you *should* also add a default value for it (in `src/backend/config.ts`). See [this comment chain](<https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3565#discussion_r1489992624>) for details on how this is done

It isn't strictly necessary to do this (we have enough type issues in the codebase right now that nothing yells at you for it), but it does add an "unknown" `undefined` into your key type ("unknown" because this is not actually reflected in the type)

Mariaboni: pushed a fix to this. Can you check again?

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
Mariaboni: fixed!

src/common/types.ts
Etaash-mathamsetty: that's not the code that I think CommandMC was talking about, look for enableFsync: isLinux in the source

Mariaboni: image
Well, in config.ts "isMac" is the default value already:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3630/files#diff-002cc40c00ed6f11f3ff2ccc3006de3bbda8a0aef5a18cc5fc29c2d1a55d59b7R316

Etaash-mathamsetty: ok neat

flavioislima (approved):
Looks good.
Thanks for implementing that :D

PR #3634, opened 2024-03-16
[Test] Add a small timeout to fix flaky webview e2e spec

arielj:
The test seems to be flaky (I guess moving between webview pages is not consistent in speed). There are many PRs failing in this test and re-running the e2e tests sometimes solves the problem.

This adds a small timeout before pressing `Go back` that seems to solve the flakyness, I tested locally and it's green consistently with this.

---


CommandMC (approved):
I noticed this myself in one of my PRs, but was too lazy to look into it
Thanks!

PR #3648, opened 2024-03-24
[Feat] Improve the "Move Game" experience a bit

CommandMC:
Nothing too significant, just 3 quick improvements to the "Move Game" functionality:
- Progress reports are dispatched to the Frontend more frequently
  - There were multiple `if (percent)` checks, which erroneously filtered out the case where percent might be 0 (aka we just started copying a new file)
- On Linux, we now tell `rsync` to delete source/old files after they're copied to the new location
  - This should help in scenarios with very little disk space being available. Before, we first copy and then delete, now we delete files while copying, and then the entire directory (which should be using virtually no space anymore) at the end
- The types for the `progress` property of the progress update are now correct (`percent` should be a number instead of a string, and `eta` and `bytes` were missing entirely)
  - Note that `eta` and `bytes` still aren't populated on Windows (`robocopy` simply doesn't give us this info), so I've not displayed them anywhere yet
   - This issue was found by type-checking `sendFrontendMessage`, a PR for that should land alongside this one

I'm still not entirely happy with this function (from a user's point of view I'd expect it to give me a progress bar for the entire move operation, knowing that one specific file is being moved is not too helpful), but that'd be a more complex refactor

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -996,41 +996,33 @@ export async function moveOnWindows(
   newInstallPath = join(newInstallPath, basename(install_path))
 
   let currentFile = ''
-  let currentPercent = ''
 
   // move using robocopy and show progress of the current file being copied
   const { code, stderr } = await spawnAsync(
     'robocopy',
-    [install_path, newInstallPath, '/MOVE', '/MIR'],
+    [install_path, newInstallPath, '/MOVE', '/MIR', '/NJH', '/NJS', '/NDL'],
     { stdio: 'pipe' },
     (data) => {
-      data = data.replaceAll(/\s/g, ' ')
+      let percent = 0
+      const percentMatch = data.match(/(\d+)(?:\.\d+)?%/)?.[1]
+      if (percentMatch) percent = Number(percentMatch)
 
-      const match = data.split(' ').filter(Boolean)
-      // current percentage
-      const percent = match.filter((m) => m.includes('%'))[0]
-      // current file
-      const file = match[match.length - 1]
-      if (percent) {
-        currentPercent = percent
-      }
+      const filenameMatch = data.match(/([\w.:\\]+)$/)?.[1]
+      if (filenameMatch) currentFile = filenameMatch
 
-      if (file && file.includes('.') && !file.includes('%')) {
-        currentPercent = '0%'
-        currentFile = file
-      }
-
-      if (match) {
-        sendFrontendMessage(`progressUpdate-${gameInfo.app_name}`, {
-          appName: gameInfo.app_name,
-          runner: gameInfo.runner,
-          status: 'moving',
-          progress: {
-            percent: currentPercent,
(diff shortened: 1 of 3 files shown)


src/backend/utils.ts
```diff
@@ -1070,38 +1061,63 @@ export async function moveOnUnix(
...
-            .find((d) => d.includes('/') && !d.includes('%'))
-            ?.split('/') || []
-        const file = split.at(-1) || ''
-
-        if (file) {
-          currentFile = file
+        let percent = 0
+        let eta = ''
+        let bytes = '0'
+        // Rsync outputs either the file currently being transferred or a
```
arielj, line 1083: one comment on this, when files are small you can get multiple files in the same `data` so it's not either one or the other. I'm printing `data` to the console and I see things like this:

```
api-ms-win-crt-conio-l1-1-0.dll
          20824 100%   36,12kB/s    0:00:00 (xfr#41, ir-chk=1049/1091)
api-ms-win-crt-convert-l1-1-0.dll
          23680 100%   41,07kB/s    0:00:00 (xfr#42, ir-chk=1048/1091)
api-ms-win-crt-environment-l1-1-0.dll
```

(this is all a single `data`)

So in a single string we can have a mix of files and progress updates (I imagine when files are really small).

src/backend/utils.ts
```diff
@@ -1070,38 +1061,63 @@ export async function moveOnUnix(
...
+          // file (we thus start at 0%, with 0 bytes moved, and with no ETA)
+
+          // Data will be a multiline string, potentially with directory names,
+          // finally the file name, and ending with an empty line
+          currentFile = data.trim().split('\n').at(-1)!
+        } else {
+          // If we got the progress update, try to read out the bytes, ETA and
+          // percent
+          const bytesMatch = data.match(/^\s*(\d+)/)?.[1]
+          const etaMatch = data.match(/(\d+:\d+:\d+)/)?.[1]
```
arielj, line 1103:
```suggestion
          const etaMatch = data.match(/(\d+:\d\d:\d\d/)?.[1]
```

since we know it's always 2 digits for the second and minute

src/backend/utils.ts
CommandMC: Ugh, and here I was thinking only robocopy did this buffering
Looking at just the last line sounds good though, I'll try that out

arielj (approved):
looks good to me!

I can't test the windows one though, and I feel like we are doing more work than needed sending so many updates to the frontend but I think that's something to refactor in another PR anyway (I imagine something like capturing the last file and progress for every data but only send messages to the frontend every X milliseconds? something like that)

CommandMC:
We should definitely modify how the progress is presented, I agree. I think going for a general "Here's how far along the entire move process is" progress bar (as Steam does it, for example), and switching to per-file progress if one takes more than a couple seconds to transfer, would be the best option

PR #3649, opened 2024-03-24
[Tech] Type-check Backend & Frontend events

CommandMC:
This adds typechecking for `backendEvents` and `sendFrontendMessage`. I've already caught a few issues with this (see #3648, some smaller issues are also corrected here)  
Typechecking is done with the usual pattern of an interface holding all possible events/messages, and their corresponding function definitions (similarly to how IPC works, for example)

This is entirely a dev change, no actual code is altered

Requires #3648 

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,6 +142,7 @@
     "ts-jest": "29.0.5",
     "ts-prune": "0.10.3",
     "type-fest": "3.6.1",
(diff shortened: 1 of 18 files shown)


arielj (approved):
looks good to me :+1:

PR #3654, opened 2024-03-29
[Fix] Make checkDiskSpace look at the correct drive again

CommandMC:
In my checkDiskSpace refactor (#3440), I used the `root` property of the object returned by Node's [`path.parse`](https://nodejs.org/api/path.html#pathparsepath), thinking that it'd, logically, give me the root folder (or in other words the mount point) of the disk the path is on. This is not the case on Linux/macOS, it's just simply always `/` there.
This issue would manifest itself in the wrong disk space info being displayed when installing a game onto a drive other than the primary one.

Resolving this was somewhat tricky. Since this function runs on both macOS and Linux, we can't use commandline tools like `findmnt` to help us here, so I went back to what we did before the refactor: Going up the directory tree and using the first path that exists. This works just fine on Unix, since there will always be a `/` path (as opposed to Windows, where entire drive letters can be missing)

I've also added a test for these functions now, so this issue can't happen again.

To test this:
1. Install a game onto another hard drive on the current main branch, see that the total/free disk space is reported incorrectly
2. Try the same thing with this PR, info will now be correct

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -563,16 +563,10 @@ ipcMain.on('unlock', () => {
 })
 
 ipcMain.handle('checkDiskSpace', async (_e, folder): Promise<DiskSpaceData> => {
-  // We only need to look at the root directory for used/free space
-  // Trying to query this for a directory that doesn't exist (which `folder`
-  // might be) will not work
-  const { root } = path.parse(folder)
-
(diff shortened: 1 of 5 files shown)


Etaash-mathamsetty (approved):
looks like it's working again

PR #3663, opened 2024-03-31
[Revert] Refactor Extraction methods #3181

flavioislima:
Revert this since its not being useful anymore and it is breaking arm64 builds.

---


src/backend/tools/index.ts
```diff
@@ -117,24 +117,29 @@ async function installOrUpdateTool(tool: Tool) {
 
   const extractDestination = join(toolsPath, tool.name, latestVersion)
   mkdirSync(extractDestination, { recursive: true })
-  try {
-    await extractFiles({
-      path: latestVersionArchivePath,
-      destination: extractDestination,
-      strip: tool.strip ?? 1
+  const extractCommand = `${tool.extractCommand} ${latestVersion} -C ${toolsPath}/${tool.name}`
+  await execAsync(extractCommand)
```
imLinguin, line 121: this begs for issues with space in the path. It may be better to use spawn here, or if we really need exec, let's at least quote variables that may contain spaces

imLinguin (dismissed):

src/backend/tools/index.ts
flavioislima: changed

src/backend/utils.ts
```diff
@@ -1483,66 +1478,6 @@ function calculateEta(
...
-    '-C',
-    destination,
-    `--strip-components=${strip}`
-  ])
-  if (code !== 0) {
-    logError(`Extracting Error: ${stderr}`, LogPrefix.Backend)
-    return { status: 'error', error: stderr }
-  }
-  return { status: 'done', installPath: destination }
-}
```
CommandMC, line 1525: This function should stay & become the new `extractFiles` IMO

Doing it that way would mean you wouldn't need all the changes in `backend/tools/index.ts`

CommandMC (changes requested):

CommandMC (approved):
My approval of course means less now, but still; LGTM

PR #3702, opened 2024-04-17
[UX] Accessibility option to always show titles in library without having to hover

arielj:
This PR adds a new accessibility toggle to always show the cards' titles without having to hover/focus the card.

This can make finding games easier and it's particularly useful when using a screen with though capabilities where hover doesn't always make sense.

image

---

### DIFF
--- a/src/frontend/screens/Library/components/GameCard/index.tsx
+++ b/src/frontend/screens/Library/components/GameCard/index.tsx
@@ -90,7 +90,6 @@ const GameCard = ({
   const {
     hiddenGames,
     favouriteGames,
-    allTilesInColor,
     showDialogModal,
     setIsSettingsModalOpen,
     activeController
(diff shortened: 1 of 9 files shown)


src/frontend/screens/Library/components/GameCard/index.tsx
```diff
@@ -367,12 +366,8 @@ const GameCard = ({
...
   const gamepadClass = activeController ? 'gamepad' : ''
   const justPlayedClass = justPlayed ? 'justPlayed' : ''
-  const imgClasses = `gameImg ${isInstalled ? 'installed' : ''} ${
-    allTilesInColor ? 'allTilesInColor' : ''
-  }`
-  const logoClasses = `gameLogo ${isInstalled ? 'installed' : ''} ${
-    allTilesInColor && 'allTilesInColor'
-  }`
+  const imgClasses = `gameImg ${isInstalled ? 'installed' : ''}`
+  const logoClasses = `gameLogo ${isInstalled ? 'installed' : ''}`
```
arielj, line 370: I moved these classes to be applied in the list of games instead of in each game, since it applies to all games

CommandMC (approved):

PR #3711, opened 2024-04-23
[GOG]: don't check for notInstallable during library sync

imLinguin:
This should resolve number of issues, but also allow players with pre-release access to actually download their games.

### DIFF
--- a/src/backend/storeManagers/gog/games.ts
+++ b/src/backend/storeManagers/gog/games.ts
@@ -119,19 +119,22 @@ export async function getExtraInfo(appName: string): Promise<ExtraInfo> {
 
   const gamesData = await getGamesData(appName)
 
-  const gogStoreUrl = gamesData?._links?.store.href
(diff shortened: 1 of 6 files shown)


arielj (approved):
looks good to me, I can't confirm the solution cause I never had the issue if games flagged as not installable in the first place

imLinguin:
There is one flaw with this approach. Demos are treated as full releases because the endpoint we use collects different game editions under one entry.

flavioislima (approved):
lgtm but I have no way of testing it

daygr:
I was experiencing this problem, I used the AppImage from this run https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/actions/runs/8850391069 and that fixed the issue for me. I shared my logs from 2.14.1 on a [discord thread](https://discord.com/channels/812703221789097985/1207885396382580846), and they are now correct and do not show any timeouts or other errors.

PR #3716, opened 2024-04-25
Disable installing games with broken anticheat (add advanced setting for override)

arielj:
Users keep installing games like Fornite on linux even after all the warnings, so I'm making it harder to install without reading that the game will not work.

By default we now disable installing games with broken or denied anticheat. Users that care to read the messages will see that this measure can be disabled in the Advanced settings (I added even more warnings).

After checking the toggle in the advanced settings, the warning that asks users to click `I understand this won't work, install anyway` is still displayed.

I know this is more aggressive, but it's only more annoying for users that don't read. Advanced users can just disable this check once in the settings and it won't bother again.

---


flavioislima (approved):
LGTM

Another approach would be to disable the install button for like 10, 20 seconds and add a countdown inside it. But this one is fine as well 👍🏽

PR #3480, opened 2024-01-30
[Exp]: ULWGL support

imLinguin:
Experimental feature with ULWGL support. The feature is enabled by default.

Requires user to manually put `ULWGL-launcher` as `$XDG_CONFIG_HOME/heroic/tools/runtimes/ulwgl`.
This may as well be a symlink

e.g
```
git clone https://github.com/open-wine-components/ULWGL-launcher ~/.config/heroic/tools/runtimes/ulwgl
```

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #3724, opened 2024-04-27
[Exp]: UMU support

Etaash-mathamsetty:
Experimental feature with UMU support. The feature is enabled by default.

~~Requires user to manually put `umu-run` inside `$XDG_CONFIG_HOME/heroic/tools/runtimes/umu`~~

based on: #3480 


R1kaB3rN:
Using this branch, I noticed the umu runtime isn't downloaded at ~/.config/heroic/tools/runtimes from the Lutris API when the checkbox _Use UMU as Proton runtime_ is checked. Is this intended or are users supposed build it and put it in that directory themselves?

src/backend/launcher.ts
```diff
@@ -864,11 +860,21 @@ async function runWineCommand({
   return new Promise<{ stderr: string; stdout: string }>((res) => {
     const wrappers = options?.wrappers || []
     let bin = ''
+    const umuSupported = isUmuSupported(wineVersion.type)
+
     if (wrappers.length) {
       bin = wrappers.shift()!
-      commandParts.unshift(...wrappers, wineBin)
+      if (umuSupported) {
+        const umuBin = join(runtimePath, 'umu', 'umu-run')
```
R1kaB3rN, line 868: Alternatively, you can execute umu_run.py instead, which is more direct and what Lutris does

src/backend/launcher.ts
```diff
@@ -761,9 +755,15 @@ export async function verifyWinePrefix(
...
-    commandParts: ['wineboot', '--init'],
+    commandParts:
+      wineVersion.type === 'proton' &&
+      GlobalConfig.get().getSettings().experimentalFeatures?.umuSupport !==
+        false
+        ? ['createprefix']
+        : ['wineboot', '--init'],
     wait: haveToWait,
     gameSettings: settings,
+    protonVerb: 'run',
```
R1kaB3rN, line 766: By the way, if you omit the verb, umu will default to using `waitforprefixandrun` verb which will block users from running more than 1 game in the same prefix. However, if heroic wants to allow users to run more than 1 game, then you could default to `runinprefix`.

src/backend/launcher.ts
```diff
@@ -545,7 +537,9 @@ function setupWineEnvVars(gameSettings: GameSettings, gameId = '0') {
     }
     case 'proton':
       ret.STEAM_COMPAT_CLIENT_INSTALL_PATH = steamInstallPath
+      ret.WINEPREFIX = winePrefix
       ret.STEAM_COMPAT_DATA_PATH = winePrefix
+      ret.PROTONPATH = dirname(gameSettings.wineVersion.bin)
```
R1kaB3rN, line 542: If you choose not to set this, it will cause umu download the latest UMU-Proton and setting PROTONPATH=GE-Proton, will cause it to download the latest GE-Proton. 

When umu is configured as the runtime, perhaps another option can be added to the dropdown list as the first option (e.g., UMU-Latest)?

Etaash-mathamsetty:
> Using this branch, I noticed the umu runtime isn't downloaded at ~/.config/heroic/tools/runtimes from the Lutris API when the checkbox _Use UMU as Proton runtime_ is checked. Is this intended or are users supposed build it and put it in that directory themselves?

that feature is not implemented yet, i simply just made it work again and haven't done anything on top of that

arielj:
I used one of the Alien Breed games that I know UMU has a fix installing physx. This is my result:

- Using Proton-GE and UMU disabled (tried with proton-ge 8.x and 9.x)
Games won't run if the prefix is not already created. I get this error:
```
Traceback (most recent call last):
  File "/home/ariel/.config/heroic/tools/proton/Proton-GE-Proton8-32/proton", line 1815, in <module>
    g_session.init_session(sys.argv[1] != "runinprefix")
  File "/home/ariel/.config/heroic/tools/proton/Proton-GE-Proton8-32/proton", line 1575, in init_session
    g_compatdata.setup_prefix()
  File "/home/ariel/.config/heroic/tools/proton/Proton-GE-Proton8-32/proton", line 770, in setup_prefix
    with self.prefix_lock:
  File "/home/ariel/.config/heroic/tools/proton/Proton-GE-Proton8-32/filelock.py", line 323, in __enter__
    self.acquire()
  File "/home/ariel/.config/heroic/tools/proton/Proton-GE-Proton8-32/filelock.py", line 271, in acquire
    self._acquire()
  File "/home/ariel/.config/heroic/tools/proton/Proton-GE-Proton8-32/filelock.py", line 384, in _acquire
    fd = os.open(self._lock_file, open_mode)
FileNotFoundError: [Errno 2] No such file or directory: '/home/ariel/Games/Heroic/Prefixes/Alien Breed Impact/pfx.lock'
Launch command: ['/home/ariel/.config/heroic/tools/proton/Proton-GE-Proton8-32/proton', 'waitforexitandrun', '/home/ariel/Games/Heroic/Alien Breed Impact/Binaries/AlienBreed-Impact.exe']
All processes exited
============= End of log =============
```

If I do the same in the main branch, the error doesn't happen. Something changed that makes running games with proton without the experimental feature fails to create a prefix (if the prefix is already there it works).

- Using Proton 8 and the UMU runtime enabled
The prefix gets created, UMU doesn't detect the game so it doesn't install physx, then the game doesn't start cause it's missing physx I guess

```
Game Log:
[34mProtonFixes[444140] INFO: Running protonfixes[0m
[34mProtonFixes[444140] INFO: Running checks[0m
[34mProtonFixes[444140] INFO: All checks successful[0m
[34mProtonFixes[444140] INFO: Non-steam game UNKNOWN (umu-22610)[0m
[34mProtonFixes[444140] INFO: Using global defaults for UNKNOWN (umu-22610)[0m
[34mProtonFixes[444140] INFO: Non-steam game UNKNOWN (umu-22610)[0m
[34mProtonFixes[444140] INFO: GOG store specified, using GOG database[0m
[34mProtonFixes[444140] INFO: No protonfix found for UNKNOWN (umu-22610)[0m
Proton: /home/ariel/Games/Heroic/Alien Breed Impact/Binaries/AlienBreed-Impact.exe
Proton: Executable a unix path, launching with /unix option.
fsync: up and running.
wine: RLIMIT_NICE is <= 20, unable to use setpriority safely
Launch command: ['/home/ariel/.config/heroic/tools/runtimes/umu/umu_run.py', '/home/ariel/Games/Heroic/Alien Breed Impact/Binaries/AlienBreed-Impact.exe']
All processes exited
============= End of log =============
```

I wonder if this is an issue with UMU not knowing the game? because with proton 9 it shows the game title just fine (and it knows it's `umu-22610`). Or if we need to do something different for proton 8.

```
Game Log:
[34mProtonFixes[456441] INFO: Running protonfixes[0m
[34mProtonFixes[456441] INFO: Running checks[0m
[34mProtonFixes[456441] INFO: All checks successful[0m
[34mProtonFixes[456441] INFO: Non-steam game Alien Breed: Impact (umu-22610)[0m
[34mProtonFixes[456441] INFO: GOG store specified, using GOG database[0m
[34mProtonFixes[456441] INFO: No global defaults found for Alien Breed: Impact (umu-22610)[0m
[34mProtonFixes[456441] INFO: Non-steam game Alien Breed: Impact (umu-22610)[0m
[34mProtonFixes[456441] INFO: GOG store specified, using GOG database[0m
[34mProtonFixes[456441] INFO: Using global protonfix for Alien Breed: Impact (umu-22610)[0m
[34mProtonFixes[456441] INFO: Checking if winetricks physx is installed[0m
Proton: /home/ariel/Games/Heroic/Alien Breed Impact/Binaries/AlienBreed-Impact.exe
...
```

- Using Proton 9 and the UMU runtime enabled
The prefix gets created, UMU detects the game and installs physx automatically, but then the game doesn't work (shows splash screen and then screen goes black and game closes instead of starting the initial credits). But this seems to be an issue of proton 9, because if I test with the main branch proton 9 + auto install known fixes I get the same crash.

src/backend/cache.ts
```diff
@@ -6,14 +6,19 @@ export default class CacheStore<ValueType, KeyType extends string = string> {
...
    * Creates a new cache store
    * @param filename
    * @param max_value_lifespan How long an individual entry in the store will
    *                           be cached (in minutes)
    */
-  constructor(filename: string, max_value_lifespan: number | null = 60 * 6) {
+  constructor(
+    filename: string,
+    max_value_lifespan: number | null = 60 * 6,
+    options?: { invalidateCheck?: (data: ValueType) => boolean }
```
arielj, line 20: we should update the JSDoc above the method with this new option so it's more clear what it's used for

at least from the usage of this in this PR I'm not sure I understand why it's needed

src/backend/storeManagers/gog/games.ts
```diff
@@ -558,6 +559,13 @@ export async function launch(
 
     const { bin: wineExec, type: wineType } = gameSettings.wineVersion
 
+    if (wineType === 'proton') {
```
arielj, line 562: Should this be?

```suggestion
    if (isUmuSupported(wineType)) {
```

to avoid the extra API request if umu is not enabled

maybe this if can also be moved somewhere else? I see it's the same code in the 3 stores

src/backend/wiki_game_info/umu/utils.ts
```diff
@@ -0,0 +1,44 @@
+import axios from 'axios'
+import { Runner } from 'common/types'
+import { umuStore } from '../electronStore'
+
+interface GameObject {
+  title: string
+  umu_id: string
+}
+const storeMapping: { [key in Runner]?: string } = {
```
arielj, line 9:
```suggestion
const storeMapping: Record<Runner, string> = {
```

src/backend/wiki_game_info/umu/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  legendary: 'egs',
+  nile: 'amazon'
+}
+
+export async function getUmuId(
+  appName: string,
+  runner: Runner
+): Promise<string | null> {
+  const store = storeMapping[runner]
+  if (!store) {
```
arielj, line 20: can this ever be falsey?

Arias800:
Hello,

I don't know if this is a known issue or not, but UMU doesn't seem to work for sideloaded games on Linux.
I get a "Proton: Need a verb" in the log.

Everything works fine on the main Heroic branch.

arielj:
> Hello,
> 
> I don't know if this is a known issue or not, but UMU doesn't seem to work for sideloaded games on Linux. I get a "Proton: Need a verb" in the log.
> 
> Everything works fine on the main Heroic branch.

hmmm I cannot reproduce this, I sideloaded a game, all the process done with proton 9.5 and the UMU support enabled and it worked just fine

arielj:
I still have the issue that, after installing a game (or deleting its prefix), if I launch it with Proton-GE without UMU it fails to launch (the first log in this message https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3724#issuecomment-2143223039)

src/backend/launcher.ts
```diff
@@ -749,7 +758,7 @@ export async function verifyWinePrefix(
     return { res: { stdout: '', stderr: '' }, updated: false }
   }
 
-  if (!existsSync(winePrefix)) {
+  if (!existsSync(winePrefix) && wineVersion.type !== 'proton') {
```
arielj, line 761: This change is the one preventing proton without UMU from creating the prefix when needed, so maybe:

```suggestion
  if (!existsSync(winePrefix) && !umuSupported(wineVersion.type)) {
```

so it still creates the prefix folder for proton when umu is disabled

Arias800:
> can you share the logs when it fails?

Log for the game : [cSENNjXteRS3ky69W8w9Ue-lastPlay.log](https://github.com/user-attachments/files/15612551/cSENNjXteRS3ky69W8w9Ue-lastPlay.log)

It's a game that I download from Itch.io if you want to test in the same condition : https://sirtartarus.itch.io/after-the-curtain-call

arielj:
I just tried it and it worked fine for me. what I did was:
- download and unzip file in /downloads
- in heroic > Add Game > enter info > select proton-ge 9.5 > select `Curtains.exe` > finish
- game gets added
- double checked the `umu` experimental feature is enabled in settings > advanced
- click play > it prompts to install VC++ runtime
- game started

Arias800:
I found the source of my problem.
I started testing this branch when installing UMU manually was mandatory.
But now it seems to create a conflict with the current code.
Removing UMU and rebooting my system seems to fix everything.

R1kaB3rN:
I think Heroic should attempt to search the user's PATH and prefer the system package if it's installed like Lutris does to avoid possible crashes. Because when using this branch and adding a DRM free game, the game will fail to launch when setting UMU as the runtime when the user already has a system umu-launcher installation that they built or packaged from their distro that is several commits ahead of the release fetched from the Lutris API.

Here's the traceback:
https://gist.github.com/R1kaB3rN/f2f521ab4d5e34029e59e0b59990030b

In the log,  `umu_run.py` in `~/.config/heroic/tools/runtimes/umu` will fail to find two files, namely the `reaper` and `umu_version.json` in the `~/.local/share/umu` directory. Since commit https://github.com/Open-Wine-Components/umu-launcher/commit/e518165395b8759c4d7412c74a03759780a5ee00 these two files will no longer exist as `reaper` is built natively into the launcher and umu does not copy its configuration file anymore.

To fix this, just search the users PATH, prefer the one installed if found and otherwise use the one provided by Heroic. Here's the relevant snippet in Lutris:

https://github.com/lutris/lutris/blob/master/lutris/util/wine/proton.py#L28

src/backend/launcher.ts
imLinguin: we use `waitforexitandrun` only for game executables to ensure we run in proper env. There is no reason for users to use same prefix for more than one game.
`waitforexitandrun` acts as a safeguard to avoid situations where wineserver instance runs with fsync and the game needs that disabled.

R1kaB3rN:
It appears the winecfg button isn't working properly for some reason as the window will not appear. Here's a snippet of the Heroic logs:

> (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> (10:01:06) DEBUG:   [Backend]:          Running Wine command: createprefix
> (10:01:06) WARNING: [Backend]:          Required prefix files are missing, running `verifyWinePrefix` to create prefix
> (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> (10:01:06) DEBUG:   [Backend]:          Running Wine command: createprefix
> (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> (10:01:06) DEBUG:   [Backend]:          Running Wine command: winecfg
> [7303:0619/100109.194551:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 3 times!

One way this can be fixed is for Heroic to run winecfg through UMU:
`GAMEID=umu-0 umu_run.py winecfg`

Also, the winetricks button shouldn't be working properly right now either as it will fail to find its dependencies:

> (10:12:07) ERROR:   [Winetricks]:       cabextract: error while loading shared libraries: libmspack.so.0: cannot open shared object file: No such file or directory

To address this, Heroic will need to simply call `GAMEID=umu-0 umu_run.py winetricks ...`, but [this can't be fixed yet until the next release of GE-Proton/UMU-Proton](https://github.com/GloriousEggroll/proton-ge-custom/pull/115). Also, Heroic shouldn't need to download its own winetricks anymore either as GE-Proton already includes.

wins1ey:
> It appears the winecfg button isn't working properly for some reason as the window will not appear.

I'm just running the latest version of the umu branch on Arch and it works fine for me.
Heroic logs:
> (18:32:07) INFO:    [Backend]:          Checking if wine version exists: Proton - proton-ge-custom
> (18:32:07) INFO:    [Backend]:          Checking if wine version exists: Proton - proton-ge-custom
> (18:32:07) INFO:    [Backend]:          Checking if wine version exists: Proton - proton-ge-custom
> (18:32:07) DEBUG:   [Backend]:          Running Wine command: createprefix
> (18:32:09) INFO:    [Backend]:          Checking if wine version exists: Proton - proton-ge-custom
> (18:32:09) DEBUG:   [Backend]:          Running Wine command: winecfg

Etaash-mathamsetty:
> It appears the winecfg button isn't working properly for some reason as the window will not appear. Here's a snippet of the Heroic logs:
> 
> > (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> > (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> > (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> > (10:01:06) DEBUG:   [Backend]:          Running Wine command: createprefix
> > (10:01:06) WARNING: [Backend]:          Required prefix files are missing, running `verifyWinePrefix` to create prefix
> > (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> > (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> > (10:01:06) DEBUG:   [Backend]:          Running Wine command: createprefix
> > (10:01:06) INFO:    [Backend]:          Checking if wine version exists: Proton - GE-Proton9-7
> > (10:01:06) DEBUG:   [Backend]:          Running Wine command: winecfg
> > [7303:0619/100109.194551:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 3 times!
> 
> One way this can be fixed is for Heroic to run winecfg through UMU: `GAMEID=umu-0 umu_run.py winecfg`
> 
> Also, the winetricks button shouldn't be working properly right now either as it will fail to find its dependencies:
> 
> > (10:12:07) ERROR:   [Winetricks]:       cabextract: error while loading shared libraries: libmspack.so.0: cannot open shared object file: No such file or directory
> 
> To address this, Heroic will need to simply call `GAMEID=umu-0 umu_run.py winetricks ...`, but [this can't be fixed yet until the next release of GE-Proton/UMU-Proton](https://github.com/GloriousEggroll/proton-ge-custom/pull/115). Also, Heroic shouldn't need to download its own winetricks anymore either as GE-Proton already includes it.

heroic already runs winecfg through umu...

R1kaB3rN:
> heroic already runs winecfg through umu...

Sorry, you're right. I just retried running `winecfg` with the latest umu after copying the files to `$HOME/.config/heroic/tools/runtimes/umu`.

R1kaB3rN:
https://github.com/GloriousEggroll/proton-ge-custom/pull/115 was just merged so Heroic should be able to run winetricks. 

I think the integration should be pretty much finished except for two things:

1. Hide UMU-Latest from the drop down menu. This is not intended to be selected by end users and it won't actually use the latest UMU-Proton. It's also not relevant to Heroic because they're only interested in the runtime.
2. Pass `GAMEID=winetricks-gui` when the user wants to use the winetricks GUI.

R1kaB3rN:
Patch for hiding `UMU-Latest` in dropdown:

```
From 80510b4bd0d391764756a40d4e95df7d4a964ff1 Mon Sep 17 00:00:00 2001
From: R1kaB3rN <100738684+R1kaB3rN@users.noreply.github.com>
Date: Sun, 30 Jun 2024 12:44:53 -0700
Subject: [PATCH] Hide UMU-Latest from dropdown

---
 src/backend/utils/compatibility_layers.ts | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/src/backend/utils/compatibility_layers.ts b/src/backend/utils/compatibility_layers.ts
index 242ab0d7..b973c700 100644
--- a/src/backend/utils/compatibility_layers.ts
+++ b/src/backend/utils/compatibility_layers.ts
@@ -167,6 +167,9 @@ export async function getLinuxWineSet(
   protonPaths.forEach((path) => {
     if (existsSync(path)) {
       readdirSync(path).forEach((version) => {
+        if (version.startsWith("UMU-Latest")) {  // File is for internal use only
+          return
+        }
         const protonBin = join(path, version, 'proton')
         // check if bin exists to avoid false positives
         if (existsSync(protonBin)) {
-- 
2.45.2
```

R1kaB3rN:
Also, do we want to address removing Wine-GE from the Wine Manager UI in this PR? Those builds are end of life and obsolete in favor of umu with GE-Proton or UMU-Proton. 

Here's a patch that replaces Wine-GE in the Wine Manager to UMU-Proton:

```
From 174c2d924d16de1022de65e2c127766bfe55b6b2 Mon Sep 17 00:00:00 2001
From: R1kaB3rN <100738684+R1kaB3rN@users.noreply.github.com>
Date: Tue, 2 Jul 2024 09:27:59 -0700
Subject: [PATCH] Remove Wine-GE from Wine Manager

---
 .../wine/manager/downloader/constants.ts       |  4 ++--
 src/backend/wine/manager/downloader/main.ts    | 10 +++++-----
 .../wine/manager/downloader/utilities.ts       |  3 +++
 src/backend/wine/manager/utils.ts              |  2 +-
 src/common/types.ts                            |  4 ++--
 src/frontend/screens/WineManager/index.tsx     | 18 +++++++++---------
 6 files changed, 22 insertions(+), 19 deletions(-)

diff --git a/src/backend/wine/manager/downloader/constants.ts b/src/backend/wine/manager/downloader/constants.ts
index d991fa23..e0ed9301 100644
--- a/src/backend/wine/manager/downloader/constants.ts
+++ b/src/backend/wine/manager/downloader/constants.ts
@@ -1,6 +1,6 @@
 /// Url to Wine GE github release page
-export const WINEGE_URL =
-  'https://api.github.com/repos/GloriousEggroll/wine-ge-custom/releases'
+export const UMUPROTON_URL =
+  'https://api.github.com/repos/Open-Wine-Components/umu-proton/releases'
 
 /// Url to Proton GE github release page
 export const PROTONGE_URL =
diff --git a/src/backend/wine/manager/downloader/main.ts b/src/backend/wine/manager/downloader/main.ts
index d37728db..affbd6c0 100644
--- a/src/backend/wine/manager/downloader/main.ts
+++ b/src/backend/wine/manager/downloader/main.ts
@@ -10,7 +10,7 @@ import {
 } from 'graceful-fs'
 
 import {
-  WINEGE_URL,
+  UMUPROTON_URL,
   PROTONGE_URL,
   PROTON_URL,
   WINELUTRIS_URL,
@@ -43,17 +43,17 @@ interface getVersionsProps {
  *          * rejects with an {@link Error}
  */
 async function getAvailableVersions({
-  repositorys = [Repositorys.WINEGE, Repositorys.PROTONGE],
+  repositorys = [Repositorys.UMUPROTON, Repositorys.PROTONGE],
   count = 100
 }: getVersionsProps): Promise<VersionInfo[]> {
   const releases: Array<VersionInfo> = []
 
   for await (const repo of repositorys) {
     switch (repo) {
-      case Repositorys.WINEGE: {
+      case Repositorys.UMUPROTON: {
         await fetchReleases({
-          url: WINEGE_URL,
-          type: 'Wine-GE',
+          url: UMUPROTON_URL,
+          type: 'UMU-Proton',
           count: count
         })
           .then((fetchedReleases: VersionInfo[]) => {
diff --git a/src/backend/wine/manager/downloader/utilities.ts b/src/backend/wine/manager/downloader/utilities.ts
index 809ba11c..46bf6bce 100644
--- a/src/backend/wine/manager/downloader/utilities.ts
+++ b/src/backend/wine/manager/downloader/utilities.ts
@@ -31,6 +31,7 @@ async function fetchReleases({
       .get(url + '?per_page=' + count)
       .then((data) => {
         for (const release of data.data) {
+          console.log(`${release}`)
           const release_data = {} as VersionInfo
           release_data.version = type.includes('Wine')
             ? `Wine-${release.tag_name}`
@@ -38,6 +39,8 @@ async function fetchReleases({
           release_data.type = type
           release_data.date = release.published_at.split('T')[0]
           release_data.disksize = 0
+          console.log(`${release_data.type}`)
+          console.log(`${release_data.date}`)
 
           for (const asset of release.assets) {
             if (asset.name.endsWith('sha512sum')) {
diff --git a/src/backend/wine/manager/utils.ts b/src/backend/wine/manager/utils.ts
index 0dc88a5a..d55682b0 100644
--- a/src/backend/wine/manager/utils.ts
+++ b/src/backend/wine/manager/utils.ts
@@ -41,7 +41,7 @@ async function updateWineVersionInfos(
 
     const repositorys = isMac
       ? [Repositorys.WINECROSSOVER, Repositorys.WINESTAGINGMACOS]
-      : [Repositorys.WINEGE, Repositorys.PROTONGE]
+      : [Repositorys.UMUPROTON, Repositorys.PROTONGE]
 
     await getAvailableVersions({
       repositorys,
diff --git a/src/common/types.ts b/src/common/types.ts
index d765e0e7..a56b957a 100644
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -679,7 +679,7 @@ export interface WikiInfo {
  * Defines from where the version comes
  */
 export type Type =
-  | 'Wine-GE'
+  | 'UMU-Proton'
   | 'Proton-GE'
   | 'Proton'
   | 'Wine-Lutris'
@@ -710,7 +710,7 @@ export interface VersionInfo {
  * Enum for the supported repositorys
  */
 export enum Repositorys {
-  WINEGE,
+  UMUPROTON,
   PROTONGE,
   PROTON,
   WINELUTRIS,
diff --git a/src/frontend/screens/WineManager/index.tsx b/src/frontend/screens/WineManager/index.tsx
index 79a72a59..ff51c8c6 100644
--- a/src/frontend/screens/WineManager/index.tsx
+++ b/src/frontend/screens/WineManager/index.tsx
@@ -46,8 +46,8 @@ export default function WineManager(): JSX.Element | null {
 
   const winege: WineManagerUISettings = {
-    type: 'Wine-GE',
-    value: 'winege',
+    type: 'UMU-Proton',
+    value: 'umuproton',
     enabled: isLinux
   }
   const winecrossover: WineManagerUISettings = {
@@ -62,7 +62,7 @@ export default function WineManager(): JSX.Element | null {
   const [wineManagerSettings, setWineManagerSettings] = useState<
     WineManagerUISettings[]
   >([
-    { type: 'Wine-GE', value: 'winege', enabled: isLinux },
+    { type: 'UMU-Proton', value: 'umuproton', enabled: isLinux },
     { type: 'Proton-GE', value: 'protonge', enabled: isLinux },
     { type: 'Wine-Crossover', value: 'winecrossover', enabled: !isLinux },
     { type: 'Wine-Staging-macOS', value: 'winestagingmacos', enabled: !isLinux }
@@ -71,8 +71,8 @@ export default function WineManager(): JSX.Element | null {
   const getWineVersions = (repo: Type) => {
     let versions = wineDownloaderInfoStore.get('wine-releases', [])
 
-    if (repo.startsWith('Wine-GE')) {
-      versions = versions.filter((version) => version.type === 'Wine-GE')
+    if (repo.startsWith('UMU-Proton')) {
+      versions = versions.filter((version) => version.type === 'UMU-Proton')
       return versions.filter((version) => !version.version.endsWith('LoL'))
     } else {
       return versions.filter((version) => version.type === repo)
@@ -111,13 +111,13 @@ export default function WineManager(): JSX.Element | null {
 
   const wineVersionExplanation = useMemo(() => {
     switch (repository.type) {
-      case 'Wine-GE':
+      case 'UMU-Proton':
         return (
           <div className="infoBox">
             <FontAwesomeIcon icon={faCheck} color={'green'} />
             {t(
-              'wineExplanation.wine-ge',
-              'Wine-GE-Proton is a Wine variant created by Glorious Eggroll. It is the recommended Wine to be used outside Steam. This provides useful logs when troubleshooting.'
+              'wineExplanation.umu-proton',
+              'UMU-Proton is a Proton variant created by Glorious Eggroll based on a stable version of Valve\'s Proton.'
             )}
           </div>
         )
@@ -127,7 +127,7 @@ export default function WineManager(): JSX.Element | null {
             <FontAwesomeIcon icon={faWarning} color={'orange'} />
             {t(
               'wineExplanation.proton-ge',
-              'Proton-GE-Proton is a Proton variant created by Glorious Eggroll. It is meant to be used in Steam, but some games outside Steam may work better with this variant. It provides mostly useless logs for troubleshooting.'
+              'Proton-GE-Proton is a Proton variant created by Glorious Eggroll. It is meant to be used in Steam or UMU, but some games outside Steam may work better with this variant. It provides mostly useless logs for troubleshooting.'
             )}
           </div>
         )
-- 
2.45.2


```

arielj:
> Also, do we want to address removing Wine-GE from the Wine Manager UI in this PR? Those builds are end of life and obsolete in favor of umu with GE-Proton or UMU-Proton.

Not sure about this, I'd rather leave it there for some time just in case and then remove it if it's not needed at all, but I don't know if we can ensure it won't be needed (what if a game doesn't work with proton for any reason?)

gabriele2000:
> (what if a game doesn't work with proton for any reason?)

I'd say, Dark Souls Remastered has trouble recognizing my Dualsense as Xbox controller, something that WINE does by default.
Proton just shits itself and it just cannot FUCKING SEE IT.

Sorry, it pisses me off too much, I have to use WINE-GE (which, as you said, it's end-of-life) just for a single game.

Etaash-mathamsetty:
> > (what if a game doesn't work with proton for any reason?)
> 
> Life the fact that Dark Souls Remastered has trouble recognizing my Dualsense as Xbox controller, something that WINE does by default (you can thank some SDL patches, or something like that). Proton just shits itself and it just cannot FUCKING SEE IT.
> 
> Sorry, it pisses me off too much, I have to use WINE-GE (which, as you said, it's end-of-life) just for a single game.

yea that's a problem with proton since it assumes that steam input exists, which it doesn't outside of steam

gabriele2000:
> yea that's a problem with proton since it assumes that steam input exists, which it doesn't outside of steam

This Steam here Steam there is getting on my nerves.
Why do I have to use Steam for everything, else something doesn't work here and there, even if Proton is WINE with patches?

hirak99:
> yea that's a problem with proton since it assumes that steam input exists, which it doesn't outside of steam

Wasn't UMU supposed to solve this thing? At least that's what my expectation was - instead of publishing a separate custom made wine-ge, which is time and resource consuming to maintain, we can continue to use proton from now on. But that is under the premise that proton+UMU will be on par.

If UMU cannot expose controllers with SDL, what exactly does it do over using proton directly?

gabriele2000:
> If UMU cannot expose controllers with SDL, what exactly does it do over using proton directly?

As far as I know, UMU would use Steam pressure vessel and the runtimes and other stuff, but it wouldn't require Steam to function at 100%, so I think it'd fix this thing...
Technically, you could uninstall Steam and you'd play every game you want without issues (not that you can't right now).

BTW yeah, the README should be even more clearer since I use Proton outside Steam for pretty much 50% of games, and everything works perfectly, except the Dualsense thing with Dark Souls.

dancaer69:
Winetricks should work now? I just tried to use it and it doesn't work. Additionally creates another prefix and after that the game crash(tried with "The Evil Within 2". I'm using chaoti aur repo and umu-launcher updated today(v: 0.1.RC4.r159.91c5f09-1). I also use GE-Proton9-9. When I try to run winetricks(I tried with and without the GAMEID=winetricks-gui) it creates a new prefix instead of using the already configured inside pfx directory(it replaces the symlink pfx directory). The game runs fine again if I delete this "pfx" directory.

Etaash-mathamsetty:
> Winetricks should work now? I just tried to use it and it doesn't work. Additionally creates another prefix and after that the game crash(tried with "The Evil Within 2". I'm using chaoti aur repo and umu-launcher updated today(v: 0.1.RC4.r159.91c5f09-1). I also use GE-Proton9-9. When I try to run winetricks(I tried with and without the GAMEID=winetricks-gui) it creates a new prefix instead of using the already configured inside pfx directory(it replaces the symlink pfx directory). The game runs fine again if I delete this "pfx" directory.

it does not

src/backend/utils/compatibility_layers.ts
```diff
@@ -474,3 +481,27 @@ export function getWineFlagsArray(
   }
   return commandArray
 }
+
+export function getUmuPath() {
+  //TODO: figure out how to use searchForExecutableOnPath here
+
+  const paths = ['/app/share', '/usr/local/share', '/usr/share', '/opt']
```
R1kaB3rN, line 488: In the next release of umu-launcher, these values need to change because the system package is now distributed as a Python module. For this case, just search PATH for `umu-run` instead of `umu_run.py`:

`const paths = ['/app/bin', '/usr/local/bin', '/usr/bin']`

src/backend/launcher.ts
```diff
@@ -228,17 +230,27 @@ async function prepareLaunch(
     }
   }
 
+  if (
+    isUmuSupported(gameSettings.wineVersion.type) &&
+    !isInstalled('umu') &&
```
CommandMC, line 235: This is missing an `await`, causing it to always be reported as `false` (Promises are truthy)
```suggestion
    !(await isInstalled('umu')) &&
```

What IDE are you using @Etaash-mathamsetty? Webstorm warns you about missing `await`s, pretty sure VSCode also does

src/backend/launcher.ts
```diff
@@ -228,17 +230,27 @@ async function prepareLaunch(
     }
   }
 
+  if (
+    isUmuSupported(gameSettings.wineVersion.type) &&
```
CommandMC, line 234: This returns `false` if the UMU path does not exist, so it'll never be downloaded

The following patch resolves this issue:
<details>

```patch
Index: src/backend/utils/compatibility_layers.ts
IDEA additional info:
<+>UTF-8
diff --git a/src/backend/utils/compatibility_layers.ts b/src/backend/utils/compatibility_layers.ts
--- a/src/backend/utils/compatibility_layers.ts	(revision Staged)
+++ b/src/backend/utils/compatibility_layers.ts	(date 1721574962214)
@@ -497,11 +497,14 @@
   return defaultUmuPath
 }
 
-export function isUmuSupported(wineType: WineInstallation['type']): boolean {
-  return (
-    wineType === 'proton' &&
-    GlobalConfig.get().getSettings().experimentalFeatures?.umuSupport !==
-      false &&
-    existsSync(getUmuPath())
-  )
+export function isUmuSupported(
+  wineType: WineInstallation['type'],
+  checkUmuInstalled = true
+): boolean {
+  const umuEnabled =
+    GlobalConfig.get().getSettings().experimentalFeatures?.umuSupport !== false
+  const wineVersionSupported = wineType === 'proton'
+  const umuInstalled = checkUmuInstalled ? existsSync(getUmuPath()) : true
+
+  return umuEnabled && wineVersionSupported && umuInstalled
 }
Index: src/backend/launcher.ts
IDEA additional info:
<+>UTF-8
diff --git a/src/backend/launcher.ts b/src/backend/launcher.ts
--- a/src/backend/launcher.ts	(revision Staged)
+++ b/src/backend/launcher.ts	(date 1721575173784)
@@ -231,7 +231,7 @@
   }
 
   if (
-    isUmuSupported(gameSettings.wineVersion.type) &&
+    isUmuSupported(gameSettings.wineVersion.type, false) &&
     !isInstalled('umu') &&
     isOnline() &&
     getUmuPath() === defaultUmuPath

```
</details>

src/backend/launcher.ts
```diff
@@ -302,14 +316,6 @@ async function prepareWineLaunch(
     }
   }
 
-  // Log warning about Proton
-  if (gameSettings.wineVersion.type === 'proton') {
```
CommandMC, line 306: We should probably still log this if UMU isn't enabled

CommandMC (changes requested):
Auto-installation is broken for two different reasons right now

It'd also be great to get Winetricks working, I'll see what can be done in that regard

src/backend/launcher.ts
Etaash-mathamsetty: Vscode does not

Etaash-mathamsetty: Vscode is probably more useless than a text editor (and eats up 30 gb disk space) so I might just switch to one of those

slagiewka: Try setting up [no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises/) rule. IIRC it might catch that.

Etaash-mathamsetty: Even if it did show the promise it's still not worth 30 gigs lol

slagiewka: Correct. However, my point is that it's good to have automated checks catch issues like so, never mind what editor a contributor uses :wink:

R1kaB3rN:
When running winetricks against a non existent prefix, Heroic will create a prefix with a `pfx` subdirectory then umu will create the prefix inside of it. 

Steps to reproduce:

1. Enable umu in Heroic
2. Add a game
3. Select GE-Proton as the wine version
4. Select an empty directory (e.g., `$HOME/Games/Heroic/Prefixes/foo`) as the wine prefix for the game
5. Click the winetricks button in the game's settings
6. Run `cd $HOME/Games/Heroic/Prefixes/foo` and inspect the directory contents

Or when the user selects a umu created prefix for the game, Heroic will create a `pfx` subdirectory inside the prefix then umu will recreate the prefix in `pfx`. 

Steps to reproduce:

1. Run `GAMEID=0 WINEPREFIX=$HOME/Games/Heroic/Prefixes/foo PROTONPATH=GE-Proton umu-run ""` to create a  wine prefix.
2. Enable umu in Heroic
3. Add a game
4. Select GE-Proton as the wine version
5. Select a prefix created by umu as the wine prefix for the game
7. Click the winetricks button in the game's settings
8. Run `cd $HOME/Games/Heroic/Prefixes/foo` and inspect the directory contents

In both of these cases, can we have Heroic not create the `pfx` subdirectory when umu is enabled?

R1kaB3rN:
Also, Heroic is currently running winetricks inside the container when getting the dll or font list which has the risk of updating the runtime.

In this case, I believe it should be safe to simply run winetricks outside the container when getting the font or dll list and only running winetricks inside the container when the user selects to install a verb. But if this is too much work, umu can have another environment variable to allow skipping runtime updates and Heroic can just set it in this case.

Etaash-mathamsetty:
> When running winetricks against a non existent prefix, Heroic will create a prefix with a `pfx` subdirectory then umu will create the prefix inside of it.
> 
> Steps to reproduce:
> 
>     1. Enable umu in Heroic
> 
>     2. Add a game
> 
>     3. Select GE-Proton as the wine version
> 
>     4. Select an empty directory (e.g., `$HOME/Games/Heroic/Prefixes/foo`) as the wine prefix for the game
> 
>     5. Click the winetricks button in the game's settings
> 
>     6. Run `cd $HOME/Games/Heroic/Prefixes/foo` and inspect the directory contents
> 
> 
> Or when the user selects a umu created prefix for the game, Heroic will create a `pfx` subdirectory inside the prefix then umu will recreate the prefix in `pfx`.
> 
> Steps to reproduce:
> 
>     1. Run `GAMEID=0 WINEPREFIX=$HOME/Games/Heroic/Prefixes/foo PROTONPATH=GE-Proton umu-run ""` to create a  wine prefix.
> 
>     2. Enable umu in Heroic
> 
>     3. Add a game
> 
>     4. Select GE-Proton as the wine version
> 
>     5. Select a prefix created by umu as the wine prefix for the game
> 
>     6. Click the winetricks button in the game's settings
> 
>     7. Run `cd $HOME/Games/Heroic/Prefixes/foo` and inspect the directory contents
> 
> 
> In both of these cases, can we have Heroic not create the `pfx` subdirectory when umu is enabled?

I pushed a potential fix, could you retest ?

R1kaB3rN:
> I pushed a potential fix, could you retest ?

Just retested and, yes, the bug is fixed.

R1kaB3rN:
Since https://github.com/Open-Wine-Components/umu-launcher/commit/7e00027c1bf9f4402318ee0530d95d34fbbc72c8 updates to the runtime can be disabled. 

So to prevent the runtime from being updated when using winetricks, just set `UMU_RUNTIME_UPDATE=0`.

R1kaB3rN:
Also, because Heroic is calling umu separately for the font and dll list, it looks like it can lead to the runtime being downloaded and installed multiple times if umu is not already installed for the user.

Steps to reproduce:

1. Delete $HOME/.local/share/umu
2. Enable umu in Heroic
3. Add a game
4. Select GE-Proton as the wine version
5. Select an empty directory (e.g., $HOME/Games/Heroic/Prefixes/foo) as the wine prefix for the game
6. Click the winetricks button in the game's settings
7. Watch the umu runtime being downloaded multiple times in the winetricks window

Like I mentioned, I think this can be avoided by not calling winetricks for the fonts and dlls through umu. Or maybe to just get dll or font list in sequence instead async?

Etaash-mathamsetty:
> Since https://github.com/Open-Wine-Components/umu-launcher/commit/7e00027c1bf9f4402318ee0530d95d34fbbc72c8 updates to the runtime can be disabled. 
> 
> So to prevent the runtime from being updated when using winetricks, just set `UMU_RUNTIME_UPDATE=0`.

I don't understand why we would need to disable runtime updates?

Etaash-mathamsetty:
> Also, because Heroic is calling umu separately for the font and dll list, it looks like it can lead to the runtime being downloaded and installed multiple times if umu is not already installed for the user.
> 
> Steps to reproduce:
> 
> 1. Delete $HOME/.local/share/umu
> 2. Enable umu in Heroic
> 3. Add a game
> 4. Select GE-Proton as the wine version
> 5. Select an empty directory (e.g., $HOME/Games/Heroic/Prefixes/foo) as the wine prefix for the game
> 6. Click the winetricks button in the game's settings
> 7. Watch the umu runtime being downloaded multiple times in the winetricks window
> 
> Like I mentioned, I think this can be avoided by not calling winetricks for the fonts and dlls through umu. Or maybe to just get dll or font list in sequence instead async?

Good find, I will fix it heroic side but it might be worth including some sort of lock file system in umu itself

Etaash-mathamsetty:
I just checked the code, they aren't running in parrallel

R1kaB3rN:
> I don't understand why we would need to disable runtime updates?

The runtime is an environment to run games, so I think the runtime should only be updated during the process of launching a game, not when running winetricks verbs. I think it would be pretty annoying/confusing if the runtime gets updated when the user just wants to modify the wine prefix.

R1kaB3rN:
A file lock system was merged in https://github.com/Open-Wine-Components/umu-launcher/commit/a505a90eaef60182516acfd86c6e643fc0e56197

imLinguin (approved):
The feature works as expected

PR #3727, opened 2024-04-28
comet support

Etaash-mathamsetty:
depends on #3849 (do not merge until that gets merged)


Etaash-mathamsetty:
doesn't cause games to crash and allows for multiplayer functionality might be a good enough initial version

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -186,7 +185,6 @@ export default React.memo(function GamePage(): JSX.Element | null {
 
         if (
           runner !== 'sideload' &&
-          !is_installed &&
```
arielj, line 189: is this intentional? it's not clear why this is removed (I understand this makes sure the install info is not fetch when the game is already installed)

src/frontend/screens/Settings/sections/SystemInfo/software.tsx
```diff
@@ -51,6 +56,14 @@ function SoftwareInfo({ software }: Props) {
             }
           )}
           <br />
+          {t(
```
arielj, line 59: I think the translation file was not updated with this change

src/backend/storeManagers/gog/presence.ts
```diff
@@ -0,0 +1,83 @@
...
+  CURRENT_GAME = game
+}
+
+async function setPresence() {
+  try {
+    if (!GOGUser.isLoggedIn() || !isOnline) return
+    const credentials = await GOGUser.getCredentials()
+    if (!credentials) {
+      return
+    }
```
arielj, line 28:
```suggestion
    if (!credentials) return
```

to match the syntax in line 24 doing a single line if/return

src/backend/storeManagers/gog/presence.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const payload: PresencePayload = {
+      application_type: 'Heroic Games Launcher',
+      force_update: false,
+      presence: 'online',
+      version: app.getVersion(),
+      game_id: undefined
+    }
+
+    if (CURRENT_GAME.length > 0) {
```
arielj, line 42:
```suggestion
    if (CURRENT_GAME !== '') {
```

src/backend/storeManagers/gog/games.ts
```diff
@@ -667,13 +670,28 @@ export async function launch(
...
 
-  sendGameStatusUpdate({
-    appName,
-    runner: 'gog',
-    status: 'playing'
-  })
+  let child = undefined
+
+  if (userData && userData.username) {
+    const path = getCometBin()
```
arielj, line 680: I'm thinking we should add this under an experimental flag feature since it seems to have caused some issues from what I saw in discord

or maybe instead of a feature flag we could make it a game setting? that would give way more control to disable it as needed if some game fails but others don't after the next release

imLinguin:
Things to address
- [ ] Make comet a experimental feature (enabled by default)
- [ ] Download and install comet's `dummy-service` (possibly with versioning in case we needed to update it in the future)
- [ ] Add an option to disable GOG online presence (maybe merge it with playtime option?)

src/backend/launcher.ts
```diff
@@ -357,14 +359,34 @@ async function prepareWineLaunch(
...
+      runtimePath,
+      'comet_dummy_service',
+      'install-dummy-service.bat'
+    )
+    if (existsSync(installerScript)) {
+      await runWineCommand({
+        commandParts: [],
+        gameSettings,
+        protonVerb: 'runinprefix', // Prefix was verified before this, it's safe
+        skipPrefixCheckIKnowWhatImDoing: true
```
CommandMC, line 381: Is there any reason you're skipping this here? Note that the function won't automatically run `wineboot -i` if this is not set, it'll first check whether prefix files are actually missing

imLinguin: we are verifying the prefix before this, I think it's safe

Etaash-mathamsetty: it's probably safe, but it just seems unnecessary to skip it

imLinguin: right, let's remove it

imLinguin (approved):

PR #3742, opened 2024-05-09
Adapt amazon games metadata to new design

imLinguin:
This makes Amazon games look better on new design, since now we are using Amazon's background and horizontal image. Additionally I made it so Heroic will display genres incoming from given provider whenever possible (GOG, Amazon).

obraz
obraz

### DIFF
--- a/src/backend/storeManagers/gog/library.ts
+++ b/src/backend/storeManagers/gog/library.ts
@@ -943,10 +943,14 @@ export async function gogToUnifiedInfo(
     // @ts-expect-error TODO: Handle this somehow
     return {}
   }
+  const background = info.game.background?.url_format
+    .replace('{formatter}', '')
+    .replace('{ext}', 'webp')
 
-  const art_cover = info.game?.logo?.url_format
-    ?.replace('{formatter}', '')
-    .replace('{ext}', 'jpg')
+  const art_cover =
+    info.game?.logo?.url_format
+      ?.replace('{formatter}', '')
+      .replace('{ext}', 'jpg') ?? background
 
   const icon = (
     info.game?.square_icon.url_format || info.game?.icon?.url_format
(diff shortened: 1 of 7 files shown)


src/backend/storeManagers/gog/library.ts
```diff
@@ -963,14 +967,13 @@ export async function gogToUnifiedInfo(
...
-      .replace('{formatter}', '')
-      .replace('{ext}', 'webp'),
+    art_background: background,
     cloud_save_enabled: false,
     art_icon: icon,
     extra: {
       about: { description: info.summary['*'], shortDescription: '' },
-      reqs: []
+      reqs: [],
+      genres: info.game.genres.map((genre) => genre.name['*'])
```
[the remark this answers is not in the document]
imLinguin: there is only `*` and `en-US`. Not much we can use here tbh. If we wanted we could add translations based on genre keys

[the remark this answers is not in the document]
CommandMC: Ah, let's leave it like this for now then

CommandMC (approved):

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -381,7 +381,13 @@ export default React.memo(function GamePage(): JSX.Element | null {
                     <DotsMenu gameInfo={gameInfo} handleUpdate={handleUpdate} />
                   </div>
                   <div className="infoWrapper">
-                    <Genres genres={wikiInfo?.pcgamingwiki?.genres || []} />
+                    <Genres
+                      genres={
+                        gameInfo.extra?.genres ||
```
arielj, line 386: could this be an empty array? because in that case it `[] || wikiInfo...` would still be `[]` even if we have genres from pc gaming wiki and for some reason the gog or amazon info has no genres

imLinguin: good point

imLinguin: As far as I can see the array is empty for some GOG titles, that aren't games and in case of `Galaxy Integration Demo`. Amazon seems to always have genres. I'd say we are safe

Issue #3691, opened 2024-04-09
Allow to pass arguments for scripts to run before and after a game is launched


casasfernando:
### Problem description

I need to use a pre-launch/post-exit scripts (https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3565 - https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2507) with my games to be able to do some work for the games to run correctly in my environment and then clean up after the game exits.
At the moment is only possible to select the pre-launch/post-exit script but I couldn't find a way to also pass some arguments to the scripts, like the game name or game prefix path, or any other arbitrary one.
Without this I will have to keep separate scripts for each game since all would require a little twist specific to the game.

### Feature description

I would like to be able to pass arguments to the scripts so I can only have on pre-launch and one post-launch or even just one script that can run differently accordingly to the arguments passed that would be different if the script is executed before launching the game or after it exits.
In terms of implementation it can be something similar to what Windows does with shortcuts, that means full script path in double quotes, followed by arguments. E.g.: "/path/to/pre-launch/script.sh" arg0 arg1 argX
Another, I think better, option would be to have something similar to the wrapper command in the same settings page, where there are two fields: wrapper and arguments. In this case it would be script and arguments.

### Alternatives

I tried to check if the application is passing any game related information in the environment variables or default arguments when running the scripts but it doesn't.
Having some information through environment variables or even default arguments, containing the game name and/or the game prefix path would be a nice to have.

casasfernando:
I've looked into the code and came up with the following patch to pass some additional information (game title and executable path) to the pre/post launch script using environment variables. It gets the job done and it is a very small code change:

```
diff --git a/src/backend/launcher.ts b/src/backend/launcher.ts
index 2835cfff..5f7f8ad1 100644
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1368,7 +1368,9 @@ async function runScriptForGame(
   scriptPath: string
 ): Promise<boolean | string> {
   return new Promise((resolve, reject) => {
-    const child = spawn(scriptPath, { cwd: gameInfo.install.install_path })
+    var scriptEnv = { GAMEINFO_TITLE: gameInfo.title , GAMEINFO_EXEC: gameInfo.install.executable };
+    Object.assign(scriptEnv, process.env);
[...]

PR #3748, opened 2024-05-12
[Feat] Pass additional game information to the pre-post launch script using environment variables

casasfernando:
This PR goal is to pass additional game information (title and executable path) to the pre-post launch script using environment variables. This would allow to use the same script with different games, taking actions based on the game title value or executable path.
Tries to fulfill the alternative solution proposed in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3691.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1368,7 +1368,7 @@ async function runScriptForGame(
   scriptPath: string
 ): Promise<boolean | string> {
   return new Promise((resolve, reject) => {
-    var scriptEnv = { GAMEINFO_TITLE: gameInfo.title , GAMEINFO_EXEC: gameInfo.install.executable };
+    let scriptEnv = { GAMEINFO_TITLE: gameInfo.title , GAMEINFO_EXEC: gameInfo.install.executable };
     Object.assign(scriptEnv, process.env);
     const child = spawn(scriptPath, { cwd: gameInfo.install.install_path , env: scriptEnv })
 
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1370,6 +1370,7 @@ async function runScriptForGame(
   return new Promise((resolve, reject) => {
     const scriptEnv = {
       HEROIC_GAMEINFO_EXEC: gameInfo.install.executable,
+      HEROIC_GAMEINFO_RUNNER: gameInfo.runner,
       HEROIC_GAMEINFO_TITLE: gameInfo.title
     }
     Object.assign(scriptEnv, process.env)
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1368,9 +1368,15 @@ async function runScriptForGame(
   scriptPath: string
 ): Promise<boolean | string> {
   return new Promise((resolve, reject) => {
-    const scriptEnv = { GAMEINFO_TITLE: gameInfo.title , GAMEINFO_EXEC: gameInfo.install.executable };
-    Object.assign(scriptEnv, process.env);
-    const child = spawn(scriptPath, { cwd: gameInfo.install.install_path , env: scriptEnv })
+    const scriptEnv = {
+      GAMEINFO_TITLE: gameInfo.title,
+      GAMEINFO_EXEC: gameInfo.install.executable
+    }
+    Object.assign(scriptEnv, process.env)
+    const child = spawn(scriptPath, {
+      cwd: gameInfo.install.install_path,
+      env: scriptEnv
+    })
 
     child.stdout.on('data', (data) => {
       appendGamePlayLog(gameInfo, data.toString())
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1368,7 +1368,9 @@ async function runScriptForGame(
   scriptPath: string
 ): Promise<boolean | string> {
   return new Promise((resolve, reject) => {
-    const child = spawn(scriptPath, { cwd: gameInfo.install.install_path })
+    var scriptEnv = { GAMEINFO_TITLE: gameInfo.title , GAMEINFO_EXEC: gameInfo.install.executable };
+    Object.assign(scriptEnv, process.env);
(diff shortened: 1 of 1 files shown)


flavioislima:
   - - -
   - - -

casasfernando:

casasfernando:
recheck

OlegAckbar:
I've just tried it and I'm pretty much content with the result. Now my scripts can create folders for recorded videos based on `GAMEINFO_TITLE` environment variable. Hope it will get merged so I won't have to compile Heroic every time.
изображение

imLinguin:
While I like the idea I see some issues that may occur.

- The game title may be set to ??? for games from Amazon when their API gets a stroke
- The `gameInfo.install.executable` is valid only for Epic, as other providers handle game launching based on the manifest file from the install directory.

I propose to expose variables like we did for wrappers. Making it clear that they come from Heroic (having a `HEROIC_` prefix for example).
Additionaly maybe add variables that inform about the runner and game id.

casasfernando:
Thanks for your comment @imLinguin.
It is true that I wrote this patch mainly for my use case (sideloaded Windows games running with Wine) so I didn't think about the points above.

- I like the idea of changing the variables name and prefix them with `HEROIC_` so if this the way to go I will change it. Just let me know.
- `gameInfo.install.executable` is not only valid for Epic, as mentioned above, sideloaded games must(?) have a title. Please correct me if I'm wrong.
- I thought about adding variables like for other information like the game id, the runner or the prefix path but I wanted to keep the PR as small as possible while at the same time maximizing the value it adds. The runner and the prefix path for example will require to pass the GameSettings object to the function (right now only GameInfo is passed) so it will require a slightly bigger change while I wasn't sure it worth it. The prefix path can be obtained from the executable path and right now I can't think of a good use for the game id (again my use case of Heroic is limited to running sideloaded Windows games). In any case I thought this was a good start and more variables could be added as needed, in the future.

Last but not least the variables are passed as strings so it should be fine if they are null/empty or have unexpected values. If the script is going to use them I would like to think that is up to it for validating the data but I would be very interested into hearing others thoughts on thise.

casasfernando:
I've checked the gameInfo object again and this is the information available in my case. I will highlight the properties that I think can be useful for pre-post launch script, but please let me know if there are other as well for other use cases and I can add them to the variables list:

- runner
- app_name
- **title**
- install:
    - **executable**
    - platform
    - is_dlc
- folder_name
- art_cover
- is_installed
- art_square
- canRunOffline
- browserUrl
- customUserAgent
- launchFullScreen

Thanks

casasfernando:
@imLinguin I just pushed two more commits, one to add the `HEROIC_` prefix to the variables and another to add a variable for the runner. Not sure which gameInfo property will match the game id. Maybe the app_name?

imLinguin:
Yes the app_name is a game id.

casasfernando:
> Yes the app_name is a game id.

Thanks for clarifying. Should I add this on a new commit? And if yes, should be named `HEROIC_GAMEINFO_APP_NAME` or `HEROIC_GAMEINFO_ID` instead?
I think it should be the former so it's easier/faster to know where the value is coming from within Heroic.

casasfernando:
recheck

arielj:
> > Yes the app_name is a game id.
> 
> Thanks for clarifying. Should I add this on a new commit? And if yes, should be named `HEROIC_GAMEINFO_APP_NAME` or `HEROIC_GAMEINFO_ID` instead? I think it should be the former so it's easier/faster to know where the value is coming from within Heroic.

I would use `HEROIC_APP_NAME`.

Maybe we can also send the prefix for games that have a prefix? (HEROIC_GAME_PREFIX for example?) I imagine it could be useful if someone needs to reference it after a game is closed (like if you want to copy some file or something)

casasfernando:
Thanks @arielj for the input. I like the idea of adding the prefix info as well.
Will prepare a new commit for the suggestions above and will also add a variable to indicate if the script is running before or after the game. This will further facilitate using one script for both stages if the user wishes to do that.

src/backend/launcher.ts
```diff
@@ -1365,10 +1365,27 @@ async function runAfterLaunchScript(
  */
 async function runScriptForGame(
   gameInfo: GameInfo,
-  scriptPath: string
+  gameSettings: GameSettings,
+  scriptStage: string
```
arielj, line 1369:
```suggestion
  scriptStage: 'before' | 'after'
```

src/backend/launcher.ts
```diff
@@ -1365,10 +1365,27 @@ async function runAfterLaunchScript(
...
 ): Promise<boolean | string> {
   return new Promise((resolve, reject) => {
-    const child = spawn(scriptPath, { cwd: gameInfo.install.install_path })
+    const scriptPath =
+      scriptStage === 'before'
+        ? gameSettings.beforeLaunchScriptPath
+        : gameSettings.afterLaunchScriptPath
```
arielj, line 1375:
```ts
const scriptPath = gameSettings[`${scriptStage}LaunchScriptPath`]
```

casasfernando: Thanks for the feedback, this is more simple and cleaner. Just pushed a new commit to address it.

casasfernando:
@arielj any chance this will be merged before the next release? Thanks

src/backend/launcher.ts
```diff
@@ -1365,10 +1365,24 @@ async function runAfterLaunchScript(
...
   gameInfo: GameInfo,
-  scriptPath: string
+  gameSettings: GameSettings,
+  scriptStage: 'before' | 'after'
 ): Promise<boolean | string> {
   return new Promise((resolve, reject) => {
-    const child = spawn(scriptPath, { cwd: gameInfo.install.install_path })
+    const scriptPath = gameSettings[`${scriptStage}LaunchScriptPath`]
+    const scriptEnv = {
+      HEROIC_GAME_APP_NAME: gameInfo.app_name,
```
Etaash-mathamsetty, line 1374: adding `...process.env` here would eliminate the need for the object.assign later is cleaner imo

arielj (approved):

Issue #3705, opened 2024-04-19
Auto Install/Update DXVK on Prefix setting is not compatible with 32-bit prefixes


casasfernando:
### Describe the bug

When checking the `Auto Install/Update DXVK on Prefix` in the game settings options Heroic tries to install DXVK to the Wine prefix, but is not checking if the prefix is 32-bit or 64-bit before doing so.
As a result, when using this option with a 32-bit Wine prefix, DXVK 64-bit libraries are copied to the system32 directory inside the prefix and an error is logged while trying to copy the 32-bit libraries to syswow64, because obviously this directory doesn't exist on 32-bit prefixes:

```
(19:52:37) ERROR:   [DXVKInstaller]:    Error when copying d3d9.dll Error: ENOENT: no such file or directory, copyfile '/home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/dxvk/dxvk-2.3.1/x32/d3d9.dll' -> '/home/deck/Games/Heroic/Prefixes/default/Call of Duty 2/drive_c/windows/syswow64/d3d9.dll'
(19:52:37) ERROR:   [DXVKInstaller]:    Error when copying dxgi.dll Error: ENOENT: no such file or directory, copyfile '/home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/dxvk/dxvk-2.3.1/x32/dxgi.dll' -> '/home/deck/Games/Heroic/Prefixes/default/Call of Duty 2/drive_c/windows/syswow64/dxgi.dll'
(19:52:37) ERROR:   [DXVKInstaller]:    Error when copying d3d10core.dll Error: ENOENT: no such file or directory, copyfile '/home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/dxvk/dxvk-2.3.1/x32/d3d10core.dll' -> '/home/deck/Games/Heroic/Prefixes/default/Call of Duty 2/drive_c/windows/syswow64/d3d10core.dll'
(19:52:37) ERROR:   [DXVKInstaller]:    Error when copying d3d11.dll Error: ENOENT: no such file or directory, copyfile '/home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/tools/dxvk/dxvk-2.3.1/x32/d3d11.dll' -> '/home/deck/Games/Heroic/Prefixes/default/Call of Duty 2/drive_c/windows/syswow64/d3d11.dll'
```

[...]

PR #3749, opened 2024-05-12
[Fix] Support DXVK installation on 32-bit prefixes

casasfernando:
Fix support for DXVK installation on 32-bit prefixes.
Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3705.

---

### DIFF
--- a/src/backend/tools/index.ts
+++ b/src/backend/tools/index.ts
@@ -209,6 +209,12 @@ export const DXVK = {
 
     tool = isMac ? 'dxvk-macOS' : tool
 
+    const is64bitPrefix = existsSync(`${winePrefix}/drive_c/windows/syswow64`)
+
+    if (!is64bitPrefix) {
+      logWarning('Installing DXVK on a 32-bit prefix!', LogPrefix.DXVKInstaller)
+    }
+
     if (!existsSync(`${toolsPath}/${tool}/latest_${tool}`)) {
       logWarning('dxvk not found!', LogPrefix.DXVKInstaller)
       await DXVK.getLatest()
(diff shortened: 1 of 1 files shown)


flavioislima:
@casasfernando thanks for your contribution!

Can you add instructions on how to test this, examples of games, etc?

casasfernando:
Thanks for reviewing this @flavioislima .

Example of games that would need or prefer a 32-bit Wine prefix:
- All the "classic" Need for Speed games (II SE, III Hot Pursuit, High Stakes, etc) that were meant run on Windows 95/98 (Wine 64-bit prefixes can't be configured to present themselves as Windows 98).
- Virtua Fighter PC and Virtua Fighter 2
- Other games that require Windows 95/98 and would fail to install or run in newer Windows versions like XP

How to test (with the patch in place):
1- Add the following environment variable in Heroic game settings before adding the new game to the library `WINEARCH=win32`. This will make the new Wine prefix to be 32-bit.
2- Go to Library -> Add game.
3- Fill in the game details and run the installer along any other prerequisites (e.g.: nglide)
4- Select the game executable
5- Run the game and it should be using DXVK

Let me know if this is the information you was looking for.
If you would like to save some time I can provide a default 32-bit Wine prefix created with Heroic (without any games) that would be enough to test the PR and confirm that with the patch in place the right architecture libraries are copied to system32 when DXVK installation is executed.

CommandMC (approved):
I don't like how much duplication is in this function, but resolving that is not a task for this PR

casasfernando:
> I don't like how much duplication is in this function, but resolving that is not a task for this PR

After/if this PR is merged I was planning to work on two additional ones. One to fix DXVK uninstallation and another one to clean up the function code in general and reduce duplication. But I prefer to address one thing at a time and keep changes as small and clean as possible.

Issue #3707, opened 2024-04-19
Auto Install/Update DXVK on Prefix setting is not correctlly removing DXVK when unchecked


casasfernando:
### Describe the bug

When trying to disable/uninstall DXVK from a Wine prefix by unchecking the `Auto Install/Update DXVK on Prefix` setting only the libraries overrides are removed from the Wine prefix configuration but the DXVK libraries in system32 are not. Due to this, next time the prefix is used Wine will continue to load DXVK DLLs instead of stock Wine libraries.

### Add logs

```text
(19:41:37) INFO:    [Legendary]:        Legendary location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary
(19:41:37) INFO:    [Gog]:              GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl
(19:41:37) INFO:    [Connection]:       Connectivity: check-online
(19:41:37) INFO:    [Connection]:       Pinging external endpoints
(19:41:37) INFO:    [Gog]:              Checking for existing gog manifests
(19:41:37) INFO:    [Backend]:          DRM module staus {
  "oimompecagnajdejgnnjijobebaeigek": {
    "status": "new",
    "title": "Widevine Content Decryption Module",
    "version": "4.10.2710.0"
  }
}
(19:41:37) WARNING: [Backend]:          Failed to register protocol with OS.
(19:41:37) INFO:    [Legendary]:        Running command: LEGENDARY_CONFIG_PATH=/home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/legendaryConfig/legendary /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/legendary --version
(19:41:37) INFO:    [Gog]:              Running command: GOGDL_CONFIG_PATH=/home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/gogdlConfig /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json --version
(19:41:37) INFO:    [Nile]:             Running command: NILE_CONFIG_PATH=/home/deck/.var/app/com.heroicgameslauncher.hgl/config/heroic/nile_config /app/bin/heroic/resources/app.asar.unpacked/build/bin/linux/nile --version
[...]

PR #3760, opened 2024-05-17
[Fix] Fix DXVK uninstallation from prefix

casasfernando:
Fix DXVK uninstallation by restoring Wine/Proton DLLs after DXVK DLL overrides are removed. Otherwise `wineboot -u` will have to be executed to achieve the same result.
Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3707 .
---

### DIFF
--- a/src/backend/tools/index.ts
+++ b/src/backend/tools/index.ts
@@ -295,30 +295,29 @@ export const DXVK = {
       logInfo('Removing DXVK DLLs', LogPrefix.DXVKInstaller)
 
       // removing DXVK dlls
+      let dllsToRemove
       if (is64bitPrefix) {
-        dlls32.forEach((dll) => {
-          rm(`${winePrefix}/drive_c/windows/syswow64/${dll}`, (err) => {
-            if (err) {
-              logError([`Error removing ${dll}`, err], LogPrefix.DXVKInstaller)
-            }
-          })
-        })
-        dlls64.forEach((dll) => {
-          rm(`${winePrefix}/drive_c/windows/system32/${dll}`, (err) => {
-            if (err) {
-              logError([`Error removing ${dll}`, err], LogPrefix.DXVKInstaller)
-            }
-          })
-        })
+        dllsToRemove = dlls64.map(
+          (dll) => `${winePrefix}/drive_c/windows/system32/${dll}`
+        )
+        dllsToRemove.concat(
+          dlls32.map((dll) => `${winePrefix}/drive_c/windows/syswow64/${dll}`)
+        )
       } else {
-        dlls32.forEach((dll) => {
-          rm(`${winePrefix}/drive_c/windows/system32/${dll}`, (err) => {
-            if (err) {
-              logError([`Error removing ${dll}`, err], LogPrefix.DXVKInstaller)
-            }
-          })
-        })
+        dllsToRemove = dlls32.map(
+          (dll) => `${winePrefix}/drive_c/windows/system32/${dll}`
+        )
       }
+      dllsToRemove.forEach((dllFile) => {
+        rm(dllFile, (err) => {
+          if (err) {
+            logError(
+              [`Error removing ${dllFile}`, err],
+              LogPrefix.DXVKInstaller
+            )
+          }
+        })
+      })
 
       // Restore stock Wine libraries
       logInfo('Restoring Wine stock DLLs', LogPrefix.DXVKInstaller)
--- a/src/backend/tools/index.ts
+++ b/src/backend/tools/index.ts
@@ -295,75 +295,40 @@ export const DXVK = {
       logInfo('Removing DXVK DLLs', LogPrefix.DXVKInstaller)
 
       // removing DXVK dlls
-      const lib32Path =
-        gameSettings.wineVersion.lib32 === undefined
-          ? ''
-          : gameSettings.wineVersion.lib32
-      const wineLib32Path = lib32Path.replace('~', userHome)
-      if (wineLib32Path === '') {
-        logError(
-          'invalid 32-bit library source directory! Will not remove DXVK DLLs',
-          LogPrefix.DXVKInstaller
-        )
-        return false
-      }
       if (is64bitPrefix) {
-        const libPath =
-          gameSettings.wineVersion.lib === undefined
-            ? ''
-            : gameSettings.wineVersion.lib
-        const wineLibPath = libPath.replace('~', userHome)
-        if (wineLibPath === '') {
-          logError(
-            'invalid 64-bit library source directory! Will not remove DXVK DLLs',
-            LogPrefix.DXVKInstaller
-          )
-          return false
-        }
         dlls32.forEach((dll) => {
-          copyFile(
-            `${wineLib32Path}/wine/i386-windows/${dll}`,
-            `${winePrefix}/drive_c/windows/syswow64/${dll}`,
-            (err) => {
-              if (err) {
-                logError(
-                  [`Error when copying ${dll}`, err],
-                  LogPrefix.DXVKInstaller
-                )
-              }
+          rm(`${winePrefix}/drive_c/windows/syswow64/${dll}`, (err) => {
+            if (err) {
+              logError([`Error removing ${dll}`, err], LogPrefix.DXVKInstaller)
             }
-          )
+          })
         })
         dlls64.forEach((dll) => {
-          copyFile(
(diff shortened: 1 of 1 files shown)


Etaash-mathamsetty:
wine will use wined3d if the overrides were removed properly, something is not working right with your wine installation as I have tested this hundreds of times at this point

casasfernando:
I’m not doing anything special to be honest. It’s also documented here: https://github.com/doitsujin/dxvk#how-to-use 🤷🏻‍♂️

For what it worth I found this issue while playing on a PC with an Intel iGPU which doesn’t have great Vulkan support. Only way to stop using DXVK and go back to WineD3D was to also restore the DLLs.
On the Steam Deck for example the problem went unnoticed for me because the APU will handle both fine.

Etaash-mathamsetty:
> I’m not doing anything special to be honest. It’s also documented here: [doitsujin/dxvk#how-to-use](https://github.com/doitsujin/dxvk#how-to-use) 🤷🏻‍♂️
> 
> For what it worth I found this issue while playing on a PC with an Intel iGPU which doesn’t have great Vulkan support. Only way to stop using DXVK and go back to WineD3D was to also restore the DLLs. On the Steam Deck for example the problem went unnoticed for me because the APU will handle both fine.

yes but I have carefully checked with the toggle on and off, wine will load dlls from where its dlls are stored not the prefix. I used mangohud and also checked logs

casasfernando:
I understand and thanks for the feedback @Etaash-mathamsetty . Maybe it is something wrong with my setup. I still think this PR adds some value and it certainly helps in my scenario.
Of course is up to the team to accept it or not. In any case I would also really appreciate feedback on the PR code itself if any since I’m always looking to improve my coding skills.

Thanks.

src/backend/tools/index.ts
```diff
@@ -291,6 +291,83 @@ export const DXVK = {
...
+        const wineLibPath = libPath.replace('~', userHome)
+        if (wineLibPath === '') {
+          logError(
+            'invalid 64-bit library source directory! Will not remove DXVK DLLs',
+            LogPrefix.DXVKInstaller
+          )
+          return false
+        }
+        dlls32.forEach((dll) => {
+          if (!isMac) {
```
Etaash-mathamsetty, line 324: btw this !isMac check is wrong, i will be making a PR to fix this in the rest of the file soon

casasfernando: Thanks for highlighting this, just pushed a new commit to address it.

casasfernando:
@arielj , @Etaash-mathamsetty , @flavioislima any chance this will be reviewed/merge before the next release?

Thanks

arielj:
I'm not really familiar with how dxvk install/uninstall works or should work so I'll let others review this, I'm really not versed in this topic

flavioislima:
> @arielj , @Etaash-mathamsetty , @flavioislima any chance this will be reviewed/merge before the next release?
> 
> Thanks

This used to be like this at some point but then we started to using `wineboot` and I dont remember the reason. @CommandMC perhaps knows the reason.

Etaash-mathamsetty:
Running wineboot -u is not good since it will reset all winetricks overrides

CommandMC:
As far as I can tell, Wineboot only replaces *missing* DLLs, it does not undo modifications done to existing ones
To test, I overwrote `drive_c/windows/system32/d3d9.dll` with nothing (`echo ""` piped into it) and ran `wineboot -u`. The file stayed the same. After that I removed the file completely. `wineboot -u` now restored it to what comes with Wine.

The approach used in this PR relies on a specific structure of Wine versions, which is not given when using, for example, the system's Wine version
Running `wineboot -u` takes about 5 seconds on my machine. Sure, not *great*, but reaching into Wine like this just to save a few seconds in an unlikely scenario doesn't sound like a good idea to me.

Etaash-mathamsetty:
> As far as I can tell, Wineboot only replaces _missing_ DLLs, it does not undo modifications done to existing ones To test, I overwrote `drive_c/windows/system32/d3d9.dll` with nothing (`echo ""` piped into it) and ran `wineboot -u`. The file stayed the same. After that I removed the file completely. `wineboot -u` now restored it to what comes with Wine.
> 
> The approach used in this PR relies on a specific structure of Wine versions, which is not given when using, for example, the system's Wine version Running `wineboot -u` takes about 5 seconds on my machine. Sure, not _great_, but reaching into Wine like this just to save a few seconds in an unlikely scenario doesn't sound like a good idea to me.

oh I thought it copies every single dll again, yeah then wineboot -u is the best option

casasfernando:
Should I go ahead and modify the PR to use wineboot instead of manually copying the stock DLLs back after removing DXVK ones or do you prefer to discard it completely and tackle this differently?

Etaash-mathamsetty:
> Should I go ahead and modify the PR to use wineboot instead of manually copying the stock DLLs back after removing DXVK ones or do you prefer to discard it completely and tackle this differently?

wineboot -u would be the best

casasfernando:
@CommandMC , @flavioislima , @Etaash-mathamsetty I just pushed a new commit using wineboot to restore the stock libraries. I must say that the code is much simpler this way.
Looking forward your review and comments.

Thanks

src/backend/tools/index.ts
```diff
@@ -291,6 +291,44 @@ export const DXVK = {
...
+        })
+      } else {
+        dlls32.forEach((dll) => {
+          rm(`${winePrefix}/drive_c/windows/system32/${dll}`, (err) => {
+            if (err) {
+              logError([`Error removing ${dll}`, err], LogPrefix.DXVKInstaller)
+            }
+          })
+        })
+      }
```
arielj, line 321: maybe this can be change to use the if/else stuff to set an array of paths to delete and then loop through the array to delete things?

something like:

```ts
const dllsPathsToDelete = dlls32.map((dll) => `${winePrefix}/drive_c/windows/system32/${dll}`)
if (is64bitPrefix) dllsPathsToDelete.concat(dlls64.map((dll) => `${winePrefix}/drive_c/windows/syswow64/${dll}`))

dllsPathsToDelete.forEach((dllPath) => {
    rm(dllPath, (err) => {
      if (err) {
        logError([`Error removing ${dllPath}`, err], LogPrefix.DXVKInstaller)
      }
    })
}
```

casasfernando: I was just trying to change the original code as little as possible to reduce the chance of introducing a new bug but if you guys prefer to try to optimise/clean up a little bit in the same PR I can certainly look into changing this. Just let me know.

arielj: my suggestion is to only refactor this new code that is added, not to change the rest of the code that was there

casasfernando:
> btw afaik rm is async, you should be using rmSync

I had the same impression but the existing code is using rm.
I will prepare a new commit to address @arielj comment above, can you confirm rmSync is needed instead? I can change that as well.

Thanks

CommandMC:
`rm` is asynchronous, that's correct. That's a problem with this function in general though, so cleaning that up could also wait until a future PR

You'd have to use either `rmSync` from `fs` or `rm` from `fs/promises`. My vote would always be to use promise-based functions where possible (as they do not [block the main process](<https://www.electronjs.org/docs/latest/tutorial/performance#3-blocking-the-main-process>)), ideally also wrapping them in a `Promise.all` to run concurrently. That might look like this here (for the 64-bit case):
```ts
await Promise.all([
  ...dlls32.map((dll) => rm(`${winePrefix}/drive_c/windows/syswow64/${dll}`)),
  ...dlls64.map((dll) => rm(`${winePrefix}/drive_c/windows/system32/${dll}`))
])

casasfernando:
I just pushed a couple of commits including the following changes:

- DLL file removal code refactoring
- Switch from `rm` to `rmSync`: for this change I used a try/catch structure because `rmSync` doesn't handle callbacks/errors as `rm`does.

Please take a look at it and let me know if I made any mistakes.

arielj (approved):
looks good to me

casasfernando:
@arielj I see you approved the PR around 3 weeks ago but yet is not merged. Does it also need approval from @Etaash-mathamsetty ?

Etaash-mathamsetty:
> @arielj I see you approved the PR around 3 weeks ago but yet is not merged. Does it also need approval from @Etaash-mathamsetty ?

Code looks good to me, one of us will merge soon, wanted to get the hotfixes out of the way first and minimize potential breakages

PR #3761, opened 2024-05-18
[Fix] Prevent empty name category from being added

DenysMb:
Currently we can add empty categories and also rename categories for an empty string.
This is just a small fix to prevent those cases.
I also did another one change to disable the "confirm rename button" when the name was not changed.

Before:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/33737137/9c392ba2-eadd-4231-9dfc-08d9f00b2a7e

After:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/assets/33737137/51b0d6f3-883e-4190-878b-b140fc923e2e

---

### DIFF
--- a/e2e/categories.spec.ts
+++ b/e2e/categories.spec.ts
@@ -15,6 +15,9 @@ electronTest('categories', async (app) => {
 
     await expect(dialog.getByText('No categories yet.')).toBeInViewport()
 
+    // add button should be disabled without a name
+    await expect(dialog.getByTitle('Add', { exact: true })).toBeDisabled()
+
     // add new category
     await dialog.getByPlaceholder('Add new category').fill('Great games')
     await dialog.getByTitle('Add', { exact: true }).click()
@@ -23,8 +26,20 @@ electronTest('categories', async (app) => {
       dialog.locator('span', { hasText: 'Great games' })
     ).toBeInViewport()
 
-    // rename category
+    // add button should be disable if trying to rename category to empty name
     await dialog.getByTitle('Rename "Great games"').click()
+    await dialog.getByLabel('Rename "Great games"').fill('Great games')
+    await expect(dialog.getByTitle('Add', { exact: true })).toBeDisabled()
+
+    // add button should be disabled if trying to rename category to empty string
+    await dialog.getByLabel('Rename "Great games"').fill('')
+    await expect(dialog.getByTitle('Add', { exact: true })).toBeDisabled()
+
+    // add button should be disabled if trying to rename category to empty spaces
+    await dialog.getByLabel('Rename "Great games"').fill('   ')
+    await expect(dialog.getByTitle('Add', { exact: true })).toBeDisabled()
+
+    // rename category
     await dialog.getByLabel('Rename "Great games"').fill('Amazing games')
     await dialog
       .getByTitle('Confirm rename of "Great games" as "Amazing games"')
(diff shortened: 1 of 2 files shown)


src/frontend/screens/Library/components/CategoriesManager/index.tsx
```diff
@@ -30,6 +30,7 @@ function CategoryItem({
   const [renameMode, setRenameMode] = useState(false)
   const [newName, setNewName] = useState(name)
   const [removeMode, setRemoveMode] = useState(false)
+  const isNewNameEmptyOrEqualsOldName = newName === '' || newName === name
```
arielj, line 33: just curious, does this prevent blank strings like a string full of white spaces too?

DenysMb: Good point. No, it doesn't prevent it.
I will change this.

DenysMb: Done.

src/frontend/screens/Library/components/CategoriesManager/index.tsx
```diff
@@ -30,6 +30,8 @@ function CategoryItem({
   const [renameMode, setRenameMode] = useState(false)
   const [newName, setNewName] = useState(name)
   const [removeMode, setRemoveMode] = useState(false)
+  const isNewNameEmptyOrEqualsOldName =
+    newName?.trim() === '' || newName === name
```
arielj, line 34: do we need the `?` here and in `newCategoryName?.trim()`? they should be strings always, and if there's any case when they are possible undefined we should update the type to reflect that

DenysMb: No, we don't need it at all.
I just added because I don't know the way if there is a bug or something somewhere that makes the input `undefined`, so I though that would be better have and don't need than need and don't have.
But I'll remove.

DenysMb: Done.

arielj (commented):
Sorry, one more thing, it would be good to update the `e2e/categories.spec.ts` to test these changes

it should be easy to update the test to check that the button is disabled with blank and unchanged values

DenysMb:
> Sorry, one more thing, it would be good to update the `e2e/categories.spec.ts` to test these changes
> 
> it should be easy to update the test to check that the button is disabled with blank and unchanged values

No worries. Everything is valid to increase the code quality.
I update the test, but I couldn't test. I receive the following error: `Error: electron.launch: Process failed to launch!`.
Probably something related to openSUSE Tumbleweed + Nvidia, I think... Because when I need to run the `start` script, I need to add `app.commandLine?.appendSwitch('disable-gpu-sandbox')` to `main.ts` to be able to run the application.
I'm trying to solve this, so I can check if is everything okay with the added scenarios.

e2e/categories.spec.ts
```diff
@@ -23,8 +26,20 @@ electronTest('categories', async (app) => {
       dialog.locator('span', { hasText: 'Great games' })
     ).toBeInViewport()
 
-    // rename category
+    // add button should be disable if trying to rename category to empty name
```
arielj, line 29: Here you are testing that the button is disabled when the name is unchanged, the comment doesn't match that

the other 2 expects are for the blank tests

I can merge this after that's fixed

DenysMb: Sorry for the delay in fixing this.
These previous days were very busy.
I fixed this now. Where it says "empty" should be "same", I didn't realize that "words swap".

arielj (approved):

arielj:
thanks! merging

PR #3302, opened 2023-12-08
[TECH] Add React Dev tools debug script

flavioislima:
Install the proper extension and loads it when using `yarn debug:react`

image

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #3765, opened 2024-05-18
[Tech] External React Dev Tools application

CommandMC:
Adopting #3302

Now that the Zustand PR is merged, this is the only thing missing to truly start battling the re-render issues :^)

---

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,4 +1,4 @@
-import { defineConfig, UserConfigExport } from 'vite'
+import { defineConfig, type Plugin, type UserConfigExport } from 'vite'
 import electron from 'vite-plugin-electron'
 import react from '@vitejs/plugin-react-swc'
 import svgr from 'vite-plugin-svgr'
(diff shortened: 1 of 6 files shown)


vite.config.ts
```diff
@@ -24,7 +24,20 @@ const electronViteConfig: UserConfigExport = {
   }
 }
 
-export default defineConfig({
+// FIXME: Potentially publish this as a dedicated plugin, if other projects
+//        run into the same issue
```
arielj, line 28: what does this mean? what is the issue this is fixing? would be nice to have that context so we know what to look for

arielj (commented):
I tried this and it works fine, the only thing I noticed is that when I close Heroic the react dev tools window stays open and I have to also close it

it would be nice if it auto-closes when closing heroic

vite.config.ts
CommandMC: The issue this PR resolves is not Heroic-specific, it's something every Electron app will run into when trying to use the React dev tools. Because of that, every project will have to do something like this (adding a plugin to their bundler that adds the `<script>` tag). Thus, it might make sense to publish just this small Vite plugin separately (instead of defining it in this config file), so everyone else can just install it instead of essentially copy-pasting this code

CommandMC:
> it would be nice if it auto-closes when closing heroic

I doubt we can implement features like that into the React dev tools (at least not without modifying the package ourselves)

arielj (approved):

Issue #3736, opened 2024-05-04
Mangohud no longer appears


a931vwewi:
### Describe the bug

Post AppImage version 2.8.0, mangohud no longer appears for applications. Within the logs for 2.8.0, I can see mangohud being called in an info backend message when an application launches, I do not see the same in the logs for 2.14.1. Launching heroic through the shell with mangohud, `mangohud Heroic.Appimage`, works fine and loads the config stored in `~/.config/MangoHud/MangoHud.conf`. Perhaps [pr#2841](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/2841/commits) may be why?

### Add logs

```text
AppImage 2.8.0
(13:23:01) INFO:    [Legendary]:        Refreshing library...
(13:23:01) INFO:    [Connection]:       Connectivity: check-online
(13:23:01) INFO:    [Connection]:       Pinging external endpoints
(13:23:01) INFO:    [Legendary]:        Running command: /tmp/.mount_HeroicEK6NZU/resources/app.asar.unpacked/build/bin/linux/legendary --version
(13:23:01) INFO:    [Legendary]:        Legendary location: /tmp/.mount_HeroicEK6NZU/resources/app.asar.unpacked/build/bin/linux/legendary
(13:23:01) INFO:    [Gog]:              GOGDL location: /tmp/.mount_HeroicEK6NZU/resources/app.asar.unpacked/build/bin/linux/gogdl
(13:23:01) INFO:    [Gog]:              GOGDL location: /tmp/.mount_HeroicEK6NZU/resources/app.asar.unpacked/build/bin/linux/gogdl
(13:23:02) WARNING: [Backend]:          Failed to register protocol with OS.
(13:23:02) INFO:    [Gog]:              Running command: /tmp/.mount_HeroicEK6NZU/resources/app.asar.unpacked/build/bin/linux/gogdl --auth-config-path /home/user/Desktop/Heroic-2.8.0.AppImage.home/.config/heroic/gog_store/auth.json --version
(13:23:02) INFO:    [Connection]:       Connectivity: online
(13:23:02) INFO:    [Backend]:          User Not Found, removing it from Store
(13:23:02) INFO:    [Backend]:          

Heroic Version: 2.8.0 Nico Robin
Legendary Version:  0.20.32 Dark Energy (hotfix #6)
GOGdl Version: 0.7.2

Electron Version: 23.1.4
[...]

PR #3767, opened 2024-05-18
[FIX] some issues with sideloads

Etaash-mathamsetty:
- fixes sideload logs (end of log was printed early)
- fixes wrappers with native sideloads (#3736)

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1022,7 +1022,7 @@ async function callRunner(
 
   //this must be a native sideload
   if (runner.name === 'sideload' && wrappers.length > 0) {
-    commandParts.unshift( ...wrappers, runner.bin)
+    commandParts.unshift(...wrappers, runner.bin)
     bin = commandParts.shift()!
     fullRunnerPath = bin
   }
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1007,6 +1007,7 @@ async function callRunner(
   options?: CallRunnerOptions
 ): Promise<ExecResult> {
   const appName = appNameFromCommandParts(commandParts, runner.name)
+  const wrappers = options?.wrappers || []
 
   // Necessary to get rid of possible undefined or null entries, else
   // TypeError is triggered
@@ -1018,6 +1019,13 @@ async function callRunner(
   // macOS/Linux: `spawn`ing an executable in the current working directory
   // requires a "./"
   if (!isWindows) bin = './' + bin
+
+  //this must be a native sideload
+  if (runner.name === 'sideload' && wrappers.length > 0) {
+    commandParts.unshift( ...wrappers, runner.bin)
+    bin = commandParts.shift()!
+    fullRunnerPath = bin
+  }
 
   // On Windows: Use PowerShell's `Start-Process` to wait for the process and
   // its children to exit, provided PowerShell is available
--- a/src/backend/storeManagers/sideload/games.ts
+++ b/src/backend/storeManagers/sideload/games.ts
@@ -71,7 +71,8 @@ export async function launch(
   /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
   launchArguments?: LaunchOption
 ): Promise<boolean> {
-  return launchGame(appName, getGameInfo(appName), 'sideload')
+  // wait for the game to launch and finish running
+  return Promise.resolve(launchGame(appName, getGameInfo(appName), 'sideload'))
 }
 
 export async function stop(appName: string): Promise<void> {


arielj:
I'm trying to reproduce what you mention of the `End of log` printed early but I can't reproduce that, I tried 3 sideloaded games and that always happens at the end

src/backend/launcher.ts
```diff
@@ -1019,6 +1020,13 @@ async function callRunner(
   // requires a "./"
   if (!isWindows) bin = './' + bin
 
+  // this must be a native sideload
```
arielj, line 1023: what does this comment mean? that at this point we are sure this is a native sideloaded game? I don't understand it

Etaash-mathamsetty: `that at this point we are sure this is a native sideloaded game` yes that's what it means :)

src/backend/storeManagers/sideload/games.ts
```diff
@@ -71,7 +72,8 @@ export async function launch(
   /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
   launchArguments?: LaunchOption
 ): Promise<boolean> {
-  return launchGame(appName, getGameInfo(appName), 'sideload')
+  // wait for the game to launch and finish running
+  return Promise.resolve(launchGame(appName, getGameInfo(appName), 'sideload'))
```
CommandMC, line 76: This does not do anything, calling `Promise.resolve` with a promise just returns that promise

src/backend/launcher.ts
```diff
@@ -1019,6 +1020,13 @@ async function callRunner(
...
+  if (runner.name === 'sideload' && wrappers.length > 0) {
+    commandParts.unshift(...wrappers, fullRunnerPath)
+    bin = commandParts.shift()!
+    fullRunnerPath = bin
```
CommandMC, line 1027:
```suggestion
    bin = fullRunnerPath = commandParts.shift()!
```

src/backend/launcher.ts
CommandMC, line 1023: I'm not sure what this comment is telling me that I don't already know. If this wasn't a native sideloaded game, we wouldn't be in this function

src/backend/storeManagers/sideload/games.ts
```diff
@@ -71,7 +72,10 @@ export async function launch(
   /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
   launchArguments?: LaunchOption
 ): Promise<boolean> {
-  return launchGame(appName, getGameInfo(appName), 'sideload')
+  // wait for the game to launch and finish running
+  return Promise.resolve(
+    await launchGame(appName, getGameInfo(appName), 'sideload')
+  )
```
CommandMC, line 78: This still does not do anything. What are you actually trying to do/fix? As long as callers of this function `await` it, they'll already `await` the Promise returned by `launchGame`

Etaash-mathamsetty: oh that's right actually, then why did this work in the first place lol

src/backend/storeManagers/storeManagerCommon/games.ts
```diff
@@ -187,6 +187,11 @@ export async function launchGame(
       return false
     }
 
+    appendGamePlayLog(
+      gameInfo,
+      `Launch Command: ${executable} ${launcherArgs ?? ''}\n`
```
arielj, line 192: One issue I noticed with this is that it doesn't really print the launch command, it prints the executable and arguments only. So for a native game this is fine, but for a game using wine/proton this doesn't really show the actual command? for the different stores we print more data also like env variables for example

Etaash-mathamsetty: I don't know how to convert the env variable record to a string, any guidance on how to do that is welcome

CommandMC:
```ts
Object.entries(envVars)
  .map(([name, value]) => `${name}=${value}`)
  .sort()
  .join(' ')
```
would be one way to do it

CommandMC (commented):
As it stands right now, this PR seems to have a few unrelated changes thrown together, with rather unclear commit messages. Could you squash commits where necessary?

src/backend/storeManagers/storeManagerCommon/games.ts
```diff
@@ -219,12 +219,16 @@ export async function launchGame(
...
         ...setupWrapperEnvVars({ appName, appRunner: runner }),
         ...setupEnvVars(gameSettings, gameInfo.install.install_path),
         ...getKnownFixesEnvVariables(appName, runner)
       }
 
+      if (wrappers.length > 0) {
+        extraArgs.unshift(...wrappers, executable)
+        executable = extraArgs.shift()!
+      }
+
```
arielj:
```suggestion

```

PR #3779, opened 2024-05-22
[Fix] Prevent uninstalling a game while it's running

arielj:
This PR disables the "Uninstall" button while a game is running so it cannot be uninstalled by mistake.

Other games can be uninstalled (though we still have the limitation if not uninstalling an Epic game while another Epic game is being installed, but that's unrelated to this fix).

Apart from disabling the buttons, I also made the Uninstall dialog no-op when the game is playing, so if we add an uninstall button and we forget to disable it, it still won't allow the user to uninstall the game while it's running.

---


Etaash-mathamsetty (approved):
looks good, i will test it later

Etaash-mathamsetty:
looks like it's working

PR #3783, opened 2024-05-26
[Fix] Account for spaces in GOG game save paths on Windows

ghoste07:
The way arguments are passed to PowerShell already handles spaces in the path and the additional quotation marks added by `getShellPath()` causes the path to be truncated at the first space (if present).

For example, 'C:\Users\User\Documents\The Witcher 3\' gets interpreted as 'C:\Users\User\Documents\The'.

Using `.replaceAll()` should be safe since files/folders in Windows can't contain quotes anyway.

(Apologies if there's a better way to do this - Most of my experience is in bash and this is the first I've looked at the heroic codebase)

---

### DIFF
--- a/src/backend/save_sync.ts
+++ b/src/backend/save_sync.ts
@@ -212,7 +212,6 @@ async function getDefaultGogSavePaths(
           )
         }
       }
-      absolutePath = absolutePath.replaceAll('"', '')
     }
 
     resolvedLocations.push({
--- a/src/backend/save_sync.ts
+++ b/src/backend/save_sync.ts
@@ -212,6 +212,7 @@ async function getDefaultGogSavePaths(
           )
         }
       }
+      absolutePath = absolutePath.replaceAll('"', '')
     }
 
     resolvedLocations.push({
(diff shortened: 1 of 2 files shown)


flavioislima:
   - - -
   - - -

ghoste07:

CommandMC (commented):
Thank you for your contribution

This change is generally fine, but it'd be great if this function wouldn't rely on knowing that `getShellPath` adds quotes around paths. Given we don't do this (and don't handle this properly) in other places either, the best course of action might be changing `getShellPath` to just not add quotes in the first place

ghoste07:
> the best course of action might be changing `getShellPath` to just not add quotes in the first place

I agree, that's better. I've updated the PR for that. I didn't notice any regressions in my limited testing.

CommandMC:
Well, I'd consider regressions rather unlikely here, considering this is the only function using `getShellPath` (not sure where everything else went, could've sworn this was used elsewhere)

CommandMC (approved):

PR #3827, opened 2024-06-17
[Tech] Remove the Global State

CommandMC:
The GlobalState / ContextProvider is now replaced with a Zustand store, which improves the re-rendering issues dramatically

The commits here are quite bite-sized, so reviewing them one at a time is recommended. The commit messages also include details in case something was out of the ordinary when migrating that specific property

Known issues:
- [ ] Library changes following user action (game install, update, uninstall, etc) take a longer time to update, I assume there was some optimistic UI here that I didn't migrate
- [ ] ...probably more! Still have to test this on a clean-install & on multiple OSs

---


arielj:
something seems to be wrong with the global state after uninstalling a game, when I try to close heroic it thinks it's still doing something and shows the `heroic is performing an operation, do you want to close it anyway` message

there's no way to make these changes more gradually? I know I always say the same but it's really hard to review 100 files changing, this changes a lot around the whole app, it's impossible for us to test/QA everything properly

a nice thing that I liked about zustand was that it would allow us to move things to zustand gradually in different PRs with more focused scope to focus the test instead of changing everything, I'm always scared of changing that much

CommandMC:
Part of why I was so granular with these commits is so we can take only a part of them and merge those. I'll create a new PR with some of the "easy" changes here, if that's easier

PR #3845, opened 2024-07-01
[GOG]: ignore port provided in adtraction url

imLinguin:
Fixes fallback issues introduced after recent changes in how GOG does redirects

### DIFF
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -226,7 +226,15 @@ export default function WebView() {
           const parsedUrl = new URL(validatedURL)
           const redirectUrl = parsedUrl.searchParams.get('url')
           const url = new URL(redirectUrl || 'https://gog.com')
-          // Remove any port definitions, they are dumb
+          // Remove any port definitions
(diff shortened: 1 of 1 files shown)


src/frontend/screens/WebView/index.tsx
```diff
@@ -225,7 +225,10 @@ export default function WebView() {
         if (validatedURL && validatedURL.match(/track\.adtraction\.com/)) {
           const parsedUrl = new URL(validatedURL)
           const redirectUrl = parsedUrl.searchParams.get('url')
-          webview.loadURL(redirectUrl || 'https://gog.com')
+          const url = new URL(redirectUrl || 'https://gog.com')
+          // Remove any port definitions, they are dumb
```
CommandMC, line 229: Maybe add the explanation you provided in the issue here? Just "they are dumb" might be humorous, but I imagine a confused me looking at this in about a year going "why is this there again?"

CommandMC (commented):
This is a fine fix, but part of me wonders why GOG does this and whether the "better" solution might just be bugging them about it

CommandMC (approved):

PR #3848, opened 2024-07-02
[Fix] Sideload: Don't update game settings if we're in editMode

CommandMC:
Clicking "Finish" in the "Edit App/Game" dialog made Heroic "forget" the Wine Version associated with that game. That's now fixed.

The problem here is that the configured Wine version is initially undefined, only being updated by the version selector. That's not rendered in editMode, so we're just always saving undefined.
It seems this issue was thought about for the Wine prefix, as *it* is set by an effect above (so the whole spiel ends up doing nothing). This is no longer needed now, so I've removed that part of the effect's functionality.

To test: Edit a sideloaded game (making sure to click "Finish" to save your changes), head to the game settings. Before, notice your Wine Version being reset to the first one in the list. Now, notice your Wine Version is not altered.

Note: The configured CrossOver bottle should also be affected by this change, although I don't have an Apple device to test this on.

---

### DIFF
--- a/src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
+++ b/src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
@@ -114,25 +114,14 @@ export default function SideloadDialog({
     }
   }, [])
 
+  // Suggest default Wine prefix if we're adding a new app
   useEffect(() => {
-    const setWine = async () => {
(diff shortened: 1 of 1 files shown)


arielj (approved):
looks good to me, tested it and I can confirm the wine config is not lost

flavioislima (approved):
Good catch

PR #3885, opened 2024-07-27
[Tech] Update to ESLint 9

CommandMC:
This upgrades to a "flat" config file as well

I've had to move to `eslint-plugin-import-x`, since `eslint-plugin-import` (what we were using before) does not support ESLint 9 yet. -x is a fork of the original plugin, so it works the same way otherwise (it claims to be a good deal faster as well, although I have not tested this)

Don't let the changed lines count intimidate you, most of those are in the pnpm lock file

---


imLinguin (approved):
Looks good. I like how GitHub makes annotations for any linting warnings (not sure if that's related to eslint 9 or not)

imLinguin (approved):

PR #3895, opened 2024-08-03
[GOG]: add support for linux native dlcs in modify install dialog

imLinguin:
also automatically run post install script after completed download

### DIFF
--- a/src/backend/storeManagers/gog/games.ts
+++ b/src/backend/storeManagers/gog/games.ts
@@ -399,11 +399,12 @@ export async function install(
   }
 
   const sizeOnDisk = await getPathDiskSize(join(path, gameInfo.folder_name))
+  const install_path = join(path, gameInfo.folder_name)
 
   const installedData: InstalledInfo = {
     platform: installPlatform,
     executable: '',
-    install_path: join(path, gameInfo.folder_name),
+    install_path,
     install_size: getFileSize(sizeOnDisk),
     is_dlc: false,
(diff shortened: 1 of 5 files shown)


flavioislima (approved):
LGTM

src/backend/storeManagers/gog/games.ts
```diff
@@ -1140,6 +1152,17 @@ export async function update(
     (isWindows || existsSync(gameSettings.winePrefix))
   ) {
     await setup(appName, gameObject, false)
+  } else if (gameObject.platform === 'linux') {
+    const installer = join(gameObject.install_path, 'support/postinst.sh')
+    if (existsSync(installer)) {
+      logInfo(`Running ${installer}`, LogPrefix.Gog)
+      await spawnAsync(installer, []).catch((err) =>
```
CommandMC, line 1159: Are we sure these scripts always have their executable bit set?

CommandMC (approved):
I don't think I have any games I could test this with, but code looks good

src/backend/storeManagers/gog/games.ts
imLinguin: pretty sure, yeah, they are automatically added to the zip and include the flag. I guess we could add a safeguard to always set it.

CommandMC: I see. Let's add that if we run into any issues, no point in adding safeguards where they aren't necessary

PR #3914, opened 2024-08-09
[chore]: add keys for fontStore consts

12ya:
add keys for fontStore consts

### DIFF
--- a/src/backend/constants.ts
+++ b/src/backend/constants.ts
@@ -13,6 +13,10 @@ import { GlobalConfig } from './config'
 import { TypeCheckedStoreBackend } from './electron_store'
 import { dirSync } from 'tmp'
 
+export enum FONTS_STORE_KEYS {
+  FONTS = 'fonts'
+}
+
 const configStore = new TypeCheckedStoreBackend('configStore', {
   cwd: 'store'
 })
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -94,7 +94,8 @@ import {
   isSnap,
   fixesPath,
   isWindows,
-  isMac
+  isMac,
+  FONTS_STORE_KEYS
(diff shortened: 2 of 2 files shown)


flavioislima:
   - - -
   - - -

arielj (commented):
I think this change is just adding complexity, it's replacing 2 strings (it could be just a single `const fontKey = 'fonts'` above the function if the idea is to reduce repetition) with:
- an enum
- an export
- an import

I don't think it adds value and it makes things more complicated

12ya:
> I think this change is just adding complexity, it's replacing 2 strings (it could be just a single `const fontKey = 'fonts'` above the function if the idea is to reduce repetition) with:
> 
> * an enum
> * an export
> * an import
> 
> I don't think it adds value and it makes things more complicated

my objective is to add safety (replace strings with enums);
let me know if it's a good idea :)

arielj:
> > I think this change is just adding complexity, it's replacing 2 strings (it could be just a single `const fontKey = 'fonts'` above the function if the idea is to reduce repetition) with:
> > 
> > * an enum
> > * an export
> > * an import
> > 
> > I don't think it adds value and it makes things more complicated
> 
> my objective is to add safety (replace strings with enums), plus adds consistency for future fontKey additions; let me know if it's a good idea :)

sorry but I don't think we need these changes

I understand where the idea is coming from (if we use an enum we just change that and it applies everywhere), but we haven't really had issues like that with the current code and these PRs are adding complexity to solve a problem that never really happened (and many of those strings will really never change)

it also creates an extra level of indirection (now the paths in this PR are harder to read because you have constants and the actual values in a different place of the code, while before each line was the actual path, of the fonts PR where the constant is in a different file and needs to be imported)

PR #3919, opened 2024-08-09
[FIX] Only use steam runtime with proton

Etaash-mathamsetty:
title

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -243,6 +243,7 @@ async function prepareLaunch(
   let steamRuntime: string[] = []
   const shouldUseRuntime =
     gameSettings.useSteamRuntime &&
+    gameSettings.wineVersion.type === 'proton' &&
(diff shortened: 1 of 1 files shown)


src/backend/launcher.ts
```diff
@@ -243,6 +243,7 @@ async function prepareLaunch(
   let steamRuntime: string[] = []
   const shouldUseRuntime =
     gameSettings.useSteamRuntime &&
+    gameSettings.wineVersion.type === 'proton' &&
```
arielj, line 246:
```suggestion
    (!isNative && gameSettings.wineVersion.type === 'proton') &&
```

maybe?

because with the current code then you wouldn't be able to use it for native games which sometimes fixes issues

arielj (approved):
tested the different combinations of native/non-native wine/proton umu on/off and it seems to work fine :+1:

PR #3924, opened 2024-08-10
[GOG]: simplify library fetching

imLinguin:
Instead of spawning so many async functions let's just run a for loop, the wait time may be a little longer, however it's already miles better than the one in 2.14 


flavioislima (approved):
tested here and wrks fine

PR #3927, opened 2024-08-11
[Tech] Patch `@types/node` to restrict our `platform` type to valid platforms

CommandMC:
`process.platform` can, under normal circumstances, only ever be `win32`, `darwin`, or `linux`. `@types/node` can't know this however, so we have to patch it to only make those values possible literal options

Technically this is incorrect (we could have someone run a development version of Heroic on any other platform), but this makes it easier to do checks on the property, or to use it in a switch case / index a Record with it
Somewhat improves the situation of #3908

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,5 +157,10 @@
   },
   "resolutions": {
     "ts-morph": "17.0.1"
+  },
+  "pnpm": {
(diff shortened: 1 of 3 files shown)


flavioislima (approved):
LGTM

PR #3941, opened 2024-08-11
[Tech] Don't compute log file locations when importing `constants.ts`

CommandMC:
Log file locations (`currentLogFile`, `lastLogFile`, `legendaryLogFile`, etc.) were computed when importing the `constants.ts` file, which in turn lead to a lot of code being ran & modules being imported, even if those values are never actually read.
We can instead just use the common `initXXX` pattern to set the log file paths, and `getLogFile` to get them where they're needed.

This doesn't have an immediate benefit, but makes testing slightly easier in the future. Instead of having to mock a lot of modules just to get a non-crashing test setup, you now need *slightly less*

I've tested all aspects of the app relating to log files. Code itself was not touched, so I only tested on Linux (since the only thing required to test was whether the functions actually ran at the correct time)

---

### DIFF
--- a/src/backend/constants.ts
+++ b/src/backend/constants.ts
@@ -5,7 +5,6 @@ import { parse } from '@node-steam/vdf'
 
 import { GameConfigVersion, GlobalConfigVersion } from 'common/types'
 import { logDebug, LogPrefix } from './logger/logger'
-import { createNewLogFileAndClearOldOnes } from './logger/logfile'
 import { env } from 'process'
 import { app } from 'electron'
 import { existsSync, mkdirSync, readFileSync } from 'graceful-fs'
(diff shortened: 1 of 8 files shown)


arielj:
I've been thinking about maybe splitting that big `constants.ts` file into smaller files inside a `constants` directory, then we could import the constants that we need in different places without node having to parse the whole constants file?

then in production code we'll still have some code importing `constants/loggerPaths.ts` and that would trigger the initialization of those paths as soon as it's imported, but then during testing if we test something that doesn't import anything from that file we don't have to worry?

Not sure why I have this feeling that the `initX` pattern is going to create some problem in the future (like we'll end up with things initialized at the wrong time/place). I haven't tried yet but I think we could solve the mocking thing and skip the need if `initX`?

CommandMC:
Closing in favor of https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4264

Issue #2556, opened 2023-03-18
Default wine prefixes folder can be left empty


arielj:
### Describe the bug

When in the General settings, the `Set Folder for new Wine Prefixes` setting can be left blank, leading to problems installing games.

This is a problem because a user may delete it by mistake and we run no validations and no fallback.

### Add logs

```text
This is the log deleting the path:

(15:24:04) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Games/Heroi"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Games/Hero"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Games/Her"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Games/He"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Games/H"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Games/"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Games"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Game"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Gam"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/Ga"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/G"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel/"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ariel"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/arie"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ari"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/ar"
(15:24:05) INFO:    [Backend]:          Heroic: Setting defaultInstallPath to "/home/a"
[...]

PR #3948, opened 2024-08-12
warn whem WinePrefixesBasePath is empty

pt8o:
Resolves https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2556

Screenshot from 2024-08-12 18-22-53

Since all of these settings are saved on edit, rather than waiting for the user to hit a "Save" button or something, I didn't think there was a great way (UX wise) to prevent the user from saving an empty folder. Instead I opted to display a warning below the input, and a button to revert to the most recent non-empty path.

If the user has already saved an empty path and then opens Settings, then there is no recent non-empty path, so the warning still appears but the undo button doesnt.

---


flavioislima:
   - - -
   - - -

src/frontend/components/UI/TextInputField/index.tsx
```diff
@@ -35,7 +35,7 @@ const TextInputField = ({
       {label && <label htmlFor={htmlId}>{label}</label>}
       {inputIcon}
       <input type="text" id={htmlId} value={value} {...inputProps} />
-      {value && warning}
+      {warning}
```
pt8o, line 38: I was nervous about the blast radius of changing this logic in the main `TextInputField` component but it actually looks like the `warning` prop is not currently being used anywhere in the app. The way I see it, the `warning` shouldn't hinge on the existence of a `value`, but if people disagree with me on that, I think I can just use `afterInput` to display the empty path warning.

pt8o:

src/frontend/screens/Settings/components/WinePrefixesBasePath.tsx
```diff
@@ -19,20 +21,59 @@ const WinePrefixesBasePath = () => {
...
+        {lastValidPrefix && (
+          <SvgButton
+            className="button button-icon-flex is-danger"
+            onClick={handleRevert}
+          >
+            <Undo />
+          </SvgButton>
+        )}
+      </Box>
+    ) : undefined
```
CommandMC, line 61: Might want to use `useMemo` here

src/frontend/screens/Settings/components/WinePrefixesBasePath.tsx
```diff
@@ -19,20 +21,59 @@ const WinePrefixesBasePath = () => {
     'defaultWinePrefix',
     ''
   )
+  const [lastValidPrefix, setLastValidPrefix] = useState(defaultWinePrefix)
+
+  function handlePathChange(val: string) {
+    setDefaultWinePrefix(val)
+    if (val) {
```
CommandMC, line 28: I'm not sure whether a simple truthiness check makes sense here. Sure, if the user highlights the entire path and then deletes it, this'll work. More likely however (especially considering virtual keyboards, e.g. on the Deck), they're gonna backspace one-by-one, causing a restore operation to just re-insert the `/`. This of course won't work as a prefix path either

The easiest way to resolve this would be to not store the "last valid" value at all, instead just restoring it to the same path a new install would use:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/47d3075ca1c53a8e009b6a504a839a3143e11d0e/src/frontend/screens/Library/components/InstallModal/WineSelector/index.tsx#L70-L73

src/frontend/components/UI/TextInputField/index.tsx
CommandMC: Yup this is fine. If the component using this has any specific preferences on when `warning` should be displayed, it can handle those itself

src/frontend/screens/Settings/components/WinePrefixesBasePath.tsx
```diff
@@ -19,20 +21,59 @@ const WinePrefixesBasePath = () => {
...
+      setLastValidPrefix(val)
+    }
+  }
```
arielj, line 31: Instead of storing `lastValidPrefix` (which the name can be a bit misleading, it doesn't mean this is valid) I imagine this logic for `handlePathChange`:

```ts
function handlePathChange(newPath: string) {
  if (newPath) {
    // not empty, all good
    setDefaultWinePrefix(newPath)
  }  else {
    // empty, show an alert/warning saying it will be reverted to the previous value
    // reset the value to the original config value
  }
}
```

But I actually think we have a problem with the PathSelectionBox component because we have no simple way of resetting the value (we would have to re-render the component cause it uses the internal tmpPath state).

I imagine the `onPathChange` function that we pass with the props could be updated to return a string (instead of void) that the component should set the field to, and instead of simply calling that function in the `onBlur` callback, we could wrap it to call it and use the return value to update with `setTmpPath(theReturnValueOfOnPathChange)`.

src/frontend/screens/Settings/components/WinePrefixesBasePath.tsx
pt8o: Sorry if this is a silly question but I don't see where to find the original `defaultWinePrefix`? In the code you linked the `suggestedWinePrefix` is built partly the `title` or `appName` from `WineSelector`'s props, which we don't have in the component we're working on... Is there something like a constants file that the initial value comes from?

PR #3958, opened 2024-08-16
[Tech] Fix some of the smaller ESLint warnings

CommandMC:
ESLint auto-fixes some of these problems, so in the interest of not making everyone *not* include these in their PRs (since it's most likely changing something unrelated), I figured I'd make a dedicated PR for them.

---

### DIFF
--- a/src/backend/utils/systeminfo/heroicVersion.ts
+++ b/src/backend/utils/systeminfo/heroicVersion.ts
@@ -3,12 +3,7 @@ import pkg_json from 'backend/../../package.json'
 
 function getHeroicVersion(): string {
   const VERSION_NUMBER = app.getVersion()
-  // FIXME: Currently ESLint doesn't know that these are type-checked. I believe
-  //        TS 5.X introduced support for this properly, so come back to this
-  //        once we update to that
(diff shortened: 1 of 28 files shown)


src/backend/utils/systeminfo/heroicVersion.ts
```diff
@@ -6,9 +6,7 @@ function getHeroicVersion(): string {
   // FIXME: Currently ESLint doesn't know that these are type-checked. I believe
```
arielj, line 6: I think this comment can be removed now? sound like it was a FIXME for that eslint-disable

src/frontend/screens/Settings/components/Tools/index.tsx
```diff
@@ -23,8 +23,7 @@ export default function Tools() {
     return <></>
   }
 
-  type Tool = 'winecfg' | string
```
arielj, line 26: I wonder if it would be better to add the different strings we use when calling `callTools` instead of using just `string`? seems like a really limited number of possible values, sounds like an easy win for this type

arielj (approved):
looks good to me, I added some comments but nothing crazy or blocker

PR #3961, opened 2024-08-17
[FIX/FEATURE] Make virtual keyboard compatible with all inputs

arielj:
This PR updates the virtual keyboard to solve a few issues and makes it compatible with all text inputs.

Before this, the virtual keyboard when using a gamepad was only valid for the search box in the library and it was partially compatible with the search box in the winetricks dialog (it would open, but behind the dialog, so it was unusable).

This PR has a few changes:
- Use `showModal` for dialogs instead of the `popover` API, that fixes the issue with the keyboard appearing behind the winetricks dialog
- Fix an issue where the `x` button of the search inputs was not working with the controller (instead of clicking the SVG icon, we now click the parent button)
- Updates the gamepad handling to open the keyboard for any text input and to send events to it, not only search inputs
- Updates the `TextInputField` to support the `input` events sent by the keyboard module (we can't use react's onChange flow)
- Because of the previous item, now the `onChange` function passed to `TextInputField` receives the new value and not a react on change event, so I had to updated all the places that used that to replace the `e.target.value` with `newValue`

There are 2 issues I can't find how to fix (both errors are present in the main branch already though, so leaving the out of the scope of this):
- when navigating the elements inside a modal, when we reach the top/bottom, if we continue moving the stick it will actually scroll the library for example, I can't find any way to prevent this (seems to be some issue with the spatial navigation feature we use)
- when displaying search results below the input (either the list of games in the library or the list of winetricks packages found), the elements are not focusable. This is also some issue with the spatial navigation that can't seem to move the focus to elements inside a container with a predefined height or max-height

---

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,6 +10,5 @@
   </head>
   <body>
     <div id="root"></div>
-    <div class="simple-keyboard"></div>
   </body>
 </html>
--- a/src/frontend/helpers/virtualKeyboard.ts
+++ b/src/frontend/helpers/virtualKeyboard.ts
@@ -2,15 +2,10 @@ import Keyboard from 'simple-keyboard'
 import 'simple-keyboard/build/css/index.css'
 
 let virtualKeyboard: Keyboard | null = null
+let targetInput: HTMLInputElement | null = null
 
 function currentElement() {
   return document.querySelector<HTMLElement>(':focus')
-}
-
-function searchInput() {
-  // only change this if you change the id of the input element
-  // in frontend/components/UI/SearchBar/index.tsx
-  return document.querySelector<HTMLInputElement>('#search')
 }
 
 function focusKeyboard() {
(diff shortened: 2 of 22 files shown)


index.html
```diff
@@ -10,6 +10,5 @@
   </head>
   <body>
     <div id="root"></div>
-    <div class="simple-keyboard"></div>
```
arielj, line 13: this was not used, we have another div for this inside the react app

src/frontend/helpers/virtualKeyboard.ts
```diff
@@ -2,47 +2,61 @@ import Keyboard from 'simple-keyboard'
...
+    targetInput.value = targetInput.value + ' '
   }
-  input.dispatchEvent(new Event('input'))
+  targetInput.dispatchEvent(new Event('input'))
+}
+
+function makeKeyboardTopLayer() {
+  const wrapper = document.querySelector(
+    '.simple-keyboard-wrapper'
+  ) as HTMLDialogElement
```
CommandMC, line 36:
```suggestion
  const wrapper = document.querySelector<HTMLDialogElement>(
    '.simple-keyboard-wrapper'
  )
```
This probably won't lint

arielj: updated

flavioislima (approved):
Code Lgtm

PR #3989, opened 2024-09-01
Check prefix folder is correct before deleting it

arielj:
This PR adds a few checks before deleting the prefix of a game to check that we don't delete the wrong thing:

- if the path is the same as the game installation path, we don't delete it (would delete all games)
- if the path is the same as the default/shared prefix path, we don't delete it (would delete all prefixes)
- if the path does NOT contain a `drive_c` folder, we don't delete it (it either not a prefix or there's something off in it)

I also used the opportunity to move some code into a new `utils/uninstaller.ts` file to make `main.ts` a bit smaller.

---

### DIFF
--- a/src/backend/utils/uninstaller.ts
+++ b/src/backend/utils/uninstaller.ts
@@ -1,0 +1,125 @@
+import { GlobalConfig } from 'backend/config'
+import { fixesPath, gamesConfigPath } from 'backend/constants'
+import { notify } from 'backend/dialog/dialog'
+import { logError, logInfo, LogPrefix } from 'backend/logger/logger'
+import { gameManagerMap } from 'backend/storeManagers'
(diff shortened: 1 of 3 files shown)


src/backend/utils/uninstaller.ts
```diff
@@ -0,0 +1,125 @@
...
+    )
+    return
+  }
+
+  const dirContent = readdirSync(winePrefix)
+
+  if (dirContent.length > 0) {
+    const driveCPath = join(winePrefix, 'drive_c')
+
+    if (!existsSync(driveCPath)) {
```
CommandMC, line 46: Note that Proton prefixes (not using umu) will never have this folder. Maybe we also want to check for a `pfx` folder?

flavioislima (approved):
It is good to have those safeguards 👍🏽

PR #3997, opened 2024-09-06
[UX] Filter settings that don't apply to current platform and game

arielj:
This PR changes the behavior of the Game logger to remove settings from the output when the settings don't make sense for a given game and platform.

For example: it will filter out all the wine-related settings for native games, or hide the gamescope settings when the parent gamescope setting is disabled.

The idea behind this is that these settings that don't make any sense for a given situation are still displayed in the logs, adding noise and confusing the users and during support.

I think this is a good enough first iteration, I think there are probably more settings that can be removed specially on windows, but I don't have a machine with windows to verify (I was looking at the code of the different settings only).

Note: I know the code is kinda verbose, but it was intentional to make it super clear which elements are being removed an in which conditions. I could refactor this to set settings in an array and looping but I kinda felt this was better. I can change this if we don't agree.

---

### DIFF
--- a/src/backend/logger/logger.ts
+++ b/src/backend/logger/logger.ts
@@ -585,6 +585,7 @@ class GameLogWriter extends LogWriter {
       delete gameSettings.enableFSR
       delete gameSettings.showMangohud
       delete gameSettings.showFps
+      delete gameSettings.disableUMU
 
       if (notNative) {
         const wineType = gameSettings.wineVersion
@@ -639,6 +640,7 @@ class GameLogWriter extends LogWriter {
       delete gameSettings.battlEyeRuntime
       delete gameSettings.eacRuntime
       delete gameSettings.nvidiaPrime
+      delete gameSettings.disableUMU
     }
 
     return gameSettings
--- a/src/backend/logger/logger.ts
+++ b/src/backend/logger/logger.ts
@@ -4,7 +4,13 @@
  *        Note that with console.log and console.warn everything will be saved too.
  *        error equals console.error
  */
-import { AppSettings, GameInfo, GameSettings, Runner } from 'common/types'
+import {
+  AppSettings,
+  GameInfo,
+  GameScopeSettings,
+  GameSettings,
+  Runner
+} from 'common/types'
 import { showDialogBoxModalAuto } from '../dialog/dialog'
 import {
   appendMessageToLogFile,
@@ -26,7 +32,7 @@ import { existsSync, mkdirSync, openSync } from 'graceful-fs'
 import { Winetricks } from 'backend/tools'
 import { gameAnticheatInfo } from 'backend/anticheat/utils'
 import { isUmuSupported } from 'backend/utils/compatibility_layers'
-import { isWindows } from 'backend/constants/environment'
+import { isLinux, isMac, isWindows } from 'backend/constants/environment'
 import { gamesConfigPath } from 'backend/constants/paths'
 
 export enum LogPrefix {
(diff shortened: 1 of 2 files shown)


Etaash-mathamsetty:
looks good, but you should reduce the amount of repeated code in the filtering function

arielj:
> looks good, but you should reduce the amount of repeated code in the filtering function

I made a comment about the repetition of the code in the PR description, I intentionally left it repeated (and I can reduce this if we agree it's not good, but I feel it's more clear this way)

Etaash-mathamsetty:
> > looks good, but you should reduce the amount of repeated code in the filtering function
> 
> I made a comment about the repetition of the code in the PR description, I intentionally left it repeated (and I can reduce this if we agree it's not good, but I feel it's more clear this way)

oh whoops I missed that

src/backend/logger/logger.ts
```diff
@@ -468,6 +474,128 @@ class GameLogWriter extends LogWriter {
...
+      }
+    }
+
+    // remove settings that are not used on Linux
+    if (isLinux) {
+      delete gameSettings['enableMsync']
+      delete gameSettings['wineCrossoverBottle']
+
+      if (notNative) {
+        const wineType = gameSettings['wineVersion']
```
CommandMC, line 505:
```suggestion
        const wineVersion = gameSettings['wineVersion']
```

src/backend/logger/logger.ts
```diff
@@ -468,6 +474,128 @@ class GameLogWriter extends LogWriter {
     this.filePath = lastPlayLogFileLocation(gameInfo.app_name)
   }
 
+  // cleanup settings for logs to avoid confusions during support requests
+  filterGameSettingsForLog(
+    gameSettings: Partial<GameSettings>,
+    notNative: boolean
+  ): Partial<GameSettings> {
+    // remove gamescope settings if it's disabled
+    const gscope = gameSettings['gamescope'] as Partial<GameScopeSettings>
```
CommandMC, line 483:
```suggestion
    const gscope: Partial<GameScopeSettings> | undefined = gameSettings['gamescope']
```
This may seem a little excessive, but *always* avoid `as` if you can do it somehow.  
Why? Try changing `'gamescope'` to `'someOtherNonexistantName'`. No errors! Alternatively, try changing the `GameSettings` type, changing or removing `gamescope`. Again, no error here

src/backend/logger/logger.ts
```diff
@@ -507,7 +635,11 @@ class GameLogWriter extends LogWriter {
 
       // log game settings
       const gameSettings = await gameManagerMap[runner].getSettings(app_name)
-      const gameSettingsString = JSON.stringify(gameSettings, null, '\t')
+      const gameSettingsString = JSON.stringify(
+        this.filterGameSettingsForLog({ ...gameSettings }, notNative),
```
CommandMC, line 639:
```suggestion
        this.filterGameSettingsForLog(structuredClone(gameSettings), notNative),
```
Right now, this will only shallow-clone the settings object

src/backend/logger/logger.ts
```diff
@@ -468,6 +474,128 @@ class GameLogWriter extends LogWriter {
...
+      delete gameSettings['showMangohud']
+      delete gameSettings['showFps']
+      delete gameSettings['preferSystemLibs']
+      delete gameSettings['wineCrossoverBottle']
+      delete gameSettings['winePrefix']
+      delete gameSettings['wineVersion']
+      delete gameSettings['battlEyeRuntime']
+      delete gameSettings['eacRuntime']
+      delete gameSettings['nvidiaPrime']
+    }
```
CommandMC, line 594: This whole function should use dot notation to access properties. Bracket notation currently hides some errors, since we have `noUncheckedIndexedAccess` turned on in our tsconfig

CommandMC (approved):

PR #4033, opened 2024-09-29
[Fix] Delete runner configurations when resetting heroic

Etaash-mathamsetty:
Some people are having issues with nile, and have to manually delete these files, so I propose making heroic do this automatically when it is reset

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -111,7 +111,7 @@
         },
         "reset-heroic": {
             "question": {
-                "message": "Are you sure you want to reset Heroic? This will remove all Settings and Caching but won't remove your Installed games or your Epic credentials. Portable versions (AppImage, WinPortable, ...) of heroic needs to be restarted manually afterwards.",
+                "message": "Are you sure you want to reset Heroic? This will remove all Settings and Caching. Portable versions (AppImage, WinPortable, ...) of heroic needs to be restarted manually afterwards.",
                 "title": "Reset Heroic"
             }
         },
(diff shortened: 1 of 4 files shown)


Etaash-mathamsetty:
I can't run i18n for some reason, if someone else could do it that would be great

arielj:
I don't think this is a good idea. Users will be confused, then they would have to import every game after logging in?

If this is meant to workaround a problem with nile I think we should solve the problem either with nile or with the integration we have with nile (not sure if the issue is heroic or nile itself).

Etaash-mathamsetty:
> I don't think this is a good idea. Users will be confused, then they would have to import every game after logging in?
> 
> If this is meant to workaround a problem with nile I think we should solve the problem either with nile or with the integration we have with nile (not sure if the issue is heroic or nile itself).

It makes users more confused that resetting heroic doesn't actually reset it imo

arielj:
> > I don't think this is a good idea. Users will be confused, then they would have to import every game after logging in?
> > If this is meant to workaround a problem with nile I think we should solve the problem either with nile or with the integration we have with nile (not sure if the issue is heroic or nile itself).
> 
> It makes users more confused that resetting heroic doesn't actually reset it imo

Maybe the solution for that (and I think it would be a good UX improvement in general) is to have a `Reset` section and allow users to select what to delete. Something like:

```
RESET
======

- [ ] game settings                       (i)
- [ ] legendary / epic                    (i)
- [ ] nile / amazon                       (i)
- [ ] gogdl / gog                         (i)
- [ ] cache                               (i)
- [ ] something else                      (i)

[ reset ]
```

Then users can hover the `(i)` next to each item to understand what that means and decide if they want to reset that particular thing.

Etaash-mathamsetty:
yea that sounds good, will implement it soon

PR #4049, opened 2024-10-09
[FIX]: gog setup regression for v1 games

imLinguin:
On Linux we were looking at the windows version of the path which was always false.


CommandMC (approved):
I always have a hard time testing these without an explicit "These games were affected by this issue" list, but I'm gonna assume you know what you're doing

imLinguin:
We were doing a getWinePath for win variant on this variable. So we were trying to check if `C:/` paths exist... on Linux. I guess it doesn't need explanation in certain games

PR #4059, opened 2024-10-14
Render SelectFields as MUI Selects

CommandMC:
This renders most drop-down selection prompts inside Heroic. In turn, these dropdowns now work inside the Gamescope session.

Current issues:
- [X] Controller navigation does not work, hitting "Down" opens the dropdown, with no way to navigate further down to other components
- [X] There is a black box in the top-left of the component, where the label should be
- [X] The dropdowns do not work while in a Dialog (they're rendered behind the Dialog)

---

### DIFF
--- a/src/frontend/components/UI/Dialog/components/Dialog.tsx
+++ b/src/frontend/components/UI/Dialog/components/Dialog.tsx
@@ -58,14 +58,16 @@ export const Dialog: React.FC<DialogProps> = ({
   return (
     <MuiDialog
       open={open}
-      onClose={close}
+      onClose={(e, reason) => {
+        if (disableDialogBackdropClose && reason === 'backdropClick') return
+        close()
+      }}
       scroll="paper"
       maxWidth="md"
       PaperComponent={StyledPaper}
       PaperProps={{
         className
       }}
-      disableEscapeKeyDown={disableDialogBackdropClose}
       sx={{
         '& .Dialog__element': {
           maxWidth: 'min(700px, 85vw)',
--- a/src/frontend/components/UI/Dialog/components/Dialog.tsx
+++ b/src/frontend/components/UI/Dialog/components/Dialog.tsx
@@ -39,20 +39,21 @@ export const Dialog: React.FC<DialogProps> = ({
   onClose
 }) => {
   const [open, setOpen] = useState(true)
-  const [focusOnClose, setFocusOnClose] = useState<HTMLElement | null>(null)
   const { disableDialogBackdropClose } = useContext(ContextProvider)
 
   useEffect(() => {
-    setFocusOnClose(document.querySelector<HTMLElement>('*:focus'))
+    // HACK: Focussing the dialog using JS does not seem to work
+    //       Instead, simulate one or two tab presses
+    // One tab to focus the dialog
+    window.api.gamepadAction({ action: 'tab' })
+    // Second tab to skip the close button if it's shown
+    if (showCloseButton) window.api.gamepadAction({ action: 'tab' })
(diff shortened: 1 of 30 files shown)


src/frontend/components/UI/Dialog/components/Dialog.tsx
```diff
@@ -19,100 +23,73 @@ interface DialogProps {
...
-
-  useEffect(() => {
-    const dialog = dialogRef.current
-    if (dialog) {
-      const cancel = () => {
-        close()
-      }
-      dialog.addEventListener('cancel', cancel)
-
-      if (disableDialogBackdropClose) {
```
arielj, line 53: this was here to support the Accessibility option to disable the `close dialog when clicking outside` behavior. In this PR that's not working

arielj (changes requested):
Looks good, but I left a comment about an accessibility feature that is not working with the new dialogs

src/frontend/components/UI/Dialog/components/Dialog.tsx
CommandMC: Whoops, somehow understood this as "Disable Esc key closing dialogs"
Fixed!

src/frontend/components/UI/SelectField/index.tsx
```diff
@@ -35,13 +36,24 @@ const SelectField = ({
...
+        id={htmlId}
+        value={value}
+        onChange={onChange}
+        disabled={disabled}
+        sx={{
+          '& .MuiSelect-icon': { display: 'none' },
+          '& .MuiOutlinedInput-notchedOutline legend': { width: 0 },
+          '& .MuiSelect-select': {
+            textAlign: isRTL ? 'right' : 'left',
+            paddingLeft: 0
```
arielj, line 49: can we use `paddingInlineStart: 0` here instead? it's better to support RTL languages since `inlineStart` is equivalent to `left` in LTR and `right` in RTL

arielj (changes requested):
There's one more thing I noticed, when using a RTL language, the arrow in the select field is in the wrong side of the element:

in `main`:
image

with this changes:
image

src/frontend/components/UI/SelectField/index.tsx
CommandMC: I've ended up just using MUI's padding instead of ours, which also resolves an issue with excessive padding on RTL text

arielj:
I've been testing a bit more, I found 1 issue and two small things:

----------
one small thing is the `categories` and `filters` style changed and it's missing some padding:
- main:
image

- this branch:
image

----------
the second small thing is that the controller hints (the instructions of what each button does at the bottom of the screen) shows incorrect hints for dropdowns, it thinks it's the context menu of library cards and it shows `( x ) Close Options` incorrectly. This one can be fixed by using `target.closest('.contextMenu')` in line 60 of src/frontend/components/UI/ControllerHints/index.tsx

---------
now the issue is that, when using a controller to navigate the UI (or a keyboard), you can get `stuck` focusing the `X` button of dialogs when the dialog's content is long and it shows a scrollbar:

- open the settings for a game
- in the `wine` tab (because it's long it will show a scrollbar), move `up` until you focus the `X`
- you can't move out of there (moving up/down scrolls the library page, moving left/right does nothing)

a smaller dialog doesn't have this problem:
- now use the mouse and change to the `other` tab (this is short and won't show a scrollbar)
- move `up` until you focus the `X`
- you can move out (moving up does nothing as expected, down/left/right moves focus to the tabs)

arielj (approved):

PR #4070, opened 2024-10-23
[Fix] Disable import button when invalid flatpak path

Etaash-mathamsetty:

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1281,7 +1281,7 @@ ipcMain.handle(
       error: 'No flatpak sandbox access to this path'
     }
 
-    if (isAccessibleWithinFlatpakSandbox(path as Path))
+    if (isAccessibleWithinFlatpakSandbox(Path.parse(path)))
       moveRes = await gameManagerMap[runner].moveInstall(appName, path)
     if (moveRes.status === 'error') {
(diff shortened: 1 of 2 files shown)


src/backend/main.ts
```diff
@@ -1275,7 +1276,13 @@ ipcMain.handle(
     const { title } = gameManagerMap[runner].getGameInfo(appName)
     notify({ title, body: i18next.t('notify.moving', 'Moving Game') })
 
-    const moveRes = await gameManagerMap[runner].moveInstall(appName, path)
+    let moveRes: InstallResult = {
+      status: 'error',
+      error: 'No flatpak sandbox access to this path'
+    }
+
+    if (isAccessibleWithinFlatpakSandbox(path as Path))
```
CommandMC, line 1284: Doing `path as Path` means that `path` won't actually be verified to be a `Path`. Use `Path.parse(path)` instead (or perhaps `safeParse`)

src/backend/main.ts
CommandMC, line 1: I am a bit confused as to why `moveInstall` is touched here. The "Choosing a path for the imported game install" logic happens solely in the frontend (see `src/frontend/helpers/library.ts#84`)

PR #4113, opened 2024-11-14
Fix logging with proton

Etaash-mathamsetty:
Only tested with UMU enabled so far

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -960,6 +960,19 @@ function setupWineEnvVars(gameSettings: GameSettings, gameId = '0') {
     // This sets the name of the log file given when setting PROTON_LOG=1
     ret.SteamGameId = `heroic-${gameId}`
     ret.PROTON_LOG_DIR = flatPakHome
+    // add back default wine/dxvk debug logging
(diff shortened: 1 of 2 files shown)


arielj:
"fix" in what sense? what was broken? can you elaborate?

CommandMC:
Proton's log don't contain the output of Wine/DXVK/VKD3D-Proton. I was under the impression that this was un-resolveable, but it seems the solution was quite simple after all

IMO this should be a change inside UMU itself. It's going to be enabled in Heroic by default at some point in the future (so logging only working with UMU enabled isn't a problem then), and I expect that Lutris (along with other UMU clients) would require the same / a similar change

Etaash-mathamsetty:
Superseded by https://github.com/Open-Wine-Components/umu-launcher/pull/279

Etaash-mathamsetty:
> Superseded by [Open-Wine-Components/umu-launcher#279](https://github.com/Open-Wine-Components/umu-launcher/pull/279)

was reverted because it reduced performance (should have been opt-in not opt-out)

arielj (approved):

flavioislima (approved):

PR #4164, opened 2024-12-07
Enable macOS universal build

LenaWil:
Creates a build that runs on both architectures.

I feel like this should break things but small tests I did it seems to work?

---

### DIFF
--- a/.github/workflows/build-base.yml
+++ b/.github/workflows/build-base.yml
@@ -75,7 +75,7 @@ jobs:
       - name: Upload macOS Universal version
         uses: actions/upload-artifact@v4
         with:
-          name: mac-arm64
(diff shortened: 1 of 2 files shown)


CommandMC:
Excuse my inexperience, but do we still want to keep the separate x86_64 and arm64 builds? Wouldn't the universal build alone suffice?

LenaWil:
Honestly could be, I don’t really care, but some people do like the smaller size and I have no idea how the auto update works; I don’t know what y’alls opinion on universal builds are and why they weren’t enabled; figured the easiest way to get an answer would be to send a pull request. 🙃

arielj:
Closing, we'll keep the builds for the different architectures to keep the files smaller

PR #4178, opened 2024-12-15
Fix: Remove sideloaded games from recent list when uninstalled

### LINKED ISSUE #4170: Recent entries in tray icon can't be removed after uninstalling a sideloaded game
### Describe the bug

The Tray icon shows the last 5 launched games. For games owned in stores and for installed games, we can remove them from that list by right clicking the element in the Recently Played area in the library > Remove from recent.

This is not an option for sideloaded games, as they are not displayed anymore in that list.

### Add logs

```text
nothing relevant in logs
```

### Steps to reproduce

1. Install a game with `Add Game`
2. Launch it
3. It should be listed in the Tray icon menu
4. Uninstall the game
5. It's still in the Tray icon menu and there's no way to remove it

### Expected behavior

The game should not be displayed in the Tray icon menu if it's not available anymore.

It should either be removed from the `recent` list during the uninstall, or it should be ignored when populating the Tray icon menu if not present anymore.

### Screenshots

_No response_

### Heroic Version

Latest Stable

### System Information

not relevant

### Additional information

_No response_

arielj:
We have an issue that we can only manually remove elements from the recent list in the tray icon when they are either installed or part of the games in the library, but uninstalled sideloaded games disappear completely.

This PR doesn't fix the full problem but I think it's good enough for most use cases:
- It does remove the item when uninstalled
- It does NOT cleanup the list if it has old items (I don't think it's worth including code to check the list and remove entries, we can handle those cases manually)

To clean the list with already old elements, we can instruct users to edit the file at `~/.config/heroic/store/config.json` and remove elements they don't want from the `"recent"` key.

Fixes #4170 

---

### DIFF
--- a/src/backend/storeManagers/sideload/games.ts
+++ b/src/backend/storeManagers/sideload/games.ts
@@ -24,6 +24,7 @@ import { launchGame } from 'backend/storeManagers/storeManagerCommon/games'
 import { GOGCloudSavesLocation } from 'common/types/gog'
 import { InstallResult, RemoveArgs } from 'common/types/game_manager'
 import { removePrefix } from 'backend/utils/uninstaller'
+import { removeRecentGame } from 'backend/recent_games/recent_games'
 
 export function getGameInfo(appName: string): GameInfo {
   const store = libraryStore.get('games', [])
@@ -124,6 +125,7 @@ export async function uninstall({
   notify({ title, body: i18next.t('notify.uninstalled') })
 
   removeShortcutsUtil(gameInfo)
+  removeRecentGame(appName)
 
   sendGameStatusUpdate({
     appName,


src/backend/storeManagers/sideload/games.ts
```diff
@@ -128,6 +129,7 @@ export async function uninstall({
   notify({ title, body: i18next.t('notify.uninstalled') })
 
   removeShortcutsUtil(gameInfo)
+  removeRecentGame(appName)
```
Etaash-mathamsetty, line 132: could you add awaits and `await removeNonSteamGame({ gameInfo })` as well ?

arielj: I understand we intentionally don't await for these things, they can happen async and makes it faster

CommandMC: Could still wrap 'em in an `await Promise.all([ ... ])` to at least run them in parallel

arielj: aren't those already running in parallel? I think we really don't care about waiting for the result of those calls at all

is there any benefit I'm missing?

Etaash-mathamsetty: I thought it would be better if everything was removed by the time this function returned, but it's not necessary

flavioislima (approved):

src/backend/storeManagers/sideload/games.ts
flavioislima: I dont think we need await there tbh, I agree with Ariel that this type of thing is fine to be async since it will take less than a second and should not block anything

PR #4185, opened 2024-12-19
[UX/Tech] Use Proton-GE instead of Wine-GE as default and as recommended

arielj:
The other day I made a PR enabling UMU by default.

With that in place we can now make Proton-GE the recommended/preferred compatibility layer to be used.

I updated:
- the default if no wine is present
- updated the wine manager to make it the first tab
- changed the copy at the top of that page to reflect the new recommendation

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -911,8 +911,8 @@
         "size": "Size"
     },
     "wineExplanation": {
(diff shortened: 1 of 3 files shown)


CommandMC (approved):
I've yet to test this on a clean-install, but code LGTM

PR #4188, opened 2024-12-20
Launch game from the backend when processing launch protocol without sending a message to the frontend

arielj:
We've always had a weird/complicated flow to start a game when using the `heroic://launch` protocol: main.ts would call the `handleProtocol` function, that would eventually call `sendFrontendMessage`, then the frontend would receive a message to launch the given game and that will trigger another call to the backend to then be handled by `ipcMain.on('launch', ...)`

We can avoid all that extra work by calling the launch code directly from the backend.

I used this opportunity to extract the logic of the `launch` handler and put it into `launcher.ts` where it makes more sense, that helps removing imports and code from `main.ts` that is already too big and complicated.

The `launchEventCallback` function is just the code from `main.ts` as a function, there's nothing to review in there cause it's literally the same code.

This should speed things a biiiiiit when launching games with the protocol, but I doubt is significant.

---

### DIFF
--- a/src/backend/api/library.ts
+++ b/src/backend/api/library.ts
@@ -64,14 +64,6 @@ export const onProgressUpdate = (
   }
 }
 
-export const handleLaunchGame = (
-  callback: (
-    event: Electron.IpcRendererEvent,
-    appName: string,
-    runner: Runner
-  ) => Promise<{ status: 'done' | 'error' | 'abort' }>
-) => ipcRenderer.on('launchGame', callback)
-
 export const handleInstallGame = (
   callback: (
     event: Electron.IpcRendererEvent,
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1712,7 +1712,5 @@ export {
   callRunner,
   getRunnerCallWithoutCredentials,
   getWinePath,
-  runAfterLaunchScript,
-  runBeforeLaunchScript,
   launchEventCallback
 }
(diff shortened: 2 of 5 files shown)


CommandMC:
The round trip actually had a reason: Prompting the user to update the game in case there is an update. It doesn't seem like this is happening now?

arielj:
> The round trip actually had a reason: Prompting the user to update the game in case there is an update. It doesn't seem like this is happening now?

The `launch` flow when using the protocol seems to always try to launch the game as if there were no updates, check this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/frontend/state/GlobalState.tsx#L793 it has `hasUpdate: false` hardcoded.

It doesn't have the same logic to ask about updates that we have when clicking the `play` button.

Etaash-mathamsetty:
> > The round trip actually had a reason: Prompting the user to update the game in case there is an update. It doesn't seem like this is happening now?
> 
> The `launch` flow when using the protocol seems to always try to launch the game as if there were no updates, check this [`main`/src/frontend/state/GlobalState.tsx#L793](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/frontend/state/GlobalState.tsx#L793) it has `hasUpdate: false` hardcoded.
> 
> It doesn't have the same logic to ask about updates that we have when clicking the `play` button.

I think I had a fix for this somewhere, I don't know what I did with it lol

arielj:
Maybe that can be a different PR? this one is just to not go to the frontend, currently we don't handle updates in that flow

CommandMC (approved):

PR #4191, opened 2024-12-21
[UX/Fix] Respect experimental UMU support disabled

arielj:
A few days ago I made this PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4175 making UMU the default when running games with Proton.

There's a problem with that PR: if a game was running with Proton and UMU disabled, updating heroic will make the game run with UMU incorrectly (affecting its prefix irreversibly, maybe breaking it if the game is incompatible with UMU).

To avoid that, we have to also check if UMU was not used before for games that are already installed.

So: for new games, the `disableUMU` setting will be `false` by default or `true` if changed by the user, so it will ignore the old experimental feature value. For games installed before this change, the `disableUMU` setting will be `undefined` so we check the experimental feature value.

How to test:
- revert some commit to have the experimental UMU and disable it
- install a game and run it once using proton without the experimental feature
- update to this commit
- run the game again
- it should NOT use UMU (even though the `Disable UMU` check is disabled in the advanced tab)

- revert some commit to have the experimental UMU and enable it
- install a game and run it once using proton with the experimental feature
- update to this commit
- run the game again
- it should use UMU

---

### DIFF
--- a/src/backend/game_config.ts
+++ b/src/backend/game_config.ts
@@ -275,6 +275,10 @@ class GameConfigV0 extends GameConfig {
       // read game's settings
       const settings = JSON.parse(readFileSync(this.path, 'utf-8'))
       gameSettings = settings[this.appName] || ({} as GameSettings)
+    } else {
+      // only set the `disableUMU` default value when getting settings for a new game
+      // we want it to be `undefined` for games that were installed already
+      defaultSettings.disableUMU = false
     }
 
     if (!isWindows) {
--- a/src/backend/utils/compatibility_layers.ts
+++ b/src/backend/utils/compatibility_layers.ts
@@ -559,7 +559,7 @@ export async function isUmuSupported(
 ): Promise<boolean> {
   if (!isLinux) return false
   if (gameSettings.wineVersion.type !== 'proton') return false
-  if (gameSettings.disableUMU === false) return false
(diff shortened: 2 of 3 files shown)


CommandMC:
Are there actually cases where games break if umu is applied after the prefix is already created? I understand not using umu on installed games would be the safer bet, but do we really expect the average user to go out and manually enable umu for all of their existing games?

In my mind, we'll eventually remove the "Disable umu" checkbox completely, and make it the only valid option of running Proton (as it's the only wrapper running it like Steam does). So we'll have to bite the "Upgrading old prefixes" bullet at some point

arielj:
users won't have to manually enable UMU, it will be enabled by default for all new games and if the users where already using the experimental feature

but at least for now we need the `Disable UMU` checkbox because some games don't work with UMU at all (there are some cases in discord)

I don't know exactly what happens if you enable UMU for a proton prefix that was not created with UMU before, but I know some games don't work with UMU so we don't want to enable UMU for those

Maybe it doesn't break the prefix and a user can then go and `Disable UMU` if needed (they'll have to know that's what broke their game), but I can't confirm the prefix doesn't change in a way that keeps the game broken after disabling UMU, but knowing some games don't work with UMU I'd rather not risk it.

flavioislima (approved):
LGTM

PR #4194, opened 2024-12-21
[Fix] Show Crossover bottle in installed info details

arielj:
We have a bug on Mac, when using Crossover, the game details page shows the wineprefix instead of the bottle name in the Installed Info area, when wine prefix makes no sense in that case.

There was code already for that but it was bugged.

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/components/InstalledInfo.tsx
+++ b/src/frontend/screens/Game/GamePage/components/InstalledInfo.tsx
@@ -65,8 +65,7 @@ const InstalledInfo = ({ gameInfo }: Props) => {
       wine = wine.split('-')[0]
     }
     wineName = wine
-    wineType =
-      wineVersion.type === 'crossover' ? wineCrossoverBottle : winePrefix
+    wineType = wineVersion.type
   }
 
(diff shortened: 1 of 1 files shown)


src/frontend/screens/Game/GamePage/components/InstalledInfo.tsx
```diff
@@ -111,10 +110,10 @@ const InstalledInfo = ({ gameInfo }: Props) => {
           <div>
             <b>Wine:</b> {wineName}
           </div>
-          {wineVersion && wineType === 'crossover' ? (
+          {wineType === 'crossover' ? (
             <div>
               <b>{t2('setting.winecrossoverbottle', 'Bottle')}:</b>{' '}
-              <div className="truncatedPath">{winePrefix}</div>
+              <div>{wineCrossoverBottle}</div>
```
CommandMC, line 116: This isn't really related to this PR, but I think we should do these types of translations a little differently. Not every language will follow the "Label: Value" format

CommandMC (approved):
Can't test this, but sounds & looks fine to me

PR #4214, opened 2024-12-27
[UX] Enable/disable logs game-by-game instead of globally

arielj:
Currently, we can only enable/disable logs globally to log game output.

This has some issues:
- the setting to disable all logs is not next to the game's settings, so it's not easy to find for users
- logs can impact performance (specially for games with a lot of output) and disk usage, having to enable logs for all games to debug one game means a user can forget to disable it after debugging one game and it will silently apply to all games, logging unnecessary information for games that are already working
- we were enabling all logs by default, which meant users had to know they can disable them

This PR changes that so we can enable `verbose logs` in a game-by-game basis. Some benefits:
- If I forget the logs "on" for a game, it won't affect the other games
- The setting is now in the Advanced tab in the game's settings, so it's easier to find
- Now the log file shows a message telling the user that logs are disabled, and tells them to enable verbose logs before reporting a problem

This also enables me to do another change I want to do related to logs and UX, but I'll keep that in another PR to keep this one smaller.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -162,13 +162,6 @@ const launchEventCallback: (args: LaunchParams) => StatusPromise = async ({
 
   initGamePlayLog(game)
 
-  if (logsDisabled) {
-    appendGamePlayLog(
-      game,
-      'IMPORTANT: Logs are disabled. Enable logs before reporting an issue.'
-    )
-  }
-
   const isNative = gameManagerMap[runner].isNative(appName)
 
   // check if isNative, if not, check if wine is valid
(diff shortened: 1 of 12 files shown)


src/backend/storeManagers/storeManagerCommon/games.ts
```diff
@@ -262,7 +261,7 @@ export async function launchGame(
         logFile: lastPlayLogFileLocation(appName),
         logMessagePrefix: LogPrefix.Backend,
         onOutput: (output) => {
-          if (!logsDisabled) appendGamePlayLog(gameInfo, output)
```
Etaash-mathamsetty, line 265: probably should add that
```
  if (!gameSettings.verboseLogs) {
    appendGamePlayLog(
      gameInfo,
      'IMPORTANT: Logs are disabled. Enable verbose logs before reporting an issue.'
    )
  }
  ```
  here as well ?

**Review by Etaash-mathamsetty — COMMENTED** [review]
lgtm other than that one small thing

File: src/backend/launcher.ts [hunk 2]
```diff
@@ -1461,23 +1461,27 @@ async function runScriptForGame(
...
+
+      child.stderr.on('data', (data) => {
+        appendGamePlayLog(gameInfo, data.toString())
+      })
+    }
 
-    child.stderr.on('data', (data) => {
-      appendGamePlayLog(gameInfo, data.toString())
+    child.on('error', (err: Error) => {
+      if (gameSettings.verboseLogs) {
```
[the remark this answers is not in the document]
arielj: My only concern with this is that users may see an error message and miss the `IMPORTANT: Logs are disabled. Enable verbose logs before reporting an issue` line and send it anyway without turning on logs (I know it's going to happen anyway, but I think not having anything apart from that line would reduce it the chances)

src/backend/storeManagers/storeManagerCommon/games.ts
arielj: hmmm looks like we have some issues with how we log things with sideloaded games, for native games the logs are just broken (we log the wrong thing in the wrong file and no system info / config)

I'm trying to fix that, but feel that should be its own PR to limit the scope of this one

flavioislima (approved):
LGTM

PR #4220, opened 2024-12-29
[UX] Alt action to launch games with and without logs

arielj:
This PR adds an alternative "Play Now (with logs)" action right next to the normal `Play Now` button in the details page behind a small dropdown.

The idea is that users can easily select to launch a game with and without logs enabled for debugging purposes (and it's easier to instruct to do that and easier to discover).

The main action will be the last one used so users can clearly see if they are running the game with/without logs as a reminder that they can disable them, and both actions will change the game's settings so it's persisted (this way it can be used for things like running as a non-steam game and log things at the same time).

This depends on 

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/components/MainButton.tsx
+++ b/src/frontend/screens/Game/GamePage/components/MainButton.tsx
@@ -2,6 +2,7 @@ import React, { useContext } from 'react'
 import { useTranslation } from 'react-i18next'
 import GameContext from '../../GameContext'
 import {
+  ArrowBackIosNew,
   Cancel,
   CloudQueue,
   Download,
@@ -13,6 +14,7 @@ import {
 } from '@mui/icons-material'
 import classNames from 'classnames'
 import { GameInfo } from 'common/types'
+import useSetting from 'frontend/hooks/useSetting'
 
 interface Props {
   gameInfo: GameInfo
@@ -25,6 +27,7 @@ interface Props {
 const MainButton = ({ gameInfo, handlePlay, handleInstall }: Props) => {
   const { t } = useTranslation('gamepage')
   const { is } = useContext(GameContext)
+  const [verboseLogs, setVerboseLogs] = useSetting('verboseLogs', false)
 
   function getPlayLabel(): React.ReactNode {
     if (is.syncing) {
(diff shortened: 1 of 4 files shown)


src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -348,144 +357,86 @@ export default React.memo(function GamePage(): JSX.Element | null {
...
 
     return (
-      <div className="gameConfigContainer">
-        {!!(art_background ?? art_cover) &&
-          experimentalFeatures.enableNewDesign && (
-            <CachedImage
-              src={art_background || art_cover}
-              className="backgroundImage"
+      <SettingsContext.Provider value={settingsContextValues}>
+        <div className="gameConfigContainer">
```
arielj, line 361: the diff here is a mess, but I'm just wrapping things in a SettingsContext wrapper, nothing changed inside this

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -274,6 +274,40 @@
       max-width: 560px;
     }
 
+    & .playButtons {
```
arielj, line 277: this CSS is partially duplicated on purpose for both old and new design, eventually we'll remove the `old` styles and it's easier to do that with duplications

src/backend/config.ts
```diff
@@ -323,7 +323,7 @@ class GlobalConfigV0 extends GlobalConfig {
       beforeLaunchScriptPath: '',
       afterLaunchScriptPath: '',
       disableUMU: false,
-      verboseLogs: false
+      verboseLogs: true
```
arielj, line 326: enabled by default, since they can be disabled+play at the same time with a single action now

but by default we can get more logs for new users

src/frontend/screens/Game/GamePage/components/MainButton.tsx
```diff
@@ -62,6 +74,48 @@ const MainButton = ({ gameInfo, handlePlay, handleInstall }: Props) => {
...
+    }
+
+    return (
+      <span className="buttonWithIcon">
+        <PlayArrow data-icon="play" />
+        {t('label.playing.start_with_logs', 'Play Now (with logs)')}
+      </span>
+    )
+  }
+
```
CommandMC, line 118: Consider making these memoized values to avoid computing them every re-render

src/frontend/screens/Game/GamePage/components/MainButton.tsx
```diff
@@ -62,6 +74,48 @@ const MainButton = ({ gameInfo, handlePlay, handleInstall }: Props) => {
     )
   }
 
+  function showAltPlayAction() {
+    if (
+      is.syncing ||
+      is.installingRedist ||
+      is.installingWinetricksPackages ||
+      is.launching ||
+      is.playing
```
CommandMC, line 83: Maybe factor this out into its own variable? Seems like this condition (although with some variation on which `is` values are considered) is used 3 times now

src/frontend/screens/Game/GamePage/components/MainButton.tsx
```diff
@@ -109,45 +163,60 @@ const MainButton = ({ gameInfo, handlePlay, handleInstall }: Props) => {
...
+              is.moving ||
+              is.updating ||
+              is.uninstalling ||
+              is.syncing ||
+              is.launching ||
+              is.installingWinetricksPackages ||
+              is.installingRedist
+            }
+            autoFocus={true}
+            onClick={async () => handlePlay(gameInfo)}
```
CommandMC, line 207: These moved from before the `className` prop to after it. Not a problem, just wanted to mention it since it seems unintentional (nothing apart from the position was changed)

src/frontend/screens/Game/GamePage/index.tsx
CommandMC: The "Hide whitespace" checkbox in the settings of GitHub's diff view is helpful here

CommandMC (commented):
A couple of nitpicks
The feature itself sounds & looks good to me!

flavioislima (approved):
Tested here, looks good.

PR #4221, opened 2024-12-30
Stop automatically downloading dxvk

Etaash-mathamsetty:
It's not needed for people purely using proton, so avoid installing it for everyone. 


arielj (commented):
looks good to me

this means it's not downloaded at boot but it will be downloaded when needed the first time, right?

arielj (approved):

PR #4224, opened 2024-12-30
Introduce Enable Gamescope toggle

Etaash-mathamsetty:
Should fix some of the confusion that was caused by the additional options text field

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -590,8 +590,10 @@ async function prepareLaunch(
   }
 
   if (
-    (gameSettings.gamescope?.enableLimiter ||
-      gameSettings.gamescope?.enableUpscaling) &&
+    (gameSettings.gamescope?.enableGamescope === true ||
+      (gameSettings.gamescope?.enableGamescope === undefined &&
+        (gameSettings.gamescope?.enableUpscaling ||
+          gameSettings.gamescope?.enableLimiter))) &&
     !isSteamDeckGameMode
   ) {
     const gameScopeBin = await searchForExecutableOnPath('gamescope')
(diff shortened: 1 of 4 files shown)


src/backend/launcher.ts
```diff
@@ -143,11 +143,7 @@ async function prepareLaunch(
     }
   }
 
-  if (
-    (gameSettings.gamescope?.enableLimiter ||
-      gameSettings.gamescope?.enableUpscaling) &&
-    !isSteamDeckGameMode
-  ) {
+  if (gameSettings.gamescope?.enableGamescope && !isSteamDeckGameMode) {
```
arielj, line 146: with this change we'll be disabling gamescope for all users that currently have it configured wit the other checkboxes until they go and manually enable it

Maybe we need some code setting the initial `enableGamescope` setting to be the value of `gameSettings.gamescope?.enableLimiter || gameSettings.gamescope?.enableUpscaling`. Or maybe we can type it as `boolean | null` with `null` meaning `it was never set, use the previous logic`?

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -115,20 +116,36 @@ const Gamescope = () => {
...
+                ...gamescope,
+                enableUpscaling: !gamescope.enableUpscaling
+              })
+            }
+            title={t('setting.gamescope.enableUpscaling', 'Enables Upscaling')}
+          />
+        </div>
+      )}
       {/* Upscale Settings */}
       {gamescope.enableUpscaling && (
```
arielj, line 150: I wonder if the upscaling/limiter options should also be conditionally rendered with with gamescope enable check

because with the new code I think you can:
- enable gamescope
- enable upscaling
- disable gamescop
- upscaling checkbox is gone but upscaling options will still be visible

arielj (changes requested):

src/backend/launcher.ts
BananaWorks07: I used undefined in the new version

PR #4226, opened 2024-12-31
[UX] Wait until winetricks lists before allowing the user to search

arielj:
When opening the Winetricks dialog, if a user starts typing too fast in the input field before we get the list of valid components/fonts then it will look like nothing is found. Many users reported that in discord.

This PR changes the dialog to show a message while we are loading the list of things available so users can't type while it's loading, and then we allow searching only once it's ready.

This should make it less confusing for users.

---


Etaash-mathamsetty (approved):
lgtm (and it works when I tested it)

CommandMC (approved):

PR #4235, opened 2025-01-03
Ass support for Automatic creation of steam shortcuts for uninstalled games

lostb1t:
continuation of ##4234

### DIFF
--- a/src/backend/storeManagers/gog/library.ts
+++ b/src/backend/storeManagers/gog/library.ts
@@ -28,6 +28,10 @@ import {
 import { dirname, join } from 'node:path'
 import { existsSync, readFileSync } from 'graceful-fs'
 import { app } from 'electron'
+import {
+  addShortcuts,
+  removeShortcuts
+} from '../../shortcuts/shortcuts/shortcuts'
 
 import {
   logDebug,
@@ -385,6 +389,7 @@ export async function refresh(): Promise<ExecResult> {
   const filteredApiArray = gameApiArray.filter(
     (entry) => entry.platform_id === 'gog'
   )
+  console.log(filteredApiArray.length)
 
   const gamesObjects: GameInfo[] = [redistGameInfo]
   apiInfoCache.use_in_memory() // Prevent blocking operations
@@ -441,6 +446,10 @@ export async function refresh(): Promise<ExecResult> {
       library.set(copyObject.app_name, copyObject)
       break
     }
+  }
+
+  for (const game of gamesObjects) {
+    await addShortcuts(game, false)
   }
 
   apiInfoCache.commit() // Sync cache to drive
--- a/src/frontend/screens/Settings/components/Shortcuts.tsx
+++ b/src/frontend/screens/Settings/components/Shortcuts.tsx
@@ -26,6 +26,9 @@ const Shortcuts = () => {
     false
   )
 
+  const [addSteamShortcutsUninstalled, setAddSteamShortcutsUninstalled] =
+    useSetting('addSteamShortcutsUninstalled', false)
+
   if (!isDefault) {
     return <></>
   }
@@ -67,6 +70,17 @@ const Shortcuts = () => {
         value={addSteamShortcuts}
         handleChange={() => setAddSteamShortcuts(!addSteamShortcuts)}
         title={t('setting.addgamestosteam', 'Add games to Steam automatically')}
+      />
+      <ToggleSwitch
+        htmlId="shortcutsToSteamUninstalled"
+        value={addSteamShortcutsUninstalled}
+        handleChange={() =>
+          setAddSteamShortcutsUninstalled(!addSteamShortcutsUninstalled)
+        }
+        title={t(
+          'setting.addgamestosteamuninstalled',
+          'Add uninstalled games to Steam automatically'
+        )}
       />
     </>
   )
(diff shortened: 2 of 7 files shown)


src/backend/storeManagers/gog/library.ts
```diff
@@ -443,6 +445,10 @@ export async function refresh(): Promise<ExecResult> {
     }
   }
 
+  for (const game of gamesObjects) {
```
arielj, line 448: maybe, instead of checking `addSteamShortcutsUninstalled` for every game in `addShortcuts`, you could check that once here and then if it's try you do the loop

src/backend/storeManagers/gog/library.ts
```diff
@@ -385,6 +386,7 @@ export async function refresh(): Promise<ExecResult> {
   const filteredApiArray = gameApiArray.filter(
     (entry) => entry.platform_id === 'gog'
   )
+  console.log(filteredApiArray.length)
```
arielj, line 389:
```suggestion
```

src/backend/storeManagers/gog/library.ts
arielj: also, won't this trigger one system notification per game? it can be really annoying to turn this on, restart heroic and start seeing maybe hundreds of notifications popping up

arielj: there's also A LOT of work happening for every game: getting wiki game info, downloading the images, checking if steam files exist, etc

I think using `addShortcuts` in a loop is not scalable (there's users with 1000+ games), we would need a different method, something that adds all games and then saves the shortcuts.vdf once

src/frontend/screens/Settings/components/Shortcuts.tsx
```diff
@@ -68,6 +71,17 @@ const Shortcuts = () => {
...
       />
+      <ToggleSwitch
+        htmlId="shortcutsToSteamUninstalled"
+        value={addSteamShortcutsUninstalled}
+        handleChange={() =>
+          setAddSteamShortcutsUninstalled(!addSteamShortcutsUninstalled)
+        }
+        title={t(
+          'setting.addgamestosteamuninstalled',
+          'Add uninstalled games to Steam automatically'
```
arielj, line 82: this feature seems to add `all games` to steam, even installed ones that were not added before

src/backend/storeManagers/gog/library.ts
arielj: also, having each library running its own `for` loop and all of them writing to the `shortcuts.vdf` file at the same time in parallel can create a bunch of race conditions (what if while the gog loops is trying to add a game, the legendary loop saved a file, then the gog loops overrides it?)

or if I refresh the library while the games are being added, then this would run again for the same store in parallel

arielj: finally, I don't think it's scalable to wait for all the shortcuts to be added to continue with the refresh call, this will delay showing the library in heroic

arielj (commented):
This process is something that can take quiet a lot of time. For example, I have over 600 games, turning this on would start a loop doing adding the games one by one in the background and if a game takes 0.5s to be added (we have to get pcgamingwiki data, then download the images, edit steam files, save) it would take 5 minutes to complete with no visible indication of the progress (and no way to stop it)

I think a better approach to do something like this at least as an initial version of this feature would be to add a button `Add all games to Steam` that opens a dialog and shows progress. That way it's more manageable and can solve some of the issues I mentioned in other comments. I imagine a dialog shows up telling you `You are about to add these games to Steam` or something like that, add some information about not closing the dialog maybe, and a `confirm` button. then that action could loop through the libraries and start adding games in batches maybe and show some progress indicator.

I think this would still need some rework, would the pcgamingwiki and steamgrid be ok with this amount of requests in a short period of time? would they stop us with some rate limiting? what happens if some of those requests fail?

lostb1t:
I have around 600 games aswell (jej epic games ;)

running this sync doesnt cause any rate limit issues but to speed up the process running this async would be a good idea.

I still like this to be part of the refresh process to be honest. Don't see the value of being it a separate thing and making it extra work for the user 

Heres what we can do:

1. Make shortcut request async (will look into rate limits to determine how many in parallel)
2. Optimize the shortcut process, maybe have a seperate function for it that ditches things howlongtobeat etc
3. I assumed there would already be a process indicator for import and refresh (something somethings assumptions). If this isn't the case i could look into giving these processes an process indicator.

kmturley:
I am also very keen for this solution!

Looking through the code, it is better than my prototype solution, but I think we can improve it much more.

**1) During the library refresh, loop through every game with an await method which blocks the thread and delays the library display**
Create a new backend method which is called independently from the library refresh.

**2) Checking if a game has already been added to steam, occurs after getting metadata from APIs**
Ideally this should happen as soon as possible to reduce the number of operations and requests.

**3) When getting data for the shortcuts file it is getting more data than needed, and might not be cached/stored locally**
Only request the data required for a shortcut to be added. Cache/store the data for the next time the library is refreshed

**4) When reading and writing shortcuts, it performs the operation for every game separately**
Batch the shortcuts together and perform one operation to reduce.

kmturley:
I've created a new version following the above list of guidance as best I could:
https://github.com/kmturley/HeroicGamesLauncher/commit/7e4d5eb37e818007a619635429f686cc569258e7

arielj:
Closing due to inactivity. We can reopen if the comments are addressed

PR #4237, opened 2025-01-04
Remove "Use Steam Runtime"

CommandMC:
This option had two use cases:
- You're running a Windows game with Proton
- You're running a Linux-native game

Umu now covers both of these use cases. Running Proton with umu is already implemented & done by default, while Linux-native apps still require work in Heroic.
I would however argue that almost no-one used Steam RT for Linux-native games (Linux native games themselves are rare, them *working* is even rarer, and this option being found by the user & helping virtually does not happen). If demand for this option is indeed there, we should look into using umu for it.

---


arielj:
If there are users playing games that were fixed using this, those games will break.

Maybe it's better to only support the steam runtime for linux native games instead of completely removing it? or is UMU capable of handling those cases? in that case we would need a way to migrate those users from using the steam runtime to using UMU automatically

without that this PR can break games that were already working, I know many discord support threads with native games are solved by enabling this so there are users using this options

CommandMC:
> Maybe it's better to only support the steam runtime for linux native games instead of completely removing it? or is UMU capable of handling those cases?

Umu can run arbitrary Linux executables in the runtime, yes. This is however only a feature in 1.1.4, while the version in Lutris' API (which we use) is 1.1.3. So a first step would have to be to update that.

> in that case we would need a way to migrate those users from using the steam runtime to using UMU automatically

That would be ideal, yes

> I know many discord support threads with native games are solved by enabling this so there are users using this options

Huh, must've missed those. I honestly never saw anyone use this option (at most using it with Proton)

---

I'll go poke the umu guys to update the release then. After that I'll update this PR to migrate Linux-native apps from Steam RT to umu

arielj:
> Huh, must've missed those. I honestly never saw anyone use this option (at most using it with Proton)

well, when I say `many` I meant `many of the linux native reports`, not that there are many reports like that but I do remember seeing that

I have one in my notes that got fixed with the steam runtime too, Mana Spark linux native, so I could test this once updated

CommandMC:
I've now updated this approach to remove Steam Runtime & prefer umu. To make this work for native games, we have to wait on umu to implement the Steam Runtime 1.0 (`scout-on-soldier`), which requires more work than one might expect, since it's one runtime (`scout`) inside another (`soldier`). https://github.com/Open-Wine-Components/umu-launcher/pull/293 is working on layered runtime support

aleex5:
Ragnarok Online an old MMORPG. They are launching a new server in Latin America. In its beta, it worked fine with umu-Launcher; without umu-Launcher, it didn't work due to the anticheat it has, which is GameGuard. I don't know why it works with these runtimes. I don't know if it was an anticheat update or the latest version of Heroic, because both coincided when it started crashing. So I tried Steam-Runtime and it worked again.

CommandMC:
Closing in favour of https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5832

Issue #2239, opened 2022-12-20
Download progress / pause issue


d3d43:
### Describe the bug

I paused a download due to changing vpn server. However after this, i cannot tell if the download restarted from the beginning or what. 

The download progress bar as of now appears that the download restarted from the beginning, but viewing previous threads suggests otherwise and that heroic has the ability to pause a download in between and resuming after. 

Despite this, the completed section appears that the previous download which i left overnight for has been cancelled, and that the current download says 59% complete which is representative of the currently downloaded file size. All this happened despite the fact that i pressed "Keep downloaded files" when attempting to pause. 

The issue is quite frustrating as i feel like i have wasted 6 hours trying to download a file only to be wiped.

### Add logs

```text
not available. comment too long
```

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Expected behavior

Display correct percentage of download, if the pause download was successful.

Display real pause button that functions properly and does not result in confusion for the end user.

### Screenshots

<img width="1447" alt="image" src="https://user-images.githubusercontent.com/47383653/208586150-6cde25e2-6b85-4dcc-b1f3-272cc9ce4341.png">

### Heroic Version

Latest Stable

### System Information

Macos monterey 12.3.1
Apple m1 macbook air

ElmeriVincent:
You are right. It is a bit confusing currently.

If the user chooses to keep the files, it should not say (canceled). It should say (Paused). Currently when the user clicks on the "canceled" game it will display **Install** option afterwards but instead it should display **Continue Download**.  Paused game shouldn't be displayed in the completed section but rather in it own (paused section).

AlexBoXY:
I think you need another section. It will be clear, when you rename the cancel-button to Pause and if paused, put the game back into the queue section with a continue and abort/remove button.
Less effort and understandable for everybody...

arielj:
Is the current approach better? Now we have a `Pause Download`/`Resume Download` button.

If that's enough I could close this issue.

d3d43:
> Is the current approach better? Now we have a `Pause Download`/`Resume Download` button.
> 
> If that's enough I could close this issue.

Unfortunately I am no longer using the application. Please feel free to consult other users if the thread is suited to be left up, or close the issue at your discretion.

cookiEoverdose:
It is still happening, you lose a lot of download progress on pause + resume.

jackman105:
[...]

PR #4240, opened 2025-01-04
Fix legendary download progress bar percentage for updates

### LINKED ISSUE #4190: Legendary update progress bar shows incorrect percentage
### Describe the bug

I was updating fortnite and noticed that the progress bar was stuck at 90% since half hour.

I went ahead and inspected the legendary logs and found the percentage to be vastly different from the one shown in ui. (this is shown in my screenshot down below where i paused the update and screenshotted the last entry which corresponds with what is displayed in the ui)

I tried to look into why this could be when other metrics are exactly the same as displayed in logs but different for percentage and found this snippet in the following file:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/17526abcb39738c93711d7257dfba92f9194906d/src/backend/storeManagers/legendary/games.ts#L465-L473

this is in function `onInstallOrUpdateOutput`, and from what it appears looking through rest of the code, we are parsing all other metrics such as download speed, eta, disk write speed, downloaded bytes, etc, directly from the logs, but for percentage for some reason it is being calculated in the code only.

I am not sure what is wrong with the calculation here, but I do think if we are parsing all other metrics from legendary itself, we should also parse the progress from the log rather than calculating ourselves as the legendary binary would have a better context and calculation for the correct progress, reducing the effort done on heroic side.

### Add logs

```text
< logs added as attachment as there are more than 2k lines and 12k lines for heroic & legendary respectively >
```

### Steps to reproduce

1. put a game on update
2. the progress starts from ~80% (see logs)
3. compare with legendary logs
4. you will see a mismatch between progress reported by heroic & legendary, with legendary one being more correct

### Expected behavior

the percentage value should match exactly what is outputted from legendary

### Screenshots

image_2024-12-21_02-14-52

### Heroic Version

Latest Stable

### System Information

```
System Information:
CPU: 12x 13th Gen Intel(R) Core(TM) i5-13420H
Memory: 16.89 GB (used: 8.61 GB)
GPUs:
  GPU 0:
    Name: Intel Corporation Raptor Lake-P [UHD Graphics]
    IDs: D=a7a8 V=8086 SD=3c90 SV=17aa
    Driver: 31.0.101.4255
OS: Microsoft Windows 11 Home Single Language 10.0.22631 (win32)

The current system is not a Steam Deck
We are not running inside a Flatpak container

Software Versions:
  Heroic: 2.15.2 Emeth
  Legendary: 0.20.34 Direct Intervention
  gogdl: 1.1.2
  comet: comet 0.1.2

  Nile: 1.1.2 Will A. Zeppeli
```

### Additional information

log files which corroborate the same:
[2024-12-20T19_41_48-heroic.log](https://github.com/user-attachments/files/18215387/2024-12-20T19_41_48-heroic.log)

[2024-12-20T19_41_48-legendary.log](https://github.com/user-attachments/files/18215388/2024-12-20T19_41_48-legendary.log)

RoguedBear:
closes #4190 

### Issue:
as discussed in #4190, the current method of calculating progress percentage for updates provides a misleading output. 

This PR attempts to fix this by:
- preserving the existing functionality of using already downloaded files into progress bar calculation for installs
- and ignoring download cache for updates to better represent the game update progress.

For more context and clarity, a detailed discussion is present in the linked issue

### Testing:
I am using Rocket League as an example here. I redownloaded the game from an old manifest that legendary had saved locally, then triggered an update from Heroic.

#### Before (Heroic 2.15.2)
- image_2025-01-05_01-14-15

logs: 
  - image_2025-01-05_01-16-52

#### After (my dev branch)

- image_2025-01-05_01-25-34

logs:
  - image_2025-01-05_01-25-52

---

### DIFF
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -467,8 +467,14 @@ export function onInstallOrUpdateOutput(
     const downloaded = parseFloat(progress.bytes)
     const downloadCache = totalDownloadSize - currentDownloadSize[appName]
     const totalDownloaded = downloaded + downloadCache
-    const newPercent =
-      Math.round((totalDownloaded / totalDownloadSize) * 10000) / 100
+    let newPercent: number
+    if (action === 'installing') {
+      newPercent =
+        Math.round((totalDownloaded / totalDownloadSize) * 10000) / 100
+    } else {
+      newPercent =
+        Math.round((downloaded / currentDownloadSize[appName]) * 10000) / 100
+    }
     progress.percent = newPercent >= 0 ? newPercent : undefined
   }
 


flavioislima:
   - - -
   - - -

RoguedBear:

arielj (commented):
To be sure I understand, this will match the % displayed by legendary in its output in the different cases of downloading, updating, and pausing/resuming? or it's still there something different?

because if that's the case I guess we could just change this to read the % from the output

something like:

```ts
progress.percent = data.match(/Progress: (\d+\.\d\d)%/m)
```

and replace all the calculation?

RoguedBear:
> To be sure I understand, this will match the % displayed by legendary in its output in the different cases of downloading, updating, and pausing/resuming? or it's still there something different?

From what ive seen experimenting with legendary by default all downloads are resumable regardless of it being installation or update. However with each download session the "GBs to download" decreases as it takes into account what is already downloaded, and the progress starts from 0%.


In the beginning I was thinking of adding  `progress.percent = data.match(/Progress: (\d+\.\d\d)%/m)` too but I found that the reason this calculation was introduced (in this commit https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/commit/223b91510fdd473c87cc3996eef8a755162b9ba4 ) was as a way to include already downloaded files as part of the progress percentage.

This approach works really well for new installs as the progress bar correctly reflects how much is downloaded in each pause/resume session.
But when updating a game you can see from the screenshots i posted here and in the linked issue that we get a very misleading percentage bar (as the calculation is including the already downloaded game size as part of the update, the moment actual update beings to download)


I figured a middle ground approach could work fine here where the existing behavior is preserved for new installs, but for updates do what legendary does -> `current downloaded in this session / total to download in this session` %

---



I still think that for users not familiar with legendary it _could_ cause confusion even after this PR is merged as for updates the progress starts from 0%, but I think that can be resolved if we display total size to download on the client side. 
Then the users would notice that the remaining download size decreases each time they pause/resume, and get reassurance it is actually resuming.


Also this doesn't "_exactly_" match legendary's percentage, but is very close to it. I am doing `size downloaded / total size to download`, whereas legendary does `number of files downloaded / number of files to download`.
You can verify this from the legendary logs screenshot in PR description

RoguedBear:
If there is more experimentation we can do regarding new installs, updates and pause/resuming, I would be open to help with that.

arielj:
> Also this doesn't "_exactly_" match legendary's percentage, but is very close to it. I am doing `size downloaded / total size to download`, whereas legendary does `number of files downloaded / number of files to download`. You can verify this from the legendary logs screenshot in PR description

Yeah, I didn't mean exactly, but wanted to understand is: if we are showing a similar % than legendary in the 3 different scenarios (start from beginning, resume paused/canceled, update), then we don't need the calculation, if we have some consideration to do something different I guess it's fine to keep the calculation.

One thing that I know is a problem with the calculation from Heroic's side is that since it's based on size downloaded and not in files downloaded, stopping/cancelling a download will actually discard the current files that didn't finish downloading (at least for epic games), which can be a problem with games that come with big files (I understand that's the main issue in #2239), so a user will see a % (based on size) when they want to pause, legendary will discard whatever it is downloading at this point, and when resumed the % will be different. With small files downloads it's not that much of a difference, but it's a problem in many cases with big files.

I really don't know what's the best solution, but I do think that screen is missing some information in general:
- progress based on files downloaded (and making it clear that this is the % to reference for pause/resume)
- total to be downloaded (now we have a % and amount downloaded, but it's not great UX)
- disk split into write/read

I honestly don't know what's the right answer, personally I would err on the side of showing as much information as possible, and maybe improve the `pause` message to say "THIS is the % where we will restart the download, are you sure?????" with a `(?)` to explain why it's that % (based on files) and not the other % (based on size).

arielj:
> I was doing some exploring of my own, and i couldnt find any way to know how many files are already downloaded for a resume. if we want to switch to using file count based % then we lose including download cache in the total progress we display on resume of an install.

The issue here is that we track the wrong % when resuming an install, because we track size % while legendary will actually throw away the current file being downloaded and start from that file, so if the file is big the resume % is actually lower.

> However, if for updates we want to switch to file based %, that I can try and implement that

Sorry, I honestly don't know what's better to say which one we want, maybe we just need to show all the %s we have with relevant detail of what they mean instead of picking one

> Regardless, I too would like to see as much information as is relevant. Adding to your points maybe we can even have a second % bar to keep track for number of files downloaded separate from the size % and utilise that on the warning message

Yeah, I it would be better to show show both.

> I was also thinking of maybe estimating how much data could be lost by tracking the size for the file currently being downloaded and displaying that in the pause/resume warning dialogue but it could be prone to errors as legendary might use multiple workers to download files parallely.

> But until a clear idea of solving this comes up or if legendary can export in detail about the download cache, i think a simple solution such as:
> 
> * communicating to the user in a warning as you mentioned, something like "Warning: some download data might be lost which will be redownloaded next time"
> * and displaying total size to download as mentioned in my comment and yours.
> 
> ...could work well in conveying this meaning

I don't think we have that information (about the current file being downloaded -because we can be downloading multiple files at once- and about the size of the file -I don't think legendary exposes that in the output-) so we can't really do that

I'm sure there's always room for improvement after we get the UI to show all the information, but having that displayed is a good start (also, one thing to consider is GOG/Amazon games report based on size I think, so might be weird that some downloads have more info than others, not sure if that's great)

Maybe that's for another PR, I really don't know what's the best answer. Maybe this is just to improve the update % and then something else is about showing more information in general.

RoguedBear:
I am also confused. Initially my scope of the PR was correcting only the misleading update % value. But now it feels like the entire progress bar itself is a bit misleading for legendary only for the case if pause/resume happens.

I guess until there is a proper idea and proper metrics available, we can a display warning in the UI.

I am not sure if this PR will be merged or not, but I'd be willing to work on improving progress bar in the future.

arielj:
closing, the idea is to address this in legendary directly to not have to fix the progress in heroic

PR #4244, opened 2025-01-05
[Fix] Correct names for GE-Proton and umu

CommandMC:
Proton-GE got renamed to GE-Proton [a bit ago](<https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/GE-Proton7-1>)
Secondly, "umu" is not an abbreviation of some kind, it is just [a (rather uncommon) word](<https://www.collinsdictionary.com/dictionary/english/umu>)

---


public/locales/ar/translation.json
```diff
@@ -911,7 +911,7 @@
         "noResults": "لم تسفر التصنيفات الحالية عن أية نتائج."
     },
     "wineExplanation": {
-        "proton-ge": "Proton-GE-Proton هو نوع Proton تم إنشاؤه بواسطة Glorious Eggroll. من المفترض استخدامه في Steam، لكن بعض الألعاب خارج Steam قد تعمل بشكل أفضل مع هذا المتغير. ويوفر سجلات غير مجدية في الغالب لاستكشاف الأخطاء وإصلاحها.",
+        "proton-ge": "GE-Proton هو نوع Proton تم إنشاؤه بواسطة Glorious Eggroll. من المفترض استخدامه في Steam، لكن بعض الألعاب خارج Steam قد تعمل بشكل أفضل مع هذا المتغير. ويوفر سجلات غير مجدية في الغالب لاستكشاف الأخطاء وإصلاحها.",
```
arielj, line 914: hmmm I wonder how these changes work with weblate

will this update weblate? or weblate will have the old stuff and the next PR it will override these changes again?

src/backend/utils.ts
```diff
@@ -917,10 +917,10 @@ export async function downloadDefaultWine() {
   await updateWineVersionInfos(true)
   // get list of wines on wineDownloaderInfoStore
   const availableWine = wineDownloaderInfoStore.get('wine-releases', [])
-  // use Proton-GE type if on Linux and GamePortingToolkit if on Mac
+  // use GE-Proton type if on Linux and GamePortingToolkit if on Mac
   const release = availableWine.filter((version) => {
     if (isLinux) {
-      return version.version.includes('Proton-GE-Proton')
+      return version.version.includes('GE-Proton')
```
arielj, line 923: wine-ge versions are `Wine-GE-Proton` so this `includes` will match those incorrectly

the idea for this filter is to ignore all `wine-ge` and pick `proton-ge` as a default wine/proton to suggest

maybe this needs to be changed to `startsWith`? or maybe `!version.version.includes('Wine')`?

CommandMC: I see
I wonder why we aren't just filtering on `version.type` actually

public/locales/ar/translation.json
CommandMC: I'm pretty sure that Weblate will update with the changes from here. The relation of repo<->Weblate goes both ways (at least it does for the English translations, and I don't see why other languages should be different)

src/backend/utils.ts
arielj: I know at some point  the code was filtering for `wine-ge` but not `-LoL` I think? that's all kinda all now though

arielj (approved):
Changes look good to me, I only have one comment but not a blocker (I don't know if it can be fixed): In the `Wine version` selector for games settings and default settings, it's showing the old names for protons, but the same thing shows the new name in the wine manager page

Not sure if the selectors can be updated to display the new names (I understand it gets the folder names from the tools folders, maybe the proton ones can be re-formatted during load just for visuals?).

I think it's not a problem for versions in general cause new proton releases will happen and the old names will go away eventually, but the `Proton-GE-Latest` shows `Proton-GE-Latest` unless you uninstall > install it. I can imagine a user updating their `GE-Proton-Latest` in the wine manager and then thinking it didn't work because they se `Proton-GE-Latest` in the selector.

CommandMC:
Ah yeah, I noticed the folders keeping their own names on update, but I didn't consider that the version detection also reads those folder names.
We can't really rename the actual folders, as that'll break configs, so I guess a visual solution (as you mentioned) it is.

Alternatively, we could also move away from detecting Wine Versions outside of Heroic (especially with umu & GE-Proton becoming the recommended way to run Wine) and take all data from the Wine Manager instead. But that'd involve integrating it a bit more into Heroic logic as a whole (it's quite separate at the moment, which is understandable considering it started out as its own thing)

arielj:
I pushed an updated to the PR to rename the select option text to replace `Proton-GE-Proton` to `GE-Proton` and `Proton-GE-latest` to `GE-Proton-latest`.

The change is only visual, internally the setting has the correct name and matching folder with the disk.

PR #4251, opened 2025-01-07
fix(gog): make sure game directory is mounted in container

imLinguin:

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1100,6 +1100,7 @@ async function runWineCommand({
   wait,
   protonVerb = 'run',
   installFolderName,
+  gameInstallPath,
   options,
   startFolder,
   skipPrefixCheckIKnowWhatImDoing = false,
@@ -1153,7 +1154,7 @@ async function runWineCommand({
   const env_vars: Record<string, string> = {
     ...process.env,
     GAMEID: 'umu-0',
-    ...setupEnvVars(settings),
+    ...setupEnvVars(settings, gameInstallPath),
     ...setupWineEnvVars(settings, installFolderName),
     PROTON_VERB: protonVerb
   }
--- a/src/backend/storeManagers/gog/setup.ts
+++ b/src/backend/storeManagers/gog/setup.ts
@@ -207,6 +207,7 @@ async function setup(
             gameSettings,
             wait: false,
             protonVerb: 'run',
+            installFolderName: gameInfo.install.install_path,
             skipPrefixCheckIKnowWhatImDoing: true,
             startFolder: path.join(
               gogSupportPath,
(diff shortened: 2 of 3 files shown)


arielj:
how can we test this? like, how can we reproduce the problem this is fixing to know what to look for?

Etaash-mathamsetty:
code looks fine, but is this only an issue with gog ?

imLinguin:
it might affect others as well, I haven't experienced any issues though...

imLinguin:
what this essentially does is allows to mount game directory during setup. This is crucial especially when mount point is in non-standard location that is not handled by pressure vessel out of the box

arielj:
Sorry, not sure what you mean by `mount`, you mean using the game's install folder as the current working directory?

imLinguin:
soldier and sniper runtimes are container based and have limited list of paths they expose in container by default. Adding `STEAM_COMPAT_INSTALL_PATH` env variable does exactly that, makes sure container exposes game directory.   
Current working directory was another addition I had to make in order for setup to succeed.

arielj:
One more question, I see the `installFolderName` argument in `runWineCommand` is passed to `setupWineEnvVars` as the `gameId` argument (?) and then that's used as `ret.SteamGameId = 'heroic-${gameId}'` https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/launcher.ts#L924, that one doesn't set `STEAM_COMPAT_INSTALL_PATH`.

The one that sets `STEAM_COMPAT_INSTALL_PATH` is the line above `setupEnvVars` https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/launcher.ts#L767 but we are NOT passing the `installFolderName` there https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/launcher.ts#L1156 (the second argument, which is the install dir, is not used)

so I'm really confused how this even works, but I guess it's just me not understanding the steam runtimes (they use the `SteamGameId` variable as a path? too).


I think there's a problem, that we have to pass the install path to the correct method, I think that's a bug

imLinguin:
Hmm, you are right. Guess that shows I looked into Heroic code long time ago....

arielj:
Is this PR meant to fix this https://github.com/Open-Wine-Components/umu-launcher/issues/315 ? because if that's the case it's not working for me

I can run the game with Wine-GE creating the prefix, but I can't run the game with proton-ge+umu creating the prefix

imLinguin:
Are you sure? For me a clean prefix with UMU + Proton-GE-latest works as expected and shows the launcher.   

obraz

arielj:
I'm using umu 1.1.3, maybe that's the issue for me? can heroic download 1.1.4? or I need to do it manually?

arielj (approved):
ok, now that I got 1.1.4 it does work as expected for both witcher 2 and ghost master that I knew had issues :+1:

PR #4254, opened 2025-01-10
[UX] Ignore Proton - Experimental by default, but allow enabling it for advanced usage

arielj:
This PR changes how we handle the `Proton - Experimental` option to run games.

We currently have a problem, that, for some users, `Proton - Experimental` might be the only wine/proton they have when they use Heroic for the first time and it gets assigned as a default game setting. From there, all their games ends up using `Proton - Experimental` with the risks that creates: it's experimental and not tested as much as stable versions, so an updated in steam without the user noticing it can break games in Heroic and it's hard to know what changed (users usually think nothing changed).

With this change, `Proton - Experimental` won't show up by default in the `Wine Version` selectors throughout the app, so inexperienced users can't select it by mistake and shouldn't be automatically selected by Heroic either (there's also this idea that, because it fixes many games, it's always better, so checking the option shows a warning for the user with some information about what experimental means).

For users that know what they are doing and want to be intentional about using `Proton - Experimental`, there's now a checkbox in Settings > Advances to stop ignoring it.

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -584,6 +584,14 @@
             },
             "label": "Allow installation of games with broken or denied anticheat"
         },
+        "allow_proton_experimental": {
+            "confirmation": {
+                "message": "'Proton - Experimental' is, by nature, meant for experimentation and not for stable use. Using it as a default for all games can unexpectedly break games previously working during an upgrade. Use it only if you know what you are doing.",
+                "title": "Are you sure?",
+                "understand": "Don't allow it"
+            },
+            "label": "Allow using \"Proton - Experimental\" to run games"
+        },
         "alt-gogdl-bin": "Choose an Alternative GOGDL Binary to use",
         "alt-legendary-bin": "Choose an Alternative Legendary Binary",
         "alt-nile-bin": "Choose an Alternative Nile Binary",
--- a/src/backend/utils/compatibility_layers.ts
+++ b/src/backend/utils/compatibility_layers.ts
@@ -169,6 +169,15 @@ export async function getLinuxWineSet(
   protonPaths.forEach((path) => {
     if (existsSync(path)) {
       readdirSync(path).forEach((version) => {
+        if (version.includes('Experimental')) {
+          if (!GlobalConfig.get().getSettings().allowProtonExperimental) {
+            logInfo(
+              'Ignoring "Proton - Experimental". You can change this in Settings > Advanced'
+            )
+            return
+          }
+        }
+
         // Only relevant to Lutris
         if (version.startsWith('UMU-Latest')) {
           return
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -102,6 +102,7 @@ export interface AppSettings extends GameSettings {
   startInTray: boolean
   allowInstallationBrokenAnticheat: boolean
   disableUMU: boolean
+  allowProtonExperimental: boolean
 }
 
 export type LibraryTopSectionOptions =
--- a/src/frontend/screens/Settings/components/AllowProtonExperimental.tsx
+++ b/src/frontend/screens/Settings/components/AllowProtonExperimental.tsx
@@ -1,0 +1,63 @@
+import React, { useContext } from 'react'
+import { useTranslation } from 'react-i18next'
+import useSetting from 'frontend/hooks/useSetting'
+import { ToggleSwitch } from 'frontend/components/UI'
+import ContextProvider from 'frontend/state/ContextProvider'
+
+const AllowProtonExperimental = () => {
+  const { t } = useTranslation()
+  const { showDialogModal } = useContext(ContextProvider)
+  const [allowProtonExperimental, setAllowProtonExperimental] = useSetting(
+    'allowProtonExperimental',
+    false
+  )
+
+  const toggleConfig = () => {
+    if (allowProtonExperimental) {
+      setAllowProtonExperimental(false)
+    } else {
+      showDialogModal({
+        title: t(
+          'setting.allow_proton_experimental.confirmation.title',
+          'Are you sure?'
+        ),
+        message: t(
+          'setting.allow_proton_experimental.confirmation.message',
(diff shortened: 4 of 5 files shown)


CommandMC:
Honestly I'd vote for ignoring all Proton versions found in that path by default (or at least everything non-GE-Proton). Regular Proton lacks Winetricks (and protonfixes as a whole), so you already lose a lot of games

arielj:
> Honestly I'd vote for ignoring all Proton versions found in that path by default (or at least everything non-GE-Proton). Regular Proton lacks Winetricks (and protonfixes as a whole), so you already lose a lot of games

I'm not against that, personally I just use whatever I can get with heroic, I don't know what users expect but I don't see the point of using stock proton when ge-proton exists haha

We still have the `Add custom Wine/Proton Paths` for people that wants those I guess?

Etaash-mathamsetty:
> Honestly I'd vote for ignoring all Proton versions found in that path by default (or at least everything non-GE-Proton). Regular Proton lacks Winetricks (and protonfixes as a whole), so you already lose a lot of games

I save storage by installing stuff to the steam compatibilitytools.d folder, so +1 for ignoring everything that's not GE-proton by default

arielj:
Do you think the checkbox to enable those things is needed? I didn't want to remove things without having an option to have them back cause I imagine some users wanting that, but if there's no need for proton nor proton - experimental I wonder if we should just ignore them and that's it?

arielj:
The main thing I want to avoid with this PR is users using something unstable by default or by mistake, but let them use experimental if they need to (would experimental have fixes before ge-proton for example that users may want to try?).

Etaash-mathamsetty:
> The main thing I want to avoid with this PR is users using something unstable by default or by mistake, but let them use experimental if they need to (would experimental have fixes before ge-proton for example that users may want to try?).

maybe add an option that allows showing all the proton versions (and it's off by default)

gabriele2000:
I mean, Proton Experimental isn't so experimental... the real experimental thing is proton-experimental-bleeding-edge or something like that.

Proton experimental is just updated every few weeks, Proton stable, uh...

Etaash-mathamsetty:
> I mean, Proton Experimental isn't so experimental... the real experimental thing is proton-experimental-bleeding-edge or something like that.
> 
> Proton experimental is just updated every few weeks, Proton stable, uh...

proton experimental isn't recommended to be run with umu because it doesn't contain protonfixes + some other patches

gabriele2000:
> proton experimental isn't recommended to be run with umu because it doesn't contain protonfixes + some other patches

Then I suggest you to negate Proton Experimental only if UMU is enabled (enable UMU by default but let the user choose to disable it)

arielj:
The changes here are aimed to the general user to not, by default, with a with a setup that is not recommended (the second problem with experimental is that an update breaks something you can't easily rollback to the working version, but if using `proton-ge-latest` for example and an update breaks something you can easily change to another fixed version using the wine manager).

The intention of this is to prevent users who don't know what they are doing to use something they don't know how to handle properly. The option to continue using experimental and enable/disable umu as needed will be there for any user who wants to try that.

gabriele2000:
> The changes here are aimed to the general user to not, by default, end with a setup that is not recommended (the second problem with experimental is that an update breaks something you can't easily rollback to the working version, but if using `proton-ge-latest` for example and an update breaks something you can easily change to another fixed version using the wine manager).
> 
> The intention of this is to prevent users who don't know what they are doing to use something they don't know how to handle properly. The option to continue using experimental and enable/disable umu as needed will be there for any user who wants to try that.

Awesome, thanks for the clarification!

arielj:
I updated the PR to ignore all non-GE protons by default instead of just Proton - Experimental.

And things can be allowed with the settings in advanced

arielj:
one thing I'm not sure is how to handle cases where users already have Experimental or stock Proton selected for a game

this doesn't remove the setting and doesn't prevent those games from being launched with whatever is the current setting, but opening the game settings shows the wine version blank cause the selected proton is not present in the results unless the user enables the advanced setting

I wonder if we should have it enabled for current users and only disabled for new installations?

arielj:
> one thing I'm not sure is how to handle cases where users already have Experimental or stock Proton selected for a game
> 
> this doesn't remove the setting and doesn't prevent those games from being launched with whatever is the current setting, but opening the game settings shows the wine version blank cause the selected proton is not present in the results unless the user enables the advanced setting
> 
> I wonder if we should have it enabled for current users and only disabled for new installations?

I changed the logic to filter in the frontend instead of the backend, so we can not ignore the currently selected proton to not show a blank selector

src/frontend/screens/Settings/components/WineVersionSelector.tsx
```diff
@@ -42,7 +60,7 @@ export default function WineVersionSelector() {
       }
       setRefreshing(false)
     }
-    getAltWine()
+    void getAltWine()
```
arielj, line 63: to make the linter happy

src/frontend/screens/Settings/components/WineVersionSelector.tsx
```diff
@@ -102,15 +128,13 @@ export default function WineVersionSelector() {
             <InfoBox text="infobox.help">
               <span>{t('help.wine.part1')}</span>
               <ul>
-                <i>
```
arielj, line 105: this `<i>` tag here was invalid HTML

flavioislima: I think this supposed to make the text _italic_

arielj: the problem is not the tag itself, it's that `<i>` can't be a direct children of `<ul>` (and the browser is fixing that for us), I'll update this to put the `<i>` inside each `<li>`

actually, I'll check if I can do this with CSS, I don't like tags for style purposes

EDIT: the CSS selector was too ugly, I just added the `<i>` inside the `<li>` tags

PR #4255, opened 2025-01-10
[Feat/Tech] Migration system

CommandMC:
Originally part of my components branch, this is a generic system for running *something* only once per machine. Say we change something persistent (game settings format, file locations, etc.) and now need to update old instances of *thing*: We'd add a new migration that does the actual upgrade, give it a unique name, add it to the list of migrations, and everything else is taken care of.

I've not gone hunting for use cases of this in the codebase yet, but I can already think of a couple (which I'll add to this PR once implemented, if nothing else than to give an idea on how one might use the system, and to not merge a PR doing nothing)
I would still like a review of the system itself though

---


src/backend/main.ts
```diff
@@ -301,6 +301,9 @@ if (!gotTheLock) {
     handleProtocol(argv)
   })
   app.whenReady().then(async () => {
+    const migrationComponent = new MigrationComponent()
+    await migrationComponent.applyMigrations()
+
```
CommandMC, line 306: This is a little awkward, and is a result of porting this to "non-components-system" Heroic. If components would be in main right now, you'd not see any of this here (the component would instead be initialized automatically)

arielj: I have some questions about the components, but there's no PR for that right? to not ask the questions here

arielj (commented):
How do you imagine migrations working for a user the opens Heroic for the first time and there are migrations in the list? from the code I understand it will try to run the migrations and then the migration itself has to check if it needs to do something or not? 

Because most migration systems I've used tend to have an option that, if you start the project for the first time, you can get the correct state right away and all migrations are marked as already applied since they are not really needed when there's nothing to migrate.

src/backend/main.ts
CommandMC: There's no PR yet, no. I do have a WIP branch, but it's a little outdated
The reason I haven't made a PR yet is that to properly show what components can do (and to get them some actual use, instead of just being an unused system), I need some preliminary steps (which this PR is one of)

CommandMC:
> How do you imagine migrations working for a user the opens Heroic for the first time and there are migrations in the list? from the code I understand it will try to run the migrations and then the migration itself has to check if it needs to do something or not?

Yes, correct.

I don't think this will actually be that much of an issue. Heroic is already so modular that basically everything a migration might do would require a check for whether it has to do that task first.
Say we're moving the config folder of one of our runners, we'd first have to make sure the old config folder exists (even if Heroic is only updated, the user might still never have used that runner). Update game settings? You already have to loop over all stored settings, on a new install that'll just do nothing. You get the idea.
I can't imagine a use case where a check like that *wouldn't* be necessary, so new installs should already be covered without extra work

src/backend/main.ts
CommandMC: I ended up uncoupling this from the "components" model now. I'm not sure whether I'll continue with components, at least for now. Lots of other things to do first

flavioislima (approved):
looks good. 
there is any way to test this already?

CommandMC:
You can test the one migration I've implemented, yes.
- In case you ran this PR before, remove `~/.config/heroic/store/migrations.json`
- Move your current Legendary config folder from `~/.config/heroic/legendaryConfig/legendary` to `~/.config/legendary`
- Launch Heroic
You'll see the folder get copied to its Heroic-specific location again, and this printed in the logs:
```
(16:16:34) INFO:    [Backend]:  Applying migration "legendary-move-global-config-folder"
(16:16:34) INFO:    [Backend]:  Migration "legendary-move-global-config-folder" successfully applied
```

PR #4264, opened 2025-01-11
[Refactor] Constants cleanup

arielj:
The `backend/constants.ts` file was getting out of hands since a long time ago, it was a bucket to put things from all types, paths, urls, helper booleans, it also did initialization of the logger, had some helper functions related to steam, it was a long and complicated file.

This also created problems when wanting to import one thing from that file could lead into issues with the file doing unexpected things.

This PR splits the file into different concerns and also moves some constants to other places where they make more sense:
- store related paths where moved into `storeManagers/<store>/constants.ts`, so all gog paths are next to the gog file, nile paths next to the nile paths, etc
- some constants where only used in one single place, like the anticheat json location, or the vulkan helper location, images cache path, etc, those were moved to the places they are used instead
- I divided the constants in different groups: paths, urls, environment (all the `isXyz` booleans), key_value_stores for the electron stores (I think these can be moved somewhere else though), and others (to collect some things I didn't know where to put)
- I also took the opportunity to rename some constants: some where related to legendary but were too generic like `installed` (renamed to `legendaryInstalled`), or `flatPakHome` renamed as `flatpakHome`
- I added a new const `isAppImage` that could be useful in the future
- I also moved all the logger constants (and initialization) to happen in the `initLogger` module, makes more sense there

I can split this into smaller PRs if it's easier to check and test, but I think the changes are pretty repetitive through all the files, just changing imports in most of them.

---


CommandMC (approved):

flavioislima (approved):
Good stuff! 
Things looks more organized this way.

PR #4268, opened 2025-01-12
[Refactor] Make InstallModal global with state moved to zustand

arielj:
This PR moves how we handle the `InstallModal` rendering and moves the state to Zustand. This helps removing some repeated code and makes opening and closing the modal to produce fewer re-renderings.

This PR changes:
- Currently we have the component rendered inside different components: Library, GamePage, SubMenu, etc. This was like this I imagine to avoid having it global to avoid too many re-rendering, but with the cost of duplicating code for the render and the state
- Even with the previous consideration, the state was still producing unintentional re-renderings, like... clicking the install button on a game in the library would re-render the whole library, or clicking the install button in the game page would re-render the whole page, or clicking `Add Game` in the library would re-render the whole library. With this PR it won't re-render extra components
- Similar to the game settings PR, I took the opportunity to make the method names and state more declarative, so instead of doing something like `setShowModal(false)` we can now call `closeInstallGameModal()`

I tested all the places I could see:
- Install in library
- Install in game page from body
- Install in game page from 3-dots menu
- Edit sideloaded app
- Add sideloaded app
- Install EA game

I think they all work as expected, not sure if I'm missing something to test.

---

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -18,6 +18,7 @@ import classNames from 'classnames'
 import { ThemeProvider, createTheme } from '@mui/material/styles'
 import LogFileUploadDialog from './components/UI/LogFileUploadDialog'
 import UploadedLogFilesList from './screens/Settings/sections/LogSettings/components/UploadedLogFilesList'
+import { InstallGameWrapper } from './screens/Library/components/InstallModal'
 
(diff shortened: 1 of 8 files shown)


flavioislima (approved):
LGTM

PR #4276, opened 2025-01-14
[GOG]: Support overlay

imLinguin:
This PR will cover Linux and maybe Windows  
PR plan:
- [x] Add option to download/update comet's overlay and web components
- [ ] Update gogdl itself to inject the overlay on Windows
- [x] Ship galaxy-helper and add support for it in gogdl (requires https://github.com/Heroic-Games-Launcher/heroic-gogdl/commit/9079373d6fd7904f6c639d531b9c1877ea4a03c1)

As I don't own a Mac and can't confidently write support for usage of native overlay build over there


arielj:
I'm looking at the code and this would be enabled with the same comet experimental feature flag but there's no way to disable it on a game-by-game basis, right?

I think it would be a good idea to make this independent, if comet is enabled (we should also make this a game-by-game config instead of global) we let users enable or disable the gog overlay.

I'm just worried that knowing some games don't work with comet maybe some games work with comet but don't work with the overlay, and also it could be good to have the option to disable it like we have for the EOS overlay for more consistency.

src/frontend/screens/Settings/components/CometComponents.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <>
+      <h3>{t('setting.comet.components', 'Comet Components')}</h3>
+
+      <div className="footerFlex">
+        <button className="button is-primary" onClick={updateCometComponents}>
+          <CachedOutlined />
+          <span>
+            {isUpdating
+              ? t('gamepage:status.updating')
```
arielj: this prints `UPDATING GAME` here

there's also no feedback to know if there was anything to update or if the update was successful, there's output in the logs so maybe that can be returned in the updateCometComponents call and show it in the frontend for the user?

src/frontend/screens/Settings/components/CometComponents.tsx
```diff
@@ -0,0 +1,34 @@
+import { CachedOutlined } from '@mui/icons-material'
+import { useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+const CometComponents = () => {
+  const { t } = useTranslation()
+
+  const [isUpdating, setUpdating] = useState(false)
+
+  async function updateCometComponents() {
```
arielj, line 10: we should only run this `if (!isUpdating)` or you can trigger multiple updates at the same time

src/frontend/screens/Settings/components/CometComponents.tsx
```diff
@@ -0,0 +1,34 @@
...
+    setUpdating(true)
+    await window.api.updateCometComponents()
+    setUpdating(false)
```
arielj:
```suggestion
    window.api.updateCometComponents().finally(() => setUpdating(false))
```

maybe the `setUpdating(false)` can be done in a `.finally`, so whatever happens with the promise we set the correct state

src/frontend/screens/Settings/components/CometComponents.tsx
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  return (
+    <>
+      <h3>{t('setting.comet.components', 'Comet Components')}</h3>
```
arielj: maybe we can have some context here of what this is? "Comet Components" might make sense only for people familiar with Comet, but we could have some reference to the GOG Overlay for example.

src/backend/storeManagers/gog/library.ts
```diff
@@ -1291,6 +1257,63 @@ export async function getProductApi(
...
+  }
+  remoteConfigStore.set(clientId, response)
+  return response
+}
+
+export async function getClientId(
+  appName: string,
+  install_path: string
+): Promise<string | undefined> {
+  let clientId
```
arielj:
```suggestion
  let clientId = readInfoFile(appName, install_path)?.clientId
```

to not need the `else` blocks after, if the `if (existsSync...)` returns false we already have the default

src/backend/storeManagers/gog/games.ts
```diff
@@ -568,6 +575,10 @@ export async function launch(
       return false
     }
 
+    if (!isOnline() && wineEnvVars) {
+      delete wineEnvVars['HEROIC_GOGDL_WRAPPER_EXE']
```
arielj: in the prepareWineLaunch we are always adding this env variable, and then removing it here if not online

maybe we can have this check there and just not add the variable if offline? even skip the whole code for comet if offline? to not do the work to add this and then skip it

arielj (changes requested):

CommandMC (commented):
I have had some issues with the overlay, testing with Witcher 3 (specifically with the EE Redux mod). The game no longer takes mouse input at all (the overlay receives it just fine though). An option to disable the overlay per-game would thus make sense

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -226,6 +228,7 @@ export default function GamesSettings() {
         )}
         <VerboseLogs />
         <DisableUMU />
+        {isGog && <Overlay />}
```
arielj, line 232: The reason for this as a select is because some games don't support the overlay? because it looks weird that the gog overlay setting is here and the epic overlay setting is in the 3-dots menu, and it might not be clear to the user what's the difference between default and force enabled

arielj: I think we should also add some message here if Comet is disabled, because some users have to disable comet for some games to work and that will disable it globally, but then see this setting not knowing it won't work without comet

Maybe we should move the comet experimental option also here and make it a per-game config (it would be weird to have this new option here not experimental when it depends on an experimental feature in another place), so users that can't user comet for some games but can for other can still use the overlay without having to switch the experimental feature

src/frontend/screens/Settings/components/Overlay.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const { isDefault } = useContext(SettingsContext)
+  const { platform } = useContext(ContextProvider)
+
+  const [forceDisableOverlay, setForceDisableOverlay] = useSetting(
+    'forceDisableOverlay',
+    false
+  )
+  const [forceEnableOverlay, setForceEnableOverlay] = useSetting(
+    'forceEnableOverlay',
+    false
```
arielj, line 22: is there a reason to have 2 different settings that are controlled by the same select input? can't we have a single setting with 3 possible values?

arielj:
I tested earning some achievements and the popup shows up :+1:

arielj:
one more question, is the overlay supposed to work with linux native games? I tried one and it does not show up for me, even with the force enable setting

if it doesn't work for linux native games we should disable the overlay setting

src/frontend/screens/Settings/components/Overlay.tsx
arielj: maybe also include `GOG` as part of the name of the setting? since it's specific to GOG, to avoid confusions with the epic overlay

src/backend/storeManagers/gog/library.ts
```diff
@@ -1295,6 +1261,60 @@ export async function getProductApi(
...
+  clientId: string
+): Promise<GOGClientsResponse | null> {
+  const resp = remoteConfigStore.get(clientId)
+  if (resp) return resp
+
+  let response: GOGClientsResponse | undefined
+  try {
+    response = (
+      await axiosClient.get<GOGClientsResponse>(
+        `https://remote-config.gog.com/components/galaxy_client/clients/${clientId}?component_version=2.0.80`
```
arielj: is this ok that this `component_version` is harcoded? shouldn't this match the version we have locally from this?

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4276/changes#diff-7e6e78ae2155f41a0efc1c2fceccf5dd35614cae66fdfe01def49b06c7fe8503R8

src/frontend/screens/Settings/components/CometComponents.tsx
```diff
@@ -0,0 +1,35 @@
+import { CachedOutlined } from '@mui/icons-material'
+import { useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+const CometComponents = () => {
```
arielj, line 5: we should hide this component on windows and mac until that's supported

src/backend/launcher.ts
```diff
@@ -945,34 +949,64 @@ async function prepareWineLaunch(
 
   try {
     if (runner === 'gog' && experimentalFeatures?.cometSupport !== false) {
```
arielj: I think some of this code is being applied for mac and windows making things crash when launching a game (at least on mac, I don't have windows to confirm)

On mac we can enable the comet support and get achievements (without the overlay popup), but with the changes of this PR the game fails to launch since the default is to try to enable the overlay if the game supports it, the error in the logs is `[galaxy_helper] Failed to load unix functions for sockets` (nothing more, just that line, to stack trace or anything)

arielj (changes requested):

src/backend/launcher.ts
imLinguin: Right, the wrapper is only for Linux at the moment, as I didnt write a mac native unixlib.

We should add a platform specific guard and for Mac support a native overlay instead. That implies the wrapper would setup the other end of pipe forwarding

src/backend/storeManagers/gog/library.ts
imLinguin: This is fine, the component_version only adds / changes certain fields in the response. We are fine with both varying

src/frontend/screens/Settings/components/Overlay.tsx
imLinguin: We could have one setting, thats true. Just my first thought was that this way we have a clear definition of what the config can hold. I can change that

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
imLinguin: Adding a note makes sense. I dont think having comet per game will be useful. The v0.3.2 should be more stable than v0.2.0 and I think people experiencing these issues wont see them again

imLinguin: And regarding the 3 dots menu discrepancy, I think we should address that to the point where we always install EOS overlay. There is no reason not to at this point as more and more games rely on it.

arielj: > Adding a note makes sense. I dont think having comet per game will be useful. The v0.3.2 should be more stable than v0.2.0 and I think people experiencing these issues wont see them again

I'm thinking about the users that have issues and have to disable Comet and we don't know the cause, maybe they have to disable it for some games but it works fine for others?

anyway, I do think we need to move that out of experimental if we are adding the overlay as not experimental, will be confusing to users

arielj: > And regarding the 3 dots menu discrepancy, I think we should address that to the point where we always install EOS overlay. There is no reason not to at this point as more and more games rely on it.

I think we already do, if the EOS overlay is installed I think we enable it automatically for games, I don't know if there's a reason to disable it though (does the official client have that option? I know some launchers do have it)

src/frontend/screens/Settings/components/Overlay.tsx
arielj: > We could have one setting, thats true. Just my first thought was that this way we have a clear definition of what the config can hold. I can change that

do we need the force enable option? I mean, if we enable the overlay and the game does not support it, does that create any issue? because I feel like we could do something like with epic, we either enable or disable the overlay, independent of if the game supports it or not, it would be simpler for the logic and for the user and consistent with epic

I just don't know if the overlay can cause issues if enabled when not supported

Sevenseraphim:
Is the Gog overlay working now? will it be added to the next heroic release?

(I'm not sure if this is the right place to follow up on this or not)

arielj:
> Is the Gog overlay working now? will it be added to the next heroic release?
> 
> (I'm not sure if this is the right place to follow up on this or not)

with this PR it will work on linux, but not on windows or macos

but the PR as is has some issues and also conflicts so it cannot be merged yet, it might be included in a future release once the issues are solved, but there's no guarantee it will be in the next release, it depends on when these issues are solved and when the next release happens

iofhua:
Hello I am using heroic games launcher on Nobara linux to play GOG games and I want to voice my support for a UI pop-up when I earn an achievement. A way to look at each game and see my list of GOG achievements in the launcher would also be nice.

Dwarsen:
I checked this branch out, built it locally and tried the overlay out with two games, with pretty poor results.

First, Tainted Grail: Fall of Avalon. With the overlay active (it pops up in the corner and I can bring it up with shift-tab), the mouse cursor works in the menus but not when in-game. Keyboard still moves my character, but moving the mouse doesn't change where I'm looking or do anything else. I tried different proton versions, no change. I tried with Wine-Wayland activated and that actually made things worse, now the mouse cursor doesn't work in the menus either. Disabling the overlay immediately restores mouse functionality in the game.

Then I tried The Sinking City Remastered. Same issues as with above, but with the added bonus that if the overlay is active there is no sound outside of the intro splash screens. Disabling the overlay makes the game run as it should, with both sound and controls working.

I looked at the logs and didn't really find anything that looked like an error.

I'm running Manjaro with KDE on an AMD GPU, but can of course provide more info if necessary. If I'm doing something wrong or if this isn't the right place to post this then please correct me :)

imLinguin:
Would you be able to compare same games by installing them with GOG Galaxy in Proton? I'm curious if we are doing something wrong or if it's just how Overlay works.

Dwarsen:
> Would you be able to compare same games by installing them with GOG Galaxy in Proton? I'm curious if we are doing something wrong or if it's just how Overlay works.

Unfortunately I can't seem to install any games with GOG Galaxy running through Proton. It downloads the game, gets to 97% and then pop ups "installation failed".

Dwarsen:
Okay, I managed to make it install a game by reverting to Proton 9, where it does seem to work. It's not ideal that Gog chose to put their overlay on the same combo as steam so I can't bring it up in-game, but I can see that it loads when starting the Sinking City Remastered in the bottom corner. And yeah, same issue via GOG Galaxy it seems, at least I get no sound and the mouse doesn't work with the overlay active. I have no idea how to troubleshoot this, but at least it doesn't seem like it's a Heroic issue. :)

Arias800:
I tested the GOG overlay on A Plague Tale: Requiem and everything worked great. Only issue is that the OptiScaler GUI won't open while the GOG overlay is active. It also crashed the game once, but I didn't have logs enabled at the time.

Sevenseraphim:
I'm confused. It says the checks are successful but then it says that there are conflicts that must be resolved...

I can't tell if that's good or bad...

src/backend/launcher.ts
```diff
@@ -894,34 +897,65 @@ async function prepareWineLaunch(
...
+        mkdirSync(galaxyOverlay, { recursive: true })
+        await copyFile(
+          galaxyOverlayShimExePath,
+          join(galaxyOverlay, 'galaxy.exe')
+        )
+        await copyFile(
+          galaxyOverlayShimUnixlibExePath,
+          join(galaxyOverlay, 'libgalaxyunixlib.dll.so')
+        )
+      }
```
arielj, line 937: I understand we are copying these files even before knowing if we should use them?

because I think it's probably better to only try to copy them if the gog overlay is enabled

src/frontend/screens/Settings/components/CometComponents.tsx
arielj, line 5: this still needs the check to only show this button on linux, now it's showing on all platforms

src/frontend/screens/Settings/components/Overlay.tsx
```diff
@@ -0,0 +1,77 @@
...
+        setForceEnableOverlay(false)
+        break
+      case 'enable':
+        setForceDisableOverlay(false)
+        setForceEnableOverlay(true)
+    }
+  }
+
+  // When we add support for GOG overlay on other platforms we can remove platform check
+  if (isDefault || platform !== 'linux') {
```
arielj, line 52: should we also consider native games in this check?

I understand the overlay code depends on copying some .exe files inside the game prefix, so I understand it does not work for linux native games, is that correct?

arielj (changes requested):

PR #4280, opened 2025-01-15
[UX] Show tab and message when Cloud Saves are not supported

arielj:
Currently, we do not display the `Cloud Saves Sync` tab in the game's settings if the game shows that it doesn't support cloud saves in the metadata and if the game is linux native.

It had happened many times in the past that users can't find the `Cloud Saves Sync` tab and they think there's some bug when, in reality, they may be hitting one of those conditions.

So, instead of hiding the tab and letting the user figure out why they can't see it, it's more user-friendly to show the tab and explain why it doesn't work. And in the case of a GOG linux-native game we can also instruct the user to use the Windows version instead so they don't need to figure that out.

I did leave the tab ignored for Amazon games and Sideloaded apps/games because I understand Amazon doesn't cloud saves at all and for sideloaded games it makes no sense.

This is similar to what we do with the Gamescope tab, that we show it with an error message if the `gamescope` binary cannot be found.

The style is not great but I used the same as the Gamescope tab for consistency (just a div with red text (this is not great for themes though))

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -818,6 +818,8 @@
         "open-config-file": "Open Config File",
         "reset-heroic": "Reset Heroic",
         "saves": {
+            "gog_linux_native_warning": "Linux native games do not support GOG's Cloud Saves feature. Use the Windows version instead.",
+            "not_supported": "This game does not support Cloud Saves.",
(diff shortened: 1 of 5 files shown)


flavioislima (approved):
lgtm

Issue #4242, opened 2025-01-05
Allow hiding sidebar items


ghost:
### Problem description

If you don't use some items you should be able to hide them.

### Feature description

Add toggles in the settings to hide the items individually.

### Alternatives

Themes, but they aren't an easy solution for most people.

PR #4282, opened 2025-01-16
[UX] CustomCSS field for advance persistent styling

arielj:
This PR aims to fix 2 problems:
- When customizing heroic, themes can change the style of the app but users may want to apply changes that are persisted between different themes (like hiding parts of the app, making different elements bigger, moving things around, etc). Currently, the only option to do that is to edit all themes to include the same style.
- The second problem to solve is to allow users the flexibility to customize things without the need of a full CSS theme and without the need for a specific configuration

For example, this feature can be used to support this feature request https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4242 (which was also requested in Discord) by applying this custom CSS

image

(or `display: none` to fully remove those icons)

Then, any theme that's selected will still have this custom CSS style applied.

Since this can be dangerous (a user could `body { display: none }` and break everything), there's a warning with an indication on where to update the config files manually, and that's the reason why I put this in Advanced, it's not meant for inexperienced users.

---


flavioislima (approved):
Nice!

Issue #4037, opened 2024-09-30
"Browse files" option


hfc2x:
### Problem description

I needed to change some text configuration for a game, and I needed to go into the install folder. I realized Heroic has no such option, so I had to go to the install folder manually.

### Feature description

This should be a simple option like Lutris has when wanting to open the install folder of a game.

### Additional information

I'm on Fedora 40. Heroic is installed via the COPR repository.

redromnon:
Hi, I believe you can open the game's install folder from a game page: click on _Installed Information_ -> _Install Path_ and Heroic will open the install folder using your native file manager.

arielj:
Yes, clicking the install path in the game page information opens the file explorer in the installation folder. Also clicking the wine prefix path opens the file explorer in the wine prefix folder.

Closing, feel free tor reply and we can reopen if this is not what you meant.

Coolumin:
> Yes, clicking the install path in the game page information opens the file explorer in the installation folder. Also clicking the wine prefix path opens the file explorer in the wine prefix folder.
> 
> Closing, feel free tor reply and we can reopen if this is not what you meant.

This isn't really what OP is looking for (and frankly not what I'm looking for either). Clicking the game's option for the changing the install path is for changing where it's installed from, and thus opens an in-app pop-up window rather than a normal file explorer window. 

What's desired here is just a shortcut button that takes you directly inside the game's folder by opening a new window of the file explorer used by the system, just like on Steam or Lutris.

arielj:
> > Yes, clicking the install path in the game page information opens the file explorer in the installation folder. Also clicking the wine prefix path opens the file explorer in the wine prefix folder.
> > Closing, feel free tor reply and we can reopen if this is not what you meant.
> 
> This isn't really what OP is looking for (and frankly not what I'm looking for either). Clicking the game's option for the changing the install path is for changing where it's installed from, and thus opens an in-app pop-up window rather than a normal file explorer window.
> 
> What's desired here is just a shortcut button that takes you directly inside the game's folder by opening a new window of the file explorer used by the system, just like on Steam or Lutris.

I wasn't talking about the option to change the installation path, I was talking about the install path and prefix path displayed under `Installed Information`:

![Image](https://github.com/user-attachments/assets/3055b171-ae2b-4589-992f-6f4408927d9a)

those 2 paths open those 2 paths in the file explorer

hfc2x:
[...]

PR #4285, opened 2025-01-16
Implement menu options for opening the install file path and Wine prefix in the default file manager

luluco250:
Implements #4037

This was a personal nitpick of mine after migrating to Heroic from Lutris since I often need to open the game folder or prefix (changing/installing mods, editing save files in the prefix etc).

I have added menu options for `Browse Files` and `Browse Wine Prefix` to game cards context menus on the game list page as well as in the three dot menu of the game details page.

When clicked they will open the default file manager using Electron's `shell.openPath` from the backend. The first option opens the install path of the game (the directory portion of the executable path) while the second option opens the path to the Wine prefix used by the game.

I've added checks to ensure the options only show up if the game is installed and if it's a non-native game for the browse prefix option.

I've added English entries for both options in the gamepage translation file so that it could be pushed, not sure if that was the correct way.

Let me know if there are any issues with the PR and thanks for your work on Heroic!

<details>
<summary>Screenshots</summary>

</details>

---

### DIFF
--- a/src/frontend/screens/Game/GameSubMenu/index.tsx
+++ b/src/frontend/screens/Game/GameSubMenu/index.tsx
@@ -1,10 +1,10 @@
 import './index.css'
 
-import React, { useContext, useEffect, useState } from 'react'
+import React, { useCallback, useContext, useEffect, useState } from 'react'
 
 import { GameInfo, GameStatus, Runner } from 'common/types'
 
-import { createNewWindow, repair } from 'frontend/helpers'
+import { createNewWindow, isNativeGame, repair } from 'frontend/helpers'
 import { useTranslation } from 'react-i18next'
 import ContextProvider from 'frontend/state/ContextProvider'
 import { NavLink } from 'react-router-dom'
@@ -233,6 +233,16 @@ export default function GamesSubmenu({
     isInstalled &&
     !isThirdPartyManaged
 
+  const isNative = isNativeGame(gameInfo, platform)
+
+  const onBrowseFiles = useCallback(async () => {
+    await window.api.browseInstallPath(gameInfo)
+  }, [gameInfo])
+
+  const onBrowsePrefix = useCallback(async () => {
+    await window.api.browsePrefixPath(gameInfo)
+  }, [gameInfo])
+
   return (
     <>
       <div className="gameTools subMenuContainer">
@@ -375,6 +385,22 @@ export default function GamesSubmenu({
               className="link button is-text is-link"
             >
               {t('game.modify', 'Modify Installation')}
+            </button>
+          )}
+          {isInstalled && (
+            <button
+              onClick={async () => onBrowseFiles()}
+              className="link button is-text is-link"
+            >
+              {t('button.browse_files', 'Browse Files')}
+            </button>
+          )}
+          {!isNative && (
+            <button
+              onClick={async () => onBrowsePrefix()}
+              className="link button is-text is-link"
+            >
+              {t('button.browse_wine_prefix', 'Browse Wine Prefix')}
             </button>
           )}
         </div>
--- a/src/frontend/screens/Game/GameSubMenu/index.tsx
+++ b/src/frontend/screens/Game/GameSubMenu/index.tsx
@@ -4,7 +4,7 @@ import React, { useCallback, useContext, useEffect, useState } from 'react'
 
 import { GameInfo, GameStatus, Runner } from 'common/types'
 
-import { createNewWindow, isNativeGame, repair } from 'frontend/helpers'
+import { createNewWindow, repair } from 'frontend/helpers'
 import { useTranslation } from 'react-i18next'
 import ContextProvider from 'frontend/state/ContextProvider'
 import { NavLink } from 'react-router-dom'
@@ -50,7 +50,7 @@ export default function GamesSubmenu({
     showDialogModal,
     setIsSettingsModalOpen
   } = useContext(ContextProvider)
-  const { is } = useContext(GameContext)
+  const { is, gameSettings } = useContext(GameContext)
   const isWin = platform === 'win32'
   const isLinux = platform === 'linux'
 
@@ -233,15 +233,22 @@ export default function GamesSubmenu({
     isInstalled &&
     !isThirdPartyManaged
 
-  const isNative = isNativeGame(gameInfo, platform)
+  const hasWine = !is.win && !is.native
 
-  const onBrowseFiles = useCallback(async () => {
-    await window.api.browseInstallPath(gameInfo)
+  const onBrowseFiles = useCallback(() => {
+    const path = gameInfo.install.install_path || gameInfo.folder_name
+    if (path) {
+      window.api.openFolder(path)
+    }
   }, [gameInfo])
 
-  const onBrowsePrefix = useCallback(async () => {
-    await window.api.browsePrefixPath(gameInfo)
-  }, [gameInfo])
+  const onBrowsePrefix = useCallback(() => {
+    const path = gameSettings?.winePrefix
+
+    if (path) {
+      window.api.openFolder(path)
+    }
+  }, [gameSettings])
 
   return (
     <>
@@ -395,7 +402,7 @@ export default function GamesSubmenu({
               {t('button.browse_files', 'Browse Files')}
             </button>
           )}
-          {!isNative && (
+          {hasWine && (
             <button
               onClick={async () => onBrowsePrefix()}
               className="link button is-text is-link"
--- a/src/frontend/screens/Game/GameSubMenu/index.tsx
+++ b/src/frontend/screens/Game/GameSubMenu/index.tsx
@@ -237,6 +237,7 @@ export default function GamesSubmenu({
 
   const onBrowseFiles = useCallback(() => {
     const path = gameInfo.install.install_path || gameInfo.folder_name
+
     if (path) {
       window.api.openFolder(path)
     }
--- a/src/frontend/screens/Game/GameSubMenu/index.tsx
+++ b/src/frontend/screens/Game/GameSubMenu/index.tsx
@@ -233,9 +233,11 @@ export default function GamesSubmenu({
     isInstalled &&
     !isThirdPartyManaged
 
-  const hasWine = !is.win && !is.native
+  const hasWine =
+    !is.win && !is.native && gameSettings?.wineVersion.type !== 'crossover'
 
   const onBrowseFiles = useCallback(() => {
+    console.debug({ gameInfo })
     const path = gameInfo.install.install_path || gameInfo.folder_name
 
     if (path) {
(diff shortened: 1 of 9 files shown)


flavioislima:
   - - -
   - - -

luluco250:

arielj:
Sorry, to be sure, is there a reason to not use the options we already have by clicking the paths in the `installed information`? or is it that it's not easy to find? or at least to use the same backend calls

image

luluco250:
@arielj My bad, I didn't even know you could click that "Installed Information" text, it doesn't look like a button at all.

I would still personally prefer having those options be more accessible though, I don't often open the game detail page when I'm launching games but I do mod games often so having quick access to the game folder is nice.

Maybe the current UI/UX is a little too obscure?

arielj:
I agree the links are hard to find, and it's not obvious they can be clicked, I think it's a good idea to have those links too in the 3-dots menu (the new design have those paths more visible, but it's still not obvious then can be clicked)

About the card context menu, I understand that it would be good for your use case, but I don't think that's a general use case (if we add that cause one user wants that, we have to add options for anything other users want). I think the argument there should be: is this something the average user needs to have quick access from the library card? without thinking about how one personally uses Heroic. I can see Play/Logs/Settings/Library modifications/Uninstall to be general use for example.

You should rever the package.yml file, yes, in general you don't want to include that file unless you are changing something that would explicitly change node packages.

luluco250:
@arielj Thanks for your feedback!

I get your point about not cluttering the context menu for game cards. It's a personal preference of mine, but having in the three dots menu is good enough to improve discoverability IMO.

I wonder if something can be improved about the look of the `Installed Information` menu, maybe shade it in a lighter color or have a `V` icon to imply it's a clickable dropdown?

So in summary, here's what I'm planning to do:
- `Browse Files` and `Browse Wine Prefix` are available in the three dots menu.
- The context menu will not be changed from how it is in `main`, so no extra options there.
- Reuse backend code that already exists for the `Installed Information` dropdown for browsing paths.

Can I go ahead and make those changes?

PS.: I've also manually reverted the lockfile to the one in `main` and it seems to have fixed the pipeline issues.

arielj:
> @arielj Thanks for your feedback!
> 
> I get your point about not cluttering the context menu for game cards. It's a personal preference of mine, but having in the three dots menu is good enough to improve discoverability IMO.
> 
> I wonder if something can be improved about the look of the `Installed Information` menu, maybe shade it in a lighter color or have a `V` icon to imply it's a clickable dropdown?
> 
> So in summary, here's what I'm planning to do:
> 
> * `Browse Files` and `Browse Wine Prefix` are available in the three dots menu.
> * The context menu will not be changed from how it is in `main`, so no extra options there.
> * Reuse backend code that already exists for the `Installed Information` dropdown for browsing paths.
> 
> Can I go ahead and make those changes?
> 
> PS.: I've also manually reverted the lockfile to the one in `main` and it seems to have fixed the pipeline issues.

yes, those changes sound good

re: the look of `installed information`, the new design already solves that by displaying the information without having to click that line (you can try it in settings > advanced > experimental features > new design)

luluco250:
I just noticed a bug in the logic used to determine if the game has Wine btw:
image

It seems to stem from this part of the code (`src/frontend/screens/Game/GamePage/components/InstalledInfo.tsx:108`):
```ts
{!is.win && !is.native && (
  <>
    <div>
      <b>Wine:</b> {wineName}
    </div>
    {wineType === 'crossover' ? (
      <div>
        <b>{t2('setting.winecrossoverbottle', 'Bottle')}:</b>{' '}
        <div>{wineCrossoverBottle}</div>
      </div>
    ) : (
      <div
        className="clickable"
        onClick={() => window.api.openFolder(winePrefix)}
      >
        <b>{t2('setting.wineprefix', 'WinePrefix')}:</b>{' '}
        <div className="truncatedPath">{winePrefix}</div>
      </div>
    )}
  </>
)}
```

I think I can fix that, I got it to work on the `Browse Wine Prefix` though I did it in a more roundabout way:
```ts
function isNativeGame(
  gameInfo: GameInfo,
  platform: NodeJS.Platform | 'unknown'
) {
  const isInstallPlatformWindows =
    gameInfo.install.platform?.toLowerCase().startsWith('win') ?? false
  return (
    platform === 'win32' ||
    !isInstallPlatformWindows ||
    ((platform === 'linux') === gameInfo.is_linux_native &&
      (platform === 'darwin') === gameInfo.is_mac_native)
  )
}
```

That aside I'm not sure why it's reporting the existence of a Wine prefix, there is no such folder in my prefixes folder which can be verified by the fact an error is thrown when I click it:
image

So maybe it's safe to just fix the condition that checks for the presence of Wine and there's no other deep rooted issue with it?

arielj:
I'm not sure I understand what's the issue exactly, the prefix path is wrong? or you are saying it shouldn't report that's using wine? I think I'm missing some details or context

luluco250:
It's reporting the presence of a Wine prefix for a native Linux game that I installed on Heroic manually (`sideload` runner I think).

arielj:
ok so the issue is the value of `is.native`, that's happening here https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/frontend/screens/Game/GamePage/index.tsx#L323

so I'm guessing the issue is with the `isLinuxNative` variable a few lines above

luluco250:
I'm assuming the same issue could affect macOS due to the missing properties,

Here's what the `gameInfo` for the game looks like:
image

I've noticed that `gameInfo.install.platform` is not consistent, I've seen `Windows`, `windows`, `Win32` and `Linux`, hence the slightly odd check I used in my code:
```ts
gameInfo.install.platform?.toLowerCase().startsWith('win') ?? false
```

arielj:
You can do `const isLinuxNative = installPlatform?.toLocaleLowerCase() === 'linux'` for now I guess, fixing the installPlatform is probably out of scope for this

luluco250:
I guess a better check can be implemented to cover all the possible install platforms, given the `InstallPlatform` type:
image
image
image

So the full range of platform strings that could appear are `Windows`, `windows`, `Win32`, `Linux`, `linux`, `Mac`, `osx` and `Browser`.

Quite tricky in how inconsistent it is, but we can cover our bases properly by just checking thoroughly.

arielj:
ideally, we want to make sure that the `installPlatform` property is normalized at some point so we don't need to know the different possible values

The code is already checking `['osx', 'Mac'].includes(installPlatform ?? '')` so we could have `['linux', 'Linux'].includes(installPlatform ?? '')` for now to address this problem

I don't know what's the best solution for the whole app, I'd prefer to do something like: when reading the information of the games from disk, we normalize the value, then from that point it's all `win` `linux` `mac` (to say something)

but I think fixing that should be a different issue and a different PR

luluco250:
Updated the branch, removing redundant API methods as well as the `isNativeGame` helper function, instead relying on `is.win` and `is.native` from `GameContext` while fixing the check for native Linux games.

Removed the menu options from the game card context menu, now they show up only on the game details page's submenu.

src/frontend/screens/Game/GameSubMenu/index.tsx
```diff
@@ -233,6 +233,24 @@ export default function GamesSubmenu({
     isInstalled &&
     !isThirdPartyManaged
 
+  const hasWine = !is.win && !is.native
```
arielj, line 236: We also need to check if that `gameSettings.wineVersion.type !== "crossover"` cause in that case it uses a crossover bottle and not a prefix

same check is done on the Installed Information paths https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/3f7033e7b37f4ccddec73f5187a98c9956909920/src/frontend/screens/Game/GamePage/components/InstalledInfo.tsx#L113

src/frontend/screens/Game/GameSubMenu/index.tsx
```diff
@@ -233,6 +233,24 @@ export default function GamesSubmenu({
...
+
+  const onBrowseFiles = useCallback(() => {
+    const path = gameInfo.install.install_path || gameInfo.folder_name
```
arielj, line 239: not sure if `gameInfo.folder_name` would help here as a fallback, this is just the folder name, not a path that can be opened

src/frontend/screens/Game/GameSubMenu/index.tsx
luluco250: In that case would it be appropriate to rename the option to `Browse CrossOver Bottle` when we're dealing with a game that uses CrossOver? Referencing this string:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/3f7033e7b37f4ccddec73f5187a98c9956909920/public/locales/en/gamepage.json#L210

src/frontend/screens/Game/GameSubMenu/index.tsx
luluco250: I've copied the way it seems to be done here:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/3f7033e7b37f4ccddec73f5187a98c9956909920/src/frontend/screens/Game/GamePage/components/InstalledInfo.tsx#L55
Is this incorrect?

src/frontend/screens/Game/GameSubMenu/index.tsx
arielj: I don't think users need that, bottles should be handle with Crossover's UI so it's better to not let the user mess with the bottle directly

src/frontend/screens/Game/GameSubMenu/index.tsx
arielj: hmmm that's weird cause we can't really open just a folder name, it wouldn't be a path, maybe when that was first implemented it was not meant to be clicked, I don't really know, I do think it won't work (you can try it, just remove the `gameInfo.instlal.install_path || ` part of that assignment and click the `Browse Files` link, I imagine it won't open at all)

src/frontend/screens/Game/GameSubMenu/index.tsx
luluco250: I've added a check for the wine type so that it doesn't show the "Browse Wine Prefix" option when the game uses Crossover.

src/frontend/screens/Game/GameSubMenu/index.tsx
luluco250: You're right about `folder_name` not leading to a valid path in the case of a GOG game:
image

However it looks like it is a valid path for sideloaded games, which do not seem to list an install path (other than the executable):
image

So it looks like the existing code from `InstalledInfo.tsx` is handling this logic correctly, shouldn't we do the same thing here?

arielj: oh ok, I didn't think it was something for the sideloaded games, yes we should do that in both places then to account for that case

arielj (approved):

PR #4312, opened 2025-01-28
[UX] Improve empty library instructions and denied anticheat copy

arielj:
This PR improves two messages that were not super clear based on the number of questions in discord.

The message displayed in the library when Heroic is started the first time now explains a bit what the `Log in` and `Add Game` options do, so users can understand what to do to see their games.

image

The other message I changed is the one that we display when a user tries to install a game with broken/denied anticheat. Currently, the terms used in the copy are based on what AreWeAnticheatYet uses, but it's not clear for new users. Instead, now it explains what that actually means and also tells the user that there's nothing to do about the multiplayer/online features.

image

---

### DIFF
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -149,7 +149,7 @@
     "install": {
         "anticheat-warning": {
             "cancel": "No",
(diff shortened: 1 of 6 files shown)


flavioislima (approved):
LGTM

PR #4318, opened 2025-01-31
[FIX] Delete GOG manifest if present before install

arielj:
This PR prevents an issue that happens when a user tries to install a GOG game but a manifest file already exists in the `manifests` folder. The installation finishes immediately and there's no clear reason for the user.

The solution on discord is always to tell the user to delete this file, we can delete the file automatically before installation just in case.

---

### DIFF
--- a/src/backend/downloadmanager/utils.ts
+++ b/src/backend/downloadmanager/utils.ts
@@ -9,9 +9,9 @@ import { DMStatus, InstallParams, Runner } from 'common/types'
 import i18next from 'i18next'
 import { notify, showDialogBoxModalAuto } from '../dialog/dialog'
 import { isOnline } from '../online_monitor'
-import { fixesPath, isWindows } from 'backend/constants'
-import path from 'path'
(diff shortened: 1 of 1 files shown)


src/backend/downloadmanager/utils.ts
```diff
@@ -9,9 +9,9 @@ import { DMStatus, InstallParams, Runner } from 'common/types'
 import i18next from 'i18next'
 import { notify, showDialogBoxModalAuto } from '../dialog/dialog'
 import { isOnline } from '../online_monitor'
-import { fixesPath, isWindows } from 'backend/constants'
-import path from 'path'
-import { existsSync, mkdirSync } from 'graceful-fs'
+import { fixesPath, gogdlConfigPath, isWindows } from 'backend/constants'
+import pathModule from 'path'
```
arielj, line 13: I had to rename this to `pathModule` cause `path` already exists inside the `installQueueElement` function defined as a variable out of the install params

Etaash-mathamsetty (approved):
lgtm

PR #4321, opened 2025-02-02
[FIX] Enable dxvk-nvapi by default

Etaash-mathamsetty:
This has been the default since Proton 9


arielj (commented):
Just to be sure, this does not affect users that don't have an NVIDIA GPU, right?

Etaash-mathamsetty:
> Just to be sure, this does not affect users that don't have an NVIDIA GPU, right?

no it does not (at least with proton)

PR #4328, opened 2025-02-06
[UX] Warn when Shader Pre-Cache is disabled and using umu in SD's gaming mode

arielj:
This PR adds new information in the logs to help both users and support, this is more important now that we are going to default to umu and we already recommend umu to users over discord.

When using umu on the SteamDeck's gaming mode, it looks for the `steam app id` in different environment variables that are only defined when the Steam's `Shader Pre-Cache` is enabled.

When all those conditions are met, the logs will now include a warning message to instruct the users that this is not enabled and the game may not work properly.

More details: https://github.com/Open-Wine-Components/umu-launcher/wiki/Frequently-asked-questions-(FAQ)#why-am-i-not-able-to-see-my-game-when-using-my-launcher-from-steam-mode-gaming-mode

---

### DIFF
--- a/src/backend/logger/logger.ts
+++ b/src/backend/logger/logger.ts
@@ -11,13 +11,18 @@ import { backendEvents } from 'backend/backend_events'
 import { GlobalConfig } from 'backend/config'
 import { getGOGdlBin, getLegendaryBin } from 'backend/utils'
 import { dirname, join } from 'path'
-import { formatSystemInfo, getSystemInfo } from '../utils/systeminfo'
+import {
+  formatSystemInfo,
+  getSystemInfo,
+  SystemInformation
+} from '../utils/systeminfo'
 import { appendFile, writeFile } from 'fs/promises'
 import { gamesConfigPath, isWindows } from 'backend/constants'
 import { gameManagerMap } from 'backend/storeManagers'
 import { existsSync, mkdirSync, openSync } from 'graceful-fs'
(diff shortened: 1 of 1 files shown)


Etaash-mathamsetty:
I dug through the original issue on gamescope's GitHub and it seems that this has been fixed in UMU already. Probably should check though

arielj:
> I dug through the original issue on gamescope's GitHub and it seems that this has been fixed in UMU already. Probably should double check though

The umu FAQ says it's needed, and this is a message from last week saying it's something that needs to be fixed by the OS, not by umu https://github.com/Open-Wine-Components/umu-launcher/issues/92#issuecomment-2622404445

Etaash-mathamsetty:
Hmm I guess GE was referring to something else then ?

arielj:
> Hmm I guess GE was referring to some other issue being fixed with python-xlib then ? I'll test it with actual hardware tmrw, but either way I feel like this needs to be fixed gamescope side

it doesn't seem to be a gamescope issue either? umu needs the steam app id to properly handle the screens changing, and it gets the id from those env variables

the issue seems to be that the only way to get those env variables to be populated is with the Shader Pre-Caching enabled

I understand the issue would be solved if the SteamOS or its variants would just add an ENV variable or a simpler way to just get the steam app id directly without unrelated env variables parsing

flavioislima (approved):
lgtm

PR #4375, opened 2025-03-03
Fix/game page library styles and breakpoints

biliesilva:

Update for the new Game Page: some styles and grid bugs;
- Midnight Mirage theme colors updated
- Button colors updated
- colors update and new colors addition
- Game requirements style refactor
- Badge colors rafactor

---


biliesilva:

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -1,454 +1,27 @@
...
+  padding: var(--space-md);
+  gap: var(--space-md);
+  border-radius: 10px;
+  max-width: 1920px;
+  max-height: 100vh;
   grid-template-areas:
-    'top top'
-    'main extra'
+    'header header'
+    'left right'
```
arielj, line 11: if we can keep `main` and `extra` it makes it more semantic and makes more sense when we think about RTL layouts because in those cases the `main` ends up on the right side of the screen and the extra info ends up on the left side

image

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -480,31 +53,59 @@
     }
 
     &:hover {
-      color: var(--navbar-active);
+      color: var(--);
```
arielj, line 56: this doesn't look right

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -480,31 +53,59 @@
...
+      display: flex;
+      justify-content: flex-start;
+      gap: var(--space-sm);
+      align-self: start;
     }
   }
 
+  .topRowWapperInner {
+    display: flex;
+    grid-area: header2;
```
arielj, line 83: there's no grid area with the name `header2`

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -907,8 +526,6 @@
     object-fit: cover;
     filter: blur(15px);
     opacity: 0.3;
-    /* scale is so there will be no ugly edges because of the blur */
-    transform: scale(1.1);
```
arielj, line 911: I don't know why this was there, but from the comment above it seems like it shouldn't be removed (I don't know what was the original issue)

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -1,454 +1,27 @@
...
-  }
-
-  @keyframes animate-stripes {
-    100% {
-      background-position: -100px 0px;
-    }
-  }
-}
-
 .App:not(.oldDesign) .gameConfigContainer {
```
arielj, line 1: if the removal of the old design is intentional, we can remove this `.App:not(.oldDesign)` selector

src/frontend/screens/Game/GamePage/index.scss
```diff
@@ -480,31 +53,59 @@
...
     & > * {
       filter: drop-shadow(4px 4px 8px rgba(0, 0, 0, 0.25));
     }
 
     .backButton {
-      margin-inline-end: auto;
+      display: flex;
+      justify-content: flex-start;
+      gap: var(--space-sm);
+      align-self: start;
```
arielj, line 77: something in this new style is making the arrow uncentered in the button

image

src/frontend/screens/Game/GamePage/index.scss
biliesilva: I removed because in the new grid config it keeps letting a space in the bottom because of the picture in the background.

src/frontend/screens/Game/GamePage/index.scss
biliesilva: I will remove it, thanks for the catch

src/frontend/screens/Game/GamePage/index.scss
biliesilva: Thanks for the tips! I will return back to this in next weekend to polish it more and style the tabs for the ExtraInfoWrapper a little more too (it's when I have more free time). After that I will request review from you, Flavio and the other devs! 😊

Issue #4374, opened 2025-03-02
Add "disable_umu" as a global setting to avoid confusion


gabriele2000:
### Feature description

I've updated Heroic today and with horror I discovered that the option "disable_umu" is unchecked for all installed games.
I had to inspect the relevant PR to understand that the actual value is "undefined" and "tied to the previously existent experimental setting", that's not good from a "normal" user point of view.

Now there's no global option to disable it in Heroic's settings.
Reasons to not want umu in all games:
- I'd like to decide what components to install in a wineprefix
- Maybe a setting in umu-database is completely useless with future updates, I also like to use as little proprietary components as possible, even if that means waiting a month to fully play something
- Full control over what gets added to every prefix, maybe I'm using a specific .dll and I don't like when something breaks in a Windows OS style after an update.

### Alternatives

The solution would be to put in settings->advanced the option "disable_umu" to enforce it globally.
The user could then disable that checkbox manually for every game.

imLinguin:
While you raised a vaild point to be able to disable umu globally. As per what @CommandMC mentioned on Discord, it may be good idea to keep it per-game, as ideally you should never disable UMU. 

You are fundamentally mistaking on what umu is for.  It's **not**  to provide fixes - the database is *just* for mapping ids so protonfixes and potentially other patches included in Proton get triggered on non-steam releases of games.
What you want to achieve here is not tied to umu itself, but to **GE and UMU Proton** builds. 
umu is in charge of running Proton inside of a container environment ensuring all dependencies of wine are in place, thus it should only be disabled when you really know what you are doing.

And if you really want to disable protonfixes for those Proton builds, you should simply use `PROTONFIXES_DISABLE=1` environment variable.

gabriele2000:
> as ideally you should never disable UMU

I used Heroic for years using Proton-GE, when UMU wasn't even in mind

CommandMC:
- Please read the last comment again. Disabling umu will not disable Protonfixes
- I understand your position of wanting to know what exactly happens when running a game, and being in control of what does & doesn't happen in a prefix. Please understand that none of the tools you are using are intended for this use case.  
[...]

PR #4378, opened 2025-03-03
[UX] Allow setting the `Disable UMU` default in General Settings > Game Defaults

arielj:
This PR adds the `Disable UMU` setting in the `defaults` in general settings so it can be configured for all new games added instead of disabling it manually for each game if the user doesn't want to use umu at all.

This doesn't globally disable umu though, for games that are already installed it has to be manually disabled.

I also added some tests for the global config and game config to start adding tests for these things when we change defaults and that (it's not covering all settings and many cases, but it's a start)

Since it doesn't exactly disable UMU globally, it doesn't fully solve https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4374, but I think it's good enough since the option to change the default value aligns more with how the other settings work.

---

### DIFF
--- a/src/backend/__tests__/config.test.ts
+++ b/src/backend/__tests__/config.test.ts
@@ -6,7 +6,7 @@ import {
   heroicInstallPath,
   setPlatformConstants
 } from 'backend/constants'
-import { existsSync, rmSync } from 'graceful-fs'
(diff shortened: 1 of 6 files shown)


arielj:
I'll close this

as discussed in the original issue users shouldn't really disable UMU and the reasons why it was requested in the first place were clarified in that issue

PR #4379, opened 2025-03-03
fix: use egs's flow for epic login

imLinguin:
Instead of routing through legendary.gl/epiclogin we will call different url directly, this should prevent any issues with corrective actions etc..

I have also modified our useragent to contain mention about Heroic

### DIFF
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -61,7 +61,7 @@ export default function WebView() {
     lang = 'pt-BR'
   }
 
-  const epicLoginUrl = 'https://legendary.gl/epiclogin'
+  const epicLoginUrl = 'https://www.epicgames.com/id/login?responseType=code'
 
   const epicStore = `https://www.epicgames.com/store/${lang}/`
   const gogStore = `https://af.gog.com?as=1838482841`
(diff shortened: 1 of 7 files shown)


arielj (approved):
Works as expected and way more simple than before

src/frontend/screens/WebView/index.tsx
```diff
@@ -318,9 +292,10 @@ export default function WebView() {
     setShowLoginWarningFor(null)
   }
 
-  if (!preloadPath && isEpicLogin) {
-    return <></>
-  }
+  const userAgent =
+    startUrl === epicLoginUrl
+      ? 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) EpicGamesLauncher'
+      : 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/200.0 HeroicGamesLauncher'
```
flavioislima, line 298: Hmm, maybe we don't need those double UA tbh, I only added this one to try to fix some issues with random errors on epic login, server errors.

flavioislima (approved):
Good stuff with this refactor

src/frontend/screens/WebView/index.tsx
imLinguin: I'm thinking it may be worth keeping for some browser content. Or in the future let users configure user agent strings for browser apps

Issue #4377, opened 2025-03-03
"Minified React error #137" after upgrade to 2.16.0 (Flatpak version)


lovenemesis:
### Describe the bug

After upgrading to 2.16.0 release from flathub, a "Minified React error" is immediately encountered upon launching HGL.

### Add logs

```text
Cannot be provided since the setting menu is not accessible.
```

### Steps to reproduce

1. Upgrade HGL to 2.16.0 via Flathub
2. Launch HGL

![Image](https://github.com/user-attachments/assets/9dd45743-6f4c-4174-996a-98c18f7fc73c)

### Expected behavior

See the HGL main UI as usual.

### Heroic Version

Latest Stable

### System Information

- OS [e. g. "Ubuntu"]:

edelweiss819:
Same for AUR version.

victorsouzaleal:
Same, installed by .rpm file on Fedora 41.

arielj:
There's a PR to fix this for the next hotfix release

In the meantime, you can edit the config files of Heroic in `~/.config/heroic/config.json` and `~/.config/heroic/store/config.json` and set `en` as the language.

After you log into a store and something is added to your library, you should be able to change back to the language of your choice.

outerwinnie:
This is happening to me in Windows 11 too.

arielj:
For windows, the config files to edit are in `C:\Users\<your user>\AppData\Local\Programs\heroic`

arielj:
On mac: `/Users/<your user>/Library/Application Support/heroic`

crtmitchn:
It is actually C:\Users\<user>\AppData\Roaming\heroic for Windows

lovenemesis:
Thanks for the quick feedback. I hope the hotfix will arrive, soon as the configuration file change does NOT work for me:

```
{
  "defaultSettings": {
    "checkUpdatesInterval": 10,
    "enableUpdates": false,
    "addDesktopShortcuts": false,
    "addStartMenuShortcuts": false,
    "autoInstallDxvk": true,
    "autoInstallVkd3d": true,
    "autoInstallDxvkNvapi": false,
    "addSteamShortcuts": false,
    "preferSystemLibs": false,
    "checkForUpdatesOnStartup": false,
    "autoUpdateGames": true,
    "customWinePaths": [],
    "defaultInstallPath": "/home/username/Games/Heroic",
    "libraryTopSection": "disabled",
[...]

PR #4380, opened 2025-03-03
[DX] Allow using `br` tags in translations to avoid confusion

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4377

There were many translations that were not respecting the indexed placeholders from i18next, adding content between `br` tags that don't support content.

This PR changes a few things:
- I removed all the translations that were wrong
- I updated the ones that were correct to use `<br />` instead of `<X></X>` to avoid confusion in weblate
- I fixed the config of i18next-parser to generate `<br />` tags in the english file
- I added a script that we can run to run some checks on the translations, I didn't make this a linter step in the CI actions because I don't know if it's correct the way it's implemented now to make it a block (in some languages maybe it makes sense to skip tags), so for now it's a helper script that we can run before releases for a sanity check

---


flavioislima (approved):
LGTM

PR #4385, opened 2025-03-04
[UX] Copy fake EpicGamesLauncher.exe for games flagged with env variable (like Rockstar games)

arielj:
This PR depends on having the legendary binary from: https://github.com/Heroic-Games-Launcher/legendary/pull/13

How to test:
- Install GTAV or RDR2 or any rockstar game
- configure the custom legendary binary from the link above
- make sure the game is NOT configure with the `fix.bat` workaround
- set an env variable for the game `USE_FAKE_EPIC_EXE=true`
- launch the game

---


public/bin/EpicGamesLauncher.exe
arielj, line 1: I think this file is small enough and shouldn't really require change in the future, I think it's safe to include it in the build directly to make it simpler

CommandMC: IMO we should still download this with the usual `meta/downloadHelperBinaries.ts` script, if nothing else than to track where this EXE comes from (I assume right now it's <https://github.com/Etaash-mathamsetty/heroic-epic-integration/releases/tag/v0.3>?)

Etaash-mathamsetty: > IMO we should still download this with the usual `meta/downloadHelperBinaries.ts` script, if nothing else than to track where this EXE comes from (I assume right now it's [Etaash-mathamsetty/heroic-epic-integration@`v0.3` (release)](https://github.com/Etaash-mathamsetty/heroic-epic-integration/releases/tag/v0.3)?)

I agree it would make things easier overall

meta/downloadHelperBinaries.ts
```diff
@@ -59,6 +65,23 @@ async function downloadAsset(
   }
 }
 
+//FIXME: better name for this function
+async function downloadAsset2(
```
Etaash-mathamsetty, line 69: What do I call this function, I don't want to use the same name as downloadAsset although that's what it does

CommandMC: Why not just use the existing function & change the path used in Heroic?

Etaash-mathamsetty: good idea

flavioislima (approved):
Code looks good to me

PR #4397, opened 2025-03-08
[UX] Default to WineCrossover and DXVK on for Intel macs

arielj:
Currently, Heroic defaults to using the Game Porting Toolkit for any mac machine, even though it doesn't support Intel mac. Also, because of that default, DXVK is turned off by default.

The default experience for Intel mac users is really poor because Heroic was setting an invalid configuration from the beginning, leading to a lot of support threads.

This PR changes that to default to Wine-Crossover and DXVK turned on when the computer is an Intel mac.

---

### DIFF
--- a/src/frontend/screens/WineManager/index.tsx
+++ b/src/frontend/screens/WineManager/index.tsx
@@ -41,7 +41,7 @@ export default function WineManager(): JSX.Element | null {
     </p>
   )
 
-  const { refreshWineVersionInfo, refreshing, platform } =
+  const { refreshWineVersionInfo, refreshing, platform, isIntelMac } =
     useContext(ContextProvider)
   const isLinux = platform === 'linux'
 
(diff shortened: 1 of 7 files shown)


src/frontend/screens/WineManager/index.tsx
```diff
@@ -55,17 +55,22 @@ export default function WineManager(): JSX.Element | null {
...
   }
+  const wineCrossover: WineManagerUISettings = {
+    type: 'Wine-Crossover',
+    value: 'winecrossover',
+    enabled: !isLinux
+  }
 
   const [repository, setRepository] = useState<WineManagerUISettings>(
-    isLinux ? protonge : gamePortingToolkit
+    isLinux ? protonge : isIntelMac ? wineCrossover : gamePortingToolkit
```
flavioislima, line 65: I think those double ternary are not clean, maybe a method for that would be better.

flavioislima (approved):
Looks good, left a small comment

PR #4402, opened 2025-03-09
[Feat] macOS: Check if Rosetta is available on Startup

flavioislima:
This will check if Rosetta is available on Macs with Apple Silicon and show a warning with a message on how to install it.
Also checks if macOS is update to Sonoma or Higher for better GPTK support..

### AI Summary

This pull request includes several important changes to improve compatibility and functionality on macOS systems, particularly for Apple Silicon chips. The changes involve adding checks for macOS version and Rosetta installation, as well as updating the logic for selecting the appropriate Wine version based on the system architecture.

Improvements for macOS compatibility:

* [`src/backend/constants.ts`](diffhunk://#diff-a69d71242c1711201614ab806056a02d4a8c610532e65740433e90bafdfdcfb5L31-R31): Added type annotation for `isIntelMac` to ensure type safety.
* [`src/backend/launcher.ts`](diffhunk://#diff-f435a4df94baea61fbede2b6e8016c233830b8330d6104bd7ad0dec8563422ceL38-R39): Added checks to ensure compatibility with macOS Sonoma or higher and to verify if Rosetta is installed on Apple Silicon Macs. [[1]](diffhunk://#diff-f435a4df94baea61fbede2b6e8016c233830b8330d6104bd7ad0dec8563422ceL38-R39) [[2]](diffhunk://#diff-f435a4df94baea61fbede2b6e8016c233830b8330d6104bd7ad0dec8563422ceR1031-R1038)
* [`src/backend/main.ts`](diffhunk://#diff-859af3c776da914ebff60fa13adfc96e79d99ea563fe1e4e790de14656442accL55-R56): Integrated Rosetta installation check during the initialization process for macOS systems. [[1]](diffhunk://#diff-859af3c776da914ebff60fa13adfc96e79d99ea563fe1e4e790de14656442accL55-R56) [[2]](diffhunk://#diff-859af3c776da914ebff60fa13adfc96e79d99ea563fe1e4e790de14656442accR172-R174)
* [`src/backend/utils.ts`](diffhunk://#diff-222385e29788efeec1fea5c8210b2121c8f42dff67759bd5137211e9a82a3012L922-R927): Added functions `checkRosettaInstall` and `isMacSonomaOrHigher` to handle Rosetta installation verification and macOS version checks. [[1]](diffhunk://#diff-222385e29788efeec1fea5c8210b2121c8f42dff67759bd5137211e9a82a3012L922-R927) [[2]](diffhunk://#diff-222385e29788efeec1fea5c8210b2121c8f42dff67759bd5137211e9a82a3012R1292-R1354)
* [`src/frontend/screens/WineManager/index.tsx`](diffhunk://#diff-9948b95402f27a7ae54e95d998362c91d1dd3bb71a2f6944eb1d6d1f6adc886fR64-R77): Refactored logic for setting the default Wine repository based on the system's architecture, simplifying the code and improving readability.

Version update:

* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L3-R3): Updated the version from `2.16.0` to `2.16.1`.

---

### DIFF
--- a/src/backend/constants.ts
+++ b/src/backend/constants.ts
@@ -28,7 +28,7 @@ const fontsStore = new TypeCheckedStoreBackend('fontsStore', {
 })
 
 const isMac = process.platform === 'darwin'
-const isIntelMac = isMac && cpus()[0].model.includes('Intel') // so we can have different behavior for Intel Mac
+const isIntelMac: boolean = isMac && cpus()[0].model.includes('Intel') // so we can have different behavior for Intel Mac
 const isWindows = process.platform === 'win32'
 const isLinux = process.platform === 'linux'
 const isSteamDeckGameMode = process.env.XDG_CURRENT_DESKTOP === 'gamescope'
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -52,7 +52,8 @@ import {
   getCurrentChangelog,
   removeFolder,
   downloadDefaultWine,
-  sendGameStatusUpdate
+  sendGameStatusUpdate,
+  checkRosettaInstall
 } from './utils'
 import { uninstallGameCallback } from './utils/uninstaller'
 import {
@@ -168,6 +169,9 @@ async function initializeWindow(): Promise<BrowserWindow> {
     Winetricks.download()
     if (!availableWine.length) {
       downloadDefaultWine()
+      if (isMac) {
+        checkRosettaInstall()
+      }
     }
   }, 2500)
 
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -919,11 +919,12 @@ export async function downloadDefaultWine() {
   // get list of wines on wineDownloaderInfoStore
   const availableWine = wineDownloaderInfoStore.get('wine-releases', [])
   // use GE-Proton type if on Linux and GamePortingToolkit if on Mac
-  const release = availableWine.find((version) => {
+  const release = availableWine.find(async (version) => {
     if (isLinux) {
       return version.type === 'GE-Proton'
     } else if (isMac) {
-      if (isIntelMac) {
+      const isMacOSUpToDate = await isMacSonomaOrHigher()
+      if (isIntelMac && isMacOSUpToDate) {
         return version.type === 'Wine-Crossover'
       } else {
         return version.type === 'Game-Porting-Toolkit'
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -1283,6 +1283,69 @@ async function getPathDiskSize(path: string): Promise<number> {
   return statData.size
 }
 
+export async function checkRosettaInstall() {
+  if (!isMac) {
+    return
+  }
+
+  // check if on arm64 macOS
+  const { stdout: archCheck } = await execAsync('arch')
+  const isArm64 = archCheck.trim() === 'arm64'
+
+  if (!isArm64) {
+    return
+  }
+
+  const { stdout: rosettaCheck } = await execAsync(
+    'arch -x86_64 /usr/sbin/sysctl sysctl.proc_translated'
+  )
+
+  const result = rosettaCheck.split(':')[1].trim() === '1'
+
+  logInfo(
+    `Rosetta is ${result ? 'available' : 'not available'} on this system.`,
+    LogPrefix.Backend
+  )
+
+  if (!result) {
+    // show a dialog saying that hyperplay wont run without rosetta and add information on how to install it
+    await dialog.showMessageBox({
+      title: i18next.t('box.warning.rosetta.title', 'Rosetta not found'),
+      message: i18next.t(
+        'box.warning.rosetta.message',
+        'Heroic requires Rosetta to run correctly on macOS with Apple Silicon chips. Please install it from the macOS terminal using the following command: "softwareupdate --install-rosetta" and restart HyperPlay. '
+      ),
+      buttons: ['OK'],
+      icon: icon
+    })
+
+    logInfo(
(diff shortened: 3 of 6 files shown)


src/backend/utils.ts
```diff
@@ -1288,6 +1290,69 @@ async function getPathDiskSize(path: string): Promise<number> {
...
+
+  const result = rosettaCheck.split(':')[1].trim() === '1'
+
+  logInfo(
+    `Rosetta is ${result ? 'available' : 'not available'} on this system.`,
+    LogPrefix.Backend
+  )
+
+  if (!result) {
+    // show a dialog saying that hyperplay wont run without rosetta and add information on how to install it
```
CommandMC, line 1318:
```suggestion
    // show a dialog saying that Heroic wont run without rosetta and add information on how to install it
```

src/backend/utils.ts
```diff
@@ -1288,6 +1290,69 @@ async function getPathDiskSize(path: string): Promise<number> {
...
+    await dialog.showMessageBox({
+      title: i18next.t('box.warning.rosetta.title', 'Rosetta not found'),
+      message: i18next.t(
+        'box.warning.rosetta.message',
+        'Heroic requires Rosetta to run correctly on macOS with Apple Silicon chips. Please install it from the macOS terminal using the following command: "softwareupdate --install-rosetta" and restart HyperPlay. '
```
CommandMC, line 1323:
```suggestion
        'Heroic requires Rosetta to run correctly on macOS with Apple Silicon chips. Please install it from the macOS terminal using the following command: "softwareupdate --install-rosetta" and restart Heroic. '
```

src/backend/constants.ts
```diff
@@ -28,7 +28,7 @@ const fontsStore = new TypeCheckedStoreBackend('fontsStore', {
 })
 
 const isMac = process.platform === 'darwin'
-const isIntelMac = isMac && cpus()[0].model.includes('Intel') // so we can have different behavior for Intel Mac
+const isIntelMac: boolean = isMac && cpus()[0].model.includes('Intel') // so we can have different behavior for Intel Mac
```
arielj, line 31: weird, shouldn't this be inferred as boolean? `isMac` is boolean and `includes` returns a boolean

src/backend/utils.ts
```diff
@@ -1288,6 +1290,69 @@ async function getPathDiskSize(path: string): Promise<number> {
   return statData.size
 }
 
+export async function checkRosettaInstall() {
+  if (!isMac) {
```
arielj, line 1294: this is only called inside an `if (isMac)` check, so we don't need to check again (or we don't need the other check)

src/backend/utils.ts
```diff
@@ -1288,6 +1290,69 @@ async function getPathDiskSize(path: string): Promise<number> {
...
+    return
+  }
+
+  // check if on arm64 macOS
+  const { stdout: archCheck } = await execAsync('arch')
+  const isArm64 = archCheck.trim() === 'arm64'
```
arielj, line 1300: why not use the `isIntelMac` constant here? then we don't need to execute a process

if it's not intel mac then it's arm

src/backend/utils.ts
```diff
@@ -919,11 +920,12 @@ export async function downloadDefaultWine() {
...
   const availableWine = wineDownloaderInfoStore.get('wine-releases', [])
   // use GE-Proton type if on Linux and GamePortingToolkit if on Mac
-  const release = availableWine.find((version) => {
+  const release = availableWine.find(async (version) => {
     if (isLinux) {
       return version.type === 'GE-Proton'
     } else if (isMac) {
-      if (isIntelMac) {
+      const isMacOSUpToDate = await isMacSonomaOrHigher()
+      if (isIntelMac && isMacOSUpToDate) {
```
arielj, line 928: is this check correct? because this means that we'd set wine-crossover only if intel mac and new OS version, when I think we want to only set game porting toolkit if NOT intel mac and up-to-date os

we actually want wine-crossover always for intel macs

src/backend/utils.ts
```diff
@@ -1288,6 +1290,69 @@ async function getPathDiskSize(path: string): Promise<number> {
...
+    logInfo(
+      'Rosetta is not available, install it with: softwareupdate --install-rosetta from the terminal',
+      LogPrefix.Backend
+    )
+  }
+}
+
+export async function isMacSonomaOrHigher() {
+  if (!isMac) {
+    return false
```
arielj, line 1338: I think it's better to return `null` here because it's not even a macos at this point, it's not that it's not sonoma or higher, and then we can be more explicit on how we handle this

src/backend/main.ts
```diff
@@ -168,6 +169,9 @@ async function initializeWindow(): Promise<BrowserWindow> {
     Winetricks.download()
     if (!availableWine.length) {
       downloadDefaultWine()
+      if (isMac) {
```
arielj, line 172: I don't know if this is the right place to check this though, cause the first time a user opens heroic, it will download a wine and run this check, but the second time it won't run this cause `availableWine.length` will be more than 0

I think this should be done somewhere else so we can show it in the logs always if needed

src/backend/utils.ts
flavioislima: TRue, should be !isMacOsUpToDate then. good catch.

src/backend/utils.ts
flavioislima: We dont need this though, same case as you pointed above, it is only called inside isMac checks

Issue #2932, opened 2023-07-29
Confusing behavior of `Don't ask again` when opening external link


arielj:
### Describe the bug

When clicking one of the links in the sidebar (discord/kofi/patreon), a dialog appears warning the user that the action will open an external link.

If you click `Don't ask again` and CANCEL the action, the next time you click one of those buttons it opens the external link without asking.

It is confusing that cancelling the action actually applies the `Don't ask again` value.

### Add logs

```text
nothing relevant, this configuration is stored in the local storage of the browser
```

### Steps to reproduce

1. Click any link in the sidebar (ie: Discord)
2. Check `Don't ask again`
3. Click Cancel
4. Click any link in the sidebar
5. The dialog does not appear anymore

### Expected behavior

At least for me, I would expect cancelling to have NOT effect at all.

### Heroic Version

Latest Stable

### System Information

not relevant

PR #4414, opened 2025-03-13
[FIX] Confusing behavior of Don't ask again when opening external link

KaranDuggal:
<--- Confusing behavior of Don't ask again when opening external link fix--->

---

### DIFF
--- a/src/frontend/components/UI/ExternalLinkDialog/index.tsx
+++ b/src/frontend/components/UI/ExternalLinkDialog/index.tsx
@@ -13,20 +13,21 @@ export default function ExternalLinkDialog() {
     useContext(ContextProvider)
 
   function onClose() {
+    setShowDialog(false)
     handleExternalLinkDialog({ showDialog: false, linkCallback: undefined })
   }
 
   function onContinue() {
+    onHideDialogChange()
     if (externalLinkDialogOptions.linkCallback !== undefined)
       externalLinkDialogOptions.linkCallback()
     onClose()
   }
 
   function onHideDialogChange() {
-    setShowDialog(!showDialog)
     localStorage.setItem(
       SHOW_EXTERNAL_LINK_DIALOG_STORAGE_KEY,
-      showDialog.toString()
+      showDialog ? 'false' : 'true'
     )
   }
 
@@ -38,7 +39,7 @@ export default function ExternalLinkDialog() {
         <br></br>
         <ToggleSwitch
           htmlId="externalLinkDialog"
-          handleChange={onHideDialogChange}
+          handleChange={() => setShowDialog(!showDialog)}
           value={showDialog}
           title={t('externalLink.dontAskAgain', "Don't ask again")}
         ></ToggleSwitch>


flavioislima:

KaranDuggal:

arielj:
can you explain a bit what you are changing here? a before/after example would be great

src/frontend/components/UI/ExternalLinkDialog/index.tsx
```diff
@@ -38,7 +39,7 @@ export default function ExternalLinkDialog() {
         <br></br>
         <ToggleSwitch
           htmlId="externalLinkDialog"
-          handleChange={onHideDialogChange}
+          handleChange={() => setShowDialog(!showDialog)}
```
KaranDuggal, line 42: Previously, when the user changed the value of the "Don't ask again" checkbox, we were calling the onHideDialogChange function. In this function, we were immediately updating the value in local storage. This behavior was incorrect because the user had not yet confirmed their choice by clicking the "Continue" button.

If the user checked the "Don't ask again" checkbox and then clicked the "Cancel" button, the change in local storage was not reverted. This caused confusion for the user.

With my changes, when the user toggles the "Don't ask again" checkbox, the value is only updated in a variable, not in local storage. The local storage is now updated only when the user clicks the "Continue" button.

src/frontend/components/UI/ExternalLinkDialog/index.tsx
```diff
@@ -13,20 +13,21 @@ export default function ExternalLinkDialog() {
...
     if (externalLinkDialogOptions.linkCallback !== undefined)
       externalLinkDialogOptions.linkCallback()
     onClose()
   }
 
   function onHideDialogChange() {
-    setShowDialog(!showDialog)
     localStorage.setItem(
       SHOW_EXTERNAL_LINK_DIALOG_STORAGE_KEY,
-      showDialog.toString()
```
KaranDuggal, line 29: This line of code is not correct because when the user checks the "Don't ask again" checkbox, the showDialog variable is set to true, and we are storing this true value in the SHOW_EXTERNAL_LINK_DIALOG_STORAGE_KEY, which is incorrect. It should be false because we don't want to show the popup when the "Don't ask again" checkbox is checked.

KaranDuggal:
Hi @arielj @flavioislima  Apologies for the delayed response. I was tied up with some work. Please review my comment, and feel free to let me know if you have any questions or suggestions.

flavioislima:
@KaranDuggal I cannot replicate the issue. For me checking the do not show again option makes the dialog to not show again even after restarting the app.

arielj:
> @KaranDuggal I cannot replicate the issue. For me checking the do not show again option makes the dialog to not show again even after restarting the app.

@flavioislima I understand this is fixing this issue https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2932

the thing is that we currently apply the `Don't ask again` checkbox even if the user CANCELS the action

we should only apply the `Don't ask again` change if the user accepted the dialog, not when cancelled, because if I click `don't ask again` and cancel the action I would expect to be asked again the next time

arielj (approved):
tested it and it works as expected:
- click external link => dialog is displayed
- check "Don't ask me again"
- click Cancel
- click external link => dialog is displayed again cause we cancelled the action
- check "Don't ask me again"
- click Continue
- click external link => dialog is not displayed now cause we continued

flavioislima:
Ah ok, got it now

PR #4423, opened 2025-03-16
[DX] Test that the empty library and settings page renders correctly for all languages

arielj:
After the issue we had in the 2.16.0 release I think it's a good idea to add this test to make sure the empty library and the settings can be accessed for all languages. It should help catch broken translations.

I'll try to find a way to test the library with content too since that's also a valid initial state, but that will be another PR cause I don't think we have a way to do that currently in tests.

---


flavioislima (approved):
good stuff

PR #4462, opened 2025-04-06
[macOS - Feature] Install Windows Steam with one Click

flavioislima:
This ports HyperPlay feature of installing the Steam for Windows on macOS with one click:
https://github.com/HyperPlay-Gaming/hyperplay-desktop-client/pull/1051

It needs to use a modified version of Steam that still works on Windows 7 since new CEF on new Steam version won't work even with GPTK.

The new Install Steam button is located on the side of Add Game:

image

Install dialog (ignore Proton since this was screenshoted from linux):

image

Next step would be to automatically import the games installed on this version of Steam to the Heroic Library.

---


src/frontend/screens/Library/components/SteamInstall/SteamInstallDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+          </span>
+          {t('Steam Installation')}
+        </h6>
+      </div>
+      <div>
+        <ul>
+          <li>
+            {t(
+              'Recommended specs for a better experience',
+              'Recommended specs for a better experience'
```
arielj, line 55: these `t` calls with the key being the same as the english string doesn't match what we do in the rest of the app

src/backend/tools/index.ts
```diff
@@ -835,3 +840,182 @@ export async function runWineCommandOnGame(
...
+    writeConfig('steam', {
+      ...gameSettings,
+      winePrefix,
+      wineVersion,
+      launcherArgs:
+        '-noverifyfiles -nobootstrapupdate -skipinitialbootstrap -norepairfiles -overridepackageurl'
+    })
+
+    try {
+      await runWineCommand({
```
arielj, line 927: would this be a problem if the default wine is crossover? we don't want to run wineboot -init in those cases, we won't have a prefix

package.json
```diff
@@ -65,6 +65,7 @@
     "@mui/material": "5.11.12",
     "@node-steam/vdf": "2.2.0",
     "@shockpkg/icon-encoder": "2.1.3",
+    "@tanstack/react-query": "^5.71.10",
```
arielj, line 68: do we HAVE to add this new dependency? I'm always worried about adding new dependencies and I see it's used to fetch the list of wines, and we already do that in other places without this dependency, can't we re-use what we did in other places instead?

flavioislima: I didn't like to use React Query at first, until I started to doing it. 
Their error handling, loading state for any request we want to and it is also not a heavy dependency. I think we should start using it more, it might improve a lot the way we handle request on the frontend, especially handling errors and refreshes.

src/backend/tools/index.ts
```diff
@@ -835,3 +840,182 @@ export async function runWineCommandOnGame(
...
   })
 }
+
+/**
+ * Download and Install the Windows version of Steam on a Wine Prefix on macOS only
+ */
+export const SteamWindows = {
+  downloadSteam: async () => {
+    const steamURL =
+      'https://drive.usercontent.google.com/download?id=1MXowtPBihbljXFs_5cM1AfyOjnC3raaT&export=download&authuser=0&confirm=t&uuid=e159346a-5836-4f7d-8aed-af3d8340fef7&at=AENtkXa2d83g4dE_UZYQKKLoEUje%3A1732794843031'
```
arielj, line 850: this looks.... weird

we'll have to be really clear to the user why are we getting a steam zip file from a random google drive account

why would they trust that we are installing something safe if they can't even see where is this coming from?

I understand it's because of the thing you mentioned about steam not supporting windows 7 and GPTK, but we need to be really clear and let the user know what's going on before they confirm anything

src/frontend/hooks/hasStatus.ts
```diff
@@ -24,9 +27,25 @@ export function hasStatus(
   const {
     thirdPartyManagedApp = undefined,
     is_installed,
-    runner,
+    runner = 'sideload',
```
arielj, line 30: I'm confused about this change, why we default to `sideload` for this custom hook?

package.json
CommandMC: I second this. Looking at how this dependency is used right now, I don't see anything that *can't* be implemented with regular React hooks or our `GlobalStateV2`.  
I'm willing to compromise on this staying in for now, but in that case I'll look into removing it again in a future PR

arielj: can we introduce this in a different PR maybe? I think it will help keep this PR simpler (I think it will actually solve your issues with the workflow if the lockfile is not changed?)

I've never used react query, not really against it but I think it's better if it's done in another PR, and I know there are alternatives  and also Suspense

src/backend/tools/index.ts
CommandMC: Other than this looking sketchy, it also won't work since GDrive makes a file unavailable if it's downloaded too frequently (which I assume will happen sometimes). Can we use an `archive.org` link for this?

src/frontend/screens/Library/components/SteamInstall/SteamInstallButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+    : t(
+        'steam-install.tooltip',
+        `This will install the Windows version of Steam and it will run using a Heroic Compatibility Layer. \n
+    Some games especially older titles, newer 3D games, or those with anti-cheat software may not work properly. \n
+    The Recommended specs are a Mac with Apple Silicon CPU (M1 or above) and at least 16GB of RAM.`
+      )
+
+  const compatibilityLayerNotAvailableTooltip = t(
+    'steam-install.compatibility-layer-not-available',
+    `Compatibility layer not available. Please install one from "Settings > Wine Manager" first.`
```
arielj, line 93: I was wondering, can this actually happen? we auto install one at boot if none is available

I can only imagine a user uninstalling all wines while heroic is running to end up in that state

or heroic failing to install a wine during boot

src/backend/tools/index.ts
flavioislima: Yeah, I can upload it to archive.org

package.json
flavioislima: they serve diferent porpuses @arielj 
I can remove ofc, but would be something good to have on Heroic since we need to do a lot of requests and the way react-query deals with errors and loading state for all requests is nice.
Normally we would need to use a setState and a useEffect to control the loading state of a request, with react-query we won't need that.

src/frontend/hooks/hasStatus.ts
flavioislima: If I dont change this I need the gameinfo to get the status. The problem is that there is no gameinfo yet since Steam was not installed. It is a small change that won't change anything that uses this hook. I tested already all the states on the gamecard and gamepage. It is just a default and the other hooks passes the gameinfo so its totally safe to do that.

src/backend/tools/index.ts
flavioislima: should we just ignore the command args in this case? I thought runWineCommand already dealt with crossover.

flavioislima: I will just ignore this command then, I think I understood what you meant

src/backend/utils.ts
```diff
@@ -1438,7 +1439,7 @@ export async function downloadFile({
             )}  @${bytesToSize(speed)}/s (${percentage.toFixed(2)}%)`,
             LogPrefix.Backend
           )
-          progressCallback(bytes, speed, percentage, writingSpeed)
+          progressCallback(bytes, speed, writingSpeed, percentage)
```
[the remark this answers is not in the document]
flavioislima: this was broken before, it was in the wrong order above.

src/backend/logger/logger.ts
```diff
@@ -340,8 +340,8 @@ export function logChangedSetting(
     if (
       Array.isArray(config[changedSetting]) &&
       Array.isArray(oldConfig[changedSetting]) &&
-      config[changedSetting].length === 0 &&
-      oldConfig[changedSetting].length === 0
+      config[changedSetting] &&
+      oldConfig[changedSetting]
```
CommandMC, line 344: These conditions will now no longer do anything, arrays are objects and thus always truthy

src/backend/tools/index.ts
```diff
@@ -830,3 +834,183 @@ export async function runWineCommandOnGame(
...
+    }
+
+    const abortController = createAbortController('steam-download')
+
+    function handleProgress(
+      downloadedBytes: number,
+      downloadSpeed: number,
+      progress: number
+    ) {
+      window?.webContents.send(`progressUpdate-steam`, {
```
CommandMC, line 864: Prefer `sendFrontendMessage` here

src/backend/tools/index.ts
```diff
@@ -830,3 +834,183 @@ export async function runWineCommandOnGame(
...
+      })
+    } catch (error) {
+      logError(['Error downloading Steam', error], LogPrefix.Backend)
+      throw new Error('Error downloading Steam')
+    }
+  },
+  installSteam: async (path: string) => {
+    const steamCoverArt =
+      'https://cdn2.steamgriddb.com/file/sgdb-cdn/grid/a7e8ba67562ea4d4ca0421066466ece4.png'
+    const steamSetupPath = `${toolsPath}/steam/Steam.zip`
```
CommandMC, line 891: This path is defined multiple times (once in the above function, once in this, once again further down). Ideally keep it as one variable in the top level

src/backend/tools/index.ts
```diff
@@ -830,3 +834,183 @@ export async function runWineCommandOnGame(
...
+    let { wineVersion } = GlobalConfig.get().getSettings()
+
+    if (wineVersion.type !== 'toolkit') {
+      const wineList = await GlobalConfig.get().getAlternativeWine()
+      const gptk = wineList.find((wine) => wine.type === 'toolkit')
+      if (gptk) {
+        wineVersion = gptk
+      }
+    }
```
CommandMC, line 900: As I understand, we always want to use GPTK. Shouldn't we then abort here if we don't have a Wine version with that type?

src/backend/tools/index.ts
```diff
@@ -830,3 +834,183 @@ export async function runWineCommandOnGame(
...
+        await runWineCommand({
+          commandParts: ['wineboot', '-init'],
+          wait: true,
+          gameSettings: {
+            ...gameSettings,
+            winePrefix: path,
+            wineVersion
+          }
+        })
+      }
```
CommandMC, line 930: Prefer `verifyWinePrefix` here

src/frontend/screens/Library/components/SteamInstall/SteamInstallDialog.tsx
```diff
@@ -0,0 +1,208 @@
...
+}) => {
+  const { t } = useTranslation('gamepage')
+  const [progress] = hasProgress('steam')
+  const [isHovering, setIsHovering] = useState(false)
+  const [installPath, setInstallPath] = useState('')
+
+  useEffect(() => {
+    const getDefaultWinePrefix = async () => {
+      const { defaultWinePrefix } = await window.api.requestAppSettings()
+      console.log('prefix', defaultWinePrefix)
```
CommandMC, line 42: Probably want to remove that

Issue #4464, opened 2025-04-08
Allow picking of local image for game's banner/picture


wantija:
### Problem description

When you add a game to the launcher, the 'App Image' only allows you to paste a URL of the image you want to use, so the image has to be on the web.

### Feature description

Allow users to pick an image from their storage for the game's display image, a 'Browse' button that brings up a filepicker would be ideal.

### Additional information

I shouldn't have to upload an image to the internet of the game I want to add when I have it on my local storage.

imLinguin:
try using `file://<location to image>`

wantija:
Well, that does work, but it should be more intuitive, I would have never figured it out without reporting this issue.

imLinguin:
I agree it should, I'm just showing the workaround for the time being

sandepten:
Hey, I would like to work on this!

I am thinking of adding a simple upload button on the right of the input which will add the file url

PR #4468, opened 2025-04-11
Pick local files with file picker for Game Image

sandepten:
Hey, I have added the ability to pick local files with file picker for Game Image (issue #4464)

Screenshot From 2025-04-11 22-28-49

### DIFF
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -22,7 +22,8 @@
             "title": "Select EXE to Run"
         },
         "select": {
-            "button": "Select"
+            "button": "Select",
+            "image": "Select Image"
         },
         "sideload": {
             "exe": "Select Executable"
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -237,7 +237,7 @@
             "useragent": "Custom User Agent"
         },
         "placeholder": {
-            "image": "Paste an Image URL here",
+            "image": "Paste an URL of an Image or select one from your computer",
             "title": "Add a title to your Game/App",
             "url": "Paste the Game URL here",
             "useragent": "Write a custom user agent here to be used on this browser app/game"
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -150,6 +150,10 @@
                 "install": "Epic Servers are having major outage right now, the game cannot be installed!",
                 "update": "Epic Servers are having major outage right now, the game cannot be updated!"
             },
+            "rosetta": {
+                "message": "Heroic requires Rosetta to run correctly on macOS with Apple Silicon chips. Please install it from the macOS terminal using the following command: \"softwareupdate --install-rosetta\" and restart Heroic. ",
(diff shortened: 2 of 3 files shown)


sandepten:

flavioislima:
Hi! Thanks for tackling this issue.

One small thing, can you update the text inside the input?
Maybe to something like "Paste an URL of an Image or select one from your computer".

sandepten:
Hi @flavioislima, I have updated the text inside the input

imLinguin:
It may be a good idea to explicitly copy the image to some Heroic owned location. I imagine in cases like flatpak, Heroic may not have constant access to the location user selected

flavioislima:
> It may be a good idea to explicitly copy the image to some Heroic owned location. I imagine in cases like flatpak, Heroic may not have constant access to the location user selected

I wonder if this could happen if the app had access to pick the file at first. So unless that permission is revoked manually, I don't think that could happen, right?

flavioislima (approved):
Code Looks good, still need testing though.

imLinguin:
> > It may be a good idea to explicitly copy the image to some Heroic owned location. I imagine in cases like flatpak, Heroic may not have constant access to the location user selected
> 
> I wonder if this could happen if the app had access to pick the file at first. So unless that permission is revoked manually, I don't think that could happen, right?

@flavioislima Not really, this can happen in every-day use. If the location is not accessible, the file picker will temporarily grant access to it through the portal, and make the file accessible in runtime location like `$XDG_RUNTIME_DIR/doc/....` 
I think we should address cases like that

public/locales/en/gamepage.json
```diff
@@ -237,7 +238,7 @@
             "useragent": "Custom User Agent"
         },
         "placeholder": {
-            "image": "Paste an Image URL here",
+            "image": "Paste an URL of an Image or select one from your computer",
```
CommandMC, line 241: Whenever you *change* the text for a key, you have to:
1. Remove the key completely
2. Update translation files (`pnpm i18n`, this'll remove the key and its old translations for all languages)
3. Re-add the key & its new translation

As-is, you changed the English text, but other languages will still use their outdated translations

CommandMC (commented):
The feature itself LGTM, just needs a proper translation update

public/locales/en/gamepage.json
sandepten: hey @CommandMC, I've tried this several times but it's not working. I removed the key from the sideload dialog page and ran pnpm i18n, but it only removed it from the English file. Then, I tried removing it from en/gamepage.json and ran pnpm i18n again, but it re-added the key. Am I doing something wrong?

flavioislima: This is fine, it should only change the english one, the others will be updated on weblate later.

CommandMC: Ah I see, no translations were added for the key yet, hence there being nothing to remove
Sorry!

flavioislima:
Tested here and works fine. Thanks for the contribution @sandepten

PR #4471, opened 2025-04-12
[Tech] Update Typescript to 5.8.3

CommandMC:
Also updates related packages, including `@types/node` and our patch for it

No functional changes were made here, so testing should hopefully be fairly easy

---

### DIFF
--- a/src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
+++ b/src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
@@ -17,7 +17,7 @@ import {
   removeSpecialcharacters,
   writeConfig
 } from 'frontend/helpers'
-import React, { useContext, useEffect, useState } from 'react'
+import React, { useCallback, useContext, useEffect, useState } from 'react'
 import { Trans, useTranslation } from 'react-i18next'
 import { AvailablePlatforms } from '..'
 import fallbackImage from 'frontend/assets/heroic_card.jpg'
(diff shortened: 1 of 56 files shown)


CommandMC:
Marking this as a draft until https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4488 is merged

src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
```diff
@@ -208,31 +208,43 @@ export default function SideloadDialog({
...
+        ]
+      case 'osx':
+      case 'Mac':
+        return [
+          { name: 'Apps', extensions: ['App'] },
+          { name: 'Other Binaries', extensions: ['sh', 'py', 'bin'] },
+          { name: 'All', extensions: ['*'] }
+        ]
+      // FIXME: Can these happen?
+      case 'Android':
```
flavioislima, line 235: Maybe just use `default` here

flavioislima (approved):
LGTM, just some small comments

src/common/types/gog.ts
```diff
@@ -140,15 +140,6 @@ interface FeatureSupport {
   }
 }
 
-export type SaveFolderVariable =
```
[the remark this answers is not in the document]
CommandMC: This type isn't necessary, since it's only used to declare a Record type in `backend/save_sync.ts`. Although now that you mention it, might want to add a `satisfies` clause to that to make sure we never accidentally erase one of the variables

src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
CommandMC: I don't like `default` here: In case we ever add a new install platform, I'd rather have this switch statement error out until that platform is explicitly added to the case list than it getting silently ignored

PR #4496, opened 2025-04-27
fix steam remote-play, remote-play together and gamerecording

nodscher:
#4493 closed for some reason and I can't reopen it.

On Linux, Steam expects the `SteamGameId` environment variable to stay unchanged in order for anything using Video-Capture to work, like remote-play, remote-play together and gamerecording. Thus we should leave it alone if it's set. 

Furthermore umu has to do the same: https://github.com/Open-Wine-Components/umu-launcher/pull/447

I have only tested this Change on Linux.


flavioislima (approved):
thanks for fixing this!

We can merge it, you just need to fix the conflicts now. ⚔️

Issue #4492, opened 2025-04-26
[BUG]: Installation settings popup is not scrollable


jiriks74:
### Describe the bug

When installing a game with a large number of DLCs (see Dying Light on GoG as en example) the installation options popup is too big for the window. 

Since the popup is not scrollable the game can't be installed (without shortcuts).

### Add logs

```text
UI Issue
```

### Steps to reproduce

1. Find a game with a lot of DLCs
2. Click install

### Expected behavior

The popup being scrollable 

### Screenshots

<details>
<summary>Dying Light installation popup</summary>

![Image](https://github.com/user-attachments/assets/1693ccde-857a-42f2-91e2-725d86082541)

</details>

### Heroic Version

Latest Stable

### System Information

- OS: NixOS 25.05

### Additional information

(For future reference)
- Heroic version: 2.16.1 Hotfix

CommandMC:
I've seen this issue once on Reddit a few months ago: <https://reddit.com/r/HeroicGamesLauncher/comments/1irbv9b/cant_install_game_because_install_button_appears/mdzkkg0/>
As mentioned in the post, the modal *should* be scrollable, and indeed *is* on my end. Are you using Nix' own Heroic package? If so, could you try reproducing this issue with our official Flatpak, or perhaps the AppImage version?

jiriks74:
Hi,
I'm using the Nix package, I'll try the appimage.

jiriks74:
I see that the popup is scrollable in the AppImage so I'll raise this issue in Nixpkgs.

jiriks74:
The issue is that Electron 31 is EOL and is not available in Nixpkgs. Because of that Heroic is built with Electron 34 in Nix.

Electron versions 33 and 34 have this bug while Electron version 32 is working. But v 32 is EOL as well:

```
       error: Package ‘electron-32.3.3’ in /nix/store/rsnglxqpys0i40ffl76ppbppq3ffyx9h-source/pkgs/development/tools/electron/binary/generic.nix:43 is marked as insecure, refusing to evaluate.


       Known issues:
        - Electron version 32.3.3 is EOL
```

PR #4497, opened 2025-04-29
tech: update electron to 36.0.0

imLinguin:
This PR updates Electron to v36.0.0
and comes with fixes related to that update.

Notably fixed:
- rendering of lists
  - game library in list mode
  - download manager
  - wine manager
- install modal scrolling
- imagecache invalid URL failures

Additionally, removed file drag and drop on `RUN EXE ON PREFIX` button

---

Some before and afters (fixing the issues after the update):

---

---


jiriks74:
I just did a quick test of this PR and everything seems to be working. If you'd like someone to test things feel free to reach out (I'll be using this as a fix for https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4492).

patches/@types__node.patch
CommandMC, line 1: Hm, I'm wondering how this patch being deleted *doesn't* break anything. I'd expect this line to be problematic for example:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/10ac2239b5334bfc67f3e17144a23830b5e364b1/src/backend/constants.ts#L226

The patch restricts `process.platform` to just `'win32' | 'darwin' | 'linux'`, so it not being there would mean `necessaryFoldersByPlatform[process.platform]` looks like it can be `undefined` to TypeScript and/or ESLint. But evidently that error isn't showing up in codecheck

electron.vite.config.ts
```diff
@@ -66,6 +66,13 @@ export default defineConfig(({ mode }) => ({
       minify: true,
       sourcemap: mode === 'development' ? 'inline' : false
     },
+    css: {
+      preprocessorOptions: {
+        scss: {
+          api: 'modern'
+        }
+      }
+    },
```
CommandMC, line 75: Is this required? According to [the docs](<https://vite.dev/config/shared-options#css-preprocessoroptions>), it should default to the `modern` API (we don't have `sass-embedded` installed; at least on `main` we don't

src/backend/images_cache.ts
```diff
@@ -34,8 +34,11 @@ const getImageFromCache = (url: string) => {
     pending.set(
       digest,
       new Promise((res) => {
-        axios
-          .get(realUrl, { responseType: 'stream' })
+        axios({
+          method: 'get',
+          url: realUrl,
+          responseType: 'stream'
+        })
```
CommandMC, line 41: Huh why was *this* necessary? Shouldn't `axios.get` and `axios` with `method: 'get'` do the same thing?

electron.vite.config.ts
imLinguin: it defaults to modern starting vite 6 afaik

src/backend/images_cache.ts
imLinguin: It is the same thing. I was trying different things for an exception that was thrown here. I forgot to revert the change. I'll do just that

patches/@types__node.patch
flavioislima: It shows this to me:
image

flavioislima (changes requested):
We need to have this line on main.ts to avoid breaking the app on Gnome on some distros:
`[app.commandLine?.appendSwitch('--gtk-version', '3')
](https://github.com/electron/electron/issues/46538)`

https://github.com/electron/electron/issues/46538
https://www.electronjs.org/blog/electron-36-0#behavior-changed-gtk-4-is-the-default-when-running-on-gnome

patches/@types__node.patch
CommandMC: Yes, that's precisely the error I would've expected

electron.vite.config.ts
CommandMC: Note: We are on Vite 6 now; I'm fine with keeping this regardless, but might as well remove it if it's no longer required

flavioislima (approved):
LGTM and its working fine now.

PR #4508, opened 2025-05-05
[Feat/Ref] Make our logger promise-based, circular-import-free and auto-rotating

CommandMC:
Originally, I just set out to get rid of the circular dependencies between `logger/logger.ts` and `logger/logfile.ts`. Then this kind of turned into a "I wonder if *thing* would be easy to do". Luckily (or sadly, depending on your interpretation), "thing" *did* turn out easy to do most of the time, and before I knew it, I re-wrote our logger

Apologies for the large commit, but I don't think incrementally making these changes would make sense. You can focus on the changes in `src/backend/logger` and `src/backend/storeManagers`

- The logger is now promise-based
  - logging a message returns a Promise, meaning logging no longer holds up the main thread
  - logging promises pauses the log write queue until that promise is resolved. As such, you'll now see the system information as the first thing in any Heroic log  
    The log queue got simplified in the process; instead of using a home-grown timing approach, we can simply rely on promises & `await`
- Logs are auto-rotating: When the first write to a log file is submitted, the logger checks whether there's an old log file & backs it up (renaming it to `<name>.old`)
- LogWriter code is shared between Heroic, runners, and games: This was somewhat the case before (with one base class and subclasses), but it's fully embraced now. Game logs now use standard `logInfo`/`logWarning`/etc. functions to log their header, the same code writing the runner log also writes the game log
- The folder structure for log files got a lot clearer
  - logs are stored in a more sensible place by default
  - logs are divided into (1) Heroic itself, (2) runners and (3) game logs (each game having its own folder for logs)
  - game logs are further divided into several types (`launch`/`install`/`import`/etc.)  
    Only `launch` logs are accessible in the UI at the moment, but the other types are right next to the launch log and simply titled `install.log`/similar

---


CommandMC:
Huh, this is a new one. I hadn't included the migration system commit in this PR, yet the checks have it (& fail because of it)
Wonder how that happened

PR #4534, opened 2025-05-18
[Fix] Protocol handling on Windows

CommandMC:
When creating a shortcut, Windows automatically adds a trailing `/` to the path. This caused Heroic to assume its parsing the old-style protocol URL, while it was actually getting a new-style one. Because of this, shortcuts created on Windows since Heroic 2.16 wouldn't actually work

Note that this is a change on the handling side, meaning previously-non-working shortcuts will now just work again

---


flavioislima (approved):
lgtm

PR #4538, opened 2025-05-19
[UI - Linux] Update FPS limit Text 

flavioislima:
Removes the info about FPS limit only working on DXVK games since now it works for VKD3D as well

---

### DIFF
--- a/src/frontend/screens/Settings/components/EnableDXVKFpsLimit.tsx
+++ b/src/frontend/screens/Settings/components/EnableDXVKFpsLimit.tsx
@@ -36,13 +36,16 @@ const EnableDXVKFpsLimit = () => {
           htmlId="enableDXVKFpsLimit"
           value={enableDXVKFpsLimit || false}
           handleChange={() => setDXVKFpsLimit(!enableDXVKFpsLimit)}
-          title={t('setting.dxvkfpslimit', 'Limit FPS (DX9, 10 and 11)')}
+          title={t('setting.fpslimit', 'Limit FPS')}
         />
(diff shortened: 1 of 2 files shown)


src/frontend/screens/Settings/components/EnableDXVKFpsLimit.tsx
```diff
@@ -36,13 +36,16 @@ const EnableDXVKFpsLimit = () => {
...
+          title={t('setting.fpslimit', 'Limit FPS')}
         />
 
         <FontAwesomeIcon
           className="helpIcon"
           icon={faCircleInfo}
-          title={t('help.dxvkfpslimit', 'Sets a frame rate cap for DXVK games')}
+          title={t(
+            'help.fpslimit',
+            'Sets a frame rate cap for DXVK/VKD3D games'
```
CommandMC, line 47: I realize you're just adapting the original wording, but I'd change this to say "DirectX games" (or "DX9-12 games" if you want to be really accurate)

Etaash-mathamsetty:
It should be DX8 to DX12 btw, and if we edit that we should change the DXVK text to go from dx8 to dx12 as well

biliesilva (approved):

flavioislima:
> It should be DX8 to DX12 btw, and if we edit that we should change the DXVK text to go from dx8 to dx12 as well

doesn't it work for DX 1-7?

It is funny because maybe the generic version is better because one day we might have DX13 and it works OOTB and then we would need a PR for that alone.

PR #4540, opened 2025-05-20
[Feat] Improve launch options

flavioislima:
---


arielj:
I I'm worried users will be confused that the launch option is persisted when they see the select in the page, but it's not when they open the game for example from steam (which is the main use I can imagine for persisting the launch option).

I think it will add a new set of reports because of that confusion.

I did something like that in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4220 for example to change the `verbose logs` setting from the game details page, so a similar approach could be used to change the launch option setting, then it can be easily used when launching a game with the protocol

flavioislima:
closing this one in favor of #4579

PR #4546, opened 2025-05-22
[Feat] Add icons to a few select menus

CommandMC:
Now that our `SelectField`s are HTML elements, we can decorate its items. I've used that opportunity to add icons to
- the 3-dot menu on the game page
- the Wine Version selector
- the game platform selector
- context menu entries (game card & settings)

TODO:
- [X] Remove "Wine - " / "Proton - " from the start of wine versions since they're now redundant
- [X] Add more icons for special wine versions (perhaps one with a little arrow for the `-latest` variants, or a GE icon for GE-Proton?)

| Before  | After |
| ------------- | ------------- |
| image  | image  |
| image | image |
| image | image |
| image | image
| image | image

---


roland-5:
Would it be a problem to add different icon for Proton from Steam and ge-proton from UMU?

CommandMC:
> Would it be a problem to add different icon for Proton from Steam and ge-proton from UMU?

That's planned, yes

src/frontend/screens/Library/components/InstallModal/index.tsx
```diff
@@ -145,7 +146,12 @@ export default React.memo(function InstallModal({
       >
         {availablePlatforms.map((p, i) => (
           <MenuItem value={p.value} key={i}>
-            {p.name}
+            <Box sx={{ display: 'flex', placeItems: 'center' }}>
+              <SvgIcon sx={{ mr: 1 }}>
```
arielj: I wonder if there's a way to set `margin-inline-end` instead of `margin-right` with these MUI syntax for styling, so it works for RTL languages

because we need the margin on the left for those

there's too much space on the right and no space between the text and the icon in RTL language
image

CommandMC: MUI's `sx` prop is essentially a better `style` prop, so we can set any CSS property there

flavioislima:
I am not sure about these update icons on top of the other icon. It looks out of place and goes above the input itself. Maybe its better to just not add it:
Screenshot 2025-07-16 at 14 32 11

Other thing that could be better is that to have a fixed width div to have the same spacing between the icons and the text. The way it is now some icons are closer to the text than others, for instance, the Steam and Browse Files in comparison with the others:
Screenshot 2025-07-16 at 14 33 09

And last but not least we could have icons on the GameCard Submenu to keep the consistency:
Screenshot 2025-07-16 at 14 34 49

CommandMC:
> I am not sure about these update icons on top of the other icon. It looks out of place and goes above the input itself. Maybe its better to just not add it:

Hm, honestly I like that it stands out like that. To clarify, the icon is added to any auto-updating Wine version (anything ending in `-latest`), to make it clear that they're special and you may (or may not) want to prefer them

> Other thing that could be better is that to have a fixed width div to have the same spacing between the icons and the text. The way it is now some icons are closer to the text than others, for instance, the Steam and Browse Files in comparison with the others:

The icons all have the same size (24x20). They might look closer to / farther away from the text, but that's caused by the icon shape instead of its actual size

> And last but not least we could have icons on the GameCard Submenu to keep the consistency:

Right, missed that. Will do

PR #4549, opened 2025-05-22
[Release] 2.17.0 Franky

flavioislima:
- Update release version name
- fixes regression for games with launch options
- Fixes regresion on game log screen
- fixes ui of update game dialog
- fixes launchoptions component position on gamepage
- introduces a new eslint rule to avoid adding conditional hooks.

---


src/frontend/screens/Game/GamePage/components/LaunchOptions.tsx
```diff
@@ -54,7 +54,6 @@ const LaunchOptions = ({ gameInfo, setLaunchArguments }: Props) => {
         }
       }}
       value={selectedLaunchOptionIndex.toString()}
-      prompt={t('launch.options', 'Launch Options...')}
```
flavioislima, line 57: Removed this to avoid showing an empty selection. It will show directly the first option now.

eslint.config.mjs
```diff
@@ -14,6 +15,9 @@ export default tseslint.config(
   {
     files: ['**/*.ts', '**/*.tsx'],
     rules: {
+      // React hooks rules
+      'react-hooks/rules-of-hooks': 'warn',
```
flavioislima, line 19: Should be error but then we need to fix all errors related with hooks first and this should be its own pr

PR #4579, opened 2025-05-28
[Feat] Add Launch options setting and show dialog when first launch

flavioislima:
Different approach than #4540 now using settings instead.
Added a bit more complexity to the launch command since we now need to check for launch options there as well to show the dialog the first time.

Dialog only shows the first time, so no need to "dont show this again" checkbox:
image

On game settings can be changed anytime:
image

---


src/frontend/screens/Settings/components/LaunchOptionSelector.tsx
```diff
@@ -0,0 +1,137 @@
+import React, { useCallback, useContext, useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { SelectField } from 'frontend/components/UI'
+import { MenuItem, SelectChangeEvent } from '@mui/material'
+import SettingsContext from '../SettingsContext'
+import { LaunchOption } from 'common/types'
+import useSetting from 'frontend/hooks/useSetting'
+
+const LaunchOptionSelector = () => {
```
arielj, line 9: I wonder if there's a way to re-use this component in the game page?

this LaunchOptionSelector and the one in GamePage/components/LaunchOptions look really similar except that one is not using the useSetting hook?

arielj: one thing I did in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4220/files#diff-98f5370ea40ad29c287d5e98482338cb852eafbdd3efd99927cb45d379bb8f14R378 was to wrap the game page in a SettingsContext so then you can use useSetting

src/backend/game_config.ts
```diff
@@ -308,6 +308,11 @@ class GameConfigV0 extends GameConfig {
   }
 
   public setSetting(key: keyof GameSettings, value: unknown) {
+    // check if the key exists, if not, initialize it with a undefined value
+    if (!(key in this.config)) {
+      this.config[key] = undefined as never
+    }
```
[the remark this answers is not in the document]
flavioislima: Yeah, for some reason it was not creating the new key so was returning a JS error saying that it cannot update a key that doesn't exists.

src/frontend/screens/Settings/components/LaunchOptionSelector.tsx
flavioislima: Yeah, I thought about doing that, would make our lives a lot easier if we could get the settings on the game page.

src/frontend/screens/Game/GamePage/components/LaunchOptions.tsx
```diff
@@ -2,34 +2,119 @@ import React, { useCallback, useContext, useEffect, useState } from 'react'
...
+
+    void fetchLaunchOptions()
+  }, [gameInfo, appName, runner, gameSettings])
 
   const labelForLaunchOption = useCallback((option: LaunchOption) => {
     switch (option.type) {
       case undefined:
       case 'basic':
-        return option.name
+        return 'name' in option ? option.name : 'Launch Option'
```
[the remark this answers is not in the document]
flavioislima: This was to fix a TS error, not sure why the other ones are not complaining.

arielj:
Following up with the comments on https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4591

I'm testing this change with Ghostrunner and we have some issues (the second issue is unrelated to this game though):
- because the game only reports 1 launch option, the launch option selector in the settings is not displayed (it has the `launchOptions.length <= 1` check), but then we wouldn't have a way to pick between using the DX12 flag or not using it

- there seems to be an infinite loop in this component, the terminal gets flooded with this when opening the Advanced tab:
```
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
.....
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
(18:56:56) DEBUG:   [Legendary]:        Using cached install info
```

And removing the component from the Advanced tab removed the loop, I'm not sure why the infinite loop is happening, but it's calling the `useEffect` non-stop.

Maybe we should have the useEffect to fetch the launchOptions and a different useEffect to process the last selected index only when launchOptions changes, that might fix the infinite loop.

src/frontend/screens/Settings/components/LaunchOptionSelector.tsx
flavioislima: Ok, refactored the components and now we have a hook to deal with the launch options.
Also added a 10s timer to autoselect so people wont be stuck on steamdeck, for instance.

PR #4588, opened 2025-05-31
[UX] Show Flatpak's runtime version when missing mangohud/gamescope

arielj:
When using Heroic as a Flatpak and mangohud or gamescope are not found, the message only told the user to install them and add them to the PATH, when the solution is to actually install the flatpak packages of those tools (because users might see they actually have the tools installed and in the PATH, but they might not know they need the flatpak version)

Since they also have to install the correct version matching the Flatpak runtime, I included that in the message to make it more obvious.

This should solve many issues of users not knowing how to solve the missing mangohud/gamescope problems.

Also, since `isFlatpak` is a static value that can never change, we don't need an api endpoint, we can just inject the value into the window object at boot using the preload script (we are already sending some static data there), this removes a little bit of complexity in a few places, which I included in this PR.

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -672,7 +672,8 @@
             "enableForceGrabCursor": "Enable Force Grab Cursor",
             "enableLimiter": "Enable FPS Limiter",
             "enableUpscaling": "Enables Upscaling",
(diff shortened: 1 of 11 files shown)


flavioislima (approved):
lgtm

Issue #4583, opened 2025-05-29
GamePage launch options dropdown doesn't show all options


warriormaster12:
### Describe the bug

Tested with Ghostrunner 1. The game can be launched in DX11 or DX12 mode but the dropdown menu shows DX12 option only.

### Add logs

```text
-
```

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Expected behavior

The dropdown menu should have an option for both dx11 and dx12

### Screenshots

![Image](https://github.com/user-attachments/assets/a8ee582b-a873-47b8-a20a-734c3b656f3d)

### Heroic Version

Latest Stable (Flatpak)

### System Information

Operating System: Fedora Linux 42
KDE Plasma Version: 6.3.5
KDE Frameworks Version: 6.14.0
Qt Version: 6.9.0
Kernel Version: 6.14.8-300.fc42.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 16 × AMD Ryzen 7 7735HS with Radeon Graphics
Memory: 30.6 GiB of RAM
Graphics Processor 1: AMD Radeon 680M
Graphics Processor 2: AMD Radeon RX 7600S
Manufacturer: ASUSTeK COMPUTER INC.
Product Name: ASUS TUF Gaming A16 FA617NS_FA617NS
System Version: 1.0

Issue #3751, opened 2024-05-14
Ghostrunner wrong launch options


nonoche2:
### Describe the bug

- Ghostrunner only has one launch option (DirectX 12) in the local menu which can't be unselected
- when launched with CrossOver and D3DMetal (which supports DirectX 12), an error alert says my setup doesn't support DirectX 12 games and the game quits
- the game runs fine with the Steam version and CrossOver

### Add logs

```text
n/a
```

### Steps to reproduce

1) launch Ghostrunner

### Expected behavior

the game should work like the Steam version does

### Heroic Version

Latest Stable

### System Information

- macOS Sonoma 14.4.1

imLinguin:
Could you provide more context like game log or details as to where the game was bought

nonoche2:
my apologies, the game was baught on the Epic Games Store.

When I install it without touching the launch options menu, the game runs fine. But if I click on the launch options menu, only "Ghostrunner DirectX 12" is available (Steam version offers an additional DirectX 11 launch option). If I select that, then the game fails to launch from then on, and I can't unselect the launch option (I have to uninstall the game and reinstall it for it to work again).

If I run it after selecting the Direct X 12 launch option, I get an error alert saying "DX12 is not supported on your system, try running without the -dx12 or -d3d12 command line argument" (which is odd since D3DMetal does support DirectX 12)

Here's Heroic's log :

Launching "Ghostrunner" (legendary)
Native? No
Installed in: /Volumes/Transcend/Heroic/Ghostrunner

System Info:
CPU: 8x Apple M1
Memory: 17.18 GB (used: 4.37 GB)
GPUs:

OS:  14.4.1 (darwin)

The current system is not a Steam Deck
We are not running inside a Flatpak container

Software Versions:
  Heroic: 2.14.1 Flame Emperor Sabo
  Legendary: 0.20.34 Direct Intervention
  gogdl: 1.0.1
  Nile: 1.0.2 Jonathan Joestar

Game Settings: {
	"autoInstallDxvk": true,
	"autoInstallDxvkNvapi": false,
	"autoInstallVkd3d": true,
	"preferSystemLibs": false,
	"enableEsync": true,
	"enableMsync": true,
	"enableFsync": true,
	"launcherArgs": "",
	"nvidiaPrime": false,
	"enviromentOptions": [],
	"wrapperOptions": [],
	"showFps": false,
	"useGameMode": false,
	"battlEyeRuntime": false,
[...]

PR #4591, opened 2025-06-01
[Fix] Add back option to unselect `Launch option`

arielj:
The empty prompt select element got lost in the last version. This PR adds that back so users can unselect any `Launch option`.

Some games like Ghostrunner only show the Launch Option to run with DX12 but no option to NOT launch with DX12, so we need a generic/empty option to be able to run the game without an extra game argument

This closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4583 and closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3751

---

### DIFF
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -210,6 +210,9 @@
         },
         "winetricks": "Installing Winetricks Packages"
     },
+    "launch": {
+        "options": "Launch Options..."
+    },
     "modifyInstall": {
         "dlcsCollapsible": "DLC",
         "nodlcs": "No DLC available",
--- a/src/frontend/screens/Game/GamePage/components/LaunchOptions.tsx
+++ b/src/frontend/screens/Game/GamePage/components/LaunchOptions.tsx
@@ -3,6 +3,7 @@ import GameContext from '../../GameContext'
 import { GameInfo, LaunchOption } from 'common/types'
 import { SelectField } from 'frontend/components/UI'
 import { MenuItem } from '@mui/material'
+import { useTranslation } from 'react-i18next'
 
 interface Props {
   gameInfo: GameInfo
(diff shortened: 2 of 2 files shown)


flavioislima:
@imLinguin  @arielj does it make sense to have a empty option though?
I removed the empty one and I am showing only the actual options on my PR here: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4579

Unless is not possible to launch without an option, I think it is counter intuitive to not select anything. So setting the initial value from -1 to 0 fixes this and shows only the actual options, no?

arielj:
> @imLinguin @arielj does it make sense to have a empty option though? I removed the empty one and I am showing only the actual options on my PR here: #4579
> 
> Unless is not possible to launch without an option, I think it is counter intuitive to not select anything. So setting the initial value from -1 to 0 fixes this and shows only the actual options, no?

The problem is that we need a way to launch with no arguments, in the reported issue for example there's the case of Ghostrunner: no option => runs DX11, the only provided launch option => runs DX12. If we don't provide an empty option then the users have no way to run the DX11 version cause they would only have the DX12 option and always selected.

Maybe the content of the option should be something different? I don't know what would be the correct text. This PR was just bringing back the old behavior of having a way to unselect a launch option that got lost in the last release.

This might not be needed for all games when the metadata provides launch options for all cases, but Ghostrunner only provides the option to launch with DX12 and no option to launch with DX11.

flavioislima:
> > @imLinguin @arielj does it make sense to have a empty option though? I removed the empty one and I am showing only the actual options on my PR here: #4579
> > Unless is not possible to launch without an option, I think it is counter intuitive to not select anything. So setting the initial value from -1 to 0 fixes this and shows only the actual options, no?
> 
> The problem is that we need a way to launch with no arguments, in the reported issue for example there's the case of Ghostrunner: no option => runs DX11, the only provided launch option => runs DX12. If we don't provide an empty option then the users have no way to run the DX11 version cause they would only have the DX12 option and always selected.
> 
> Maybe the content of the option should be something different? I don't know what would be the correct text. This PR was just bringing back the old behavior of having a way to unselect a launch option that got lost in the last release.
> 
> This might not be needed for all games when the metadata provides launch options for all cases, but Ghostrunner only provides the option to launch with DX12 and no option to launch with DX11.

ok, sounds good. I think my PR fixes this issue since it shows the dialog to select the option and store it in a setting. But I am not sure I will include it on the next hotfix.

arielj:
> ok, sounds good. I think my PR fixes this issue since it shows the dialog to select the option and store it in a setting. But I am not sure I will include it on the next hotfix.

but even for you PR we would need a way to select no option, right? I'll try that PR later with ghostrunner to see how it works, but the issue is the game has no option to set dx11, so we need to provide a `no game arguments` option just in case

Issue #3861, opened 2024-07-14
Hide/Disable NVAPI option when we know it's a steam deck


arielj:
### Problem description

NVAPI makes no sense on a steam deck but some users end up enabling it by mistake.

### Feature description

We know steamdeck is just AMD so we can hide/disable that option to avoid problems.

### Alternatives

We could also show a warning when no nvidia GPU is detected and a user tries to enable that option? so they can still use that if for some reason we fail to identify an nvidia GPU but it's there.

PR #4594, opened 2025-06-01
[UX] Hide DXVK-NVAPI option on the SteamDeck

arielj:
The `DXVK-NVAPI` option makes no sense on a SteamDeck since it's always an AMD GPU. We can hide it to remove noise from the settings.

Not sure what's the best way to detect if the devise is a steamdeck (I just checked a few env variables looking at my steamdeck's `printenv` output). We have the `getSteamDeckInfo` function that checks CPU/GPU information against some predefined values but I can't really use it cause it's an async function and I need the code to be sync to use it in the preload/index.ts file.

Closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3861

---

### DIFF
--- a/src/backend/constants/environment.ts
+++ b/src/backend/constants/environment.ts
@@ -7,6 +7,11 @@ export const isWindows = process.platform === 'win32'
 export const isLinux = process.platform === 'linux'
 export const isSteamDeckGameMode =
   process.env.XDG_CURRENT_DESKTOP === 'gamescope'
+export const isSteamDeckDesktopMode =
+  env.SESSION_MANAGER?.includes('unix/steamdeck') &&
+  env.HOME === '/home/deck' &&
+  env.DESKTOP_SESSION?.includes('steamos')
+export const isSteamDeck = isSteamDeckGameMode || isSteamDeckDesktopMode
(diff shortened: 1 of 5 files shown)


CommandMC:
> We have the `getSteamDeckInfo` function that checks CPU/GPU information against some predefined values but I can't really use it cause it's an async function and I need the code to be sync to use it in the preload/index.ts file

You can use `window.api.systemInfo.get()` with the `useAwaited` hook to get the system info in the component itself, it then has a `steamDeckInfo.isDeck` property

arielj:
> > We have the `getSteamDeckInfo` function that checks CPU/GPU information against some predefined values but I can't really use it cause it's an async function and I need the code to be sync to use it in the preload/index.ts file
> 
> You can use `window.api.systemInfo.get()` with the `useAwaited` hook to get the system info in the component itself, it then has a `steamDeckInfo.isDeck` property

yeah, I was trying to avoid doing an async call to the backend if we can just have that info once when the app boots, since it's static information that doesn't really change we can avoid the complexity of async stuff and hooks and react in the middle if it's a `window` property, that's why I was trying to avoid an async way of detecting the steamdeck

flavioislima (approved):
code looks good

PR #4623, opened 2025-06-06
[FIX] Add warning dialog for disabled Shader Pre-caching on SteamDeck

flavioislima:
image

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -224,9 +224,18 @@ const launchEventCallback: (args: LaunchParams) => StatusPromise = async ({
           runner,
           status: 'done'
         })
+        logInfo(
+          `User aborted the launch of ${title} because Shader Pre-Caching is disabled`,
+          LogPrefix.Backend
+        )
         stopLogger(appName)
         return { status: 'abort' }
       }
+
+      logWarning(
+        `User launched ${title} with Shader Pre-Caching disabled. Issues might occur.`,
+        LogPrefix.Backend
+      )
     }
   }
 


arielj (approved):

src/backend/logger/logger.ts
```diff
@@ -514,12 +516,31 @@ const shouldToggleShaderPreCacheOn = async (
...
+      ),
+      message: i18next.t(
+        'box.shaderPreCachingDisabledMessage',
+        "Steam's Shader Pre-cache is disabled. Please enable it on the Steam Settings in Desktop Mode to ensure the game works properly with UMU."
+      ),
+      buttons: [i18next.t('box.launchAnyway', 'Launch Game Anyway')]
+    })
+
+    return true
+  }
```
CommandMC: I'd move this to somewhere in `launchEventCallback`. This function's name suggests that it only *checks* whether the option is toggled, actually doing something based on that status should be left to somewhere else

CommandMC (commented):
Have you checked whether this actually stops the game launch? `shouldToggleShaderPreCacheOn` is called by the GameLogWriter's `initLog` method:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/f47d5761d3de02adf993fb458aad1af8e0e55746/src/backend/logger/logger.ts#L724

which is called in the launchEventCallback here:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/f47d5761d3de02adf993fb458aad1af8e0e55746/src/backend/launcher.ts#L169

Note that it's *not* awaited, so I'd expect `launchEventCallback` to not wait for the user to confirm this dialog

flavioislima:
@CommandMC I can confirm that this stops the game to be launched until the user clicks the OK button.
But I will check the other place you mentioned.

src/backend/logger/logger.ts
flavioislima: Moved there and it works better. Also added a Cancel button.

PR #4631, opened 2025-06-07
[UX] Support Env variables on windows and rockstar tweak

arielj:
I tried this in a VM and it seems to work kind of, the game fails to open but I think it's because I'm inside a VM, but it seems to be doing the correct workflow (not showing the `buy now` screen of the rockstar launcher)

I included a fix for the fake exe integration and also enabled env variables to be configurable and applied in windows.

Please, if anyone has a windows machine not in a VM, can you try this with either GTAV, GTAV enhanced or RDR2?

Steps:
- go to the game's settings > advanced tab
- set the env variable `USE_FAKE_EPIC_EXE = true`
- launch the game

---

### DIFF
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -871,14 +871,12 @@ export async function launch(
   let commandEnv = {
     ...process.env,
     ...setupWrapperEnvVars({ appName, appRunner: 'legendary' }),
-    ...(isWindows
-      ? {}
-      : setupEnvVars(gameSettings, gameInfo.install.install_path)),
+    ...setupEnvVars(gameSettings, gameInfo.install.install_path),
     ...getKnownFixesEnvVariables(appName, 'legendary')
   }
 
   // We can get this env variable either from the game's settings or from known fixes
-  if (commandEnv['USE_FAKE_EPIC_EXE'] && !isWindows) {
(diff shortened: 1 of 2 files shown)


src/backend/storeManagers/legendary/games.ts
```diff
@@ -871,14 +871,12 @@ export async function launch(
...
     ...setupWrapperEnvVars({ appName, appRunner: 'legendary' }),
-    ...(isWindows
-      ? {}
-      : setupEnvVars(gameSettings, gameInfo.install.install_path)),
+    ...setupEnvVars(gameSettings, gameInfo.install.install_path),
     ...getKnownFixesEnvVariables(appName, 'legendary')
   }
 
   // We can get this env variable either from the game's settings or from known fixes
-  if (commandEnv['USE_FAKE_EPIC_EXE'] && !isWindows) {
```
arielj, line 881: no wonder why it was not being applied on windows xD

flavioislima: Yeah, I saw that when I did the review of the previous PR and was thinking that it would not work but then I looked again and thought that it would pass since it was checking for 2 different things so to be true it needed for both not be windows and also check fo the env var. But in fact that commandEnv is doing another thing.
anyway, should work now 😅

flavioislima (approved):
lgtm

Issue #4639, opened 2025-06-08
Allow the user to filter game entries using the first letter in the title


Schweppesale:
### Problem description

It can become difficult for the user to find a game once their library exceeds a certain size without using either the search feature or category filters which as of today still need to be manually setup by the user.

### Feature description

The idea is to populate a sidebar with # and letters A-Z.
If a user selects "E" then all games which begin with the letter E will be filtered.
If they click on the letter "E" a second time then the filter is removed.

If the user selects # then all titles which begin with ^[0-9] will be filtered, etc.

### Additional information

Jellyfin has a similar feature to the one I'm describing:
https://us1.discourse-cdn.com/firecore/optimized/3X/b/a/ba2597addfb97e9711184ce22dafa863ca4d5157_2_1380x814.jpeg

arielj:
implemented in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4643

PR #4643, opened 2025-06-09
alphabet filter

theguy000:
[issue](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4639)

Adding a video here to better overview:

[Gravação de tela de 2025-06-09 14-43-24.webm](https://github.com/user-attachments/assets/65df2a0f-1722-4351-9167-7e9c8594b972)

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -417,7 +417,6 @@
         "releaseDate": "Release Date"
     },
     "library": {
-        "refresh": "Refresh",
         "hideAlphabetFilter": "Hide Alphabet Filter",
         "showAlphabetFilter": "Show Alphabet Filter",
         "sortAscending": "Sort Ascending",


flavioislima (approved):
Code looks good.
Tested here and it works pretty well, respecting all the filters as well. 
Good stuff

flavioislima:
@theguy000 

ACtually found only one small issue on small screens:
image


Maybe make it wrap so it can break into 2 lines.

arielj:
One comment after a quick test, games like `[REDACTED]` or `>observer_` won't match any of the letters

maybe the title should be cleaned before comparing with the selected letter removing anything that is not `A-Z` or a digit, I would expect `[REDACTED]` to be in the R and `>observer_` to be in the O

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,97 @@
...
+  onFilterChange: (filter: string | null) => void
+  allGames: GameInfo[] // 1. Add allGames prop
+}
+
+const AlphabetFilter: React.FC<AlphabetFilterProps> = ({
+  currentFilter,
+  onFilterChange,
+  allGames
+}) => {
+  const alphabet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('')
```
arielj: maybe move this out as a constant? there's really no need to do this split in every render

src/frontend/screens/Library/index.tsx
```diff
@@ -407,9 +421,135 @@ export default React.memo(function Library(): JSX.Element {
   }
 
   // select library
-  const libraryToShow = useMemo(() => {
+  const gamesForAlphabetFilter = useMemo(() => {
```
arielj, line 423: this whole function seems to be a complete copy of the `libraryToShow` function which will be a pain in the ass to maintain all this duplicated code, please find a way to re-use the old function instead of copying over all its content

arielj (changes requested):

arielj:
there seems to be on indicator of the currently selected letter, here I'm showing letter W and there's no difference between the W and the other letters
image

arielj:
something is off with the games used for the letters enable/disable and the actual games in the results when using the search

here you can see that if I search `2023`, I should have the `W` enabled because I have that game, but the letters are using the results of some previous search and not the current search, cause it shows `M` for metro exodus 2033 (which doesn't match the `2023` I typed, but probably matched the previous `2` character, or related to the fuzzy search feature?)

image

EDIT: I think this one can be ignored, I don't know if there's a way to fix this since the search uses the whole library and the list only uses what's being filtered/categorized/etc, and there's also some fuzzy search going on

arielj:
if you are filtering by a letter (let's say X for example) and you make a search that does not return any result starting with that letter, that letter can't be clicked to unselect it because it's disabled, so you have to click another letter and unclick that again to reset the letter filter if you want to see all results

arielj:
maybe we can make this visible with a toggle? like adding a button here to show/hide the letters

image

I can imagine people wanting to keep the current behavior and save that vertical space (not just for the space, but also to remove content for spatial navigation with a controller if you don't use it)

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,97 @@
+import React, { useMemo } from 'react'
+import './index.css' // Import the CSS file
+import { GameInfo } from 'common/types' // 1. Import GameInfo
```
arielj: there's no need for these extra comments, it adds noise saying the same thing the code does

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <button
+            key={symbol}
+            onClick={() => handleClick(symbol, isEnabled)}
+            className={getButtonClassName(symbol, isEnabled)}
+            disabled={!isEnabled} // Add HTML disabled attribute for accessibility
+          >
+            {symbol}
+          </button>
+        )
+      })}
```
arielj: this can be a single `map` loop if you concatenate `alphabet` and `symbols` in a single array, it's the same content basically

src/frontend/screens/Library/components/AlphabetFilter/index.css
```diff
@@ -0,0 +1,98 @@
+/*
+  Modern Alphabet Filter
+  - Uses CSS variables for easy theming.
+  - Aligned with a dark UI.
+  - Adds subtle animations and accessibility features.
+*/
+
+:root {
+  /* --- THEME COLORS (Adjust these to perfectly match your app) --- */
```
arielj: this comment looks off

arielj:
contrast of letters looks off when using the Nord Light theme (enabled letters look disabled, disabled letters look enabled, and hovering turns text white and can't be read)

Here you can see the Z is actually disabled cause I have no game starting with Z, and the L is where my mouse pointer is an you can barely see it
image

src/frontend/screens/Library/components/AlphabetFilter/index.css
```diff
@@ -0,0 +1,98 @@
...
+  transform: translateY(-2px); /* Subtle lift effect */
+}
+
+/* Accessibility: Show a clear outline only for keyboard navigation */
+.alphabet-filter-button:focus-visible {
+  box-shadow: 0 0 0 3px var(--filter-accent-color);
+}
+
+/* Active "Pill" Style */
+.alphabet-filter-button--active {
```
arielj, line 65: this class is incorrect, the code adds `.alphabet-filter-button-active`, not `.alphabet-filter-button--active`

arielj: actually the other part of the code is wrong setting `-active` instead of `--active` cause `--active` is used in more places

src/frontend/screens/Library/components/AlphabetFilter/index.css
```diff
@@ -0,0 +1,98 @@
+/*
+  Modern Alphabet Filter
+  - Uses CSS variables for easy theming.
+  - Aligned with a dark UI.
+  - Adds subtle animations and accessibility features.
+*/
+
+:root {
+  /* --- THEME COLORS (Adjust these to perfectly match your app) --- */
+  --filter-accent-color: #00aeef; /* A vibrant teal/cyan from your 'ADD GAME' button */
```
arielj: we probably want to include some colors for the different themes in this PR, cause now all the themes would use  teal/cyan color when maybe that color is not the accent color of the theme

src/frontend/screens/Library/index.tsx
```diff
@@ -211,6 +217,14 @@ export default React.memo(function Library(): JSX.Element {
     setSortInstalled(value)
   }
 
+  // 3. Create Handler Function for AlphabetFilter
```
arielj: please remove these comments that are saying the same thing as the code

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,97 @@
...
+    let className = 'alphabet-filter-button'
+    if (!isEnabled) {
+      className += ' alphabet-filter-button--disabled'
+    } else if (value === currentFilter) {
+      // Active class should only apply if enabled, but active implies enabled.
+      className += ' alphabet-filter-button-active'
+    }
+    // Specific styling for '#' can remain if needed, independent of active/disabled state for base style
+    if (value === '#') {
+      className += ' alphabet-filter-symbol'
```
arielj: do we need a different style for the symbol character? I think it's fine to leave it the same as the rest

src/frontend/components/UI/ActionIcons/index.tsx
```diff
@@ -21,10 +23,14 @@ import { LIBRARY_TOUR_ID } from 'frontend/screens/Library/components/LibraryTour
...
   'data-tour'?: string
+  showAlphabetFilter?: boolean
+  onToggleAlphabetFilter?: () => void
 }
 
 export default React.memo(function ActionIcons({
-  'data-tour': dataTour
+  'data-tour': dataTour,
+  showAlphabetFilter,
+  onToggleAlphabetFilter
```
arielj: the other actions are passed using the library context, not with props, let's keep the same pattern of using the library context

theguy000:
hey @arielj thank you for your response! I have added the changes you requested. please let me know if everything works. 

also i have a question. i am a noob using git. can you please tell me why test:e2e is failing?

src/frontend/components/UI/ThemeSelector/index.tsx
```diff
@@ -24,6 +24,24 @@ export const defaultThemes: Record<string, string> = {
   sweet: 'Sweet'
 }
 
+// Tag system for dark/light theme
+export const themeType: Record<string, 'dark' | 'light'> = {
```
arielj: I think it's kinda overkill to add this, I don't think it's really needed since we can just define css variables for each theme as needed already, I don't think we need the `data-theme-light` attribute added to the body, we already handle differences for theme throughout the app without this

src/frontend/components/UI/ThemeSelector/index.tsx
```diff
@@ -53,18 +71,22 @@ export const ThemeSelector = () => {
...
-      setAppConfig(config)
-      setThemesPath(config.customThemesPath || '')
-      loadThemes()
+      try {
+        const config = await window.api.requestAppSettings()
+        setAppConfig(config)
+        setThemesPath(config.customThemesPath || '')
+        await loadThemes()
+      } catch (error) {
+        console.error('Failed to load theme settings:', error)
```
arielj: this seems unrelated to the changes of the PR itself

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,96 @@
...
+      }
+    })
+    return chars
+  }, [allGames])
+
+  const getButtonClassName = (value: string, isEnabled: boolean) => {
+    let className = 'alphabet-filter-button'
+    if (!isEnabled) {
+      className += ' alphabet-filter-button--disabled'
+      if (isLightTheme) className += ' alphabet-filter-button--disabled-light'
```
arielj: let's not do that, the same `alphabet-filter-button--disabled` can be used for a CSS selector in any of the themes to set a color, we don't need classes for every variant

src/frontend/themes.scss
```diff
@@ -91,6 +91,10 @@ body.midnightMirage {
   /* Others for compatibility, probably can be refactored */
   --primary: var(--brand-primary);
   --secondary: var(--brand-secondary);
+
+  /* AlphabetFilter */
+  --filter-accent-color: var(--accent);
+  --filter-accent-hover: var(--accent-overlay, var(--accent));
```
arielj: I don't think we need this generic css variables in every theme, they are all the same, can't they be defined once in the root?

src/frontend/screens/Library/components/LibraryHeader/index.tsx
```diff
@@ -6,12 +6,19 @@ import LibraryContext from '../../LibraryContext'
 import './index.css'
 import AddGameButton from '../AddGameButton'
 import SteamInstallButton from '../SteamInstall/SteamInstallButton'
+// FontAwesomeIcon, faFilter, and faFilterCircleXmark are no longer needed here
```
arielj: this comment makes no sense, this PR is not removing any font awesome code from this file

src/frontend/screens/Library/components/LibraryHeader/index.tsx
```diff
@@ -6,12 +6,19 @@ import LibraryContext from '../../LibraryContext'
...
   list: GameInfo[]
+  showAlphabetFilter: boolean
+  onToggleAlphabetFilter: () => void
 }
 
-export default React.memo(function LibraryHeader({ list }: Props) {
+export default React.memo(function LibraryHeader({
+  list,
+  showAlphabetFilter,
+  onToggleAlphabetFilter
```
arielj: similar to another comment, use the LibraryContext, don't do prop-drilling

src/frontend/screens/Library/index.tsx
```diff
@@ -374,199 +570,118 @@ export default React.memo(function Library(): JSX.Element {
...
+    showNonAvailable,
+    showInstalledOnly,
+    alphabetFilterLetter,
+    sortDescending,
+    sortInstalled,
+    favouritesIds,
+    installing
   ])
 
-  // we need this to do proper `position: sticky` of the Add Game area
```
arielj, line 560: do not remove comments you didn't add

src/frontend/screens/Library/index.tsx
```diff
@@ -34,9 +34,253 @@ import { hasHelp } from 'frontend/hooks/hasHelp'
 import EmptyLibraryMessage from './components/EmptyLibrary'
 import CategoriesManager from './components/CategoriesManager'
 import LibraryTour from './components/LibraryTour'
+import AlphabetFilter from './components/AlphabetFilter/AlphabetFilter'
 
 const storage = window.localStorage
 
+const filterByPlatformHelper = (
```
arielj: this refactor in this file look really overkill to reuse some function, this is not right

arielj:
> also i have a question. i am a noob using git. can you please tell me why test:e2e is failing?

e2e tests are flaky, they fail sometimes and in general it can be ignored, you can always run those tests locally to double check if they fail consistently or if they are just flaky

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,96 @@
...
+  currentFilter,
+  onFilterChange,
+  allGames
+}) => {
+  const { theme } = useContext(ContextProvider)
+  useEffect(() => {
+    if (themeType[theme] === 'light') {
+      document.body.setAttribute('data-theme-light', 'true')
+    } else {
+      document.body.removeAttribute('data-theme-light')
```
arielj: I don't think this is needed, but if we ever do need this, the `AlphabetFilter` component shouldn't be the one adding/removing attributes of the body element, components should only be concerned about it's internal state, not change the global scope

theguy000:
- better refactoring of src/frontend/screens/Library/index.tsx
- removed dark light themeing logic. instead, used css vars.
- used library context instead of prop drilling for the alphabetfilter btn

.vscode/easycode.ignore
arielj, line 1: please remove this file

src/frontend/components/UI/ActionIcons/index.tsx
```diff
@@ -92,6 +96,22 @@ export default React.memo(function ActionIcons({
             data-tour="library-sort-installed"
           />
         </button>
+        {showAlphabetFilter !== undefined && onToggleAlphabetFilter && (
```
arielj:
```suggestion
        {showAlphabetFilter !== undefined && (
```

I don't think `onToggleAlphabetFilter` can ever be undefined/false/null

src/frontend/components/UI/ThemeSelector/index.tsx
```diff
@@ -53,7 +53,7 @@ export const ThemeSelector = () => {
     setThemesPath(path)
     await writeConfig({ appName: 'default', config: newAppConfig })
     setAppConfig(newAppConfig)
-    loadThemes()
+    await loadThemes()
```
arielj: this change is unrelated to the alphabet filter

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { useMemo, useContext } from 'react'
+import './index.css'
+import LibraryContext from '../../LibraryContext'
+
+const ALPHABET = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('')
+const SYMBOLS = ['#']
+const ALPHABET_AND_SYMBOLS = [...ALPHABET, ...SYMBOLS]
```
arielj:
```suggestion
const CHARS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ#'.split('')
```

the `ALPHABET` and `SYMBOLS` arrays are not really used, only to generate the final array with all characters

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+    setAlphabetFilterLetter: onFilterChange,
+    gamesForAlphabetFilter: allGames
+  } = useContext(LibraryContext)
+
+  const availableChars = useMemo(() => {
+    const chars = new Set<string>()
+    if (!allGames) return chars
+    allGames.forEach((game) => {
+      if (game?.title) {
+        const firstCharMatch = game.title.match(/[a-zA-Z0-9]/)
```
arielj: maybe we should remove the substring `the ` from the title before comparing, because in the library we remove `the` for the sorting (as in "The Elder Scrolls ...." is in the same group as ll the games starting with `E` when you scroll)

src/frontend/screens/Library/index.tsx
```diff
@@ -382,7 +396,7 @@ export default React.memo(function Library(): JSX.Element {
     if (storesFilters['legendary'] && epic.username) {
       displayedStores.push('legendary')
     }
-    if (storesFilters['nile'] && amazon.username) {
+    if (storesFilters['nile'] && amazon.user_id) {
```
arielj: this change is unrelated to the alphabet filter

src/frontend/themes.scss
```diff
@@ -2,6 +2,7 @@ body {
   --text-default: #eae8e5;
   --text-secondary: #b1b1b1;
   --text-tertiary: var(--neutral-01);
+  --titlebar-symbol-color: var(--text-default);
```
arielj: this css variable is never used in this PR, why is it added here?

src/frontend/themes.scss
```diff
@@ -261,7 +267,6 @@ body.dracula-classic {
   --background-darker: var(--body-background);
   --current-background: var(--body-background);
   --navbar-background: #44475a;
-  --navbar-active: var(--accent);
```
arielj: why is this variable removed?

src/frontend/themes.scss
```diff
@@ -14,6 +15,10 @@ body {
   --anticheat-supported: var(--status-success);
   --anticheat-planned: var(--status-info);
   --anticheat-unknown: var(--status-denied);
+
+  /* AlphabetFilter */
+  --filter-accent-color: var(--accent);
```
arielj: maybe name these css variables `--alphabet-filter-....` so it's clear they are related the alphabet filter and not for the `filters` selector at the top right

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
arielj: this `alphabet-filter-symbol` class is not used for anything, there's no need to add it

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
arielj, line 1: for all the other components we name the file `index.tsx`

src/frontend/screens/Library/components/AlphabetFilter/index.css
```diff
@@ -0,0 +1,95 @@
...
+  );
+}
+
+body.zombie .alphabet-filter-button--active,
+body.zombie-classic .alphabet-filter-button--active {
+  --filter-accent-color: var(--primary-button);
+  --filter-accent-hover: var(--primary-button-hover, var(--primary-button));
+}
+
+body.old-school .alphabet-filter-button--active {
```
arielj: CSS that is specific for themes should be defined in the themes.css file, so it's all in one place

src/frontend/screens/Library/index.tsx
```diff
@@ -515,6 +528,55 @@ export default React.memo(function Library(): JSX.Element {
...
+        library = library.filter(
+          (game) => game.title && startsWithNumber.test(game.title)
+        )
+      } else {
+        library = library.filter(
+          (game) =>
+            game.title &&
+            game.title
+              .toLowerCase()
+              .startsWith(alphabetFilterLetter.toLowerCase())
```
arielj: this still fails to match something like `[REDACTED]` for the R, even though `[REDACTED]` is counted when checking if the `R` is an available character

src/frontend/screens/Library/components/AlphabetFilter/AlphabetFilter.tsx
arielj: and same thing when filtering

src/frontend/screens/Library/index.tsx
```diff
@@ -515,6 +528,55 @@ export default React.memo(function Library(): JSX.Element {
...
+    if (alphabetFilterLetter) {
+      if (alphabetFilterLetter === '#') {
+        const startsWithNumber = /^[0-9]/
+        library = library.filter(
+          (game) => game.title && startsWithNumber.test(game.title)
+        )
+      } else {
+        library = library.filter(
+          (game) =>
+            game.title &&
```
arielj: I don't think `game.title` can ever be other than a string

src/frontend/screens/Library/index.tsx
arielj:
```suggestion
            game.title
              .toUpperCase()
              .startsWith(alphabetFilterLetter)
```
but we should probably do something like what's done in the code that detects the available letters so it's always consistent (get the match with `/[a-zA-Z0-9]/`, upcase that character and compare with `alphabetFilterLetter`)

arielj (changes requested):
I added more comments

the 2 issues mentioned above are still present:
- I feel like a game like `[REDACTED]` should be shown in the `R` filter
- if you select a letter and use the search and get results that don't start with that letter, there's no way to unselect that letter because the button is disabled

src/frontend/screens/Library/components/AlphabetFilter/index.css
```diff
@@ -0,0 +1,77 @@
+:root {
+  --alphabet-filter-accent-color: var(--primary, var(--brand-primary));
+  --alphabet-filter-accent-hover: var(
```
arielj, line 3: when I mentioned the idea to rename the css variables, I meant rename all of them, not just those 2, now we have `--alphabet-filter-*` and `--filter-*`

src/frontend/screens/Library/index.tsx
```diff
@@ -515,6 +528,55 @@ export default React.memo(function Library(): JSX.Element {
...
+        let normalizedTitle = game.title.replace(/^the\s/i, '')
+        normalizedTitle = normalizedTitle.trim().toUpperCase()
+        normalizedTitle = normalizedTitle.replace(/^[^A-Z0-9]+/, '')
+
+        if (alphabetFilterLetter === '#') {
+          const startsWithNumber = /^[0-9]/
+          return startsWithNumber.test(normalizedTitle)
+        } else {
+          return normalizedTitle.startsWith(alphabetFilterLetter)
+        }
```
arielj:
```suggestion
        const processedTitle = game.title.replace(/^the\s/i, '')
        const firstCharMatch = processedTitle.match(/[a-zA-Z0-9]/)
        const firstChar = firstCharMatch[0].toUpperCase()

        if (alphabetFilterLetter === '#') {
          return /[0-9]/.test(firstChar)
        } else {
          return firstChar === alphabetFilterLetter
        }
```

this is probably faster than upcasing the complete titles for all games when we only care about the first relevant character

and at the same time the code is closer to what it's used to check the availability so things should be easier to understand in the future

src/frontend/screens/Library/components/AlphabetFilter/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+    } else if (!availableChars.has(value)) {
+      className += ' alphabet-filter-button--disabled'
+    }
+    return className
+  }
+
+  const handleClick = (value: string) => {
+    if (value === currentFilter) {
+      onFilterChange(null)
+      handleSearch('')
```
arielj: I don't think we should clear the search when unselecting a letter, just unselect the letter

public/locales/en/translation.json
```diff
@@ -417,6 +417,9 @@
         "releaseDate": "Release Date"
     },
     "library": {
+        "hideAlphabetFilter": "Hide Alphabet Filter",
+        "refresh": "Refresh",
```
arielj: where is this `"Refresh"` string coming from? is this added automatically when you run `pnpm i18next`?

src/frontend/components/UI/ThemeSelector/index.tsx
```diff
@@ -63,7 +63,6 @@ export const ThemeSelector = () => {
       setThemesPath(config.customThemesPath || '')
       loadThemes()
     }
-
```
arielj: let's remove these extra changes, please make sure the PR doesn't touch files unrelated to this feature

src/frontend/screens/Library/index.tsx
```diff
@@ -189,6 +190,18 @@ export default React.memo(function Library(): JSX.Element {
 
   const [showCategories, setShowCategories] = useState(false)
 
+  const [isAlphabetFilterVisible, setIsAlphabetFilterVisible] = useState(
```
arielj:
```suggestion
  const [showAlphabetFilter, setShowAlphabetFilter] = useState(
```

so it's consistent with the other useState calls that are named `show...` and `setShow...` above this one

src/frontend/components/UI/ActionIcons/index.tsx
```diff
@@ -92,6 +96,22 @@ export default React.memo(function ActionIcons({
             data-tour="library-sort-installed"
           />
         </button>
+        {showAlphabetFilter !== undefined && (
```
arielj: I don't think we need this `showAlphabetFilter !== undefined &&` check here, we want to always show the button to enable/disable the alphabet filter

I don't think it's even possible to have `undefined` here, I think the type of that context property might be wrong, cause I think we are always passing a value there

src/frontend/screens/Library/index.tsx
```diff
@@ -515,6 +528,55 @@ export default React.memo(function Library(): JSX.Element {
       )
     }
 
+    return library
+  }, [
+    storesFilters,
+    platformsFilters,
+    epic.library,
+    epic.username,
```
arielj: why are these `.username` and `.user_id` dependencies added here? they were not in the original list of dependencies in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4643/files#diff-f4c668b672834abb8870d392c22f649737bdea028101f149cc0264991f8fb8faL542-L544

arielj (changes requested):

public/locales/en/translation.json
theguy000: yes. `pnpm i18next` is the reason this line exists.

src/frontend/screens/Library/components/AlphabetFilter/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+    alphabetFilterLetter: currentFilter,
+    setAlphabetFilterLetter: onFilterChange,
+    gamesForAlphabetFilter: allGames
+  } = useContext(LibraryContext)
+
+  const availableChars = useMemo(() => {
+    const chars = new Set<string>()
+    if (!allGames) return chars
+    allGames.forEach((game) => {
+      if (game?.title) {
```
arielj: not sure I understand this `game?` part, can a `game` variable be undefined? `allGames` is an array of GameInfo objects

also can `game.title` ever be something other than a string? I don't think so, I don't think we need this if at all

src/frontend/screens/Library/components/AlphabetFilter/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+const AlphabetFilter: React.FC = () => {
+  const {
+    alphabetFilterLetter: currentFilter,
+    setAlphabetFilterLetter: onFilterChange,
+    gamesForAlphabetFilter: allGames
+  } = useContext(LibraryContext)
+
+  const availableChars = useMemo(() => {
+    const chars = new Set<string>()
+    if (!allGames) return chars
```
arielj: can this ever happen? `allGames` is defined as `GameInfo[]`, so it would never be `false` nor `undefined`

are you trying to check if it's empty?

src/frontend/types.ts
```diff
@@ -241,6 +241,11 @@ export interface LibraryContextType {
   setShowUpdatesOnly: (value: boolean) => void
   handleAddGameButtonClick: () => void
   setShowCategories: (value: boolean) => void
+  showAlphabetFilter?: boolean
+  onToggleAlphabetFilter?: () => void
```
arielj: can these 2 properties ever be undefined? I think we are always passing a value for both

src/frontend/screens/Library/components/AlphabetFilter/index.tsx
arielj: if you mark a comment as resolved with no change at all you have to explain why

you haven't answered my questions

arielj (changes requested):
@theguy000 , do not mark my questions as resolved if you don't answer them

I marked my questions as unresolved, if the code needs no change please answer the questions explaining why

theguy000:
My bad @arielj , I thought I pushed to the dev branch in my fork with the changes and solved these issues (which I did). It seems like removing "Refresh" string from [public/locales/en/translation.json](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4643/files/b60c424cbdb9255cec65dc88ccbf421996f81ca8#diff-8e77946ee485378b13babd5fe6bdc1707b77dce605ac6723d7a5b4494dc94925) caused my push to be failed (which I didn't notice)

public/locales/en/translation.json
arielj: hmmm I checked your branch locally, removed this line and ran `pnpm i18next`, and this line is not added back

There's no place in the code using the `'Refresh'` string nor using the `library.refresh` translation key for it to be added here

theguy000: Seems like it. I opened a new environment and ran `pnpm i18next`, but it didn’t add the line back.

arielj (approved):
I think it's working all good now

thanks!

PR #4658, opened 2025-06-12
[Fix] Fixup getDiskInfo on Windows when only a single drive is connected

CommandMC:
With a single drive connected, the PowerShell command we run will return a single object instead of an array. Account for this by allowing both a single `Win32_LogicalDisk` or an array of them to be the parse result, then handle both cases accordingly

---

### DIFF
--- a/src/backend/utils/filesystem/windows.ts
+++ b/src/backend/utils/filesystem/windows.ts
@@ -39,7 +39,11 @@ async function getDiskInfo_windows(path: Path): Promise<DiskInfo> {
 
   let parsedDisks: Win32_LogicalDisk[]
   try {
-    parsedDisks = Win32_LogicalDisk.array().parse(JSON.parse(stdout))
+    const parsed = Win32_LogicalDisk.or(Win32_LogicalDisk.array()).parse(
+      JSON.parse(stdout)
+    )
+    if (Array.isArray(parsed)) parsedDisks = parsed
+    else parsedDisks = [parsed]
   } catch {
     parsedDisks = []
   }


src/backend/utils/filesystem/windows.ts
```diff
@@ -39,7 +39,11 @@ async function getDiskInfo_windows(path: Path): Promise<DiskInfo> {
 
   let parsedDisks: Win32_LogicalDisk[]
   try {
-    parsedDisks = Win32_LogicalDisk.array().parse(JSON.parse(stdout))
+    const parsed = Win32_LogicalDisk.or(Win32_LogicalDisk.array()).parse(
+      JSON.parse(stdout)
+    )
+    if (Array.isArray(parsed)) parsedDisks = parsed
+    else parsedDisks = [parsed]
```
arielj, line 46:
```suggestion
    parsedDisks = [parsed].flat()
```

just a suggestion, I think it's simpler than the if/else, but feel free to ignore

CommandMC: I've somewhat copied this handling from `Win32_VideoController` here:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/b7cd9d373249c00c83f58f292963fa53eccccc47/src/backend/utils/systeminfo/gpu/windows.ts#L38-L39

On Discord the OP for this issue expressed interest in creating a generic `Get-CimInstance` wrapper, so I figured I'd leave this code as obviously-a-duplicate as I could

PR #4659, opened 2025-06-13
[Feat] Replace font detection with browser-based approach

### LINKED ISSUE #4654: List of fonts breaks them up into two groups, one with quotes and one without
### Describe the bug

In the list of fonts in the accessibility menu, they are broken up into two groups. Ones with spaces in the name are listed as "Roboto Medium" with quotes and the ones without spaces in the name are listed without quotes, Roboto.

I feel from a nice-to-have UI standpoint, the quotes should be removed from display, and the two groups should be sorted alphabetically instead of having them in separate clusters.

With the Roboto example, you have to scroll up and down as the Roboto family appears in both clusters.

### Add logs

```text
n/a
```

### Steps to reproduce

1. Scroll through the fonts in the accessibility menu

### Expected behavior

No quotes, sorted alphabetically.

### Screenshots

_No response_

### Heroic Version

Latest Stable (Flatpak)

### System Information

SteamOS 3.6.24

Heroic 2.17.2

### Additional information

_No response_

CommandMC:
Originally I just set out to fix #4654. In doing that, I thought "Huh, our font detection code is quite complex, surely there's an easier way to do this". There is!

This replaces `font-list` with a call to [`queryLocalFonts`](<https://developer.mozilla.org/en-US/docs/Web/API/Window/queryLocalFonts>). In turn it also fixes the weird formatting, and is a great deal faster too.  
I opted to remove the "Refresh" button along the way, since I didn't see much sense in it now

Closes #4654

---


arielj (approved):
This is great, remove all the cooooodeee

Issue #1364, opened 2022-05-23
Game importing validation bug


JordanPlayz158:
### Describe the bug

When importing a game, whether or not the game is actually imported/the directory being proper. This is only for Epic/Legendary Games

Here is the log but as explained in discord, it doesn't give any meaningful issue because the way game importing is validated is a bit flawed: https://paste.jordanplayz158.xyz/?8bc5a65f02de52c3#ASQZu5iNiE4e4EhLzeZWmaXe8TzcJcTuhnqRyKkUGxqr (here is the start of the discord discussion: https://discord.com/channels/812703221789097985/812703452085223434/977591817220542514)

Here is a video on how to do it: https://dl.jordanplayz158.xyz/2022-05-22%2009-25-11.mp4

### Add logs

https://paste.jordanplayz158.xyz/?8bc5a65f02de52c3#ASQZu5iNiE4e4EhLzeZWmaXe8TzcJcTuhnqRyKkUGxqr

### Steps to reproduce

1. Go to the download button icon on any game in Epic Games section (must be not imported/installed to have the import option)
2. Go to any path and hit import (as long as it's not the game directory)
3. You will now get a Game Imported notification indicating success but the frontend will tell you it's not installed properly but the backend in the logs will say it imported successfully incorrectly

Watch the video or follow above: https://dl.jordanplayz158.xyz/2022-05-22%2009-25-11.mp4

### Expected behavior

Either no notification or Game Import Failure

### Screenshots

Even better than a screenshot: https://dl.jordanplayz158.xyz/2022-05-22%2009-25-11.mp4

### System Information

- OS [e. g. "Ubuntu"]: Manjaro Linux
- Heroic Version [e. g. 2.1.1]: v2.3.3

krablord:
got same issue trying to import cyberpunk from downloading it before with heroic on different linux distro

JordanPlayz158:
Issue is still present on Version 2.4.1

flavioislima:
@JordanPlayz158 what happens if you try importing it directly with Legendary? It gives some error message?

JordanPlayz158:
This is the output that occurs from incorrectly importing a game (importing a game with the wrong directory)
```legendary import Duckbill /mnt/HDD

[Core] INFO: Trying to re-use existing login session...
[Core] INFO: Downloading latest manifest for "Duckbill"
[...]

PR #4667, opened 2025-06-15
[Fix]: Improve error handling for game import failures

KaranDuggal:
It is the fix of #1364 bug.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -980,7 +980,10 @@ addHandler(
     })
 
     const abortMessage = () => {
-      notify({ title, body: i18next.t('notify.install.canceled') })
+      notify({
+        title,
+        body: i18next.t('notify.import.failed', 'Importing Failed')
+      })
       sendGameStatusUpdate({
         appName,
         runner,
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -785,7 +785,8 @@ export async function importGame(
 
   const res = await runLegendaryCommand(command, { abortId: appName })
   addShortcuts(appName)
-
+  const errorMatch = res.stderr.match(/^.*ERROR:.*$/gm)?.join('') ?? ''
+  res.error = (res.error ?? '') + errorMatch
   if (res.error) {
     logError(
       ['Failed to import', `${appName}:`, res.error],
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -470,6 +470,9 @@
             },
             "reparing": "Finished Repairing"
         },
+        "import": {
+            "failed": "Importing Failed"
+        },
         "install": {
             "canceled": "Installation Canceled",
             "error": "Installation Failed",


tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "module": "esnext",
-    "target": "es2017",
-    "lib": ["esnext", "dom", "dom.iterable"],
+    "target": "es2018",
+    "lib": ["esnext", "dom", "dom.iterable", "es2018"],
```
arielj: why is this change needed?

src/backend/main.ts
```diff
@@ -980,7 +980,7 @@ addHandler(
     })
 
     const abortMessage = () => {
-      notify({ title, body: i18next.t('notify.install.canceled') })
+      notify({ title, body: i18next.t('notify.importing.failed') })
```
arielj:
```suggestion
      notify({ title, body: i18next.t('notify.import.failed', 'Import failed') })
```

the other keys under `notify` don't end up with `ing`, and it's a good idea to add the default value when adding a new key

src/backend/storeManagers/legendary/games.ts
```diff
@@ -785,7 +787,8 @@ export async function importGame(
 
   const res = await runLegendaryCommand(command, { abortId: appName })
   addShortcuts(appName)
-
+  const stdLogs = convertStdLogsIntoObj(res.stderr)
+  res.error = stdLogs.ERROR.join('')
```
arielj: I think the convertStdLogsIntoObj might be a bit too much for what we need here, in other places we simply check if the output matches `ERROR` like in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/25247219fbc38547675990ebdc07df8317a84d85/src/backend/storeManagers/nile/games.ts#L134

it should make the code simpler without the need of this new object

arielj (changes requested):

tsconfig.json
KaranDuggal: I had upgraded it because the regex I was using included the s (dotAll) flag, which allows the dot (.) in the regex to match newline characters as well. This feature was introduced in ES2018. However, I have now modified the code and reverted that change.

src/backend/main.ts
KaranDuggal: Done

src/backend/storeManagers/legendary/games.ts
KaranDuggal: This code is not working as expected. I'm currently using the regex stderr.match(/^.*ERROR:.*$/gm) to extract error messages from the stderr string. I believe the logic is fine now, but please let me know if any changes are needed.

You can test it using the following stderr data, which I'm receiving from the runLegendaryCommand function:
`stderr = '[Core] INFO: Trying to re-use existing login session...\n[Core] INFO: Downloading latest manifest for "Hornbill"\n[cli] ERROR: No files belonging to Game "The Alto Collection" (Hornbill) found in the specified location, please verify that the path is correct.\n[cli] INFO: Did you mean "/home/duggal/Downloads/KaranFase2/TheAltoCollection"?';`

arielj (approved):
It works as expected, thanks!

PR #4677, opened 2025-06-19
[DX] Fix locale keys getting deleted

arielj:
After merging https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4642, the translations check complains when trying to push about translations out of sync, running the command actually removes some keys (not sure why, cause the `t('status.installing', 'Installing')` code exists in the codebase :shrug:)

Anyway, I'm adding the translations as comments to be sure they are processed correctly.

---


CommandMC:
`i18next`'s process for finding translation keys is not very smart. Changing this line

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/97afbf1ed10d3d606b6d8ad2cc9f604a870fc509/src/frontend/screens/Game/GamePage/components/GameStatus.tsx#L94

to say

```ts
      return `${is.importing ? t('status.importing') : t('status.installing')} ${currentProgress}`
```

resolves the issue

arielj:
just pushed an update, based on that but also adding the english fallbacks, cause the `Importing` string was nowhere in the code

it looked weird to have the long ternary operator inside the string interpolation so I moved that out

PR #4680, opened 2025-06-19
[FIX] Prevent error if corrupted installed_games file

arielj:
I'm not sure if this could be related to https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4678

but a user reported this error in discord:
```
An exception occurred when launching the game:
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at vn (/tmp/.mount_HeroicnfhRHc/resources/app.asar/build/main/main.js:145:1410)
    at Module.Fd [as launch] (/tmp/.mount_HeroicnfhRHc/resources/app.asar/build/main/main.js:105:7143)
    at Zo (/tmp/.mount_HeroicnfhRHc/resources/app.asar/build/main/main.js:143:1914)
    at Session.<anonymous> (node:electron/js2c/browser_init:2:106823)
```

And did some digging and the issue is the `installed_games` file seems to exist but has invalid json content (I can't confirm yet what's in the invalid file, I asked the user).

So this adds a try/catch around the `JSON.parse` call (which is always a good practice anyway cause it can throw).

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -674,9 +674,16 @@ async function prepareWineLaunch(
       writeFileSync(appsNamesPath, JSON.stringify([appName]), 'utf-8')
       hasUpdated = true
     } else {
-      const installedGames: string[] = JSON.parse(
-        readFileSync(appsNamesPath, 'utf-8')
-      )
+      let installedGames: string[] = []
+
+      try {
+        installedGames = JSON.parse(
+          readFileSync(appsNamesPath, 'utf-8')
+        ) as string[]
+      } catch (error) {
+        logError(error)
+      }
+
       if (!installedGames.includes(appName)) {
         installedGames.push(appName)
         writeFileSync(appsNamesPath, JSON.stringify(installedGames), 'utf-8')


src/backend/launcher.ts
```diff
@@ -674,9 +674,16 @@ async function prepareWineLaunch(
...
     } else {
-      const installedGames: string[] = JSON.parse(
-        readFileSync(appsNamesPath, 'utf-8')
-      )
+      let installedGames: string[] = []
+
+      try {
+        installedGames = JSON.parse(
+          readFileSync(appsNamesPath, 'utf-8')
+        ) as string[]
```
CommandMC: In cases like this, where it's easy to add a runtime type-check, I'd say we should do so.
```suggestion
        installedGames = z.string().array().parse(JSON.parse(
          readFileSync(appsNamesPath, 'utf-8')
        ))
```
(missing `import z from 'zod'` and probably won't lint, but you get the idea)

src/backend/launcher.ts
```diff
@@ -674,9 +674,16 @@ async function prepareWineLaunch(
...
+      } catch (error) {
+        logError(error)
```
CommandMC: Do we really want to just continue on here? My thinking is
- user has many different games sharing the same prefix
- `installed_games` file gets corrupted somehow, perhaps in a way that's fixable with some inspection
- Heroic overwrites `installed_games`, removing all previously-recorded AppNames

arielj: yeah, that's a good point, I was thinking that most of the times users have a prefix per game so it's kinda safe to just regenerate the file to not prevent launching the game, but I guess a better error message to fix the file can also help us understand over time why it can even get corrupted

devpool007: Btw do you guys also use AI to review the small changes or is it done just manually always?

arielj: I don't use AI in general

devpool007: Ah ok nice, was just trying to understand the contribution process and review process. As some open source projects use AI bots for initial reviews these days.

arielj:
I'll wait until I see this issue again in support threads and asks for the installed_games file to understand how it gets corrupted

src/backend/launcher.ts
```diff
@@ -855,9 +856,19 @@ async function prepareWineLaunch(
...
+
+      try {
+        installedGames = z
+          .string()
+          .array()
+          .parse(JSON.parse(readFileSync(appsNamesPath, 'utf-8')))
+      } catch (error) {
+        copyFileSync(appsNamesPath, appsNamesPath + '.old')
+        logError(
+          `Error reading installed_games file: ${error}. Invalid file backed up as ${appsNamesPath}.old`
```
arielj, line 869: When the file is corrupted, we make a copy instead of simply overriding it, so users with this issue can report this back if some of their games inside the single prefix disappear.

I imagine this won't be a common case since we use different prefixes by default.

CommandMC (approved):

PR #4681, opened 2025-06-20
[Tech] DX improvements to `useGlobalState`/`GlobalStateV2`

CommandMC:
@arielj recently asked me some questions about `useGlobalState` usage, which revealed some potential DX traps one can fall into when using it

- `useGlobalState` now automatically wraps the provided selector in `useShallow`
- `useShallowGlobalState` was renamed to `useGlobalStateKeys` to clarify what it does
- Documentation for both functions was added
- Export is done as a single default export. This is done (1) to preserve the `setState` function on `useGlobalState` (technically that's possible by just assigning a property, but that seems like a bad idea to me) and (2) since most state consumers need all methods anyways
  - `useGlobalStateKeys` is exported as `keys`, to be used as `useGlobalState.keys`

---


arielj (approved):
looks good!

PR #4684, opened 2025-06-22
Update ESLint and related packages

CommandMC:
No major functionality changes here, just a few small things
- The `typescript-eslint` update meant it found a few `try { ... } catch (error) { ... }` blocks where `error` was unused. I either used `error` (passing it along to a logging function or similar), or just removed it
- Operations in the form of `booleanIshValue ? function1() : function2()` were converted to standard `if (booleanIshValue) { function1() } else { function2() }` form. `typescript-eslint` has a rule for this, and it's more readable IMO
- `throw`n values / Promise rejection values are now always `Error`s
- Importing `React` is not required to use JSX, so a lot of unnecessary `import React from 'react'`s were removed
- Some misc changes to imports (`eslint-plugin-import-x` seems to have gotten smarter)

In turn I also moved as many packages as possible to ESLint flat configs

---


src/backend/utils.ts
```diff
@@ -889,13 +889,7 @@ export const spawnAsync = async (
...
 
   return new Promise((resolve, reject) => {
-    child.on('error', (error) =>
-      reject({
-        code: 1,
-        stdout: stdout.join(''),
-        stderr: stderr.join('').concat(error.message)
-      })
-    )
+    child.on('error', reject)
```
arielj, line 892: I'm curious about this, I understand it's so the `reject` receives an Errors object, but it seems to have a different signature than the one we were sending before, is there a place that catches this rejections that should be updated somehow?

I looked for places calling `spawnAsync` and I feel like we've had this wrong all this time, from how it's used it looks like it was intended to be

```ts
child.on('error', (error) =>
  resolve({
    code: 1,
    stdout: stdout.join(''),
    stderr: stderr.join('').concat(error.message)
  })
)
```

(using `resolve` instead of `reject`)

because we use it as `const { code, stderr } = await spawnAsync(` in many places and then check the value of `code` instead of using a try/catch to catch the error

I wonder if we need to change those places to use try/catch instead now that we have this properly rejecting with an Errors object

arielj (commented):
I looks good to me, but I left a comment on some code that I think we had wrong already before

src/backend/utils.ts
CommandMC: I did come to the same conclusion as you did (`resolve` being the intended function call), but I didn't want to break any existing behavior that relied on the promise being rejected (if such code exists anywhere). Changing the type of the rejected value is relatively safe to do (since the type of it is going to be `unknown` to consumer functions), changing the behavior from "rejects on error" to "resolves on error" might not be

arielj (approved):

src/backend/utils.ts
arielj: maybe we can add a `TODO` and keep the old code commented out? to not forget about this, I don't know haha

I do think it's a bug we have, but I guess it's failing with an exception when the spawn fails, instead of handling the error, though it's an error state in both cases

maybe it's just too rare for that to actually fail and it probably means other things failed before anyway and that's why we never noticed it

flavioislima (approved):
lgtm

Issue #1826, opened 2022-09-20
Change back/forward mouse buttons while in a store to navigate pages, instead of screens in heroic


SpiceMan:
### Problem description

I'm not exactly sure what the "events" are but let's say that if you
1) open heroic and it starts in the library
2) you go to stores -> epic store
3) click on a game title
4) press your mouse "back button"

It goes to the library screen.

### Feature description

I think that the mouse back button should go back to the epic store top page, not the heroic library screen.
it's basically a browser inside the app, should probably behave like one.

### Additional information

This is on linux.
I don't know if it also happens in other platforms, or just linux.

mamasch19:
Maybe I'm a bit late to the party but I can confirm this is also happening on Windows 10 with Heroic 2.16.0.

squirrel-bat:
I'm even later to the party (sorry for that), but this issue still exists in version 2.18.1 under Linux (Ubuntu 24.04.3 LTS).

arielj:
hmmmm I can't reproduce this, maybe your mouse is sending different button codes for the back/forth buttons?

in general it should be buttons 4 and 5 but maybe your mouse is doing something different? because I implemented that feature and my mouse sends 4/5 and it works fine https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button

arielj:
you could try this to check:
- open chrome
- open devtools (ctrl+shift+i)
- click the `Console` tab
- enter this code
```
document.addEventListener('mouseup', (ev) => {
  console.log({buttonPressed: ev.button})
})
```
- then press the mouse buttons

it should show the button number

squirrel-bat:
> you could try this to check:
> 
> * open chrome
> * open devtools (ctrl+shift+i)
> * click the `Console` tab
> * enter this code
> 
> ```
> document.addEventListener('mouseup', (ev) => {
>   console.log({buttonPressed: ev.button})
> })
> ```
> 
> * then press the mouse buttons
> 
> it should show the button number

This turns out to be trickier than expected, as the browser seems to capture the buttons presses and interpret them as history forward/back and only sending those events and I feel this might be the same behaviour in Heroic.

When listening to history changes, for example by registering an event listener to the `popstate` event on the `window` object, I do get the forward/back events shown in my browser's console:
```js
window.addEventListener('popstate', function(ev) {
  console.log(ev.state)
})
```

[...]

PR #4701, opened 2025-06-29
[UX] Navigate back/forward inside webview with mouse buttons

arielj:
This PR closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/1826

I had to add back some webviewPreload.js code so we can add mouse event listener inside the webview.

---


src/frontend/screens/WebView/index.tsx
```diff
@@ -146,6 +146,16 @@ export default function WebView() {
...
 
+  const [webviewPreloadPath, setWebviewPreloadPath] = useState('')
+  useEffect(() => {
+    const fetchWebviewPreloadPath = async () => {
+      const path = await window.api.getWebviewPreloadPath()
+      setWebviewPreloadPath(path)
+    }
+
+    void fetchWebviewPreloadPath()
+  }, [])
```
arielj, line 157: I actually wanted to use the app's preload script and just put this in the window object, but I can't properly import paths there, so I'll leave that for another time as an improvement.

src/frontend/screens/WebView/index.tsx
```diff
@@ -315,6 +360,7 @@ export default function WebView() {
         partition="persist:epicstore"
         src={startUrl}
         allowpopups={trueAsStr}
+        {...{ preload: webviewPreloadPath }}
```
arielj: looks like electron types are not correct cause it complains that `preload` is not a valid prop, so doing this hacky thing

CommandMC: This isn't an issue with Electron's types (Electron apparently doesn't even author these types, it's in React itself). The ESLint React plugin thinks the property is invalid. The following resolves it:
```patch
Index: src/frontend/screens/WebView/index.tsx
IDEA additional info:
<+>UTF-8
diff --git a/src/frontend/screens/WebView/index.tsx b/src/frontend/screens/WebView/index.tsx
--- a/src/frontend/screens/WebView/index.tsx	(revision Staged)
+++ b/src/frontend/screens/WebView/index.tsx	(date 1751233108355)
@@ -360,7 +360,7 @@
         partition="persist:epicstore"
         src={startUrl}
         allowpopups={trueAsStr}
-        {...{ preload: webviewPreloadPath }}
+        preload={webviewPreloadPath}
       />
       {showLoginWarningFor && (
         <LoginWarning
Index: eslint.config.mjs
IDEA additional info:
<+>UTF-8
diff --git a/eslint.config.mjs b/eslint.config.mjs
--- a/eslint.config.mjs	(revision Staged)
+++ b/eslint.config.mjs	(date 1751233108331)
@@ -33,7 +33,7 @@
 
       'react/no-unknown-property': [
         'error',
-        { ignore: ['partition', 'allowpopups', 'useragent'] }
+        { ignore: ['partition', 'allowpopups', 'useragent', 'preload'] }
       ],
       '@typescript-eslint/no-misused-promises': [
         'error',
```

arielj: thanks! I updated it with that code and all good now

src/frontend/screens/WebView/index.tsx
flavioislima: Maybe add a TODO comment so we won't forget about it

arielj: I mean, it's not just for this, I've been thinking about that for a long time already, I know I won't forget cause it's something in my TODO for a long time haha

my idea is to try to move as much "this won't ever change after boot" info as possible to the preload script so we don't have to fetch that info with `window.api` calls

it wouldn't be just for this, we have other values that we ask the backend that we could move there

PR #4691, opened 2025-06-27
[WIP] Some logger fixes

arielj:
- Init logger before migrations: clean install and e2e were failing

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #4709, opened 2025-07-02
[TECH] Improve app loading when on Wayland on Linux

fxzzi:
use `did-finish-load` event in wayland environments for showing main window
---

Currently, when running the launcher with native wayland (`ELECTRON_OZONE_PLATFORM_HINT=auto`), the `ready-to-show` event that we normally wait for to open the window is not fired.

Instead, for environments where the `WAYLAND_DISPLAY` environment variable is set, we can use the `did-finish-load` event instead. https://www.electronjs.org/docs/latest/api/web-contents#event-did-finish-load

It is reached at a similar point as `ready-to-show`.

Other apps also seem to use this similar workaround for Wayland environments, as it seems like a bug in Electron that has existed for a while now. One example is signal messenger, who added this workaround in this commit: https://github.com/signalapp/Signal-Desktop/commit/f51ab7e7c348a3dea8006622ce369f2d4a9dae5a

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -403,15 +403,23 @@ if (!gotTheLock) {
     }
 
     const headless = isCLINoGui || settings.startInTray
+
     if (!headless) {
-      mainWindow.once('ready-to-show', () => {
+      const isWayland = Boolean(process.env.WAYLAND_DISPLAY)
+      const showWindow = () => {
         const props = configStore.get_nodefault('window-props')
         mainWindow.show()
         // Apply maximize only if we show the window
         if (props?.maximized) {
           mainWindow.maximize()
         }
-      })
+      }
+      if (isWayland) {
+        // Electron + Wayland don't send ready-to-show
+        mainWindow.webContents.once('did-finish-load', showWindow)
+      } else {
+        mainWindow.once('ready-to-show', showWindow)
+      }
     }
 
     // set initial zoom level after a moment, if set in sync the value stays as 1


flavioislima:

fxzzi:

arielj:
we intentionally NOT show the window when it's created and wait for the `ready-to-show` event to let us know when the app is loaded https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/main.ts#L409

without that, the app would open with full white and then after a moment will open heroic's interface

can you confirm if the `ready-to-show` event is or is not firing for you?

because what I'm thinking, if for any reason that's the case sounds like a bug in electron not triggering that event

a workaround we could have is to have both that event listener and a timer, and show the mainWindow whatever happens first, but always start NOT shown to avoid the white screen

fxzzi:
I guess you are right that the `ready-to-show` event is not firing. The app never reaches that codepath so the main window is never shown. It could be an electron bug. I am building with nix, using electron 36.6.0. I could try one of the nightly releases or something if you'd like? Doubt it would change anything however, as this has been a bug for a while I think.

arielj:
I do think the solution shouldn't be showing the main windows as soon as it's created, we should still wait for the event for the other environments where it works.

We could have timer or something, that if the mainWindow didn't open in... I don't know... 1 second? 3 seconds? show it, to give time for the event to happen or the page to be ready.

That should work around the issue of some environments not triggering the event and still work as it is now for all the others

fxzzi:
@arielj 

I've pushed changes. Instead of showing the window immediately, I added a 1 second fallback in case ready-to-show is never reached.

Tested `heroic --ozone-platform=x11`, heroic launches quickly
Tested `heroic --ozone-platform=wayland`, heroic has the small 1 second wait but still does launch fine.

EDIT:

scratch what I said before, no 1 second fallback needed. Instead check if the user is running in a Wayland environment, and use the `did-finish-load` event instead.

Signal messenger uses the same workaround. https://github.com/signalapp/Signal-Desktop/commit/f51ab7e7c348a3dea8006622ce369f2d4a9dae5a

fxzzi:
not sure why CI is failing. any help?

arielj:
> not sure why CI is failing. any help?

it's an issue in the `main` branch fixed by https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4691, don't worry about that

flavioislima:
Thanks for implementing that

PR #4728, opened 2025-07-11
[macOS] Add Wine-Staging-macOS to WineManager

flavioislima:
Re-adds the option since it keeps getting updated and some users said some games works better with it then with wine-crossover.

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -99,7 +99,6 @@
                 "title-finished": "Update Finished"
             }
         },
-        "launchAnyway": "Launch Game Anyway",
         "no": "NO",
         "ok": "OK",
         "postpone": "Postpone",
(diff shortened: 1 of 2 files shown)


public/locales/en/translation.json
```diff
@@ -99,7 +99,6 @@
                 "title-finished": "Update Finished"
             }
         },
-        "launchAnyway": "Launch Game Anyway",
```
arielj, line 102: I think this file includes changes unrelated to this PR

I agree we should probably remove the pre-shader warning dialog now that umu doesn't require it, but it should be a different PR

arielj (commented):
I left one comment

I'll try to test this tomorrow on my intel mac

arielj:
I tested it on my intel mac and the feature was fine, we only need to cleanup the translations file

public/locales/en/translation.json
flavioislima: this was removed on Commands's PR and it was making i18n command to fail. My PR doesnt remove the dialog, only fix the workflow.

PR #4729, opened 2025-07-11
Chore: update Gamescope and Mangohud runtime version messages

flavioislima:
Update it to use parameter instead of hardcoded runtime version.

---

### DIFF
--- a/flatpak/templates/com.heroicgameslauncher.hgl.yml.template
+++ b/flatpak/templates/com.heroicgameslauncher.hgl.yml.template
@@ -1,9 +1,9 @@
 id: com.heroicgameslauncher.hgl
 sdk: org.freedesktop.Sdk
 runtime: org.freedesktop.Platform
-runtime-version: '24.08'
+runtime-version: '23.08'
 base: org.electronjs.Electron2.BaseApp
-base-version: '24.08'
+base-version: '23.08'
 command: heroic-run
 separate-locales: false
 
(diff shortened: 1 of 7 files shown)


arielj (changes requested):
the runtime version is also mentioned here https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/chore/runtime_param/src/backend/launcher.ts#L627

I think it would be good to define the version of the flatpak runtime in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/chore/runtime_param/src/backend/constants/environment.ts so it's defined in a single place

we can expose the constant to the frontend using this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/chore/runtime_param/src/preload/index.ts and do `windows.flatpakRuntime` for example for the one place we need it in the frontend

flavioislima:
@arielj did your requested changes.

biliesilva (approved):

flavioislima:
> looks good!
> 
> I only wonder if we even need the `|| '23.08` fallback for the constant? like... if we are inside flatpak the env var should be defined, and outside flatpak we shouldn't need to use it
> 
> also, I see flatpak/templates/com.heroicgameslauncher.hgl.yml.template still mentions 24.03, but that's because this file is actually edited in the flathub repo, right?

yes, I will update on the flatpak repo, but I think the one here is used for the local flatpak build workflow and I will update it as well.

PR #4732, opened 2025-07-12
Add settings to enable Ntsync

Twig6943:
This PR Adds Ntsync

Based on https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/commit/9a9c1568e93887ea052e8843d3f5da8fa9f3a909

---

### DIFF
--- a/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
+++ b/src/frontend/screens/Settings/sections/GamesSettings/index.tsx
@@ -189,9 +189,9 @@ export default function GamesSettings() {
             )}
             <EnableEsync />
             <EnableFsync />
+            <EnableNtsync />
             <EnableWineWayland />
             <EnableMsync />
-            <EnableNtsync />
             <AdvertiseAvxForRosetta />
             <EnableFSR />
             <EnableDXVKFpsLimit />


flavioislima:

Twig6943:

src/frontend/screens/Settings/components/EnableNtsync.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useContext } from 'react'
+import { useTranslation } from 'react-i18next'
+import { ToggleSwitch } from 'frontend/components/UI'
+import useSetting from 'frontend/hooks/useSetting'
+import ContextProvider from 'frontend/state/ContextProvider'
+import SettingsContext from '../SettingsContext'
+import InfoIcon from 'frontend/components/UI/InfoIcon'
+
+const EnableNtsync = () => {
```
flavioislima, line 9: If this should work only with Proton then we should hide it for other wine versions.
Check this line to see how we do it:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/9a9c1568e93887ea052e8843d3f5da8fa9f3a909/src/frontend/screens/Settings/components/AutoDXVK.tsx#L23

Twig6943:
image


@flavioislima I forgot what it is but I think there's also an env var for wine?

src/backend/config.ts
```diff
@@ -348,6 +348,7 @@ class GlobalConfigV0 extends GlobalConfig {
       wineVersion: defaultWine,
       enableEsync: true,
       enableFsync: isLinux,
+      enableNtsync: isLinux,
```
arielj, line 351: is it safe to make this true by default? I now it's pretty new right?

src/backend/launcher.ts
```diff
@@ -455,6 +456,7 @@ function filterGameSettingsForLog(
       if (wineType) {
         if (wineType.type === 'wine') {
           delete gameSettings.wineCrossoverBottle
+          delete gameSettings.enableNtsync
```
arielj, line 459: we should delete this from the crossover and toolkit logs too

src/frontend/screens/Settings/components/EnableNtsync.tsx
```diff
@@ -0,0 +1,39 @@
...
+    return <></>
+  }
+
+  return (
+    <div className="toggleRow">
+      <ToggleSwitch
+        htmlId="NtsyncToggle"
+        value={enableNtsync || false}
+        handleChange={() => setEnableNtsync(!enableNtsync)}
+        title={t('setting.Ntsync', 'Enable Ntsync')}
```
arielj: is this the correct case to use? I think it should be `NTSync` cause it comes from NT Synchronization https://docs.kernel.org/next/userspace-api/ntsync.html

src/frontend/screens/Settings/components/EnableNtsync.tsx
```diff
@@ -0,0 +1,39 @@
...
+      />
+
+      <InfoIcon
+        text={t(
+          'help.Ntsync',
+          'Ntsync aims to reduce wineserver overhead in CPU-intensive games. Enabling may improve performance on supported Linux kernels.'
```
arielj: just curious, is this explanation correct? I see it's a copy of the Esync/Fsync/Msync options

but when I read them they all sounds like the same thing and I wonder if there's a better explanation that we can use

Twig6943: > just curious, is this explanation correct? I see it's a copy of the Esync/Fsync/Msync options
> 
> but when I read them they all sounds like the same thing and I wonder if there's a better explanation that we can use

Most it is most likely correct. I just copy pasted it as a placeholder but it doesn't need to be changed unless said otherwise by some wine folks.

Twig6943:
> image
> 
> @flavioislima I forgot what it is but I think there's also an env var for wine?

@flavioislima Might as well just disable it for wine until it gets merged to the wine gitlab. (AKA main wine repo)

Got the thing working but I gotta move ntsync option higher in the ui any idea on how to achieve that?

image

image

arielj:
you can move the code you added in this file up a few lines https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4732/files#diff-fabd2004e8b97f62791c3b71d613f3760c45fa055845451922cc1b1cee030299R191

src/backend/config.ts
Twig6943: IDK tbh, will set it to `false` if you say so

src/frontend/screens/Settings/components/EnableNtsync.tsx
Twig6943: > is this the correct case to use? I think it should be `NTSync` cause it comes from NT Synchronization https://docs.kernel.org/next/userspace-api/ntsync.html

Yes I think so. The env var toggles ntsync for proton (Not the kernel module ik but still)

Twig6943:
> you can move the code you added in this file up a few lines https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4732/files#diff-fabd2004e8b97f62791c3b71d613f3760c45fa055845451922cc1b1cee030299R191

Done

image

Twig6943:
Yo also would you guys be interested in lsfg-vk integration?

Currently we're working on a flatpak runtime https://github.com/PancakeTAS/lsfg-vk/issues/63

src/backend/config.ts
roland-5: You need kernel version 6.14 and wine (or proton) version that support it - so if someone using Linux with older kernel than 6.14, will have problem and there were few fixes in 6.15... And there are few games that don't work with ntsync, but still work with esync or fsync. At this moment wine is trying to remove esync and replace it with ntsync, but it still need polish in many places.

BananaWorks07:
Btw this will become outdated once ntsync becomes opt out rather than opt in

Twig6943:
> Btw this will become outdated once ntsync becomes opt out rather than opt in

There's a good while for that I think? 🤷

BananaWorks07:
> > Btw this will become outdated once ntsync becomes opt out rather than opt in
> 
> There's a good while for that I think? 🤷

Not really, I fixed the 32 bit issue yesterday for instance. also if/when valve chooses to add it, it might look completely different to this

Twig6943:
> > > Btw this will become outdated once ntsync becomes opt out rather than opt in
> > 
> > 
> > There's a good while for that I think? 🤷
> 
> Not really, I fixed the 32 bit issue yesterday for instance. also if/when valve chooses to add it, it might look completely different to this

Well idk what to do as of right now

flavioislima:
Due to all these open questions about NTsync future I believe it is better to hold it to the release after the next one.

imLinguin:
I think we should keep it in-line with current synchronization settings. Let's make it `Enable NTsync` - enabled by default

src/backend/launcher.ts
```diff
@@ -1207,6 +1210,9 @@ function setupWineEnvVars(gameSettings: GameSettings, gameId = '0') {
   if (isLinux && !gameSettings.enableFsync && wineVersion.type === 'proton') {
     ret.PROTON_NO_FSYNC = '1'
   }
+  if (isLinux && !gameSettings.enableNtsync && wineVersion.type === 'proton') {
+    ret.PROTON_USE_NTSYNC = '0'
+  }
```
imLinguin, line 1215: You should also handle the case for wine, like in case of `WINEFSYNC` we should have `WINENTSYNC=1`

imLinguin (changes requested):

src/backend/launcher.ts
BananaWorks07: Upstream Wine will have ntsync enabled by default with no variable to control if it's on or off

Francehelder: In the [latest release of GE-Proton](https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/GE-Proton10-10) NTsync was enabled by default.

Twig6943:
Soo wine 11 is coming up any updated on this? @Etaash-mathamsetty

BananaWorks07:
> Soo wine 11 is coming up any updated on this? @Etaash-mathamsetty

there's no mechanism to enable/disable ntsync on upstream wine, so this toggle is not very helpful unfortunately

arielj:
should we close this PR? I understand this toggle won't really work anyway

Twig6943:
> should we close this PR? I understand this toggle won't really work anyway

Yea should be closed until a way of toggling ntsync is introduced in upstream wine. Kinda ridicilous they went with this.

PR #4748, opened 2025-07-14
[UX] Fix focused frame on game sub-menu from being off center 

veryprofessionaldodo:
Noticed the focus on the lists were off-centered. This can be replicated with a keyboard, just clicking down. In the details are the before and after images:

<details>
Before: 

image

After:

image

image

</details>

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.css
+++ b/src/frontend/screens/Game/GamePage/index.css
@@ -506,11 +506,11 @@
       }
 
       &:first-child {
-        margin-top: -0.15rem;
+        margin-top: -0.1rem;
       }
 
       &:last-child {
-        margin-bottom: -0.15rem;
+        margin-bottom: -0.1 rem;
       }
     }
 
--- a/src/frontend/screens/Game/GamePage/index.css
+++ b/src/frontend/screens/Game/GamePage/index.css
@@ -494,7 +494,7 @@
 
     .gameTools .link {
       width: 100%;
-      padding: 0 0 0.5rem;
+      padding: 0.3rem;
       text-align: start;
       white-space: nowrap;
       font-size: 1rem;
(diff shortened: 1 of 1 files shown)


src/frontend/screens/Game/GamePage/index.css
```diff
@@ -494,7 +494,7 @@
 
     .gameTools .link {
       width: 100%;
-      padding: 0 0 0.5rem;
+      padding: 0.3rem;
```
arielj: I think we can just remove this line, this element already has a padding for the `.button.is-link` that sets `0.25em`

src/frontend/screens/Game/GamePage/index.css
```diff
@@ -505,8 +505,12 @@
         color: var(--navbar-active);
       }
 
+      &:first-child {
+        margin-top: -0.1rem;
```
arielj: let's do `0` here instead of a negative margin, I'd rather avoid negative margins as much as possible

src/frontend/screens/Game/GamePage/index.css
```diff
@@ -505,8 +505,12 @@
...
+      }
+
       &:last-child {
-        padding-bottom: 0;
+        margin-bottom: -0.1 rem;
```
arielj: there's a broken syntax here, the space `-0.1 rem` instead of `-0.1rem`, but I'd rather set `0` instead of negative margin

arielj (changes requested):

src/frontend/screens/Game/GamePage/index.css
veryprofessionaldodo: You're right, just checked and removing has no visible effect

src/frontend/screens/Game/GamePage/index.css
veryprofessionaldodo: Understood!

PR #4750, opened 2025-07-15
[UX] Wine Manager - Make refresh button part a tab, clearer keyboard navigation

veryprofessionaldodo:
Noticed a strange behavior when navigating with keyboard on Wine Manager. Tabs cycle between each other, making the refresh button inaccessible, so I would have to click "down" instead of "right" to access the button. Then I could navigate with "left" to the tabs again, trapping the side navigation inside it. 

To solve this I've made the refresh be a part of the tab group, meaning it can now be accessed normally using the arrow keys. Still not a fan of how tabs trap side navigation, it's much less accessible to gamepads that don't have the option of clicking elsewhere in the UI. 

Video showing how it looks below, also inadvertently fixed padding to make button nicer to look at:

<details>
Before:

https://github.com/user-attachments/assets/305e1bf4-c009-4077-b9bf-e933fa0920f6

After: 

https://github.com/user-attachments/assets/f688c7d6-e1db-462d-af5e-c7a2fe1de253

</details>

### DIFF
--- a/src/frontend/screens/WineManager/index.tsx
+++ b/src/frontend/screens/WineManager/index.tsx
@@ -18,6 +18,7 @@ import {
 } from '@fortawesome/free-solid-svg-icons'
 import { WineVersionInfo, Type, WineManagerUISettings } from 'common/types'
 import { hasHelp } from 'frontend/hooks/hasHelp'
+import classNames from 'classnames'
 
 const WineItem = lazy(
   async () => import('frontend/screens/WineManager/components/WineItem')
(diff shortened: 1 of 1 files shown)


arielj (commented):
looks good to me, I guess it's the only fix for this due to how the MUI tabs work

I think it would be good to add a comment in the code right before this new tab tag explaining why it's added as a tab, cause I can imagine the context will be lost eventually and somebody might try to refactor it out.

veryprofessionaldodo:
@arielj good call, adding that comment in and syncing with latest main version

arielj:
thanks!

PR #4754, opened 2025-07-16
Fix error creating folder while installing a wine version

Vidi0:
If a wine version was removed, the versionInfo.installDir variable was set to empty instead of undefined, so trying to install it again would fail while creating the folder. Fix that by checking if versionInfo is undefined OR empty. Now removing and installing the same wine version without closing and opening again heroic works as expected.

---

### DIFF
--- a/src/backend/wine/manager/downloader/main.ts
+++ b/src/backend/wine/manager/downloader/main.ts
@@ -199,7 +199,9 @@ async function installVersion({
   const tarFile =
     installDir + '/' + versionInfo.download.split('/').slice(-1)[0]
   const installSubDir =
-    versionInfo.installDir ?? installDir + '/' + versionInfo.version
+    versionInfo.installDir !== undefined && versionInfo.installDir !== ''
+      ? versionInfo.installDir
+      : installDir + '/' + versionInfo.version
   const sourceChecksum = versionInfo.checksum
     ? (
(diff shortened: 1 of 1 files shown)


src/backend/wine/manager/downloader/main.ts
```diff
@@ -199,7 +199,9 @@ async function installVersion({
   const tarFile =
     installDir + '/' + versionInfo.download.split('/').slice(-1)[0]
   const installSubDir =
-    versionInfo.installDir ?? installDir + '/' + versionInfo.version
+    versionInfo.installDir !== undefined && versionInfo.installDir !== ''
+      ? versionInfo.installDir
+      : installDir + '/' + versionInfo.version
```
veryprofessionaldodo: The only way this change would be different is if `versionInfo.installDir` would be `''`, but in that case this could be simplified to `versionInfo?.installDir ? versionInfo.installDir : installDir + '/' + versionInfo.version`. This would convert a nullish coalescence into a normal falsy check. 

image

Vidi0: The requested changes have been applied. Since versionInfo must exist, versionInfo.installDir is used instead of versionInfo?.installDir

Vidi0:
Simplified as requested

PR #4755, opened 2025-07-16
[UX] Add icon to notifications

arminius-smh:
Hello,

this little change adds the heroic icon to notifications which I think looks nicer and is more what I expected to see, unless there is a specific reason why this wasn't done before that I'm not aware of?

before:
2025-07-16_19-36-12_screenshot

now:
2025-07-16_19-17-36_screenshot

---

### DIFF
--- a/src/backend/dialog/dialog.ts
+++ b/src/backend/dialog/dialog.ts
@@ -4,6 +4,7 @@ import { ButtonOptions, DialogType } from 'common/types'
 import { getMainWindow } from '../main_window'
 import { sendFrontendMessage } from '../ipc'
 import { isSteamDeckGameMode } from 'backend/constants/environment'
+import { windowIcon } from 'backend/constants/paths'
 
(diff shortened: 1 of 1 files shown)


flavioislima:

arminius-smh:

arielj:
I wonder how it looks in different systems, cause for linux mint with cinnamon it looks kinda ugly with the icon being too small

image

maybe we need different icons? or maybe it's better if it's SVG so it scales better?

arminius-smh:
Oh, argh the notification uses electrons `nativeImage` which only supports png or jpeg :/

Seems like https://github.com/electron/electron/issues/9642 would be the blocker
Since the icon size can just be user configured, I don't think it's possible to properly scale this beforehand

2025-07-17_10-10-06_screenshot

^This would be a notification with svg and icon size of 24x24

Sorry, should have tested it with smaller images, I guess I'll close this since I can't think of a way to do this otherwise

PR #4759, opened 2025-07-17
[TECH] Add Plausible for basic data insights

flavioislima:
Adds basic integration with open source Plausible Analytics for 100% anonymous data colection just to give us an idea on the number of users, operating systems and countries.

More info:
https://plausible.io/

- Opt-in: A dialog is shown on startup so people can enable/disabled it.
- A new setting was also added on the general settings.
- Collected data is shown on the logs.

image
Screenshot 2025-07-17 at 17 45 32

---


src/backend/config.ts
```diff
@@ -318,6 +318,7 @@ class GlobalConfigV0 extends GlobalConfig {
     const defaultWine: WineInstallation = isWindows ? {} : getDefaultWine()
 
     const settings: Partial<AppSettings> = {
+      analyticsOptIn: true,
```
arielj: this should be false to be really opt in, now it's actually opt out cause it's enabled by default and closing the dialog keeps it enabled

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "heroic",
-  "version": "2.17.2",
+  "version": "2.18.0",
   "versionNames": {
-    "stable": "Cutty Flam Franky",
+    "stable": "\"Waterfall Beard\" Jorul",
```
arielj, line 5: we should do this version bump in a different PR

src/frontend/screens/Settings/components/AnalyticsDialog.tsx
```diff
@@ -0,0 +1,82 @@
...
+          <>
+            {t(
+              'analyticsModal.info.pt1',
+              'In order to improve the app, Heroic collects 100% anonymous data.'
+            )}
+            <ul>
+              <li>
+                {t(
+                  'analyticsModal.info.pt2',
+                  'Heroic uses the open-source Plausible Analytics platform to gather basic data like: App Version, OS, Stores and Country.'
```
arielj:
```suggestion
                  'Heroic uses the open-source Plausible Analytics platform to gather basic data: App Version, OS, Stores and Country.'
```

to make it clear it's only that and nothing more maybe? because `like ...` sounds like there are other things that we are not listing

src/frontend/screens/Settings/components/AnalyticsDialog.tsx
```diff
@@ -0,0 +1,82 @@
...
+              <li>
+                {t(
+                  'analyticsModal.info.pt3',
+                  'It will never collect any personal information, including your username, IP address or email.'
+                )}
+              </li>
+              <li>
+                {t(
+                  'analyticsModal.info.pt4',
+                  'This data is used to gives insights on what to focus on next due to our limited resources and user feedback.'
```
arielj:
```suggestion
                  'This data is used to give us insights on what to focus on next due to our limited resources and user feedback.'
```

src/backend/main.ts
```diff
@@ -360,6 +361,29 @@ if (!gotTheLock) {
...
     const settings = GlobalConfig.get().getSettings()
 
+    if (settings && settings.analyticsOptIn === true) {
+      const plausible = Plausible()
+      plausible.enableAutoPageviews()
+      const appVersion = app.getVersion()
+      const providersObject = {
+        gog: !!GOGUser.isLoggedIn(),
+        epic: !!LegendaryUser.isLoggedIn(),
+        amazon: !!NileUser.isLoggedIn()
```
arielj: maybe we can include `sideload` here and count if the number of sideloaded games is > 0? to know how much this feature is used too? maybe even divide by games/urls to have that distinction?

arielj (changes requested):
because of the popup showing up on top of the app, many e2e tests are failing (playwright can't actually click elements)

also the e2e tests I think are sending stats? we should probably disable this completely in CI

src/backend/config.ts
flavioislima: Yeah, I will change the dialog. But I think we should still at least track tha the person declined so we can get the number of users that disable it to sum to the total numbers.

package.json
flavioislima: I normally do a version bump on any PR that will go into the release, unless when I forget.

flavioislima:
> because of the popup showing up on top of the app, many e2e tests are failing (playwright can't click elements behind)
> 
> also I think the e2e tests are sending stats? we should probably disable this completely when running e2e tests, to fix the tests and not mess up the stats with every run

should be fixed now.

public/locales/en/translation.json
```diff
@@ -139,6 +140,8 @@
             "exe": "Select EXE",
             "script": "Select script ..."
         },
+        "shaderPreCachingDisabledMessage": "Steam's Shader Pre-cache is disabled. Please enable it on the Steam Settings in Desktop Mode to ensure the game works properly with UMU.",
```
flavioislima, line 143: This was removed and now added again after merging main so there were some confusion somewhere after some merge.

src/frontend/screens/Settings/components/AnalyticsOptIn.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import { ToggleSwitch } from 'frontend/components/UI'
+import useSetting from 'frontend/hooks/useSetting'
+import { useTranslation } from 'react-i18next'
+import InfoIcon from 'frontend/components/UI/InfoIcon'
+
+const AnalyticsOptIn = () => {
+  const { t } = useTranslation()
+  const [analyticsOptIn, setAnalyticsOptIn] = useSetting('analyticsOptIn', true)
```
arielj:
```suggestion
  const [analyticsOptIn, setAnalyticsOptIn] = useSetting('analyticsOptIn', false)
```

to match the default in config.ts

src/frontend/screens/Settings/components/AnalyticsOptIn.tsx
```diff
@@ -0,0 +1,38 @@
...
+        handleChange={() => setAnalyticsOptIn(!analyticsOptIn)}
+        title={t(
+          'setting.analyticsOptIn',
+          'Send anonymous data to help Heroic development'
+        )}
+      />
+      <InfoIcon
+        text={t(
+          'help.analytics',
+          'Enables Heroic to collect 100% anonymous usage data to help improve the application. Needs restart to take effect.'
```
arielj, line 28: I wonder if adding a check for the `analyticsOptIn` setting in `sendPlausible` function can be used to not need a restart?

src/frontend/screens/Settings/components/AnalyticsOptIn.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { useTranslation } from 'react-i18next'
+import InfoIcon from 'frontend/components/UI/InfoIcon'
+
+const AnalyticsOptIn = () => {
+  const { t } = useTranslation()
+  const [analyticsOptIn, setAnalyticsOptIn] = useSetting('analyticsOptIn', true)
+
+  React.useEffect(() => {
+    if (analyticsOptIn) {
+      window.api.startPlausible()
```
arielj: just an idea, we already have backend events firing when settings are changed that we can listen to https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/5d66738a17248022b17385e00de67fb40734a3a7/src/backend/logger/index.ts#L94

then we can simplify this and not need the api call (ipc/types/all that), we can listen to the `settingChanged` event and check the key to know if it's being enabled or disabled

src/frontend/screens/Settings/components/AnalyticsOptIn.tsx
flavioislima: It only needs a restart when disabling. Enabling is easier but killing the connection is not that much.

src/frontend/screens/Settings/components/AnalyticsOptIn.tsx
flavioislima: Good point, just made the change

arielj (approved):
looks good to me

PR #4767, opened 2025-07-18
[umu] Remove disabled shader pre-cache warning dialog and logs

arielj:
Since umu released a new version removing the requirement of Steam's Shader Pre-Cache on the Steamdeck's gaming mode, we can remove all this.

It was removed before the umu release, then I re-added it just in case we made a release before umu, but now we can remove it again for good.

I can confirm it's not needed anymore because I tested a game on my steamdeck with the pre-cache disabled and I get the warning and clicking `Launch Anyway` continues and it works just fine, no issues of the window hidden behind the steam animation.

---


flavioislima (approved):
LGTM
Ideally we should not touch the translations file besides the engligh one to avoid conflicts on weblate. But I merged the weblate PR already so should be fine.

PR #4771, opened 2025-07-20
Adding an Humble bundle integration 

alex-min:
Adding a new Humble Bundle integration

Capture d’écran du 2025-07-20 15-19-49
Capture d’écran du 2025-07-20 15-19-56
Capture d’écran du 2025-07-20 15-20-33

Limitations:
- Tested on Linux & Windows , I don't have a mac to test it, it probably works same as Linux.
- The humble bundle subscription isn't supported, I would need the native humble bundle windows store for those.   
- Since the humble games aren't coming from a real store like gog/epic, the code is using heuristics to determine which is the best executable or shortcut, this can fail.
- no automated tests for now, I think we could add it on another PR as it's pretty large already maybe
- some features aren't unsupported yet like updating the games.
- only humble windows games are supported for now (either natively or under wine)

---


flavioislima:

alex-min:

flavioislima:
Can you add the humble icon here?
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/8de24c757244b7f18fd45becfc35d2254e7a1b7c/src/frontend/components/UI/StoreLogos/index.tsx#L10

package.json
```diff
@@ -103,6 +103,7 @@
     "source-map-support": "^0.5.21",
     "steam-shortcut-editor": "^3.1.3",
     "tslib": "^2.8.1",
+    "unzipper": "^0.12.3",
```
arielj, line 106: do we need this dependency? I know we unarchive files in other places like when we download wine/proton, can't we unzip without an extra dependency?

jntesteves:
The API seems to return duplicates, probably because the same game was part of several bundles I bought. The most extreme example is Braid in this picture which seems to have come in 3 bundles I've purchased, and there still are other duplicates in my library.

image

Also, this is missing a library filter for the Humble Store. I can only see these games when disabling all filters, which makes it hard to find the games.

image


Last, IDK if this issue is due to the changes in this PR, but the library search is very broken, returning several times the same game and even some empty entries. The results are not being removed when the search is cleared, so the library state gets more and more broken the more I type in the search field.

image

hifron:
There is no public API https://www.humblebundle.com/developer and maybe problem on their side if not API from widget, but listing only 1 game or selecting version from which bundle seems possible, but somehow user problematic. HB is now prominent corporation and this could be their problem or store access problem inquires.

Maybe could be closed as not planned until further digging into their API request tickets or sending some mails to make it more clear if this is supported and make it proper way according their wish and not make users some harrasment mismatch of their libs, but HB has some orientation in [HB user library](https://www.humblebundle.com/home/library) itself...

flavioislima:
@alex-min changing this to WIP for now. Let us know when it can be tested again and also if you need help with something else related with this.

neuromancer:
Is anyone working on this feature?

PR #4780, opened 2025-07-22
Fix error while collecting Plausible data

Vidi0:
In the current main, when accepting the analytics pop-up on a clean install, an error occurs:
Plausible fix
This happened because libraryStore.raw_store.games is undefined, and trying to access its length property fails. Fix that by accessing the property using the "?." operator instead of "."

---


flavioislima (approved):
Good catch, thanks for the fix.

PR #4799, opened 2025-07-25
Change "Allow non-GE Proton" to only hide Valve Proton versions

### LINKED ISSUE #4798: Unhide by default other Proton versions besides of GE Proton
### Problem description

Since version 2.18, a change has been made where all Proton versions except the GE version are not visible by default, which is confusing to many people. Many users, including myself, use other versions of Proton, such as the one from CachyOS or Valve's Proton Experimental, and want to use it in HGL by default.

Furthermore, the current setting favors GE-Proton, which, as many players have learned, isn't the best and causes many problems with games.

### Feature description

Restore the default behavior of HGL, where all Proton versions are visible, and stop favoring the Glorious Eggroll version. This creates unnecessary confusion when users want to use a different version of Proton.

### Alternatives

_No response_

### Additional information

_No response_

CommandMC:
Instead of hiding all custom Proton versions instead of GE's, we can just not include Valve's versions (since those are the only ones with issues; -CachyOS, -Sarek and -EM all have umu support)

Closes #4798

---

### DIFF
--- a/src/backend/utils/compatibility_layers.ts
+++ b/src/backend/utils/compatibility_layers.ts
@@ -152,10 +152,14 @@ export async function getLinuxWineSet(
 
   const protonPaths = [`${toolsPath}/proton/`]
 
+  const { showValveProton } = GlobalConfig.get().getSettings()
+
   await getSteamLibraries().then((libs) => {
     libs.forEach((path) => {
-      protonPaths.push(`${path}/steam/steamapps/common`)
-      protonPaths.push(`${path}/steamapps/common`)
+      if (showValveProton) {
+        protonPaths.push(`${path}/steam/steamapps/common`)
+        protonPaths.push(`${path}/steamapps/common`)
+      }
       protonPaths.push(`${path}/root/compatibilitytools.d`)
       protonPaths.push(`${path}/compatibilitytools.d`)
       return
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -617,17 +617,6 @@
             },
             "label": "Allow installation of games with broken or denied anticheat"
         },
-        "allow_non_ge_proton": {
-            "confirmation": {
-                "dont_show": "Don't show them",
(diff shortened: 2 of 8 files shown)


src/backend/utils/compatibility_layers.ts
```diff
@@ -152,10 +152,14 @@ export async function getLinuxWineSet(
...
+  const { showValveProton } = GlobalConfig.get().getSettings()
+
   await getSteamLibraries().then((libs) => {
     libs.forEach((path) => {
-      protonPaths.push(`${path}/steam/steamapps/common`)
-      protonPaths.push(`${path}/steamapps/common`)
+      if (showValveProton) {
+        protonPaths.push(`${path}/steam/steamapps/common`)
+        protonPaths.push(`${path}/steamapps/common`)
+      }
```
arielj, line 162: the reason I was doing this filtering in the frontend was to not end up with an empty wine selector

if the filter happens in the backend and a user already has Proton configured for a game, the wine selector won't have that version and it will show the empty select, making it look like there's some issue

Try this:
- check the option to show valve's protons
- select that for a game
- then uncheck the option to hide them
- now open the game's settings
- the wine selector is blank, users will think the game won't run and if it was configured already it was probably working

CommandMC: I don't really see why it shouldn't work like that. Isn't an empty select a good nudge towards the user, to get them to choose another version?

arielj: but if they are already using proton for a game which is already working we don't want to make them change it, right? if it's working already they'll think they have to pick another proton and maybe the game stops working (I know for example Genshin Impact requires Proton 10 because it doesn't work with GE-Proton https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Anime-Games#genshin-impact)

CommandMC: I've now changed the selector to also add the currently set version to the list if it's still valid. That should resolve the issue while still making the two options independent

PR #4802, opened 2025-07-26
Nix package supportability

aidalgol:
Some fixes to the Nix dev shell, NixOS information in the README, and add log messages and startup warnings to try addressing the volume of support issues for the unofficial nix package.

@CommandMC 

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -130,7 +130,6 @@ import {
   isIntelMac,
   isLinux,
   isMac,
-  isNixPkg,
   isSnap,
   isSteamDeckGameMode,
   isWindows
@@ -492,7 +491,7 @@ addOneTimeListener('frontendReady', () => {
     }
   }
 
-  if (isNixPkg) {
+  if (process.env.NIX_PKG) {
     const nixPkgWarning: Electron.MessageBoxOptions = {
       title: i18next.t(
         'box.warning.nixpkg.title',
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -528,11 +528,11 @@ addOneTimeListener('frontendReady', () => {
       const nixUmuWarning: Electron.MessageBoxOptions = {
         title: i18next.t(
           'box.warning.nixumu.title',
-          'UMU running from the nix package'
+          'umu running from the nix package'
         ),
         message: i18next.t('box.warning.nixumu.message', {
           defaultValue:
-            'If you experience issues launching games using UMU, please first check that it also happens with UMU managed by Heroic, after removing UMU from your system environment.'
+            'If you experience issues launching games using umu, please first check that it also happens with umu managed by Heroic, after removing umu from your system environment.'
         }),
         checkboxLabel: i18next.t('box.warning.nixumu.checkbox', {
           defaultValue: 'Do not show this message again'
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -130,6 +130,7 @@ import {
   isIntelMac,
   isLinux,
   isMac,
+  isNixPkg,
   isSnap,
   isSteamDeckGameMode,
   isWindows
@@ -143,6 +144,7 @@ import {
   windowIcon
 } from './constants/paths'
 import { supportedLanguages } from 'common/languages'
+import { usingNixUmu } from './utils/compatibility_layers'
 import MigrationSystem from './migration'
 
 app.commandLine?.appendSwitch('ozone-platform-hint', 'auto')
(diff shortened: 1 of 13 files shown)


flavioislima:

aidalgol:

src/backend/utils/compatibility_layers.ts
```diff
@@ -562,6 +563,18 @@ export async function getWineFlagsArray(
...
+  return searchForExecutableOnPath('umu-run').then((path) => {
+    if (path) {
+      return false
+    } else {
+      return realpath(path as string).then((path) =>
+        path.startsWith('/nix/store/')
+      )
+    }
+  })
+}
```
CommandMC: This function doesn't look right
- If a path is found, you're always returning `false`
- If a path isn't found, you treat the returned `null` as a string and try to call `realpath` on it?

Perhaps there's a missing `!` in the condition?

In general I'd use `await` here to make this more readable, along with the existing `getUmuPath`:
```ts
  const umuPath = await getUmuPath()
  const realUmuPath = await realpath(umuPath)
  return realUmuPath.startsWith('/nix/store/')
```

I'm also not sure whether `realpath` is necessary here, but I assume you've confirmed that

src/backend/utils/systeminfo/index.ts
```diff
@@ -107,12 +111,14 @@ async function getSystemInfo(cache = true): Promise<SystemInformation> {
     },
     steamDeckInfo: deckInfo,
     isFlatpak: !!process.env.FLATPAK_ID,
+    isNixPkg: Boolean(process.env.NIX_PKG),
```
CommandMC: Not sure how I feel about having this in the system info, considering the package isn't official in the first place. Perhaps remove it here & just check for `process.env.NIX_PKG` in your change to `main.ts`?

src/backend/utils/systeminfo/index.ts
```diff
@@ -137,6 +143,8 @@ The current system is${info.steamDeckInfo.isDeck ? '' : ' not'} a Steam Deck${
       : ''
   }
 We are${info.isFlatpak ? '' : ' not'} running inside a Flatpak container
+We are${info.isNixPkg ? '' : ' not'} running the community nix package
```
CommandMC: Adding to my last comment: Even if we do keep the bool in `SystemInformation`, I don't think it's necessary to be printed here. If you want to make it obvious in the log that we're on NixOS, that's already visible in the "OS" entry

src/backend/main.ts
```diff
@@ -490,6 +492,71 @@ addOneTimeListener('frontendReady', () => {
...
+  usingNixUmu().then((result) => {
+    if (result) {
+      const nixUmuWarning: Electron.MessageBoxOptions = {
+        title: i18next.t(
+          'box.warning.nixumu.title',
+          'UMU running from the nix package'
+        ),
+        message: i18next.t('box.warning.nixumu.message', {
+          defaultValue:
+            'If you experience issues launching games using UMU, please first check that it also happens with UMU managed by Heroic, after removing UMU from your system environment.'
```
CommandMC: "umu" is not an acronym (we're probably making this mistake somewhere else in the codebase still, but no need to repeat it)
```suggestion
            'If you experience issues launching games using umu, please first check that it also happens with umu managed by Heroic, after removing umu from your system environment.'
```
(the title also needs changing)

CommandMC (changes requested):

src/backend/utils/compatibility_layers.ts
aidalgol: Fixed.  We do need `realpath` because `umu-run` is likely to be a symlink on NixOS.

src/backend/main.ts
aidalgol: Fixed.  (It is all caps in the UI, which is why I made this mistake.)

src/backend/utils/systeminfo/index.ts
aidalgol: Knowing that we are in NixOS does not tell us, on its own, that we are running the nix package, because we could be running the AppImage.  How about instead log a message like we do for Flatpak, but for AppImage?

arielj:
hmmm most of these things sound like they belong to a fork of heroic, not to this repo

aidalgol:
I submitted them here after discussing with @CommandMC.  I'm happy to instead maintain them as patches for the nix package instead if that is what the core team would prefer.  I think logging whether running the appimage and the path to `umu-run` would be useful to have in mainline though.

aidalgol:
Closing in favour of #4866 and #4867.

PR #4803, opened 2025-07-26
[Tech] Clean up MUI theming

CommandMC:
We can use MUI's own theming functionality to set the font family & colors to use once. Doing it like this means MUI components will look right everywhere

There are probably more rules that can be removed now. I wasn't sure on how to test most of them, so I just left them alone

---


CommandMC:
I thought that was confusing as well. FWIW other components also have this set - didn't want to touch those in this PR, but we might want to remove it from all of them

PR #4804, opened 2025-07-26
[FIX] Inputs onChange callback bug

arielj:
In the latest release I introduced a bug, I tried to fix it here https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4787 but it was not the right fix, it was just working around the actual problem.

The issue was also affecting the Gamescope settings for example.

So, this is the correct fix, I forgot to update the event listener when the `onChange` prop changed, this was the actual fix.

---

### DIFF
--- a/src/frontend/components/UI/TextInputField/index.tsx
+++ b/src/frontend/components/UI/TextInputField/index.tsx
@@ -45,7 +45,7 @@ const TextInputField = ({
       }
     }
     return
-  }, [input])
+  }, [input, onChange])
 
   return (
     <div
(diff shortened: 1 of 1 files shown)


src/frontend/components/UI/TextInputField/index.tsx
```diff
@@ -45,7 +45,7 @@ const TextInputField = ({
       }
     }
     return
-  }, [input])
+  }, [input, onChange])
```
CommandMC: While you're at it, seems like this is also the case:
```suggestion
  }, [input.current, onChange])
```

arielj: looks like neither `input` nor `input.current` should be added to useEffect dependencies

https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array

CommandMC: Huh, TIL

PR #4808, opened 2025-07-26
[Tech] Simplify UpdateComponent

arielj:
This PR removes a lot of code for the UpdateComponent component.

The main idea is to remove all the complexity of the IntersectionObserver that was there to try to size and center the component and we can use CSS's features instead that are more reliable.

It also removes a lot of the extra CSS and props and HTML elements and code that is not really needed anymore.

I tested all the loading states (if you want to try it you'll have to force loading states in every place by changing the code, because it happens just too fast in most cases), we even had some pages that were broken, like the loading state of the game details page, or the loading state of the webview screen.

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.css
+++ b/src/frontend/screens/Game/GamePage/index.css
@@ -17,6 +17,11 @@
   align-items: flex-start;
   align-content: flex-start;
 
+  &:has(.UpdateComponent) {
+    grid-template-columns: 1fr;
+    grid-template-rows: 1fr;
+  }
+
   /*
   ** Adding icons to the MUI tabs makes them at least 72px high
   ** min-height is guaranteed by the min height of the tab parent element itself
--- a/src/frontend/screens/Game/ModifyInstallModal/GOG/index.tsx
+++ b/src/frontend/screens/Game/ModifyInstallModal/GOG/index.tsx
@@ -419,6 +419,6 @@ export default function GOGModifyInstallModal({
       </button>
     </>
   ) : (
-    <UpdateComponent inline />
+    <UpdateComponent />
   )
 }
--- a/src/frontend/screens/Login/index.tsx
+++ b/src/frontend/screens/Login/index.tsx
@@ -75,13 +75,12 @@ export default React.memo(function NewLogin() {
     navigate('/')
   }
 
+  if (loading) {
+    return <UpdateComponent />
+  }
+
   return (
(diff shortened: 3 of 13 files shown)


src/frontend/screens/Game/GamePage/index.css
```diff
@@ -17,6 +17,11 @@
   align-items: flex-start;
   align-content: flex-start;
 
+  &:has(.UpdateComponent) {
+    grid-template-columns: 1fr;
+    grid-template-rows: 1fr;
+  }
```
arielj, line 23: we need this or the page still shows a 2x2 grid with the component in the top/left cell

src/frontend/screens/Game/ModifyInstallModal/GOG/index.tsx
```diff
@@ -419,6 +419,6 @@ export default function GOGModifyInstallModal({
       </button>
     </>
   ) : (
-    <UpdateComponent inline />
+    <UpdateComponent />
```
arielj, line 422: no need for toplevel vs inline distinction

src/frontend/screens/Login/index.tsx
```diff
@@ -75,13 +75,12 @@ export default React.memo(function NewLogin() {
     navigate('/')
   }
 
+  if (loading) {
+    return <UpdateComponent />
+  }
```
arielj, line 80: this was inside the other `return` value and was rendering the loading along with other elements which looked really broken

src/frontend/screens/WebView/index.css
```diff
@@ -1,27 +1,32 @@
...
+
+  .WebView__webview {
+    height: calc(100vh - 40px);
+  }
 
-.WebView .login-warning > div {
-  background: black;
-  align-self: center;
-  padding: var(--space-xl);
+  &:has(.UpdateComponent) .WebView__webview {
```
arielj, line 29: this is the only new code, the rest is just the previous code moved things to be nested (it's easier to see nesting than read the same selectors)

src/frontend/screens/WineManager/index.css
```diff
@@ -1,4 +1,7 @@
 .wineManager {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
```
arielj, line 4: the loading element in this page was always adding vertical scroll, using flex to avoid that

PR #4809, opened 2025-07-26
[DX] Initial setup for frontend unit testing

arielj:
This PR adds some setup so we can start writing frontend unit tests.

We only have the setup for backend unit tests and we are only testing a bit of the frontend with e2e tests, but it's really hard to setup e2e tests to test most of the components because it's not easy to setup all the stubs to get to the required states.

This PR includes the initial setup and an example to write frontend tests with both global state and settings state.

There are things that we'll need to move somewhere else and make more generic, but we can deal with that as needed.

This is made on top of `refactor-update-component-height` because `jsdom` does not support IntersectionObserver. It also includes the fix from https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4804 as a way to check that the tests are actually working properly and could detect issues like that in the future.

---


package.json
```diff
@@ -142,10 +142,13 @@
     "fast-xml-parser": "^4.5.3",
     "husky": "^8.0.3",
     "i18next-parser": "^9.3.0",
+    "identity-obj-proxy": "^3.0.0",
```
arielj, line 136: we need this so we can import scss files in components in tests

package.json
```diff
@@ -142,10 +142,13 @@
...
     "jest": "^29.7.0",
+    "jest-environment-jsdom": "^30.0.5",
```
arielj, line 138: we need this to test tsx/jsx components

package.json
```diff
@@ -142,10 +142,13 @@
...
     "node-gyp": "^10.3.1",
     "prettier": "^3.5.3",
     "react-devtools": "^5.3.2",
+    "react-test-renderer": "^19.1.0",
```
arielj, line 142: we need this for the renderer/screen methods for react tests

src/frontend/components/UI/TextInputField/index.tsx
```diff
@@ -45,7 +45,7 @@ const TextInputField = ({
       }
     }
     return
-  }, [input])
+  }, [input, onChange])
```
arielj: you can try removing this change to see how the frontend test fails

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -1,15 +1,13 @@
 import { ChangeEvent, useContext, useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import {
-  SelectField,
-  TextInputField,
-  ToggleSwitch,
-  UpdateComponent
-} from 'frontend/components/UI'
```
arielj, line 8: we have a problem with this pattern, importing parts of the `UI/index.ts` file makes it load all other components even if not imported by this component, this is a problem during unit tests because other components may require extra `window.api` function that we don't really need for the component we want to test

we have to move out of this `index.ts` collection of exports and use individual imports for each components so it's `unit-test-friendly`

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -37,6 +35,8 @@ const Gamescope = () => {
   )
 
   useEffect(() => {
+    if (!isLinux) return
```
arielj, line 38: we were checking the location of the `gamescope` binary for windows and mac, unnecessary and the test helped identify this issue

src/frontend/screens/Settings/components/__tests__/Gamescope.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { render, screen } from '@testing-library/react'
+import userEvents from '@testing-library/user-event'
+import '@testing-library/jest-dom/extend-expect'
+import Gamescope from '../Gamescope'
+import ContextProvider from 'frontend/state/ContextProvider'
+import { ContextType, SettingsContextType } from 'frontend/types'
+import SettingsContext from '../../SettingsContext'
+
+// TODO: Move this somewhere else once we make it more generic
+const SettingsWrapper = (props: React.PropsWithChildren) => {
```
arielj, line 11: in the future we can probably move this somewhere else and use a factory/generator to create the wrapper with the desired initial state

Issue #4783, opened 2025-07-23
Launching the exe while Heroic is already running gives a JavaScript error


Franzyd:
### Describe the bug

If I launch the program's .exe Heroic while it's running it will show up normally but also with a Javascript error window (image attached).

### Add logs

```text
(13:59:22) [INFO]:    [Connection]:      Connectivity: check-online

(13:59:22) [INFO]:    [Backend]:         System Information: CPU: 12x Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz

Memory: 17.12 GB (used: 7.16 GB)

GPUs:

  GPU 0:

    Name: NVIDIA Corporation TU104 [GeForce RTX 2080]

    IDs: D=1e82 V=10de SD=3722 SV=1462

    Driver: 577.00

OS: Microsoft Windows 10 Pro 10.0.19045 (win32)

The current system is not a Steam Deck

We are not running inside a Flatpak container

Software Versions:

  Heroic: 2.18.0 "Waterfall Beard" Jorul

  Legendary: 0.20.37 Exit 17 (Heroic)

  gogdl: 1.1.2

  comet: comet 0.2.0

  Nile: 1.1.2 Will A. Zeppeli

(13:59:22) [INFO]:    [Backend]:         Starting Plausible Analytics

(13:59:22) [INFO]:    [Backend]:         Shared Data: {"version":"2.18.0","gog":true,"epic":true,"amazon":false,"providers":"gog, epic","OS":"win32","isFlatpak":false,"isAppImage":false,"isSnap":false,"isSteamDeckGameMode":false,"isSteamDeck":false}

(13:59:22) [INFO]:    [Connection]:      Pinging external endpoints

(13:59:22) [INFO]:    [Gog]:             Checking for existing gog manifests

(13:59:22) [WARNING]: [Backend]:         Protocol already registered.

(13:59:23) [INFO]:    [Gog]:             Checking if login is valid

[...]

PR #4812, opened 2025-07-28
[FIX] Fix logger error opening second instance of Heroic

arielj:
This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4783

There's an issue on windows (and I think mac too) when opening a second instance of Heroic (from a shortcut for example)

In linux this is not an issue because it uses the `heroic://` protocol, but in window it opens the second instance and detects the previous one running, then it tries to write to the logs which are not initialized yet for that instance.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -312,8 +312,7 @@ const processZoomForScreen = (zoomFactor: number) => {
 }
 
 if (!gotTheLock) {
-  initLogger()
-  logInfo('Heroic is already running, quitting this instance')
+  console.log('Heroic is already running, quitting this instance')
   app.quit()
 } else {
   app.on('second-instance', (event, argv) => {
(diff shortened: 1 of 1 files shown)


CommandMC:
I'm wondering whether we want to do this at all, since initializing the logger & writing a message moves the current log file to `heroic.log.old`, which the other instance then won't know about
Maybe just removing the `logInfo` would be better

arielj:
> I'm wondering whether we want to do this at all, since initializing the logger & writing a message moves the current log file to `heroic.log.old`, which the other instance then won't know about Maybe just removing the `logInfo` would be better

maybe I can change this to a `console.log(....)`? so if someone runs it in a terminal they get some output about why it's getting closed?

I think it will still keep the feedback about why it's closing and not create issues with logs?

CommandMC:
Yeah, that's a good middle ground here

CommandMC (approved):

Issue #4811, opened 2025-07-28
Clicking ‘Go Back’ after viewing a game should return you to where you were, not the top of the page


paparoxo360:
### Problem description

After clicking on a game and then hitting the 'Go Back' arrow, the page returns to the top instead of where you were. This can be a bit annoying, especially if you were near the bottom of the page.

### Feature description

Instead of returning to the top of the page, it should return to where you were before clicking.

### Additional information

First, I want to say how great this software is - huge thanks to all the developers for your amazing work and for constantly improving it with new features. My request is just a small nitpick, but I think it would be a nice addition. Thanks again for all the hard work.

arielj:
hmmm this used to be a thing, looks like it broke at some point

PR #4814, opened 2025-07-28
[FIX] Remember scroll position of library when changing screens

arielj:
This fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4811

We used to have this, but it broke at some point. We scroll the body element, not the window now.

---


CommandMC (approved):
What a strange issue. I can confirm that `window` is no longer working, although I don't understand why

PR #4818, opened 2025-07-28
[FIX] Add SPDX license to AUR release

HurricanePootis:
The AUR package `heroic-games-launcher-bin` is missing `depends()` and has a non-compliant `license()` header. I didn't know if y'all specify `GPL-3.0-or-later` or `GPL-3.0-only`, so I chose `GPL-3.0-or-later` since it is more common to see than not.

The `depends()` were found using `namcap`, which is a great tool for anyone writing PKGBUILDs on Arch.
---

### DIFF
--- a/.github/workflows/release_aur.yml
+++ b/.github/workflows/release_aur.yml
@@ -47,7 +47,7 @@ jobs:
           pkgdesc="An Open source Launcher for Epic, Amazon and GOG Games"
           arch=('x86_64')
           url="$url"
-          license=('GPL3')
+          license=('GPL-3.0-or-later')
           _filename=$_filename
           source=("$source")
           noextract=("$_filename")


flavioislima:

HurricanePootis:

roland-5:
Noob question: Are those really necessary dependencies needed to add to PKGBUILD? Like systemd-libs or glibc - you can't use Arch without them - isn't it redundant to add something like that?

HurricanePootis:
I guess so, but I was just adding everything `namcap` reported. If that's a problem, I'm more than happy to remove glibc and systemd-libs

flavioislima:
I am not sure about this PR. We added dependencies like this in the past and then for some people this broke their systems because sometimes they use specific versions of a package and they might not be compatible. The license is fine but I think we should skip the depends part, especially since we have no reports of people having issues to run the app on arch due to a missed dep.

roland-5:
> I am not sure about this PR. We added dependencies like this in the past and then for some people this broke their systems because sometimes they use specific versions of a package and they might not be compatible. The license is fine but I think we should skip the depends part, especially since we have no reports of people having issues to run the app on arch due to a missed dep.

I agree with this.

HurricanePootis:
> I am not sure about this PR. We added dependencies like this in the past and then for some people this broke their systems because sometimes they use specific versions of a package and they might not be compatible. The license is fine but I think we should skip the depends part, especially since we have no reports of people having issues to run the app on arch due to a missed dep.

Done

Issue #4792, opened 2025-07-24
Library does not get updated when changing categories of a game


arielj:
### Describe the bug

When creating a category or adding categories to game, the library does not get updated

### Add logs

```text
nothing useful in logs, it's all frontend
```

### Steps to reproduce

1. Go to the Library
2. Right click on a game > Categories
3. Create a category
4. Close the dialog
5. The `Categories` dropdown does not show the new category, we have to go to another screen and come back

another way:
1. Go to the Library
2. Configure the Categories filter to not show a category
3. Right click on a game > Categories
4. Assign that hidden category for the game
5. Close the dialog
6. The game is still displayed, we have to go to another screen and come back

### Expected behavior

New categories should update the Categories selector immediately.

Categories assigned to a game should apply immediately (if the category is hidden, the game should disappear)

### Heroic Version

Latest Stable

### System Information

any system

PR #4829, opened 2025-07-31
[UI] Update library when changing games' categories

arielj:
After moving the `Settings` modal to Zustand, we removed a lot of extra re-renders. This uncovered an issue we had with the Categories dialog that was not updating the state properly to trigger a rerender of the library (it was working because of the other extra rerenders)

This PR fixes that by copying the custom categories variables and the specific list of games for the given category, changing the memory reference so react knows it changed.

This PR fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4792

---

### DIFF
--- a/src/frontend/state/GlobalState.tsx
+++ b/src/frontend/state/GlobalState.tsx
@@ -395,11 +395,11 @@ class GlobalState extends PureComponent<Props> {
   }
 
   addGameToCustomCategory = (category: string, appName: string) => {
-    const newCustomCategories = this.state.customCategories
+    const newCustomCategories = { ...this.state.customCategories }
+    const currentGamesInCategory = newCustomCategories[category] || []
 
-    if (!newCustomCategories[category]) newCustomCategories[category] = []
-
-    newCustomCategories[category].push(appName)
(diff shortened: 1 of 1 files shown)


src/frontend/state/GlobalState.tsx
```diff
@@ -395,11 +395,11 @@ class GlobalState extends PureComponent<Props> {
   }
 
   addGameToCustomCategory = (category: string, appName: string) => {
-    const newCustomCategories = this.state.customCategories
-
-    if (!newCustomCategories[category]) newCustomCategories[category] = []
+    const newCustomCategories = { ...this.state.customCategories }
```
CommandMC: Wouldn't this line's change alone be enough to resolve the issue? I understand the issue is object equality related, so creating a new object to hold the categories should result in a re-render, no?

arielj: I'll try again, I think I tried that first and it was not enough

arielj: I guess I tested it wrong haha, updated it to just change the new categories object and not the values references

CommandMC (approved):

PR #4830, opened 2025-08-01
[Feat] (Auto-)Updater improvements

CommandMC:
1. We can safely enable the auto-updater for AppImage setups, since those (usually) live in a user-writable location
2. If we don't have the auto-updater, we still want to notify the user about the update being available

Can't really test this

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -136,8 +136,7 @@ import {
   isMac,
   isSnap,
   isSteamDeckGameMode,
-  isWindows,
-  autoUpdateSupported
+  isWindows
 } from './constants/environment'
 import {
   configPath,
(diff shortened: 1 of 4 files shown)


arielj:
> Can't really test this

I think what can be done to test it is:
- make one commit downgrading the version (here or another branch)
- get the appImage from github
- it should autoupdate I think?

CommandMC:
Did it somewhat like that (basically like your suggestion, except that I just built locally)
Can confirm that both messages (auto-update message on AppImage, and update message on other Linux packages) are appearing & working properly

src/backend/main.ts
```diff
@@ -252,7 +252,7 @@ async function initializeWindow(): Promise<BrowserWindow> {
   } else {
     Menu.setApplicationMenu(null)
     mainWindow.loadFile(join(publicDir, 'index.html'))
-    if (globalConf.checkForUpdatesOnStartup && !isLinux) {
+    if (globalConf.checkForUpdatesOnStartup) {
```
CommandMC, line 255: Since the update message now has a case for auto-update not being supported, we want to always check for updates here

PR #4838, opened 2025-08-03
[UX] Print list of GOG games in GOG log

arielj:
This PR adds 2 improvements to logs:

- We now print GOG's library in the GOG's logs like we do for Legendary (this can be useful for debugging purposes)
- We don't print `Game output` in logs if it's not a game log (this was adding noise to runner logs, incorrectly)

Note that we don't have the GOG game version at the point I'm printing this to the logs, so I didn't include that (if you compare it with legendary's log)

---


flavioislima (approved):
lgtm

Issue #4817, opened 2025-07-28
Good night, the alphabet filter highlights the letter G but I have no games that begin with G


kawa-sanmyaku:
### Describe the bug

<img width="1677" height="482" alt="Image" src="https://github.com/user-attachments/assets/ff9da4a6-27d0-4ad4-ae0b-9460514e24d4" />

I've never added a game outside of GOG beginning with G, and I've already cleared the cache and reset Heroic :(

The only culprit I can think of is this...

<img width="755" height="82" alt="Image" src="https://github.com/user-attachments/assets/9310e9fc-a803-4332-b868-6369dc8d05db" />

which is sometimes downloaded whenever I update No Man's Sky

### Add logs

```text
Log contains some sensitive information, but nothing seems relevant to this issue
```

### Steps to reproduce

Download and install any Windows GOG game that requires the Galaxy Common Redistributables, making sure it doesn't begin with G

### Expected behavior

G shouldn't be highlighted? Lol this is a little silly

### Heroic Version

Latest Stable

### System Information

OS: Arch Linux x86_64
Kernel: Linux 6.15.8-arch1-1
CPU: AMD Ryzen 5 5600H (12) @ 3.30 GHz
GPU: NVIDIA GeForce RTX 3060 (Driver: nvidia-dkms 575.64.05-1)
Heroic: heroic-games-launcher-bin 2.18.0-1

arielj:
can you check your gog account on the website? maybe it's some goodie or thing heroic should ignore that is not getting ignored

is there anything that starts with `g...` or `the g...`?

I don't think the Galaxy Common Redistributables is causing that because it's never part of the library object that is used to check which letters to show

kawa-sanmyaku:
Sorry, there is nothing that begins with either `g...` or `the g...`

If I filter out the Windows games, the G is no longer highlighted; below are screenshots of the files available for download for the two Windows games I have, and only No Man's Sky is installed currently

<img width="1112" height="752" alt="Image" src="https://github.com/user-attachments/assets/a233067d-a2d0-49de-a793-bd0d98da4f7a" />

<img width="1103" height="837" alt="Image" src="https://github.com/user-attachments/assets/58893ef7-62a8-4731-b1d6-d0f95a86f0ba" />

kawa-sanmyaku:
[...]

PR #4839, opened 2025-08-03
[UX] Ignore DLCs when calculating available alphabet filter letters

arielj:
DLCs where looped through when calculating which letters should be available for the new alphabet filter feature. This was an issue because we don't show DLCs in the library, so a letter could be marked as available when we wouldn't have anything to show for that letter in the library.

This fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4817, because the `Galaxy Common Redistributables` download is added to the GOG's library and is flagged is a DLC, so it was counted for the letters but not displayed as a card.

---


flavioislima (approved):
lgtm

Issue #4847, opened 2025-08-05
Library return to top of page up arrow button has disappeared in 2.18.1


Sunspark-007:
### Describe the bug

The up arrow button that was available in the lower right corner is no longer present in 2.18.1

### Add logs

```text
n/a
```

### Steps to reproduce

1. View library

### Expected behavior

Button to be there to avoid the need to manually scroll up.

### Heroic Version

Latest Stable (Flatpak)

### System Information

Steam OS 3.6.24

PR #4848, opened 2025-08-05
[FIX] Back to top arrow

arielj:
Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4847

Not sure why we now depend on `document.body` instead of `window` to detect/calculate scroll position, but we need to change it for the back to top button.

---

### DIFF
--- a/src/frontend/screens/Library/index.tsx
+++ b/src/frontend/screens/Library/index.tsx
@@ -237,17 +237,19 @@ export default React.memo(function Library(): JSX.Element {
 
   // bind back to top button
   useEffect(() => {
-    if (backToTopElement.current) {
-      document.body.addEventListener('scroll', () => {
-        const btn = document.getElementById('backToTopBtn')
-        const topSpan = document.getElementById('top')
-        if (btn && topSpan) {
-          btn.style.visibility =
-            document.body.scrollTop > 450 ? 'visible' : 'hidden'
-        }
-      })
+    const btn = document.getElementById('backToTopBtn')
+    const topSpan = document.getElementById('top')
+
+    const scrollCallback = () => {
+      if (btn && topSpan) {
+        btn.style.visibility =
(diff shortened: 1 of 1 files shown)


src/frontend/screens/Library/index.tsx
```diff
@@ -234,16 +234,19 @@ export default React.memo(function Library(): JSX.Element {
...
+    const scrollCallback = () => {
+      if (btn && topSpan) {
+        btn.style.visibility =
+          document.body.scrollTop > 450 ? 'visible' : 'hidden'
+      }
     }
-  }, [backToTopElement])
+
+    document.body.addEventListener('scroll', scrollCallback)
+    return () => document.body.removeEventListener('scroll', scrollCallback)
```
arielj, line 251: we were not removing the event listener from the window object before, attaching new listener on every library view

src/frontend/screens/Library/index.tsx
```diff
@@ -234,16 +234,19 @@ export default React.memo(function Library(): JSX.Element {
...
+  }, [])
```
arielj, line 252: I learned the other day we shouldn't add `useRef` values in the useEffect dependencies

src/frontend/screens/Library/index.tsx
```diff
@@ -234,16 +234,19 @@ export default React.memo(function Library(): JSX.Element {
...
-    if (backToTopElement.current) {
-      window.addEventListener('scroll', () => {
-        const btn = document.getElementById('backToTopBtn')
-        const topSpan = document.getElementById('top')
-        if (btn && topSpan) {
-          btn.style.visibility = window.scrollY > 450 ? 'visible' : 'hidden'
-        }
-      })
+    const btn = document.getElementById('backToTopBtn')
+    const topSpan = document.getElementById('top')
```
arielj, line 241: we don't need to query these elements on every scroll callback

PR #4849, opened 2025-08-05
[FIX] Scroll card into view when using controller

arielj:
Another issue related to the scroll happening in document.body instead of window.

When navigating with a controller, this improves the position of the scroll so the focused card is scrolled into view. This was already a thing but it broke along with the other scroll-related features.

---


CommandMC (commented):
Somewhat related to this: Another place using `window` is the gamepad navigation, which seems to no longer get the `focus` and `blur` events. Adding them to `document.body` instead seems to *not* work in that case

arielj:
> Somewhat related to this: Another place using `window` is the gamepad navigation, which seems to no longer get the `focus` and `blur` events. Adding them to `document.body` instead seems to _not_ work in that case

I'm so confused, not sure what broke the events on the window object :S

Issue #4052, opened 2024-10-10
Cannot Disable Interface Animation


MarjaE2:
### Describe the bug

The launcher contains a great deal of animation. While most computer games do too, this makes it harder for users who pick games without as much animation, or without specific types of animation. I've got an awful headache between the animation and trouble reading some of the text.

The accessibility settings don't seem to include any options to disable unnecessary animation.

### Add logs

```text
(19:33:51) INFO:    [Legendary]:        Legendary location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/x64/linux/legendary
(19:33:51) INFO:    [Gog]:              GOGDL location: /app/bin/heroic/resources/app.asar.unpacked/build/bin/x64/linux/gogdl
(19:33:51) INFO:    [Connection]:       Connectivity: check-online
(19:33:51) INFO:    [Connection]:       Pinging external endpoints
(19:33:51) INFO:    [Gog]:              Checking for existing gog manifests
(19:33:52) WARNING: [Backend]:          Failed to register protocol with OS.
(19:33:52) INFO:    [Connection]:       Connectivity: online
(19:33:52) INFO:    [Backend]:          User Not Found, removing it from Store
(19:33:52) ERROR:   [Gog]:              Unable to syncQueued playtime, userData not present
(19:33:52) INFO:    [Gog]:              Running command: GOGDL_CONFIG_PATH=/home/marjaerwin/.var/app/com.heroicgameslauncher.hgl/config/heroic/gogdlConfig /app/bin/heroic/resources/app.asar.unpacked/build/bin/x64/linux/gogdl --auth-config-path /home/marjaerwin/.var/app/com.heroicgameslauncher.hgl/config/heroic/gog_store/auth.json auth
(19:33:52) INFO:    [Backend]:          AreWeAntiCheatYet data downloaded
(19:33:52) INFO:    [Backend]:          Frontend Ready
(19:33:52) INFO:    [Backend]:          Checking for current version changelog
(19:33:52) INFO:    [Frontend]:         Updated to a new version, cleaaning up the cache.
(19:33:52) INFO:    [Legendary]:        Running command: LEGENDARY_CONFIG_PATH=/home/marjaerwin/.var/app/com.heroicgameslauncher.hgl/config/heroic/legendaryConfig/legendary /app/bin/heroic/resources/app.asar.unpacked/build/bin/x64/linux/legendary cleanup
[...]

PR #4851, opened 2025-08-05
[UX/A11y] Disable animations

arielj:
Allow disabling all transitions and animations in the Accessibility settings.

Closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4052

---

### DIFF
--- a/src/frontend/App.css
+++ b/src/frontend/App.css
@@ -76,6 +76,9 @@ body:has(.disableAnimations) {
   .MuiTouchRipple-root {
     display: none !important;
   }
+  .Mui-focusVisible {
+    background: color(from var(--accent-overlay) srgb r g b / 0.3);
+  }
 }
 
 .HelpButton {
--- a/src/frontend/App.css
+++ b/src/frontend/App.css
@@ -65,6 +65,19 @@ body {
   grid-area: offline;
 }
 
+body:has(.disableAnimations) {
+  *,
+  *:before,
+  *:after {
+    animation: none !important;
+    transition: none !important;
+    transition-duration: 0ms !important;
+  }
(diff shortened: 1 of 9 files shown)


flavioislima (approved):
Code looks good.
Tested here and it works fine.

src/frontend/App.css
```diff
@@ -65,6 +65,19 @@ body {
...
+  *,
+  *:before,
+  *:after {
+    animation: none !important;
+    transition: none !important;
+    transition-duration: 0ms !important;
+  }
+  .MuiTouchRipple-root {
+    display: none !important;
+  }
```
CommandMC, line 78: Might want to do this with MUI theming instead:
https://mui.com/material-ui/customization/theme-components/#theme-default-props

arielj: I tried that, but it didn't work for Tabs (unless I did something wrong or it's a version thing)

CommandMC: Do we want this for tabs? Seems like the ripple is the only way to tell which tab you're selecting when using controller/keyboard navigation

arielj: hmmm I want to disable all animations (not sure what can affect people that need this accessibility feature so I'd rather go all in and just disable all), I'll check if I can set a non-animated state to show the active tab then for the controller navigation (I can probably set some different background without the animation)

arielj: I updated it to have a non-animated background (the previous code was also missing a focused state for the X buttons to close dialogs)

CommandMC: Yup, works great

CommandMC (approved):

Issue #4853, opened 2025-08-07
Heroic log prompt: no such file or directory…


UniversunGo:
### Describe the bug

I downloaded 4 games from the Epic platform, but three of them cannot be launched and show log errors indicating that there is no such file or directory. GTA can open and shows the Rockstar launcher, but it cannot connect to the account correctly and displays 'Go to purchase.'
I have tried downloading and using different wine managers, but they all encounter the same error. The homepage often shows that the games are running, but they are actually not starting.

### Add logs

```text
(12:50:24) [INFO]:    [Backend]:         Launching Brotato (e8bbb84be35640cda646233152ff3428)
(12:50:24) [INFO]:    [Backend]:         Preventing display from sleep
(12:50:24) [INFO]:    [Backend]:         Checking if wine version exists: Proton - GE-Proton10-10
(12:50:24) [INFO]:    [Backend]:         Preventing machine to sleep
(12:50:27) [INFO]:    [Backend]:         Checking if wine version exists: Proton - GE-Proton10-10
(12:50:27) [INFO]:    [Backend]:         Checking if wine version exists: Proton - GE-Proton10-10
(12:50:27) [INFO]:    [Backend]:         Checking if wine version exists: Proton - GE-Proton10-10
(12:50:27) [DEBUG]:   [Backend]:         Running Wine command: createprefix
[...]

PR #4857, opened 2025-08-07
[FIX] Catch error if no anticheat json file so launching doesn't fail

arielj:
Fixes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4853

Currently, if there's no `areweanticheatyet.json` info file, launching a game fails because we print the anticheat info in the logs.

This catches that error so launching continues as usual.

I could check for the file existence, but I don't know if there are other possible read errors so better to just try/catch it.

---

### DIFF
--- a/src/backend/anticheat/utils.ts
+++ b/src/backend/anticheat/utils.ts
@@ -37,16 +37,21 @@ async function gameAnticheatInfo(
   if (appNamespace === undefined) return null
   if (isWindows) return null
 
-  const data = await readFile(anticheatDataPath, 'utf-8')
-  const jsonData = JSON.parse(data)
-  return jsonData.find((info: AntiCheatInfo) => {
-    const namespace = info.storeIds.epic?.namespace
-    if (namespace) {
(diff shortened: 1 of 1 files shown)


src/backend/anticheat/utils.ts
```diff
@@ -37,16 +37,21 @@ async function gameAnticheatInfo(
...
+    return jsonData.find((info: AntiCheatInfo) => {
+      const namespace = info.storeIds.epic?.namespace
+      if (namespace) {
+        return namespace.toLowerCase().includes(appNamespace)
+      } else {
+        return false
+      }
+    })
+  } catch {
+    logWarning('AreWeAntiCheatYet file not present', LogPrefix.Backend)
```
CommandMC: Might want to print out the error
```suggestion
  } catch (err) {
    logWarning(['Failed to read AreWeAntiCheatYet file:', err], LogPrefix.Backend)
```

Etaash-mathamsetty (approved):

PR #4859, opened 2025-08-08
[Fix] Stop gamepad navigation on focus loss

CommandMC:
`focus` and `blur` events don't get triggered when the actual browser window gets focus (they instead fire when an *element* gains/loses focus). Not sure how this worked before, Electron bug?  
Instead now detect those changes in the Backend and pass them along

To make this modular, two new Backend events were added, `mainWindowFocussed` and `mainWindowUnfocussed`. A small handler then sends a `gamepad.setInputsEnabled` Frontend message when either event happens (either enabling or disabling gamepad inputs), which the Frontend then listens for

To test:
1. Connect gamepad
2. Unfocus Heroic
3. On `main`: Notice gamepad still controls Heroic. On this branch: Notice gamepad no longer controls Heroic.

Since I was already creating the `src/backend/gamepad/` folder for gamepad-related code, I also went ahead and moved the `gamepadAction` handler to its own file

---


arielj:
I can't reproduce this, I open heroic on main, move the controller around, then I click on a different window and the controller stops working, and I get the focus/blur events in the window object

what's your windows manager? I think one time someone reported a similar issue way back then when I first implemented this, and it was a window manager not properly unfocusing heroic or something like that

CommandMC:
I'm on KDE Plasma
```
Operating System: Arch Linux 
KDE Plasma Version: 6.4.4
KDE Frameworks Version: 6.16.0
Qt Version: 6.9.1
Kernel Version: 6.15.9-arch1-1 (64-bit)
Graphics Platform: Wayland
Processors: 16 × AMD Ryzen 7 3700X 8-Core Processor
Memory: 32 GiB of RAM (31,3 GiB usable)
Graphics Processor: NVIDIA GeForce RTX 4070
```

According to [MDN docs](<https://developer.mozilla.org/en-US/docs/Web/API/Window/focus_event>), the `focus` event should fire "when an element has received focus", so it's not (just) related to the window focus itself. We're also using it for the controller hints after all, the usage there looks more like what I'd imagine it to work  
Electron is correctly receiving focus/blur events on my end, otherwise this approach wouldn't work either

flavioislima:
Tested here on Plasma 6.5 and I also cannot reproduce the issue @CommandMC:

https://github.com/user-attachments/assets/f6de759a-a9fe-4856-8839-bea551d71321

The moment I move the mouse it removes the controls from the screen.

CommandMC:
@flavioislima Your reproduction there doesn't seem to match my steps. This isn't about Heroic getting stuck in "controller input mode" even if the mouse is used, it's about the controller inputs still being captured with Heroic being unfocussed (unless you are unfocussing Heroic & I've missed that)

I've done some more complete testing now:
- I cannot reproduce this issue at all on my laptop
- On my desktop, I can only reproduce this issue on the development version of Heroic (started with `pnpm start`), neither the Arch package nor the Flatpak have it

The one differentiating factor I could think of is that I have 3 monitors on my desktop, and just one on my laptop. Initially I thought Plasma's "Separate screen focus" may cause the issue, however I can also reproduce it on my desktop by opening Heroic & another window on the same monitor and switching between the two. So, I don't know why this is happening, but it is definitely still happening

Do the changes in this PR cause any issues for you? If they don't, we may want to include it, just in case someone else has the same strange setup I have

flavioislima:
@CommandMC yes, I'm unfocusing it and the controller icons are gone. I tried both in my plasma desktop on opensuse tw and also on the steam deck and they work fine for me.

CommandMC:
Closing this as I cannot reproduce the issue anymore now (after an OS reinstall)

CommandMC:
Update, I am again able to reproduce this. `focus` and `blur` events on `window` are not fired, so `isFocused` always stays `true`. It also seems like a few users on Discord can reproduce it as well, e.g. <https://discord.com/channels/812703221789097985/1538939768061034526>. Using this PR's changes again resolves the issue

I understand that this issue may not happen on every system, that's not my point. My question is: Do the changes here cause issues for your setups? If not, I would suggest applying them. Even if they don't help you specifically, they may help others.

PR #4863, opened 2025-08-09
Update heroic-epic-integration and use by default

Etaash-mathamsetty:
With 0.4 version, should be stable and reliable enough to use with all games.

Needs testing

### DIFF
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -888,24 +888,22 @@ export async function launch(
     ...getKnownFixesEnvVariables(appName, 'legendary')
   }
 
-  // We can get this env variable either from the game's settings or from known fixes
-  if (commandEnv['USE_FAKE_EPIC_EXE']) {
-    if (isWindows) {
-      commandEnv['LEGENDARY_WRAPPER_EXE'] = fakeEpicExePath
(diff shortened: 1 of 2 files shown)


src/backend/storeManagers/legendary/games.ts
```diff
@@ -888,24 +888,22 @@ export async function launch(
...
-    } else {
-      // on linux and mac, we copy the fake exe
-      const fakeExeWinPath = 'C:\\windows\\command\\EpicGamesLauncher.exe'
-      const fakeEpicExePathInPrefix = await getWinePath({
-        path: fakeExeWinPath,
-        gameSettings,
-        variant: 'unix'
-      })
+  // Use the wrapper EXE to launch games.
+  if (isWindows) {
```
arielj: I wonder if we can make this have a disable options in the advanced settings of a game?

like enable it by default but keep the option somewhere to disable it just in case

I'm worried that this gets flagged sometimes by antivirus software and it will break games, and having the option to disable this would be useful

or maybe check if the file exists and it can be copied before adding the env variable? then if the antivirus deleted it it will be ignored by default too

BananaWorks07: Ok sure, but the anti virus don't flag it anymore

PR #4866, opened 2025-08-11
Nix dev shell fix and NixOS README sectionfixes

aidalgol:
Some fixes to the Nix dev shell, and corrections to the NixOS information in the README.

---

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,40 +1,31 @@
-{ pkgs ? import <nixpkgs> {} }:
+# Nix shell environment suitable for running the build watch command.
+# Use like
+# $ nix-shell --run 'pnpm start'
+{
+  pkgs ? import <nixpkgs> { },
+}:
 
-(pkgs.buildFHSEnv {
-  name = "heroic-fhs-dev";
-  targetPkgs = pkgs: with pkgs; [
-    # Base requirements
-    git
-    nodejs_22
+let
+  inherit (pkgs) lib;
+in
+pkgs.mkShell {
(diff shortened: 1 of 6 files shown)


shell.nix
```diff
@@ -1,40 +1,31 @@
...
+
+  shellHook = ''
+    ln -sf ${lib.getExe pkgs.electron} node_modules/electron/dist/electron
+    ln -sf \
+      ${lib.getExe pkgs.vulkan-helper} \
+      ${lib.getExe pkgs.gogdl} \
+      ${lib.getExe pkgs.legendary-heroic} \
+      ${lib.getExe pkgs.nile} \
+      public/bin/x64/linux/
+  '';
```
CommandMC, line 30: Do we want this? Development environments of Heroic sometimes ship newer releases of these tools than what's available in the last release

Somewhat ties into my comments in https://github.com/NixOS/nixpkgs/pull/432183

aidalgol: There are trade-offs either way.  If we use the binaries fetched by the build, we would need run `patchelf` on them to fix their `RPATH`s to point to the Nix store paths.[^1]

But if you are developing on NixOS, then you probably want to be doing things the Nix way, which would be to override these packages to point to the same version that whatever commit from this repo would be pulling.

Before someone else submitted a PR to add `shell.nix`, I had been using my own file privately.  It may make more sense to remove this file until and unless there is a core project developer who is on NixOS.

[^1]: https://ryantm.github.io/nixpkgs/hooks/autopatchelf/

PR #4867, opened 2025-08-11
Log whether we are running as an AppImage

aidalgol:
Similar to the log message,
```
We are [not] running in a Flatpak container
```
Add a log message for AppImage,
```
We are [not] running from an AppImage
```

---

### DIFF
--- a/src/backend/utils/systeminfo/index.ts
+++ b/src/backend/utils/systeminfo/index.ts
@@ -55,6 +55,7 @@ interface SystemInformation {
   }
   steamDeckInfo: SteamDeckInfo
   isFlatpak: boolean
+  isAppImage: boolean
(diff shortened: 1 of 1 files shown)


arielj (commented):
I wonder if we can make all the steamdeck/flatpak/appimage block conditional to show up only for linux? it adds noise on macos and windows

CommandMC:
Note: The Steam Deck check is intentionally not Linux-specific (by detecting hardware details instead), that one can stay IMO. Yes, it's unlikely that anyone will install Windows on their Deck & use Heroic, but still; whether a device is a Steam Deck has nothing to do with the device's OS

Either way, the actual change here looks good to me

CommandMC (approved):

PR #4868, opened 2025-08-12
Feature: Binary validation using VirusTotal

kmturley:
Relates to #4861

This is a new GitHub Workflows, which only runs if triggered manually via the UI. The purpose is to run a Virus report on each of the helper binaries depended upon. If useful this could be updated to run whevenever the binary list is changed. But for now it has to be manually run.

Steps:
- Requires `VT_API_KEY` secret to be set in the repository settings.
- Installs Heroic dependencies (same step as other workflows)
- Runs https://github.com/crazy-max/ghaction-virustotal on the downloaded files at: `./public/bin/**/*`
- Logs out the links to each report in markdown format (this could be adjusted later to comment on a PR)

Example run:
https://github.com/kmturley/HeroicGamesLauncher/actions/runs/16898430747/job/47872856107

I will post the report output as a comment on this PR

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [x] Tested the feature and it's working on a current and clean install.
- [x] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [x] Created / Updated Tests (If necessary)
- [x] Created / Updated documentation (If necessary)

Issue #4861, opened 2025-08-08
Security Report


Francesco-1-7:
### Describe the bug

Hello,

About a week ago, I submitted a report via the GitHub Security tab regarding a potential vulnerability in the launcher.

The `epicgameslauncher.exe` file has been flagged by 33 antivirus engines on VirusTotal as a Trojan. While I understand the project may occasionally trigger false positives—especially given the technologies it employs—this number is unusually high and may warrant further investigation.

I'm available to collaborate on a resolution and to provide any details needed.  
Thank you for your ongoing work and your commitment to security.

Best regards,  
Francesco 

### Add logs

```text
/
```

### Steps to reproduce

/

### Expected behavior

/

### Heroic Version

Latest Stable

### System Information

/

arielj:
the file comes from this repo https://github.com/Etaash-mathamsetty/heroic-epic-integration

you can audit the code of the exe and of the build tasks if you want

FrancescoPnr-dev:
Perfect, thank you very much. I'll do the analysis and build process this week and will try to update this post.
Good work and have a great weekend

kmturley:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/meta/downloadHelperBinaries.ts

I don't see a check for file hash in this logic. This allows for a potential attack by swapping out the binaries on one of the dependency GitHub Releases with a malicious file.

GitHub Releases provides an API to get the sha256 or you can download the file and check against the expected hash.

kmturley:
If binary dependencies are bundled/shipped with the Heroic release, it's less of an issue as long as the release is scanned for viruses at some point.

I don't see a step in GitHub Actions to scan for viruses in the binaries. This could be added using
https://github.com/crazy-max/ghaction-virustotal

There isn't a great way to get the VirusTotal report in the PR, but you can add links to the report from the PR like this:
```
     - name: VirusTotal Scan
        id: virustotal
        uses: crazy-max/ghaction-virustotal@v4
        with:
          vt_api_key: ${{ secrets.VT_API_KEY }}
          request_rate: 4
          files: ./dist/**/*

      - name: VirusTotal Report
        uses: actions/github-script@v7
        with:
          script: |
            const analysis = '${{steps.virustotal.outputs.analysis}}';
            if (!analysis) return;
            let output = 'VirusTotal Report:\n';
            let lines = analysis.split(',');
            for (let line of lines) {
[...]

PR #4869, opened 2025-08-12
Feature: Validate downloaded binaries by file hash

kmturley:
Relates to #4861

A malicious actor or maintainer not following best-practices for Release versioning, could swap out a binary file on a GitHub Release without the Heroic team knowing. It could be downloaded and bundled into a Heroic Release without their knowledge. This could break functionality or worse expose Heroic users to attack vectors.

This feature ensures that downloaded binaries from other projects, match the expected file hashes in this project. If the hash changes, assume the file has changed and the file needs to be checked for Viruses again. This feature could be used in conjunction with VirusTotal checking https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4868 to validate binaries are safe for Heroic users.

When a Release version number is updated, the corresponding hash should also be updated in `meta/downloadHelperBinaries.ts` to prevent the script from throwing an Error and blocking builds.

Steps:
- Run `pnpm download-helper-binaries`
- Script will download binaries as before
- Now checks the file hash against the list of known hashes
- If there is a mismatch, that means the file has changed, throw an Error to prevent the file being bundled.

Example error thrown:
```
Error: - x64/win32/GalaxyCommunication.exe (Hash mismatch) received 'abc208076a778ee738cae8451c9be7ab33c9787b0b69b2e7e4ffc70becc39e1e' expected 'bbb64b92fd9af97c4dc020aaa2a4bbe392bac84c22d60fc6224805e119842e38'
```

---

### DIFF
--- a/meta/downloadHelperBinaries.ts
+++ b/meta/downloadHelperBinaries.ts
@@ -1,8 +1,9 @@
+import { createHash } from 'crypto'
 import { createWriteStream } from 'fs'
 import { chmod, stat, mkdir, readFile, writeFile } from 'fs/promises'
 import { dirname, join } from 'path'
-import { Readable } from 'stream'
-import { finished } from 'stream/promises'
+import { Readable, Transform } from 'stream'
+import { pipeline } from 'stream/promises'
 
 import { setGlobalDispatcher, ProxyAgent } from 'undici'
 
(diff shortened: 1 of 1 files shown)


arielj:
Closing it for now, we can discuss again in discord for a better solution

PR #4875, opened 2025-08-14
[Feat] Clarify "Use dedicated graphics card" / PRIME run option

CommandMC:
Users often toggle this option by mistake, either while only using one GPU (seems like a warning wasn't enough there), or thinking they *have to* toggle it for their dGPU to be used. To address this, we now
- only show the option if we have more than one GPU
- make it more clear that this option is likely not necessary & may cause issues

Tested on both a single-GPU and dual-GPU system, only on Linux (this option only applied to Linux to begin with)

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -769,10 +769,6 @@
         "prefered_language": "Prefered Language (Language Code)",
         "preferSystemLibs": "Prefer system libraries",
         "primerun": {
-            "confirmation": {
-                "message": "Only one graphics card was detected in this system. Please note that this option is intended for multi-GPU systems with headless GPUs (like laptops). On single-GPU systems, the GPU is automatically used & enabling this option can cause issues. Do you really want to enable this option?",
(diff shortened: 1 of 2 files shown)


arielj:
I wonder if we can just remove this setting.

Maybe it was needed before but lately I only see this causing issues. And I feel like it's some edge case that if someone knows they need it (to toggle it intentionally) they can probably just add the variable that we add?

Or maybe we can change the name of the setting to be `Set NVIDIA_PRIME=... environment variable`, so users who have no idea what that is don't turn it on.

Because I feel some users will still toggle that incorrectly when trying to make something work and leave it on by mistake sometimes.

CommandMC:
Definitely in favor of removing this in the long run, just didn't want to risk breaking existing setups

arielj:
maybe we can remove the UI but keep the setting handled in the backend if already set to true?

that way existing setups won't break and new users won't turn it on by mistake

flavioislima (approved):
I am fine with removing as well.

PR #4876, opened 2025-08-14
[Fix] Return "macOS" as the OS name on macOS

CommandMC:
Originally I thought "OS:  X.Y.Z (darwin)" would be fine to output temporarily, but since the codename detection FIXME doesn't seem easy to solve, let's at least do "OS: macOS X.Y.Z" for now

To test:
1. Look at any macOS log, note the OS part of the system info message says "OS:  X.Y.Z (darwin)"
2. Run this PR & again check the log, now the message says "OS: macOS X.Y.Z (darwin)"

---


Etaash-mathamsetty (approved):
Seems fair enough for now

arielj (approved):

PR #4889, opened 2025-08-21
[Tech] Make LibraryManagers and GameManagers actual classes

CommandMC:
> [!NOTE]  
> Do not be alarmed by the huge amount of changed lines in this PR. Most of it is whitespace changes (since indentation changes from 2 to 4 spaces). The main changes in this PR are in 6b6126ba4e331b4bc2b81314ec37a7b50ff10219, whitespace changes are in ee09713cd5bca9fd638d1d97cee6778ada2a1f62 (so ignore that one)

Library & game managers were long used as pseudo-classes (having a defined structure & being used like a class instance, but not actually being a class). This causes various issues:
- changing a manager file does not track object shape properly (there's nothing directly stopping you from renaming a "method" of one of the managers, you'll only get indirect type errors)
- "method" usages are not tracked properly
- manager state is not encapsulated
  - This is admittedly still an issue now, but at least it's now *possible* to resolve
- when adding a new manager, it's unclear which methods are actually required. Again, you only get indirect errors in the `gameManagerMap`/`libraryManagerMap` declaration (or worse, in some other file expecting some extra method), there's nothing pointing you to the error *in the specific manager implementation*

I am looking to make core changes to the game & library managers (mainly getting rid of or at least significantly slimming down the `GameInfo` interface) to make it easier to implement more store providers. This PR is the first step towards that

I did make one "functional" change here: I've removed a test file for Legendary's save sync folder computation. I don't think this was working correctly (the test was too specialized, it might've helped more now that the helper binary stubs are in), and I want to change how save sync folders work soon anyways

---


CommandMC:
Seems like the diff isn't great even with "Hide whitespace" turned on. I'll see if I can split those changes out into (1) code changes and (2) linting

src/backend/main.ts
```diff
@@ -664,8 +650,12 @@ addHandler('getMaxCpus', () => cpus().length)
 
 addHandler('getHeroicVersion', app.getVersion)
 addHandler('isFullscreen', () => isSteamDeckGameMode || isCLIFullscreen)
-addHandler('getGameOverride', async () => getGameOverride())
-addHandler('getGameSdl', async (event, appName) => getGameSdl(appName))
+addHandler('getGameOverride', async () =>
+  libraryManagerMap['legendary'].getGameOverride()
+)
+addHandler('getGameSdl', async (event, appName) =>
+  libraryManagerMap['legendary'].getGameSdl(appName)
```
arielj, line 679: I wonder if we can export each manager too, not just the hash, then we can import `legendaryLibraryManager` to not have to import the hash and access with a hardcoded `'legendary'` string, it would make the code more declarative and easier to understand in my opinion (I know it's a small difference, but I think it will help since it's used so much throughout the app, there's many places we just have hardcoded runners)

CommandMC: I agree that this would be a good idea for now. I do have a concern: I'd like to eventually get rid of every one of these hardcoded runner-specific library manager invocations (my end goal is to dynamically enable/disable library integrations at runtime; to load them as "plugins"). Obviously exporting individual library managers would go against that goal
So, I can definitely do this for now (and if you'd like me to, I would). I just don't see it as *that* much of a problem right now, and given these areas will change relatively soon-ish anyways (I have a draft of this plugins idea around already) I hadn't bothered to change it yet

src/backend/launcher.ts
```diff
@@ -1023,7 +1003,9 @@ async function installFixes(appName: string, runner: Runner) {
 
     for (const filePath of knownFixes.runInPrefix) {
       const fullPath = join(gameInfo.install.install_path!, filePath)
-      await runWineCommandOnGame(appName, {
+      // FIXME: This doesn't seem right, shouldn't we use a generic function instead
+      //        of the Legendary-specific one?
+      await gameManagerMap['legendary'].runWineCommandOnGame(appName, {
```
arielj: looks like there is a generic `runWineCommandOnGame` function from backend/tools/index.ts that receives the runner as the first param and the logic is the same as the legendary one (but uses each runner correctly)

arielj (approved):
I tested this installing, running, and uninstalling some games from the different stores and things worked fine. I think we should merge this right after we make a release so we have more time to test since it's a big refactor, but I'll approve it now so we can merge it quickly.

I do think everything is fine, but it's hard to review it thoroughly as it's kinda long even ignoring the linter white space fixes commit

flavioislima (approved):
Code looks good and I can't find any issue when testing as well.

PR #4894, opened 2025-08-21
[Tech]  No longer use GameManagers as singletons

CommandMC:
Invoking a method on one of the game managers was always a little verbose: Every method took not just its own parameters, but also parameters describing the game's state (effectively that's only an AppName right now, but that may change for other store providers). Instead, there's now one `Game` class instance for every distinct game in the user's library

In practice, code changes in the following way:
Before:
```ts
gameManagerMap[runner].foo(appName, arg1)
gameManagerMap[runner].bar(appName)
gameManagerMap[runner].baz(appName, arg1, arg2)
```
After:
```ts
const game = libraryManagerMap[runner].getGame(appName)
// NOTE: `game` could potentially be passed as a parameter instead of `appName` and `runner`, making the above line unnecessary
game.foo(arg1)
game.bar()
game.baz(arg1, arg2)
```

As I mentioned in #4889, I plan on offloading a lot of `GameInfo` properties into individual methods (so there's not as much a burden on game managers to provide *everything* in `getGameInfo` all at once). Increasing the amount of method calls quickly leads to our before approach (the current code) becoming unreadable due to there being so much duplicate work.

Requires #4889

---

### DIFF
--- a/src/backend/downloadmanager/downloadqueue.ts
+++ b/src/backend/downloadmanager/downloadqueue.ts
@@ -1,4 +1,4 @@
-import { gameManagerMap, libraryManagerMap } from 'backend/storeManagers'
+import { libraryManagerMap } from 'backend/storeManagers'
 import { TypeCheckedStoreBackend } from './../electron_store'
 import { logError, logInfo, LogPrefix, logWarning } from 'backend/logger'
 import { getFileSize, removeFolder, sendGameStatusUpdate } from '../utils'
@@ -243,7 +243,9 @@ function cancelCurrentDownload({ removeDownloaded = false }) {
 
     if (removeDownloaded) {
       const { appName, runner } = currentElement.params
-      const { folder_name } = gameManagerMap[runner].getGameInfo(appName)
+      const { folder_name } = libraryManagerMap[runner]
+        .getGame(appName)
+        .getGameInfo()
       if (folder_name) {
         removeFolder(currentElement.params.path, folder_name)
       }
@@ -273,7 +275,7 @@ function resumeCurrentDownload() {
 function stopCurrentDownload() {
   const { appName, runner } = currentElement!.params
   callAbortController(appName)
-  gameManagerMap[runner].stop(appName, false)
+  libraryManagerMap[runner].getGame(appName).stop(false)
 }
 
 // notify the user based on the status of the element and the status of the queue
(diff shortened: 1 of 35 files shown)


src/backend/downloadmanager/downloadqueue.ts
```diff
@@ -226,7 +226,9 @@ function cancelCurrentDownload({ removeDownloaded = false }) {
 
     if (removeDownloaded) {
       const { appName, runner } = currentElement.params
-      const { folder_name } = gameManagerMap[runner].getGameInfo(appName)
+      const { folder_name } = libraryManagerMap[runner]
+        .getGame(appName)
```
arielj, line 247: I wonder if we can have a helper `getGame(runner, appName)`, because we do this `libraryManagerMap[runner].getGame(appName)` a looooot

that would also help decoupling how we get the store manager throughout the app, we wouldn't need to "know" that there's a `libraryManagerMap` hash

arielj (commented):
I like this to remove a lot of boilerplate, I left a comment to hopefully remove more

src/backend/downloadmanager/downloadqueue.ts
CommandMC: I agree that we do this a lot, but I'd like to resolve this a little differently (possibly in future PRs):
- For everything in the Backend, pass around the `Game` instead of `appName, runner`
- For calls from Frontend to Backend, I was thinking that we add something like a special "type" to our IPC wrappers that, if passed from FE to BE, automatically does the equivalent of `getGame(appName, runner)` and passes that `Game` to the actual listener/handler. Call it a `GameHandle` maybe?

flavioislima (approved):
That was a good cleanup, lgtm

src/backend/shortcuts/nonesteamgame/__tests__/nonesteamgame.test.ts
```diff
@@ -121,80 +117,21 @@ describe('NonSteamGame', () => {
...
-
-    expect(logInfo).not.toBeCalled()
-    expect(logError).toBeCalledWith(
-      `Can't add "${game.title}" to Steam user "steam_user". "${shortcutFilePath}" is corrupted!\n` +
-        'One of the game entries is missing the LaunchOptions parameter!',
+      `Remove Steam images for ${game.getGameInfo().title}`,
       'Shortcuts'
     )
-    expect(showDialogBoxModalAuto).toBeCalled()
   })
```
CommandMC, line 198: I deleted these tests since the situations they're testing are not realistic (creating a game without an AppName isn't possible, for example)

PR #4895, opened 2025-08-22
[Tech] Simplified adding runners

unbelievableflavour:
This MR moves a ton of runner specific code into a single hook so that it's a LOT easier to add new runners to the project. 
I've added a hook that will just give you the required info if you need it instead of manually defining every single runner the project has in all frontend components.

---


src/common/types.ts
```diff
@@ -14,8 +14,9 @@ import type { HowLongToBeatEntry } from 'backend/wiki_game_info/howlongtobeat/ut
 import { NileInstallInfo, NileInstallPlatform } from './types/nile'
 import type { Path } from 'backend/schemas'
 import type LogWriter from 'backend/logger/log_writer'
+import { StoreConfig } from 'frontend/hooks/useStoreConfigs'
```
CommandMC: Use a type import (`import type ...`) here to not accidentally import Frontend-only code in the Backend

src/frontend/hooks/useStoreConfigs.ts
```diff
@@ -0,0 +1,83 @@
...
+    },
+    {
+      runner: 'sideload',
+      filterKey: 'sideload',
+      displayName: t('Other'),
+      store: { library: sideloadedLibrary },
+      categories: ['all', 'sideload'],
+      authCheck: () => true // sideload doesn't need authentication
+    }
+  ] as StoreConfig[]
```
CommandMC: Is this type assertion necessary? Ideally you'd use a type hint (`const storeConfigs: StoreConfig[] = ...`)

src/frontend/hooks/useStoreConfigs.ts
```diff
@@ -0,0 +1,83 @@
...
+export function createStoreConfigs(
+  state: CreateStoreConfigsState,
+  t: (key: string) => string
+) {
+  const { epic, gog, amazon, sideloadedLibrary } = state
+  const storeConfigs = [
+    {
+      runner: 'legendary',
+      filterKey: 'legendary',
+      displayName: t('Epic Games'),
```
CommandMC: Making these strings instead of functions means translations are only applied on Heroic restart
Using functions/methods instead would be better:
```ts
displayName: () => t('Epic Games')
```

CommandMC: We don't usually use literal text as translation keys. Prefer keys like `storeDisplayNames.epic` and provide the English translation as the 2nd parameter

src/backend/storeManagers/index.ts
```diff
@@ -77,8 +77,8 @@ export function autoUpdate(runner: Runner, gamesToUpdate: string[]) {
   return gamesToUpdate
 }
 
-export async function initStoreManagers() {
-  await LegendaryLibraryManager.initLegendaryLibraryManager()
-  await GOGLibraryManager.initGOGLibraryManager()
-  await NileLibraryManager.initNileLibraryManager()
+export async function initLibraryManagers() {
+  Object.values(libraryManagerMap).forEach(async (element) => {
+    await element.init?.()
```
CommandMC: Ideally use a `Promise.all` here to load library managers in parallel:
```suggestion
  return Promise.all(Object.values(libraryManagerMap).map((manager) => manager.init?.()))
```
(untested, might not work with `init` being optional)

src/frontend/types.ts
```diff
@@ -193,10 +198,7 @@ export interface SettingsContextType {
 }
 
 export interface StoresFilters {
-  legendary: boolean
-  gog: boolean
-  nile: boolean
-  sideload: boolean
+  [key: string]: boolean
 }
```
CommandMC: `string` doesn't seem correct here, don't you want `StoreConfigs['filterKey']` here?

src/frontend/state/ContextProvider.tsx
```diff
@@ -101,7 +101,11 @@ const initialContext: ContextType = {
   disableDialogBackdropClose: false,
   setDisableDialogBackdropClose: () => null,
   disableAnimations: false,
-  setDisableAnimations: () => null
+  setDisableAnimations: () => null,
+  storeConfigs: {
+    storeConfigs: [],
+    runnerToDisplayName: () => 'Other'
+  }
```
CommandMC: I'd recommend against adding new things to the global state, instead either using individual Zustand stores or adding them to the `GlobalStateV2` store

src/common/types/game_manager.ts
```diff
@@ -76,6 +76,7 @@ export interface GameManager {
 }
 
 export interface LibraryManager {
+  init?: () => Promise<void>
```
CommandMC: I think requiring this `init` method and just making it do nothing for sideloads would be easier here
Coincidentally I ended up making this same change in #4889

CommandMC (commented):
Considering your recent PRs I was practically waiting for something like this, and considering *my* recent PRs I assume it's no surprise that I'm very much in favor of this
I just have a few comments on the implementation

src/frontend/state/ContextProvider.tsx
unbelievableflavour: Moved it to a separate zustand store. Altough initiatian is still dont through  GlobalState file because I would have no idea how to initiate it otherwise (aside from creating a new provider wrapper in `App`. Which I can do if you like.

src/common/types/game_manager.ts
unbelievableflavour: init is now required.

src/backend/storeManagers/index.ts
unbelievableflavour: Updated. Runs in parallel now.

src/frontend/hooks/useStoreConfigs.ts
unbelievableflavour: It now uses translation keys with a second english param.

unbelievableflavour: Nice find! It's now a function.

src/common/types.ts
unbelievableflavour: The imports are now type imported.

src/frontend/types.ts
unbelievableflavour: We could. But that does mean other dynamically added stores will not work. So custom libraries would for instance not work since that dynamically adds multiple stores.

src/frontend/hooks/useStoreConfigs.ts
unbelievableflavour: It is necessary unless we want all our `runner` and `filterKey` to be like: `runner: 'nile' as const`.

(I've slightly updated it though since we now use a zustand store. It now looks like this:
```
    set({ storeConfigs: storeConfigs as StoreConfig[] })
```

src/frontend/types.ts
unbelievableflavour: But for now I've locked it to `StoreConfigs['filterKey']` since that's a separate MR :)

src/frontend/hooks/useStoreConfigs.ts
CommandMC: I'd still prefer const assertions over a regular type assertion, but I don't feel strongly about it

PR #4897, opened 2025-08-25
[Tech] Simplified adding runners (backend specific)

unbelievableflavour:
This MR moves backend runner specific code into a single object so that it's a LOT easier to add new runners to the backend of the project. It's basically an addition to https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4895,
But since that PR has already been reviewed I'd thought I'd offer it as a separate PR.

---

### DIFF
--- a/src/backend/protocol.ts
+++ b/src/backend/protocol.ts
@@ -10,8 +10,9 @@ import { z } from 'zod'
 import { windowIcon } from './constants/paths'
 import { Path } from './schemas'
 import { isCLINoGui } from './constants/environment'
+import { runnerMap } from 'backend/runners'
 
-const RUNNERS = z.enum(['legendary', 'gog', 'nile', 'sideload'])
+const RUNNERS = z.enum(Object.keys(runnerMap) as [Runner, ...Runner[]])
 
 export function handleProtocol(args: string[]) {
   const urlStr = args.find((arg) => arg.startsWith('heroic://'))
--- a/src/backend/runners.ts
+++ b/src/backend/runners.ts
@@ -1,0 +1,58 @@
+import { GameManager, LibraryManager } from 'common/types/game_manager'
+
+import * as SideloadGameManager from 'backend/storeManagers/sideload/games'
+import * as SideloadLibraryManager from 'backend/storeManagers/sideload/library'
+import * as GOGGameManager from 'backend/storeManagers/gog/games'
+import * as GOGLibraryManager from 'backend/storeManagers/gog/library'
+import * as LegendaryGameManager from 'backend/storeManagers/legendary/games'
+import * as LegendaryLibraryManager from 'backend/storeManagers/legendary/library'
+import * as NileGameManager from 'backend/storeManagers/nile/games'
+import * as NileLibraryManager from 'backend/storeManagers/nile/library'
+
+interface StoreBackendConfig {
+  runner: 'legendary' | 'gog' | 'nile' | 'sideload'
+}
+
+export const runnerMap: {
+  [key in StoreBackendConfig['runner']]: {
+    store: string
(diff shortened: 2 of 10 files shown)


src/backend/runners.ts
```diff
@@ -0,0 +1,58 @@
...
+export const runnerMap: {
+  [key in StoreBackendConfig['runner']]: {
+    store: string
+    umu: { isSupported: boolean; storeName?: string }
+    gamesDB: { isSupported: boolean }
+    logPrefix: string
+    library: LibraryManager
+    gameManager: GameManager
+  }
+} = {
```
CommandMC, line 25: IMO using long type definitions like this as type hints directly always looks a little awkward. I'd define the value shape as its own type/interface (`RunnerDefinition` perhaps?) and then use `Record<StoreBackendConfig['runner'], RunnerDefinition>` here

src/backend/storeManagers/index.ts
```diff
@@ -1,39 +1,35 @@
...
-  legendary: LegendaryLibraryManager,
-  nile: NileLibraryManager
-}
+export const gameManagerMap: GameManagerMap = (() =>
+  Object.fromEntries(
+    Object.entries(runnerMap).map(([runner, config]) => [
+      runner,
+      config.gameManager
+    ])
+  ) as GameManagerMap)()
```
CommandMC, line 24: I assume this [IIFE](<https://developer.mozilla.org/en-US/docs/Glossary/IIFE>) is necessary due to circular import shenanigans? If that's not the case, IMO removing it would be clearer

src/backend/runners.ts
```diff
@@ -0,0 +1,58 @@
...
+import * as NileGameManager from 'backend/storeManagers/nile/games'
+import * as NileLibraryManager from 'backend/storeManagers/nile/library'
+
+interface StoreBackendConfig {
+  runner: 'legendary' | 'gog' | 'nile' | 'sideload'
+}
+
+export const runnerMap: {
+  [key in StoreBackendConfig['runner']]: {
+    store: string
```
CommandMC, line 18: Maybe rename this to `storeName`/`storeId`? `store` alone doesn't immediately make it clear that it's a name (one could also imagine the store URL)

src/backend/protocol.ts
```diff
@@ -10,8 +10,9 @@ import { z } from 'zod'
 import { windowIcon } from './constants/paths'
 import { Path } from './schemas'
 import { isCLINoGui } from './constants/environment'
+import { runnerMap } from 'backend/runners'
 
-const RUNNERS = z.enum(['legendary', 'gog', 'nile', 'sideload'])
+const RUNNERS = z.enum(Object.keys(runnerMap) as [Runner, ...Runner[]])
```
CommandMC, line 15: This is a little awkward. Out of scope here, but we should look into making this a branded type with a refine instead then

CommandMC (commented):
In general I'm very much in favor of this change. I've toyed with an idea like this in the past (having a "runner registry" that gets populated at runtime, to allow for a plugin-like runner interface), so this feels very familiar
Got a few small nitpicks

src/backend/runners.ts
unbelievableflavour: Funny, I thought about this change as well but that. Nah it was storeMap before I'll just keep it as is. But updated it to storeId :)

src/backend/storeManagers/index.ts
unbelievableflavour: Woops. old needed hack. Removed it.

src/backend/runners.ts
unbelievableflavour: Added `RunnerDefinition`.

PR #4626, opened 2025-06-06
[UI] Game page responsiveness fixes

arielj:
This PR closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4622 and closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4567

https://github.com/user-attachments/assets/dc29b6de-6b80-4476-b098-11615c8f40cf

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #4911, opened 2025-09-01
[UI] Fix GamePage scaling on larger displays

clanps:
this is to fix scaling issues on larger displays

Either remove the lines or change them to 100% so the UI scales properly on all displays
in #4626 there were changes made to limit the size, I'm not sure why when the elements don't really spread that far away from each other.
regular examples
before
after
extreme examples
before2
after2

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.css
+++ b/src/frontend/screens/Game/GamePage/index.css
@@ -6,7 +6,6 @@
   gap: var(--space-md);
   border-radius: 10px;
   height: 100%;
-  max-width: 2400px;
   grid-template-areas:
     'header header'
     'main extra'
@@ -93,8 +92,6 @@
     flex-direction: column;
     width: 100%;
     height: 100%;
-    max-height: 1100px;
-    max-width: 1200px;
     overflow-y: auto;
     overflow-x: hidden;
     align-items: flex-start;
@@ -108,7 +105,6 @@
 
   .extraInfoWrapper {
     grid-area: extra;
-    max-width: 850px;
     gap: 1rem;
   }
 


flavioislima:

clanps:

arielj:
things being too wide hurts readability and accessibility

the left box being too wide makes the paragraphs harder to read https://www.freshconsulting.com/insights/blog/uiux-principle-46-text-box-width-should-help-users-read

and the right box being too wide makes it hard to match the left and right sides of each row together, keeping them close makes it easier to see what's what

that's why the two boxes are kinda limited in width

maybe an option for the left box is to let it grow but have a maximum width for the description to not have that issue, but I think it will look weird

at the end, I'd rather have it look a bit weird for really wide screen rather than having bad readability/accessibility

clanps:
I really dislike the boxes not expanding to fill the screen and being glued to the top left margin. I also don't think everyone will suffer from accessibility or readability issues. A toggle that allows this but is off by default would maybe be a better solution than nothing at all?

arielj:
there's a `Custom CSS` feature in general settings > advanced, where you can modify any CSS

clanps:
People who also share the same opinion as me shouldn't need to input custom css for something as basic as this. It really should be built in. That seems counterintuitive to being accessible.

veryprofessionaldodo:
Just to give my two cents, I not only agree with the @clanps solution, I had already started working on a MR for this very purpose as well :D 

In my opinion, in the future we should have a new single-pane UI for all information. The fact that it's split means that for readability we should have the text not exceed a specific width, but since it doesn't make sense to have empty space in the left pane, we set a maximum size for the entire pane. If we had just one pane, we could set a max width for the text, while also ensuring we spread out the rest of the information well across the page. But that's another discussion :)

PR #4915, opened 2025-09-03
[UI] Make game list not appear behind header

veryprofessionaldodo:
This MR is a bit bigger than what I hoped, but I'll break everything down as simply as I can:

**Motive for this MR:**
Noticed an issue with keyboard and gamepad navigation of the main library, in which if we have the list scrolled a bit down, trying to navigate in the header becomes very confusing, as games would frequently grab the focus, since they were a bit out of view. 

**Approach:** 
To begin with, I made the game list be the only element that can scroll. That way the header is static, and the gamelist can be scrolled normally. 

**Unexpected issue**:
Having the game list be scrollable broke spatial navigation. Apparently if the container is scrollable, focus isn't propagated to the children, but rather just avoided. This meant that clicking "down" in the header would navigate to the sidebar. Apparently this is due to Chromium spatial navigation being an experimental feature. 

When I added the WCIG implementation of spatial navigation, this problem just resolved itself, and it even adds quite a few possible features in the future to make navigation with a gamepad even better. Since that library is more thoroughly documented, we can understand better and make more gamepad-friendly navigation. 

Below are two videos for a before and after:

<details>

Before:

https://github.com/user-attachments/assets/3fb57740-8d03-4c3c-811f-dc4a66a2a0b6

After:

https://github.com/user-attachments/assets/eae682f2-671d-4992-aba1-9551c27c3ae3

</details>

### DIFF
--- a/src/frontend/components/UI/LibraryFilters/index.css
+++ b/src/frontend/components/UI/LibraryFilters/index.css
@@ -9,6 +9,7 @@
     position: absolute;
     top: 100%;
     right: 0px;
+    z-index: 10;
     min-width: 250px;
     display: flex;
     flex-direction: column;


veryprofessionaldodo:
I'm an idiot, but I have no idea how to re-run the pipelines. It looks like it was an external issue to this MR

arielj:
Replacing the spatial navigation with that library creates a lot of issues:
- navigating the sidebar vertically moves by 2 elements (you can see that in your recording too)
- if you open the virtual keyboard (pressing A on a controller while focusing an input), you can't move through the keyboard keys
- if I'm on the top/left element of the library and press `left`, it moves DOWN one row, if I press `left` again it takes me to the Start Tour element next to the version
- I can't focus the options inside the `Filters` dropdown, only the `only` buttons on the right side but if I press `left` it takes me out of that drop down
- in the settings I can't focus any checkbox, only text/field/select inputs
- I can't focus the `(i)` elements that show more info
- when we are on a select element, pressing down opens the select instead of moving down to the next input, you are stuck focusing select items

I didn't test all screens but that was just testing for like 5 minutes.

Also changing the scrollable element breaks some things, the settings page now doesn't scroll properly for example

I don't know if I'm confortable replacing the spatial navigation since it can break so many things, I imagine there's a simpler way of the main reason is to avoid focusing elements under the header but I think this is too big of a change for such a small problem.

I know Chromium's spatial navigation is not great, but I think it's hard in general, there's too many issues if we replace it, maybe there are better ways to improve the navigation without replacing the spatial navigation feature (like, maybe we can implement custom navigation when the focus is on a card to improve things there, but keep the rest of the app using the spatial navigation for example). Since we first detect the action on the frontend we could for example move the focus with javascript instead of sending the event to the backend.

veryprofessionaldodo:
@arielj You're right, I apologise, I should have tested this more thoroughly. Although I do still believe that that library is a net positive add, if not only for how well documented the API is. There are some things that are being done that are workarounds (like the Selects trapping input on keyboard, but having a custom clause for gamepad) that might be more straightforward to do with the new library.  We can also make less of those workaround in the `gamepad` file, and instead determine spatial navigation actions [directly in the element itself](https://drafts.csswg.org/css-nav-1/#css-property-spatialnavigationaction), for example 

What if I drop this merge request, and instead make one solely dedicated to using the WICG library,  and more thoroughly testing everything, would that be something you guys would consider approving? I can understand if you deem it as high risk and opt not to go with that approach, so I wanted to confirm before I started that :)

arielj:
I really don't know until I see more code. I don't know if it's worth it to migrate to that and fix all the issues that change creates because currently we already worked around some of the issues for the native spatial navigation and it took so long to actually find and address them (the issue with selects is not caused by the spatial navigation but by the MUI library for example). It's hard to say if at the end the integration will be good to merge it or too hacky and risky for something that won't make it an obvious improvement initially (it might be a lot of work just to get things to work how they work now).

I also noticed that library has had no commits for the last 4 years? that's not necessarily a bad thing (maybe it's completed), but there are issues reported since 2019 like performance issues, etc, and a PR about archiving the repository because it's unmaintained

I did have an idea to implement some spatial navigation targets as data-attributes but I never got into it, to workaround the limitations of automatic spatial navigation. I also think that some elements should be accessible by pressing gamepad buttons (like L/R to switch between tabs instead of having to move the cursor, or the start button to open the top-right 3-dots menu, or some shortcut combination to focus the search, etc).

veryprofessionaldodo:
That's very reasonable. I still think it might be worth it, so I'll spend some time next week trying this out, as I'm going on vacation today. For now I'll close this MR to not cause more confusion :)

Issue #3274, opened 2023-11-29
Default wine prefix settings are confusing


arielj:
### Describe the bug

We have some problems with how we handle default wine prefixes:
- during installation, when a user clicks the `Use default wine settings` option, it makes the game use the default wine prefix which is `/Users/arielj/Games/Heroic/Prefixes/default`
- if the user does NOT click the `Use default wine settings` option, the game uses a prefix like `/Users/arielj/Games/Heroic/Prefixes/default/observer_`

This automatically breaks things, because the `default` prefix is a folder that contains other prefixes.

We also have some settings that are not super clear, like the `WinePrefix` option in Settings > Game Defaults. It should be more clear that that path will be the same for other games when checking that option in the install dialog and what that means.

Finally, in a new clean install of heroic, we have these settings in the config.json:
```
    "defaultWinePrefix": "/Users/arielj/Games/Heroic/Prefixes/default",
    "winePrefix": "/Users/arielj/Games/Heroic/Prefixes/default",
```

the first one is the folder where wine prefixes should go
the second one is the prefix path used with the `Use default wine settings` option

but from the keys it's impossible to know they are supposed to be different types of paths

### Add logs

```text
nothing relevant
```

### Steps to reproduce

1. With a clean heroic install
2. Open the install dialog of a game
3. the prefix path is `.../default/game title`
4. install it
5. open another install dialog
6. check the `Use default wine settings` option
7. install the game
8. look at the game's settings and see the prefix path is `.../default`

now that game will fail to start, the prefix going to be a problem

### Expected behavior

So I'm suggesting some changes:
- we have to differentiate the default prefixes dir (folder that holds prefixes) from the default prefix dir (the prefix path used by default). We do that in the constants file https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/constants.ts#L55 but that distinction gets lost eventually in the code and in setting keys.
  - the `Set Folder for new Wine Prefixes` in the general settings can set a `defaultWinePrefixesDir` configuration instead of the `defaultWinePrefix` configuration for example. The setting name is confusing cause it's meant to be a directory to hold prefixes and not one prefix.
- we can rename the `WinePrefix Folder` setting in Settings > Game Defaults to be `Shared Wine Prefix` (maybe even move it into Settings > General settings?)
[...]

PR #4927, opened 2025-09-12
Rename "Use Default Wine Settings" -> "Use shared Wine prefix"

CommandMC:
"Use Default Wine Settings" is (1) inaccurate and (2) doesn't really convey what the option does. In my opinion it sounds too much like a positive thing you always want to turn on, while in actuality you likely *don't* want it enabled

---


arielj:
linking https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3274 as this covers a bit of what I proposed there

arielj (approved):

Issue #4951, opened 2025-09-28
Steam shortcut have wrong executable when heroic is launched with electron manually.


patlefort:
### Describe the bug

When using for example the [AUR](https://aur.archlinux.org/packages/heroic-games-launcher) package, steam shortcut to games will point to for example `/usr/lib/electron36/electron` and will lead to an incorrect command `/usr/lib/electron36/electron --no-gui --no-sandbox "heroic://launch?appName=028263f54a874e35800d142275254d2c&runner=legendary"`.

Maybe there should be an environment variable to specify how to start heroic? Or maybe it should just use `xdg-open` like desktop shortcuts?

### Add logs

```text
(06:25:32) [INFO]:    [Backend]:         System Information: CPU: 32x AMD Ryzen 9 9950X 16-Core Processor
Memory: 200.02 GB (used: 51.36 GB)
GPUs:
  GPU 0:
    Name: Advanced Micro Devices, Inc. [AMD/ATI] Granite Ridge [Radeon Graphics]
    IDs: D=13c0 V=1002 SD=7e59 SV=1462
    Driver: amdgpu
  GPU 1:
    Name: Advanced Micro Devices, Inc. [AMD/ATI] Navi 21 [Radeon RX 6900 XT]
    IDs: D=73af V=1002 SD=5213 SV=1849
    Driver: amdgpu
OS: Arch Linux 6.16.9-1-cachyos-eevdf (linux)

The current system is not a Steam Deck
We are not running inside a Flatpak container

Software Versions:
  Heroic: 2.18.1 "Waterfall Beard" Jorul
  Legendary: 0.20.37 Exit 17 (Heroic)
  gogdl: 1.1.2
  comet: comet 0.2.0
  Nile: 1.1.2 Will A. Zeppeli
(06:25:32) [INFO]:    [Gog]:             Checking for existing gog manifests
(06:25:32) [INFO]:    [Backend]:         Shared Data: {"version":"2.18.1","gog":true,"epic":true,"amazon":false,"sideloaded":false,"providers":"gog, epic","OS":"linux","isFlatpak":false,"isAppImage":false,"isSnap":false,"isSteamDeckGameMode":false,"isSteamDeck":false}
(06:25:33) [WARNING]: [Backend]:         Failed to register protocol with OS.
(06:25:32) [INFO]:    [Connection]:      Connectivity: check-online
[...]

PR #4952, opened 2025-09-28
Add new environment variable: HEROIC_EXEC.

patlefort:
Add new environment variable: HEROIC_EXEC. This variable tells heroic what executable should be used to launch itself, if defined.
---

### DIFF
--- a/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
+++ b/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts
@@ -275,6 +275,10 @@ async function addNonSteamGame(props: {
       newEntry.StartDir = `"${process.env.PORTABLE_EXECUTABLE_DIR}"`
     }
 
+    if (process.env.HEROIC_EXEC) {
+      newEntry.Exe = `"${process.env.HEROIC_EXEC}"`
+    }
+
     newEntry.appid = generateShortcutId(newEntry.Exe, newEntry.AppName)
 
     await getIcon(props.gameInfo.app_name, props.gameInfo)


flavioislima:

patlefort:

arielj:
not sure I understand what is this for, there's no context and no explanation as to why this is needed/useful

patlefort:
Related issue: #4951 . With all due respect I disagree about "packaging issue" when it requires patching source code and it's a simple 3 lines fix.

arielj:
who would be in charge of setting that environment variable? the distro maintainer? the package maintainer? the user?

I don't think the distro maintainer is the right one cause they don't know which package the user is going to install

if it's the package maintainer, they can also apply a proper fix to keep thing working without this env like it's explained in that issue

if it's expected that the user would add this, then it's really bad UX cause the user would have to first know they are affected by this problem and then manually go and apply a workaround, and it's caused by the package maintainer, the official UX is that it just works

I think this env variable is a band-aid that would allow non-official package maintainers to not do the correct thing and throws the burden of fixing it to the heroic support (that would have to figure out this is the case and then instruct users) or to the user (to do the research or look for help)

I don't agree that this is a 3-lines fix, it's 3-lines workaround that doesn't really fix the problem unless something else is changed by the user to make it work

patlefort:
I'm not aware of any other way. You can't know how exactly your software was launched which is why I used an environment variable. It's not caused by anyone in particular, it's just how it work. There could be other uses, like wrapping heroic with firejail for security, especially when it is using many binaries downloaded from github. In that case, even the `-bin` package in the AUR would break but could make use of that variable.

Patching sources is supposed to be the last resort when upstream can't or won't fix or change things. It's obviously better to have a way supported by upstream.

PR #4965, opened 2025-10-02
[CI] Update AUR Workflow

flavioislima:
Update license and dependencies.

---

### DIFF
--- a/.github/workflows/release_aur.yml
+++ b/.github/workflows/release_aur.yml
@@ -32,7 +32,7 @@ jobs:
           latest_version=$(curl -s https://api.github.com/repos/Heroic-Games-Launcher/HeroicGamesLauncher/releases/latest | jq -r .tag_name)
           pkgver=${latest_version#v}
           pkgrel=1
-          url="https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/"
+          url="https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher"
           # Updated artifact naming convention with platform and architecture
           _filename=Heroic-${pkgver}-linux-x64.pacman
           source=("$url/releases/download/$latest_version/$_filename")
@@ -47,12 +47,13 @@ jobs:
           pkgdesc="An Open source Launcher for Epic, Amazon and GOG Games"
           arch=('x86_64')
           url="$url"
-          license=('GPL3')
+          license=('GPL-3.0-only')
           _filename=$_filename
           source=("$source")
           noextract=("$_filename")
           sha256sums=(${sha256sums[@]})
           options=(!strip)
+          depends=(which)
           provides=(heroic-games-launcher)
           conflicts=(heroic-games-launcher)
 


.github/workflows/release_aur.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           latest_version=$(curl -s https://api.github.com/repos/Heroic-Games-Launcher/HeroicGamesLauncher/releases/latest | jq -r .tag_name)
           pkgver=${latest_version#v}
           pkgrel=1
-          url="https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/"
+          url="https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher"
```
CommandMC: URLs usually go to more user-friendly websites, if available

```suggestion
          url="https://heroicgameslauncher.com/"
```

.github/workflows/release_aur.yml
```diff
@@ -47,12 +47,13 @@ jobs:
...
           arch=('x86_64')
           url="$url"
-          license=('GPL3')
+          license=('GPL-3.0-only')
           _filename=$_filename
           source=("$source")
           noextract=("$_filename")
           sha256sums=(${sha256sums[@]})
           options=(!strip)
+          depends=(which)
```
CommandMC, line 56: `which` is part of `base-devel`, meaning it is already expected to be installed when using AUR packages

flavioislima: Yes, there are some discussions on our AUR page if you want to check, I also think this is not necessary:
https://aur.archlinux.org/packages/heroic-games-launcher-bin

.github/workflows/release_aur.yml
flavioislima: Yes but this would break the download since it uses the `$url` bellow with the path to download the pacman file.

CommandMC: Right, well, I don't see why that can't be changed as well

.github/workflows/release_aur.yml
CommandMC: Quoting the comments here
> >   About the which package, it's part of base-devel and base-devel should be installed when using AUR : "Ensure base-devel is installed."

> That is for building packages; not for running packages, adding which to depends is correct

Hm, interesting point. I've always assumed that `base-devel` is expected to be present at runtime, but reading through AW now, I'm not sure

CommandMC: After asking around, it seems adding the dependency is the better option. Resolved

flavioislima: Yeah, so I think its no harm to have it. I already updated the PKGINFO there and so far no complains so I think we can update it here for the next release since it overwrites the one in there on the AUR workflow.

oech3:
Would you split PKGBUILD instead? #4989

.github/workflows/release_aur.yml
flavioislima: ok, done

PR #4968, opened 2025-10-04
Add HEROIC_GAME_SETTINGS and HEROIC_GAME_INFO env var to before/after launch script

UlyssesZh:
Add this env var to give the before/after launch script more possibilities.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1970,6 +1970,8 @@ async function runScriptForGame(
       HEROIC_GAME_RUNNER: gameInfo.runner,
       HEROIC_GAME_SCRIPT_STAGE: scriptStage,
       HEROIC_GAME_TITLE: gameInfo.title,
+      HEROIC_GAME_SETTINGS: JSON.stringify(gameSettings),
+      HEROIC_GAME_INFO: JSON.stringify(gameInfo),
       ...process.env
     }
     const child = spawn(scriptPath, {


flavioislima:

UlyssesZh:

UlyssesZh:
recheck

CommandMC:
I'm not sure how I feel about pre- and post-launch scripts getting this smart. Ideally if you need *this much* info, it's better to be implemented as an actual feature in Heroic
What's your use case for this?

UlyssesZh:
Having more information is better because more things are now possible with launch scripts. It does not matter what the actual use case is.

My use case involves using the Wine binary. I have my own system of cloud game save, and usually how I synchronize game save data with the cloud is to replace the game save files with symlinks to files in the dir synchronized with my cloud drive. This works for most games, but one game uses the registry to save data, and I should not simply synchronize the whole registry because there are other data in the registry that are not meant to be the same across different machines. In order to have a cloud save, I want to write a pre-launch script to import the cloud save data into the registry and a post-exit script to export the save data to the cloud save. I use Wine to do the registry manipulation, so the script needs to know what the correct version of Wine is.

CommandMC:
I see. My point was that, if you're creating a script this complex, it might be beneficial to implement into Heroic itself (since it's likely helpful to more people than just you). Since *this* is a pretty specific use case though, I do see why you'd want to keep this as pre- and post-launch scripts

BananaWorks07:
How big is the size of that json? Some windows apps run into issues if the environment block is too big. Technically this can be easily fixed on the wine side we can just filter out those env, but this change will only affect custom proton forks like proton-GE and not wine.

UlyssesZh:
It is only given to the pre-launch script. The actual Windows application does not see the env var.

PR #5012, opened 2025-11-05
[UI] Make categories and filters be clickable, not hoverable

veryprofessionaldodo:
Noticed that the behavior of the filter and categories was a bit cumbersome on a gamepad, since it automatically expanded when hovered, and navigating was hit or miss. Since this behavior was used in several places, I made a small component to handle the expansion logic, with custom gamepad rules for entering and exiting the children. 

I am already working on adding this component to simplify some CSS as well for the DotsMenu, but I wanted to confirm with you whether you like this approach. 

The only thing I couldn't fix (more complex than I thought :( ) was clicking with a mouse on the new Dropdown component. Before clicking only made the auto-expandable container sticky, but it didn't collapse. Now it doesn't collapse as well when clicking, but it does collapse when the container loses focus. 

I also simplified some CSS for the header that used a grid, but noticed it was a bit less flexible when testing on smaller screens. Also highlighted how it works in the video below:

<details>
Before:

https://github.com/user-attachments/assets/3cc1decd-94fd-45c1-928e-feab4f0ce11d

After:

https://github.com/user-attachments/assets/19ee964e-915e-46d1-af2b-c060b373f32f

</details> 


flavioislima:
@arielj do you remember if there is a reason for it to be the way it is? if related with the joystick navigation or the accessibility?

arielj:
> @arielj do you remember if there is a reason for it to be the way it is? if related with the joystick navigation or the accessibility?

well the reason to show the content on hover was so you don't have to click it to show it haha

I can see though how it could be kinda annoying if it's focused unintentionally when using a controller though so I think it's fine to disable that for a controller (I haven't checked the code, but I'd keep the `show it on hover` when using a mouse though)

veryprofessionaldodo:
@arielj Humm, I didn't think we could do a hover only solution for mouse, but I think there might be a way... Let me try it out and come back to you in a bit!

flavioislima:
> > @arielj do you remember if there is a reason for it to be the way it is? if related with the joystick navigation or the accessibility?
> 
> well the reason to show the content on hover was so you don't have to click it to show it haha
> 
> I can see though how it could be kinda annoying if it's focused unintentionally when using a controller though so I think it's fine to disable that for a controller (I haven't checked the code, but I'd keep the `show it on hover` when using a mouse though)

I think the issue is that it will add some complexity since it uses now a new component called Dropdown. I think it is fine to having to click on them instead of hover for these elements.

veryprofessionaldodo:
@flavioislima @arielj I'll do you one better, and support both :D One of the key reasons I wanted to make this into a component is to standardise behaviour across the whole UI, and if the component gets a new feature, it's everywhere all at once :) So here's the new behavior, I now added a flag `popUpOnHover` that I made `false` by default. 

Here's with set to `true`: 

<details>
https://github.com/user-attachments/assets/ce940576-4cc3-40f9-9191-814c32ee51f5
</details>

And here's with it set to `false`: 

<details>

https://github.com/user-attachments/assets/eb42b040-fc85-4f2f-9670-a717e2fed68e

</details>

As you can see in the videos, gamepad behavior is unaffected, and we can now use the same behavior as before. Also fixed and simplified CSS :) 

Edit: Not sure why videos aren't loading correctly, might be a Quicktime thing from the Mac screen recording. Going into the links does open up the video though

arielj:
some comments:
- it looks like the popUpOnHover prop is not used in the current state of this PR, and if I turn it on it affects both mouse and gamepad, not sure if it's fully done yet?
- one thing I noticed with the gamepad is that `up` and `down` are actually `tab` and `shift+tab` now instead of spatial navigation, so if you are focusing one of the filters, pressing down moves you to the `only` at the right of it instead of moving you to the next filter

veryprofessionaldodo:
@arielj regarding the comments:
- the popUpOnHover is not used, but since it is a new component, we might want to maintain consistency with existing behavior, so we could just add the prop to the <Dropdown> in the category and library filters and it would be similar to how it was before. That isn't working is strange, the videos I did in the comment above show how it was working. I think you might need to refresh the page when changing the popUpOnHover prop though
- yup, they were change to "tab" and "shift+tab" because as you could see in the "before" video on the description, navigating with gamepad (and keyboard) wasn't really accessible, since focus was stolen by other elements in the way of the dropdown. This way we can ensure that all the options are fully accessible, even if they have to pass the "only" in the toggle

Let me know if there's anything else I can do for this MR :)

arielj:
to clarify, when I set the prop to true it does work, but then it's showing the menu on hover and focus for both controller and mouse

what I'm saying is that, based on the comments, I expected the categories and filters to still show up the popup on hover when using the mouse but not when using the controller

now the way it works is that it either enables the popup or disables the popup for both input methods (it's not just showing the popup on hover, looks like it's also doing that on focus)

arielj:
> * yup, they were change to "tab" and "shift+tab" because as you could see in the "before" video on the description, navigating with gamepad (and keyboard) wasn't really accessible, since focus was stolen by other elements in the way of the dropdown. This way we can ensure that all the options are fully accessible, even if they have to pass the "only" in the toggle

maybe the `tab-index` attributes can be changed so the tab doesn't go to `only`

veryprofessionaldodo:
@arielj In that case it might be a bug, I'll double check the behavior and report back 👀  

I don't think setting the 'tab-index' attribute is a good idea here, because in that case we would be disabling accessibility of that feature to gamepad users and keyboard navigation. I did do [another MR](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5018) for this, but that is it's own separate conversation :)

veryprofessionaldodo:
@arielj Apologies, but I can't seem to replicate your scenario. In this clip I added the `popUpOnHover` for the categories button, but not the filters, and here's what I got. Initially used a mouse, then switched to gamepad navigation. For reference, I'm testing this on MacOS, but I don't think this would make any difference whatsoever


https://github.com/user-attachments/assets/a97a6fd9-0895-4c84-80d4-256c9ac9c59c

arielj:
This is what happens for me on linux:

first mouse, then keyboard, then controller

https://github.com/user-attachments/assets/cb1ba41b-307d-429a-a4cb-f168e61070a5

I guess it's something different between mac and linux, I'll try to debug it a bit later if I have time and also test on my mac

arielj:
> @arielj In that case it might be a bug, I'll double check the behavior and report back 👀
> 
> I don't think setting the 'tab-index' attribute is a good idea here, because in that case we would be disabling accessibility of that feature to gamepad users and keyboard navigation. I did do [another MR](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5018) for this, but that is it's own separate conversation :)

you could still move to the right to focus the `only`, the tab index would only affect the order the tab moves up and down

veryprofessionaldodo:
Very strange, this might very well be platform dependent. I don't have any active Linux development environment at the moment, but let me see if I can replicate this. I'll also take a look at tweaking tab-index ordering to see if I can replicate this, and report back

veryprofessionaldodo:
Doubly strange, I also can't replicate this on Linux, just tested now. This is on EndeavorOS KDE, so I'm very confused


https://github.com/user-attachments/assets/5e5d1c2d-afe4-4bc7-b18e-f046cb63ca56

veryprofessionaldodo:
Updated to include the Zoom platform changes, is there anything else I should do for this MR?

arielj:
> Updated to include the Zoom platform changes, is there anything else I should do for this MR?

I still think we should handle the `up/down` arrows to move in the correct direction by setting a tab index or by detecting what's the element above/below instead of moving to the `only` elements. It feels weird to press down and move to the right when in other places the navigation is spatial.

And I still think we have to keep the behavior of `open the menu when you hover with the mouse`, currently it's disabled in this PR.

arielj:
one thing I just noticed when testing this again is that sometimes I get 2 `only` actions displayed, even if I'm not hovering/focusing its corresponding line

image

it seems to happen when I'm focusing one of the `only` actions and I press left instead of up/down, it's like it gets stuck in a weird state (no idea why), it doesn't seem to happen on main

veryprofessionaldodo:
@arielj I apologise for the late reply, but I bring good news! I added in the popupOnHover to replicate the previous behavior of hovering, and I also removed the changes to the gamepad acting as a tab. This was a workaround and I wasn't really proud of the solution I gave, so I managed to fix the underlying issue with focus is escaping with navigation, so now it works for keyboards as well! Here's a video of it in action, now using just the keyboard for navigation: 

https://github.com/user-attachments/assets/a7842fe0-97eb-41ef-a7d9-a8d94f7d27f2

However, I didn't get to replicate the issue you found with the "only" appearing twice, but that might've been fixed regardless. Could've been a change with the hover behavior, and I did change that part in the CSS

flavioislima:
@arielj added this one to the next milestone so would be good to try again when you have the time.

veryprofessionaldodo:
Let me know if there's anything else I can do to help testing this!

arielj:
> Let me know if there's anything else I can do to help testing this!

I see the PR has conflicts, can you check that?

veryprofessionaldodo:
@arielj Fixed now!

arielj:
thanks!

PR #5014, opened 2025-11-06
[FIX] Time container not updating after finishing playing  

flavioislima:
Small fix to the time container.
In Main it won't update the time after finishing playing since it was not doing another request after it.
---

### DIFF
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -65,6 +65,7 @@
         "import": "Import Game",
         "install": "Install",
         "no-path-selected": "No path selected",
+        "play": "PLAY",
         "queue": {
             "remove": "Remove from Queue"
         },


CommandMC (approved):
I'm not much of a fan of how timestamp updates work in general right now. Ideally we'd push changes to the Frontend from the Backend using `sendFrontendMessage` (with the Frontend then updating a Zustand store).
That would be a lot more work to implement though, so this is fine for now

veryprofessionaldodo:
Would this be the reason why I don't have accurate time data for games I played through Heroic?

CommandMC:
No, this is purely a visual issue

veryprofessionaldodo:
@CommandMC Thank you for the quick reply! In that case I'll investigate this one later today, I think I have an idea where to start looking

flavioislima:
Agree that this should be improved in the future. But for now this is a small fix to a small issue.

PR #5018, opened 2025-11-06
[UI] Simplify filters and categories

veryprofessionaldodo:
Simplified the filters and categories logic, because I think it's significantly simpler to select the filters we actively want over de-selecting those we don't want. This will also work make gamepad navigation even simpler. 

An example use case, since I use a Mac I might want to filter by only Mac games. Previously, to do that I would have to de-select "Windows" and "Browser". Now I can just select Mac. If I do want multiple filters it is more work to select them, but since they're filters, I think most people would prefer to be more restrictive by default. The same thinking applies to Categories as well :)

Here's a video of it in action, I didn't capture the categories but it's exactly the same logi:

<details>

https://github.com/user-attachments/assets/1809c560-9526-4aba-bb70-d8d6331a0512

https://github.com/user-attachments/assets/e24b6651-01e3-4461-a7fc-112de0801c60

</details>

### DIFF
--- a/src/frontend/components/UI/CategoryFilter/index.tsx
+++ b/src/frontend/components/UI/CategoryFilter/index.tsx
@@ -24,47 +24,16 @@ export default function CategoryFilter() {
     }
   }
 
-  const setCategoryOnly = (category: string) => {
-    setCurrentCustomCategories([category])
-  }
+  const resetCategories = () => setCurrentCustomCategories([])
 
-  const selectAll = () => {
-    setCurrentCustomCategories(
-      ['preset_uncategorized'].concat(customCategories.listCategories())
-    )
-  }
-
-  const toggleWithOnly = (
-    toggle: JSX.Element,
-    onOnlyClicked: () => void,
-    category: string
-  ) => {
-    return (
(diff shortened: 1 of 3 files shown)


arielj:
I think changes like this should be discussed first on discord before you change how some feature works, we have a UX channel to talk about these things.

I'm not sure I see the benefit of this change, and now the `only` option is not there so if I want to select a single category after having 4 selected I have to unselect the other 3 (or even the other 4 if the one I want is not one of those)

veryprofessionaldodo:
Apologies, wasn't sure about the UX channel, is this something on Discord?

arielj:
> Apologies, wasn't sure about the UX channel, is this something on Discord?

yes, it's called ux-ui-design

veryprofessionaldodo:
Since this was more meant to work in tandem with another MR for the [gamepad behavior](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5012), but I ended reverting the gamepad using "tab" and "shift tab" for dropdown navigation, I'm going to close this MR as it's no longer needed

PR #5028, opened 2025-11-14
Update README to include Zoom Platform support in the list of game stores

neuromancer:
This small PR updates the list game stores to include the Zoom Platform.


CommandMC:
I think it may make sense to drop explicitly naming our supported storefronts in the README now, and using something like "multi-storefront" instead

neuromancer:
As this is a small change and it is unclear how to reformulate the README right now, I will close it.

PR #5032, opened 2025-11-15
[FIX] Make sure Zoom Platform code is setting up the umu support

neuromancer:
The Zoom Platform store code was not setting correctly the umu env variables, and therefore, proton fixes failed to run. This was tested with the [Duke Nukem Manhattan Project game](https://www.zoom-platform.com/product/duke-nukem-manhattan-project), which absolutely require this fixes to run.

---

### DIFF
--- a/src/backend/storeManagers/zoom/games.ts
+++ b/src/backend/storeManagers/zoom/games.ts
@@ -691,7 +691,6 @@ export async function launch(
 
     if (await isUmuSupported(gameSettings)) {
       const umuId = await getUmuId(gameInfo.app_name, gameInfo.runner)
-      console.log('Umu ID for', appName, ':', umuId)
       if (umuId) {
         commandEnv['GAMEID'] = umuId
       }


src/backend/launcher.ts
```diff
@@ -1476,7 +1476,7 @@ async function runWineCommand({
 
   const env_vars: Record<string, string> = {
     ...process.env,
-    GAMEID: 'umu-0',
+    ...options?.env,
```
imLinguin, line 1479: Maybe we should fallback to empty object in case options are undefined, otherwise this may cause a runtime error. Correct me if Im wrong though

neuromancer: I believe umu/protonfixes handles that correctly even if there is no GAMEID variable (and using `umu-0` will not match any known game).

imLinguin: yea, new builds of umu handle that gracefully, I mean Heroic itself. if we run spread operator on null or undefined it will fail I believe

neuromancer: I think this is only for arrays, but not for objects, but I'm not a js/ts expert, so don't quote me on this :sweat_smile:

CommandMC: Using the spread operator on `undefined`/`null` is fine:
```
~/ node
Welcome to Node.js v22.21.1.
Type ".help" for more information.
> { ...undefined }
{}
> { ...null }
{}
```
If it wouldn't be fine, TypeScript would have a check for it

neuromancer:
is there any other feedback for this PR? Can it be merged otherwise?

imLinguin (approved):

src/backend/storeManagers/zoom/games.ts
```diff
@@ -687,12 +689,21 @@ export async function launch(
       return false
     }
 
+    if (await isUmuSupported(gameSettings)) {
+      const umuId = await getUmuId(gameInfo.app_name, gameInfo.runner)
+      console.log('Umu ID for', appName, ':', umuId)
```
CommandMC:
```suggestion
```

CommandMC (approved):
LGTM other than one nit

neuromancer:
> LGTM other than one nit

Done!

neuromancer:
Please don't forget to merge if this PR is ready 😇

PR #5033, opened 2025-11-15
[FIX]: address some issues with Zoom Platform implementation

imLinguin:
This is a WIP PR that addresses a few issues I encountered so far:
- incorrect paggination handling for game library
- incorrect download progress reporting
- windows installers letting users to select a different path
- override executable not being respected

I'm still in a process of fine-tuning the installer as it doesnt seem to respect these properties


neuromancer:
Do you want to include the fix from #5032 in this one to avoid any possible conflicts as well?

src/backend/storeManagers/zoom/games.ts
```diff
@@ -345,23 +302,36 @@ export async function install(
...
     // windows
     // For Windows, use runWineCommand
     const gameSettings = await getSettings(appName)
-    const winePrefix = gameSettings.winePrefix
-    const zoomPlatformPath = join(winePrefix, 'drive_c', 'ZOOM PLATFORM')
-    if (fs.existsSync(zoomPlatformPath)) {
-      confFilesBefore = await findConfFiles(zoomPlatformPath)
+    await writeFile(
+      infFilePath,
+      `[Setup]\nLang=english\nDisableWelcomePage=yes\nDisableDirPage=yes\nDisableProgramGroupPage=yes\nDisableReadyPage=yes\n`,
```
arielj, line 316: this doesn't seem to work, I understand it should, for example, not show the language selection step, but it's shown anyway

arielj (approved):
this seems to work fine, I made another PR on top of this one too

it has some conflicts but I think we should go ahead and merge this when that's fixed

PR #5043, opened 2025-11-21
Center Loading/UpdateComponent when the app is loading for the first time

arielj:
This is a small fix for an issue I noticed after merging https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4808

The initial `Loading` state while the application loads was not centered in the window, it was aligned to the top. This fixes that.

---

### DIFF
--- a/src/frontend/index.scss
+++ b/src/frontend/index.scss
@@ -99,3 +99,10 @@ body {
   margin-inline: var(--space-xs) 0;
   color: var(--text-secondary);
 }
+
+#root:has(> .UpdateComponent) {
+  height: 100vh;
+  & > .UpdateComponent {
+    height: 100vh;
+  }
+}


veryprofessionaldodo:
Not a part of the team, but tested it locally and it looks good to me :) Edit: Made just a small comment

src/frontend/index.scss
```diff
@@ -99,3 +99,10 @@ body {
   margin-inline: var(--space-xs) 0;
   color: var(--text-secondary);
 }
+
+#root:has(> .UpdateComponent) {
+  height: 100vh;
+  & > .UpdateComponent {
+    height: 100vh;
+  }
+}
```
veryprofessionaldodo: Couldn't this be simplified into [this](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5058/files#diff-0243628f3f8faa299456e10249ef062cc6f53901d4327937e04d676bcb28c63dR2) though? It looks like the 100% there is not correct, since switching to 100vh like you have it here works as well

Issue #4617, opened 2025-06-06
[Linux]Heroic Downloading glitch if client closed mid download


tr1xem:
### Describe the bug

I was trying to download from epic games tr goty and i closed client mid download and when i restarted client i click on install (be there is no sign it had been downloaded a bit) but its not downloading it 
Something like this happening 
[Video](https://cdn.arch-linux.fun/%F0%9F%93%B8/75wecn8f.mp4)

Turn out i have to kill a process named legendary

Possible fix : make heroic and legendary close at same time? or just minimize it to tray while closed when downloading (at least take a confirmation)

### Add logs

```text
- n/a -
```

### Steps to reproduce

Close the launcher mid download and then reopen it 

### Expected behavior

Should start downloading where it stopped

### Heroic Version

Latest Stable

### System Information

- OS : Cachyos
- WM : Hyprland

arielj:
Heroic should already stop the legendary process when fully closed, are you sure you were closing heroic and not just minimizing it to the tray area?

also heroic should already resume the download when open after it was downloading something (not sure if it works if you manually kill the legendary process, if it's stopped by heroic it should work)

tr1xem:
@arielj I knew you wont believe me, Well here is the proof of whole issue

[Video](https://r2.e-z.host/68f0782e-ab00-4eda-a7e2-8cdad0dd1281/1jegvagz.mp4)

PR #5050, opened 2025-11-22
[UX] Warn about installation in progress when exiting Heroic

arielj:
We used to have this feature but at some point it got lost.

When closing Heroic, if a download is in progress we warn the user first instead of just closing Heroic.

Currently, since that code we used to have is not running, closing Heroic mid download does not kill the current downloads and the continue running in the background.

This PR does NOT fix what happens if Heroic is killed (like with `kill -9`), because that's an extreme case where we can't really show a warning to the user (and I don't really know what's the solution there).

This closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4617

---

### DIFF
--- a/src/backend/downloadmanager/downloadqueue.ts
+++ b/src/backend/downloadmanager/downloadqueue.ts
@@ -319,5 +319,6 @@ export {
   getQueueInformation,
   cancelCurrentDownload,
   pauseCurrentDownload,
-  resumeCurrentDownload
+  resumeCurrentDownload,
+  isRunning
 }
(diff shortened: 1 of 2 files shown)


arielj:
> Part of me wants to look at why the download manager is not setting `isLocked` (anymore?). Another part of me just wants to refactor the download manager code (so this PR not really touching things in there is ideal)

Yeah, I also thought about refactoring how we keep track of this `locked` state many times (using some backend events to decouple things a bit), but I always end up in weird rabbit holes so for this one I just wanted a quick fix, we can decouple it properly later haha

PR #5055, opened 2025-11-28
[FIX] Remove DXVK fps limit

Etaash-mathamsetty:
DXVK frame rate limiter env variable was recently removed in DXVK


imLinguin (approved):

flavioislima:
@Etaash-mathamsetty I just wonder if this keeps working for macOS because I believe that DXVK there is still an old version, at least when using wine instead of GPTK.

arielj:
> @Etaash-mathamsetty I just wonder if this keeps working for macOS because I believe that DXVK there is still an old version, at least when using wine instead of GPTK.

yes, dxvk on mac still support this feature https://github.com/Gcenx/DXVK-macOS?tab=readme-ov-file#frame-rate-limit (and the alternative to use mangohud I think it's not an option for mac, so I think we should keep it but just for macos non-native games)

arielj:
I made https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5195 just disabling the dxvk fps limit on windows but keeping it for mac

PR #5059, opened 2025-12-02
[Input] Improved gamepad key repeat to make navigation snappier

veryprofessionaldodo:
Noticed gamepad input was repeating too sluggishly making navigating the library a bit cumbersome or spammy. Added in base activation delay to stick and pad inputs to make navigation feel snappier, while retaining precision. Videos show it better than words, before shows how it's currently being done, after shows me alternating between using pad and joystick navigation:

Before:

https://github.com/user-attachments/assets/9d49c119-13d9-418c-82f2-e0c45e3ca0d9

After:

https://github.com/user-attachments/assets/f4724542-643f-4040-b5c3-6dbb503d77ea

Let me know if there's something you want me to change  :)

---


src/frontend/helpers/gamepad.ts
```diff
@@ -35,19 +35,31 @@ export const initGamepad = () => {
...
+  // across different controller indexes
+  const basicGamepadInputRpeat = () => {
+    return JSON.parse(
+      JSON.stringify({
+        triggeredAt: {},
+        repeatDelay: KEY_REPEAT_DELAY,
+        activationDelay: ACTIVATION_DELAY
+      })
+    )
+  }
```
CommandMC: Since you construct the object in a function, you can just return the object itself
```suggestion
  const basicGamepadInputRpeat = () => ({
    triggeredAt: {},
    repeatDelay: KEY_REPEAT_DELAY,
    activationDelay: ACTIVATION_DELAY
  })
```

In cases where this isn't possible, the [`structuredClone`](<https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone>) method would still be faster than JSON

veryprofessionaldodo: Didn't know about that, thank you! Added the feedback in the latest commit :)

CommandMC (commented):
I wonder whether we want to make the repeat delay configurable (that doesn't need to be in this PR, just thinking out loud). Before, it definitely was too slow. With the changes here, it's a *little* too fast for my taste (but I assume for other people it'll be fine)

I don't entirely understand the gamepad input handling code, so I don't feel comfortable giving this an approval just yet. @arielj if you could spare some time?

veryprofessionaldodo:
Interesting, you're right, I could have considered that as well! I can try to find out gamepad repeat defaults in other places like the PS5 to make it match things other people might be more used to

veryprofessionaldodo:
@CommandMC I took your suggestion on, and added a slider to customize the delays. I've also created a new component that I plan on using on another MR related to Game Card UI scalings. Here's it looking in action, let me know if you want anything else! I do intend on adding the translations for other files, but after you agree that it's a good enough approach:


https://github.com/user-attachments/assets/80ad95e4-4377-43f2-b2c4-90d18bb81067

veryprofessionaldodo:
@CommandMC Apologies for having abandoned this MR! Fixed the lint, synced with main, should now be good to go!

flavioislima:
> @CommandMC I took your suggestion on, and added a slider to customize the delays. I've also created a new component that I plan on using on another MR related to Game Card UI scalings. Here's it looking in action, let me know if you want anything else! I do intend on adding the translations for other files, but after you agree that it's a good enough approach:
> 
>  Gravacao.do.ecra.2026-01-26.as.20.08.58.mov

can you move those settings to the advanced settings?

veryprofessionaldodo:
@flavioislima Good suggestion, moved! 

imagem

veryprofessionaldodo:
@flavioislima Could this be merged? I fixed the conflicts with the main branch, and when using console mode especially I really miss the snappier input feel

CommandMC (approved):

PR #5064, opened 2025-12-03
[UI] Tabs in Game Page now more accessible with gamepad and keyboard

veryprofessionaldodo:
Noticed that the tabs in the game page were simultaneously not displaying focus correctly, and also not handling input very well. This is particularly bad when navigating with a keyboard or gamepad where you kinda lose track of where you are and what you're doing. Apparently, this was being caused because of the "scrollable" variant not playing nicely with the tabs. I also modified the tabs to add the `selectionFollowFocus` because I feel like it lends itself very well to navigation / gamepad driven inputs. 

Before: 

https://github.com/user-attachments/assets/885965af-8c7c-446b-9c14-4acdce83a794

After:

https://github.com/user-attachments/assets/185a83bd-3d95-447b-afb4-7c96b6b23b7f

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.css
+++ b/src/frontend/screens/Game/GamePage/index.css
@@ -30,7 +30,7 @@
 
   .MuiTab-labelIcon {
     min-height: unset;
-    align-items: flex-start;
+    align-items: center;
   }
 
   & > p {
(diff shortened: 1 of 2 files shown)


veryprofessionaldodo:
@CommandMC Yeah, I later found out that you need the window to be a bit smaller for the issue to arise. If you zoom the page in a bit it should happen, or if you try it on a Steam Deck for example

PR #5067, opened 2025-12-04
[Steam]: Show steam library in Heroic

imLinguin:
Well.. people asked for it.

This feature relies on Steam itself to handle games installations. All it does is show games in Heroic's UI.


a5ehren:
Not a maintainer, but it doesn't really feel like "read the VDF from a Steam installation and show the list" is really in the spirit of this project?

Especially when there are things like [node-steam-user](https://github.com/DoctorMcKay/node-steam-user/tree/master) that can be leveraged to do authentication and fetch the library from the server without the helper programs that HGL has to use for everything else.

imLinguin:
The idea for integrations in Heroic was to always let people interact with given game store without a need for official client, and on platforms such as Linux or Mac where often the support is limited or not existent.

However given the fact how some people use Heroic in a library aggregation use-case, expecting to see all their games in one place, and we were receiving requests to support Steam as well - this PR was born.

As per your point
> Especially when there are things like [node-steam-user](https://github.com/DoctorMcKay/node-steam-user/tree/master) that can be leveraged to do authentication and fetch the library from the server without the helper programs that HGL has to use for everything else.

Given how complex Steam client is and that it works perfectly fine on Linux systems, I still see no real incentive to pursue completely replacing the official client or parts of it. I think handling Steam login in Heroic is an unnecessary step, as we would still delegate installation and launch process of Steam games to the official client.

That said, this Pull Request is quite mature proof of concept, I'm open to discussion on this topic whether other maintainers and community feel strongly about supporting Steam or if it doesn't fit the project that well.
I'm personally torn between two sides
1. Seeing Steam games among others in one library is a valid use-case
2. Steam integration really stands out from the rest of supported platforms in terms of supported features and UX

Po11da:
It's a tough case, since Steam is much more feature-rich and it greatly depends on the user. One may be achievement maxer, the other may not care about achievements at all but their favourite games have rich, Workshop-only modding community.
I think UX consistency should be a priority, with backend playing secondary role. To recall, the biggest selling points are (in my opinion, at least):
- pleasing, intuitive, consistent interface
- access to multiple game libraries in single view, with support for extra features such as cloud saves or achievements
- lightweight, doesn't have to run full launchets.
The main approaches to integrate Steam are:
1. Don't do it. Maintain status quo. But many players rely solely on Heroic ans Steam to launch games and would benefit from even simple integration.
2. Use native launcher in the background. Conflicts with current approach but offers full features, though they have to be accessed directlt from Steam. Should be easiest to implement.
3. Use community reverse-engineered launcher. It will lack support for some features, however.
4. Let users choose. Win-win except it introduces huge maintenance burden.
5. Use community cli, but work on the same directories Steam uses by default (or just make a symlink for entire steamapps?) The main use case would be Workshop, can't really think of any other.
Looking at the state of Open Steam Client, I'm inclined to option 2 and reconsider 3 once it (or another community client) matures.

teodorgross:
Since this PR has been sitting with merge conflicts for a while and the rebase seemed to be the main blocker, here is one: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/compare/main...teodorgross:HeroicGamesLauncher:steam-integration

It is your exp/steam branch squashed and rebased onto current main. The original predates the class based store manager refactor, so I wrapped the existing module functions in SteamGame and SteamLibraryManager classes, with the sideload manager as reference. Your original work is preserved in the first commit with you as the author.

I tested it on Nobara with a real account and a library of 194 games. Library detection, covers, installs and launches through the Steam client, and the game page integrations like ProtonDB and the Steam Deck rating all work. While testing I found and fixed a few things:

1. Installs from the library and the game page opened the regular install modal, which waits forever on a download size that never comes. The steam check now lives in openInstallGameModal so every entry point routes to the Steam client.
2. The empty cache fallback in GlobalState refreshed zoom instead of steam.
3. Playtime now comes from the enabled users' localconfig.vdf, wired through the same handler GOG already uses. Installed version and platform come from the manifest buildid and the appinfo oslist.
4. Game descriptions, genres and release dates come from the public appdetails endpoint, with caching.
5. The launch log now explains that Steam owns the process instead of showing an empty log.
6. One find is unrelated to Steam: handleExperimentalFeatures assigns the whole features object to zoom.enabled, so flipping any toggle in the experimental settings enabled the Zoom UI until the next restart. This exists on current main and is fixed in its own commit.

Following what arielj did for Zoom, the whole integration now sits behind a new Experimental Features toggle, default off. I also did a consistency pass: everything logs through the logger with the Steam prefix, refresh survives appids that are missing from appinfo.vdf, the library map is cleared on resync so revoked licenses drop out, offline starts resync once the app comes online, and uninstalls are handed off to Steam without wiping local settings. prettier, ts-prune, tsc, eslint and the jest suite all pass.

I am happy to keep testing this on my setup and to help maintain it while it lands, in whatever form is useful to you. If you would rather cherry pick parts of it into your own branch, everything is split into small commits.

For transparency: I built this together with Claude, the coding assistant from Anthropic. I reviewed and tested everything myself on real hardware.

PR #5075, opened 2025-12-08
Ignore special characters while searching and filtering

AMS21:
Previously when typing "Baldurs Gate" into the search bar you would get no result. That's because the Game is actually called "Baldur's Gate" the same is true for "Stalker" since the games start with "S.T.A.L.K.E.R.".
So with this patch we ignore all non word characters while searching or filtering.

---

### DIFF
--- a/src/frontend/components/UI/LibrarySearchBar/index.tsx
+++ b/src/frontend/components/UI/LibrarySearchBar/index.tsx
@@ -5,6 +5,7 @@ import { GameInfo, Runner } from 'common/types'
 import SearchBar from '../SearchBar'
 import { useTranslation } from 'react-i18next'
 import LibraryContext from 'frontend/screens/Library/LibraryContext'
+import { normalizeTitle } from 'frontend/helpers/library'
 
 function fixFilter(text: string) {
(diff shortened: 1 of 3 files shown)


flavioislima:

AMS21:

src/frontend/components/UI/LibrarySearchBar/index.tsx
```diff
@@ -11,6 +11,14 @@ function fixFilter(text: string) {
   return text.replaceAll(regex, '')
 }
 
+const normalizeInput = (text: string) =>
+  fixFilter(text)
+    .replace(/[-.,:']/g, '')
+    .toLowerCase()
+
+const normalizeTitle = (text: string) =>
+  text.replace(/[-.,:']/g, '').toLowerCase()
```
CommandMC: Please factor out the `.replace(...).toLowerCase()` part of these functions into its own helper function & put it into `frontend/helpers/`.

Right now, we have the same RegEx 3 times. In case this feature is changed in the future, it's easy to forget to update one of them

CommandMC (approved):

PR #5078, opened 2025-12-08
Update workflow OS versions

CommandMC:
Updating to
- Windows Server 2025
- macOS 15 (this also changes the runner's CPU architecture from x86_64 to ARM, although both versions are still built)
- Ubuntu 24.04

This resolves an issue with the macos-13 runner no longer being available

---


imLinguin:
I think we could be able to just use -latest os, but either way is fine

PR #5083, opened 2025-12-11
[Fix] Update package overrides to address vulnerabilities

flavioislima:
This pull request updates the `pnpm-workspace.yaml` file to add an `overrides` section, specifying minimum versions for several dependencies. This helps ensure that the workspace uses secure and compatible versions of important packages.

Dependency version enforcement:

* Added an `overrides` section to `pnpm-workspace.yaml` to enforce minimum versions for key dependencies such as `ts-morph`, `got`, `cross-spawn`, `brace-expansion`, `electron`, `@eslint/plugin-kit`, `vite`, `axios`, `playwright`, `tmp`, `form-data`, `js-yaml`, and `glob`. This helps address security vulnerabilities and compatibility issues.

Those were generated using `pnpm audit --fix`

---

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -13,3 +13,22 @@ onlyBuiltDependencies:
   - esbuild
   - register-scheme
   - unrs-resolver
+
+overrides:
+  ts-morph: 17.0.1
+  got@<11.8.5: '>=11.8.5'
+  cross-spawn@<6.0.6: '>=6.0.6'
+  brace-expansion@>=1.0.0 <=1.1.11: '>=1.1.12'
+  brace-expansion@>=2.0.0 <=2.0.1: '>=2.0.2'
+  electron@<28.3.2: '>=28.3.2'
+  '@eslint/plugin-kit@<0.3.4': '>=0.3.4'
+  electron@<35.7.5: '>=35.7.5'
+  vite@>=6.0.0 <=6.3.5: '>=6.3.6'
+  axios@>=1.0.0 <1.12.0: '>=1.12.0'
+  vite@>=6.0.0 <=6.4.0: '>=6.4.1'
+  playwright@<1.55.1: '>=1.55.1'
+  tmp@<=0.2.3: '>=0.2.4'
+  form-data@>=4.0.0 <4.0.4: '>=4.0.4'
+  js-yaml@<3.14.2: '>=3.14.2'
+  js-yaml@>=4.0.0 <4.1.1: '>=4.1.1'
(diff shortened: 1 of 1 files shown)


imLinguin (approved):

pnpm-workspace.yaml
```diff
@@ -13,3 +13,22 @@ onlyBuiltDependencies:
...
+
+overrides:
+  ts-morph: 17.0.1
+  got@<11.8.5: '>=11.8.5'
+  cross-spawn@<6.0.6: '>=6.0.6'
+  brace-expansion@>=1.0.0 <=1.1.11: '>=1.1.12'
+  brace-expansion@>=2.0.0 <=2.0.1: '>=2.0.2'
+  electron@<28.3.2: '>=28.3.2'
+  '@eslint/plugin-kit@<0.3.4': '>=0.3.4'
+  electron@<35.7.5: '>=35.7.5'
```
arielj, line 25: I'm not sure I understand exactly what this is doing, why do we have some packages listed twice? and if we are already defining `"electron": "^36.9.4",` in the `package.json` file, are these 2 electron overrides needed?

arielj: maybe there's something I don't understand about how pnpm resolves which version of each package should be used (I think it allows having more than one version of the same package during the build process?)

flavioislima: Good question, I wonder if it is because of this dependency of the dev tools package (which we can probably remove since it wont work anymore):
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/a95fde1a31a07de1f39f3cc11c9f2cd6c7c7a8d5/pnpm-lock.yaml#L11566

arielj: I'm not sure if this is the right thing to do with overrides

like... we could just update axios in the package.json declaration (and the lockfile too) and be explicit in our dependencies instead of having `^1.9.0` in the package.json and then using an override in another file to get it to `>=1.12.0`

arielj: I think the electron dependency of react-devtools only affects the little popup with the react devtools that shows up with the `pnpm debug:react` command, I don't think it's dangerous that it's using an older electron (it's just a dev tool and only executed on demand, there's no risk) and I'd rather keep the react debugger than removing it if this overrides breaks it for example

a5ehren: I have a shorter version of this in another branch that I haven't made a PR for yet, and I suspect that `pnpm dedupe` would eliminate the need for a few of those too.

PR #5087, opened 2025-12-12
[chore] Cleanup some obsolete/unused devDepends

a5ehren:
`react-beautiful-dnd` and `unimported` have been abandoned by their creators.

`@hello-pangea/dnd` is a direct fork with some minor React 18 updates, so switch to that. 

`unimported` is not actually used anywhere in this project, so just delete it.

`react-fontawesome` says that the 0.2.2 version is obsolete and that anyone on v6 or v7 of fontAwesome should move to 3.1.1, so do that.

Also fixes a typescript compile error with the time-played commit right before this one.
---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ x ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #5094, opened 2025-12-15
[Fix] TimeContainer conflicts

arielj:
This PR didn't have any conflict https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/commit/a95fde1a31a07de1f39f3cc11c9f2cd6c7c7a8d5

But it was merged after my PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5048/files that changed the signature of the `hasStatus` hook.

I wonder if we should have a git rule to ensure the PRs are always rebased on top of the latest main to allow merging them? that could catch this type of issue.

This is also addressed in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5087 but I'd rather have this fix in one PR that is not changing the dependencies so we can merge this independently and unlock all other PRs

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -421,9 +421,7 @@ export default React.memo(function GamePage(): JSX.Element | null {
                       />
 
                       <Description />
-                      {!notInstallable && (
-                        <TimeContainer runner={runner} game={appName} />
-                      )}
+                      {!notInstallable && <TimeContainer gameInfo={gameInfo} />}
                       <GameStatus
                         gameInfo={gameInfo}
                         progress={progress}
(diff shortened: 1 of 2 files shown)


CommandMC:
> I wonder if we should have a git rule to ensure the PRs are always rebased on top of the latest main to allow merging them? that could catch this type of issue.

This would get annoying fast. Imagine every commit to main introducing essentially a manually-resolvable merge conflict on every PR

We could use a custom GitHub Action to verify that the commands used in the codecheck hook also succeed after the PR changes are rebased on main. I believe that would more accurately cover what we want here

arielj:
> > I wonder if we should have a git rule to ensure the PRs are always rebased on top of the latest main to allow merging them? that could catch this type of issue.
> 
> This would get annoying fast. Imagine every commit to main introducing essentially a manually-resolvable merge conflict on every PR
> 
> We could use a custom GitHub Action to verify that the commands used in the codecheck hook also succeed after the PR changes are rebased on main. I believe that would more accurately cover what we want here

yeah I agree it would be annoying

any alternative would be fine for me, it's just that I've seen this issue happening many times

a5ehren:
Yeah, my fix is just a hack to let me pass the pre-commit checks. Yours is better

PR #5097, opened 2025-12-16
[chore] Update all github actions to latest versions

a5ehren:
The vast majority of the updates are to ensure compatibility with the latest github runner images, which default to NodeJS 24 instead of 20.

There is also a minor change to the install-deps action to install pnpm before node as that is what the node action recommends. Updated the versions of NodeJS and pnpm in package.json to make sure the setup actions (and developer tools) grab versions that match Electron 36.

Also enabled pnpm package caching to take advantage of new features in the node action and hopefully reduce build times.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,9 +19,9 @@
     "email": "heroicgameslauncher@protonmail.com"
   },
   "engines": {
-    "node": ">= 16"
+    "node": ">=22"
   },
-  "packageManager": "pnpm@10.9.0",
+  "packageManager": "pnpm@10.26.2",
   "scripts": {
     "start": "electron-vite dev --watch",
     "debug:react": "pnpm start & react-devtools",
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -3,12 +3,13 @@ description: 'Installs dependencies (Node itself, node_modules, node-gyp, helper
 runs:
   using: 'composite'
   steps:
-    - uses: actions/setup-node@v3
-      with:
-        node-version: '20'
-    - uses: pnpm/action-setup@v3
+    - uses: pnpm/action-setup@v4
       with:
         standalone: ${{ runner.os == 'Windows' }}
+    - uses: actions/setup-node@v6
+      with:
+        node-version-file: 'package.json'
+        cache: 'pnpm'
     - name: Install node-gyp
       run: pnpm add --global node-gyp
       shell: bash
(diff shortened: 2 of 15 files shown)


a5ehren:
The CI errors are download problems in the pre-req steps. They need to re-run

package.json
```diff
@@ -19,9 +19,9 @@
     "email": "heroicgameslauncher@protonmail.com"
   },
   "engines": {
-    "node": ">= 16"
+    "node": "^22.14.0"
```
flavioislima: Do we need this change? maybe lets do it like before but for 22:

```suggestion
    "node": ">= 22"
```

a5ehren: The thinking is that Electron 36 shipped with Node 22.14, and they won’t be moving to 24.x until Electron 40, so I restricted the maximum to make sure people don’t make changes that are not compatible.

CommandMC: A bit of an uninformed comment: How is this field enforced? Given Electron ships with its own Node version, will this field prevent me from working on the project if my package manager sees my system installation of Node 24?
In other words, why do we need this field if Electron has its own Node & our `@types/node` package is (I assume) at the correct version?

niStee:
Just to clarify, couldn't Dependabot or Renovatebot do this?

package.json
a5ehren: That's a good point. I don't know what mechanism (if any) Electron has for this kind of thing.

You're correct that this change would block your system install of Node 24 from being used to build/lint/whatever. It looks like best practice is to have the engines field set to >= the Major NodeJS version bundled with the version of electron used in the project, so I will change this to >=22 as suggested.

a5ehren:
> Just to clarify, couldn't Dependabot or Renovatebot do this?

Yes, but this project doesn't have either of those configured.

.gitignore
```diff
@@ -37,3 +37,9 @@ flatpak-build
 .env
 
 playwright-report
+
+# AI workspace settings?
+CLAUDE.md
+.claude/
+.cursor/
+
```
CommandMC: These changes seem unrelated to the rest of the PR. Personally I just wouldn't make them (if people dump files into their tree, or use programs that do, that's up to them to clean up). If you do want them included, please add them in a separate PR

CommandMC (changes requested):

.gitignore
a5ehren: k

imLinguin (approved):

CommandMC (approved):

PR #5113, opened 2025-12-23
Simplifying snapcraft.yaml

Kyuyrii:
I simplified the snapcraft.yaml file; it was absurdly large and complex before.

Now it's easier to build the Snap version, as it will use the .deb version as a base.

I was able to build, install, log into my account, download a game, and launch it.

---

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -51,6 +51,7 @@ apps:
       - bluez
       - udisks2
       - system-backup
+      - umu-support
 
 plugs:
   shared-memory:
@@ -93,12 +94,9 @@ parts:
     prime:
       - etc/security/limits.d/10-gamemode.conf
       - usr/bin/gamemoded
-      - usr/games/gamemodelist
-      - usr/games/gamemoderun
-      - usr/games/gamemode-simulate-game
-      - usr/lib/*/libgamemode.so*
-      - usr/lib/*/libgamemodeauto.so*
-      - usr/lib/*/libinih.so*
+      - usr/games/gamemode*
+      - usr/lib/*/libgamemode*
+      - usr/lib/*/libinih*
       - usr/lib/systemd/user/gamemoded.service
       - usr/lib/sysusers.d/gamemode.conf
       - usr/libexec/cpucorectl
@@ -117,11 +115,9 @@ parts:
     stage-packages: [mangohud, libspdlog1.12, libfmt9]
     prime:
       - usr/bin/mangohud
-      - usr/lib/x86_64-linux-gnu/mangohud/libMangoHud.so*
-      - usr/lib/x86_64-linux-gnu/mangohud/libMangoHud_opengl.so*
-      - usr/lib/x86_64-linux-gnu/mangohud/libMangoHud_shim.so*
-      - usr/lib/x86_64-linux-gnu/libspdlog.so*
-      - usr/lib/x86_64-linux-gnu/libfmt.so*
+      - usr/lib/x86_64-linux-gnu/mangohud
+      - usr/lib/x86_64-linux-gnu/libspdlog*
+      - usr/lib/x86_64-linux-gnu/libfmt*
       - usr/share/metainfo/io.github.flightlessmango.mangohud.metainfo.xml
       - usr/share/vulkan/implicit_layer.d/MangoHud.x86_64.json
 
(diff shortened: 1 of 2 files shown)


Kyuyrii:
This is just proof that it's working.

At the beginning of the video, all the existing plugs and those that are connected are shown; the steam-support plug is essential for the Proton/UMU-launcher to function.

https://github.com/user-attachments/assets/1fa4bd50-6942-4ed3-92ba-bcbdc768aaa8

Kyuyrii:
Regarding the game mode and mangohud, I removed them because for now the focus should be on testing if the games work and the main functions of Heroic, and after that I'll try to add the game mode and mangohud back.

Kyuyrii:
Testing the mangohud

https://github.com/user-attachments/assets/0650857d-a2f9-4bdc-bfa6-abb71daeac3d

Kyuyrii:
The feature to create game shortcuts in the application menu is working.

https://github.com/user-attachments/assets/fd02cacc-fda2-40ce-b386-44bd07776f5e

snap/snapcraft.yaml
```diff
@@ -4,13 +4,15 @@ summary: An Open Source Launcher for GOG, Epic Games and Amazon Games
 description: |
   Heroic is an Open Source Games Launcher. Right now it supports launching games from the Epic Games
   Store using Legendary, Amazon Games using Nile and GOG Games using our custom implementations Nile and gogdl.
-version: 2.16.1
+version: 2.18.1
```
CommandMC: Can we load this version number from `package.json`? Ideally we'd just have one version number to bump each update

snap/snapcraft.yaml
```diff
@@ -25,449 +27,114 @@ donation:
...
-    bind: $SNAP/etc/fonts
+apps:
+  heroic:
+    command-chain: [snap/command-chain/desktop-launch-heroic]
+    command: opt/Heroic/heroic --no-sandbox
+    desktop: usr/share/applications/heroic.desktop
+    common-id: com.heroicgameslauncher.hgl
+    environment:
+      HOME: $SNAP_USER_COMMON
+      PATH: $SNAP/usr/games:$PATH
```
CommandMC, line 39: Why is `$SNAP/usr/games` added to PATH here?

CommandMC (commented):
I have no clue what half of the changes in this file do, but as I've mentioned elsewhere, I'm willing to try a Snap build again as long as it doesn't cause any major issues

snap/snapcraft.yaml
Kyuyrii: In the kde-neon-6 extension, this is set via desktop-launch, but this isn't done in the gnome extension, so I had to set it in snapcraft.yaml.

snap/snapcraft.yaml
a5ehren: definitely possible, but [takes some script wrangling](https://documentation.ubuntu.com/snapcraft/stable/how-to/crafting/configure-package-information/#with-a-script ).

The other option would be to take another crack at using electron-build's native snap target, which brings in the version automatically.

flavioislima:
@Kyuyrii I will try to build this and publish it on snap store, would it work? Or this is just for the manual installation? for the manual installation we can distribute the .snap using the elctron builder we already have, would just need to adjust the release command: https://www.electron.build/snap

Kyuyrii:
> @Kyuyrii I will try to build this and publish it on snap store, would it work? Or this is just for the manual installation? for the manual installation we can distribute the .snap using the elctron builder we already have, would just need to adjust the release command: https://www.electron.build/snap

To publish to the Snap Store, you need to request automatic connection of the steam-support plug; it is extremely important for UMU to function.

Kyuyrii:
Generally, plugs added to Snap apps that aren't connecting automatically don't prevent them from being published, but there are some exceptions for security reasons.

flavioislima:
> > @Kyuyrii I will try to build this and publish it on snap store, would it work? Or this is just for the manual installation? for the manual installation we can distribute the .snap using the elctron builder we already have, would just need to adjust the release command: https://www.electron.build/snap
> 
> To publish to the Snap Store, you need to request automatic connection of the steam-support plug; it is extremely important for UMU to function.

I see, I think we had a thread somewhere at some point but I think I would need to start a new one.
This is the error log for the build on snapcraft:
https://launchpadlibrarian.net/841753634/buildlog_snap_ubuntu_noble_amd64_018236f3314e147646868ec87cc16db2_BUILDING.txt.gz

Kyuyrii:
> I see, I think we had a thread somewhere at some point but I think I would need to start a new one. This is the error log for the build on snapcraft: https://launchpadlibrarian.net/841753634/buildlog_snap_ubuntu_noble_amd64_018236f3314e147646868ec87cc16db2_BUILDING.txt.gz

The error was about not finding the desktop-launch-heroic file. Locally, this error didn't occur here, but I modified the command-chain source to see if it fixes the problem.

a5ehren:
I'm not a snap expert, but if the output works, then its certainly better than what's there now.

snap/snapcraft.yaml
a5ehren: ok, electron-builder can't do core24 snaps right now, and the dev is working on it. But apparently they have to totally replace their snap backend to do it. https://github.com/electron-userland/electron-builder/issues/9452

Kyuyrii: I made a change to try and always use the .deb package from the most recent tag.

arielj:
Closing for now until we have better support from the Snap team

PR #5116, opened 2025-12-25
Add conservative Renovate configuration with gated majors

niStee:
Add production-ready Renovate configuration (Optimized for Electron)

Adds a comprehensive Renovate configuration tailored for the Heroic codebase.

### Key Features
*   **Electron Strategy:** Split into 3 logical groups to prevent version mismatches:
    *   **Core:** Isolated PRs (High safety).
    *   **Helpers:** (`electron-store`, `updater`) Grouped together (Safe prod deps).
    *   **Build Tools:** (`builder`, `vite`) Grouped together (Dev deps).
*   **Safety Gates:**
    *   **Major Updates:** Gated behind "Dependency Dashboard" approval (Manual trigger required).
    *   **Patched Deps:** `howlongtobeat-js` is explicitly **ignored** to prevent breaking local pnpm patches.
*   **Noise Reduction:**
    *   **Prod Deps:** "Boring" minor/patch updates are bundled into one weekly PR.
    *   **Dev Deps:** Split into logical buckets (`Types`, `Linting`, `Testing`) for easier review, rather than one massive PR.
    *   **UI Frameworks:** `@mui`, `@emotion`, etc. are grouped to ensure coupled packages update in lockstep.
*   **Schedule:** Runs weekly (Sunday 3am UTC) to prevent notification fatigue.
*   **Security:** Pins GitHub Actions to SHA digests.

### Maintainers Note
Please ensure the [Renovate GitHub App](https://github.com/apps/renovate) is installed on the repository. Once merged, the bot will create a "Dependency Dashboard" issue where you can manage pending updates.

---
*Verified against `package.json` to ensure correct grouping of internal dependencies.*


flavioislima:

niStee:

renovate.json
```diff
@@ -0,0 +1,59 @@
...
+      "matchDepTypes": ["devDependencies"],
+      "groupName": "devDependencies"
+    },
+    {
+      "matchPaths": ["e2e/**", "tests/**"],
+      "groupName": "test-tooling"
+    },
+    {
+      "description": "Patched deps stay manual",
+      "matchPackageNames": ["@types/node", "howlongtobeat-js"],
```
CommandMC: Is it possible to pull this in from `pnpm-workspace.yaml`?

CommandMC (commented):
The link in your PR description (<https://docs.renovatebot.com/getting-started/installing/>) returns a 404 error. I assume you intended to link to <https://docs.renovatebot.com/getting-started/installing-onboarding/>

How is Renovate more helpful than GitHub's Dependabot? In the past, we've had issues with Dependabot attempting to bump us to incompatible packages, justifying it with security issues that did not affect us

arielj:
to be honest, I don't really like renovate (like I don't like dependabot)

those tools sound good, but they become really annoying

renovate.json
niStee: > Is it possible to pull this in from `pnpm-workspace.yaml`?

Theoretically yeah, we could probably use matchJsonata to parse the package.json at runtime. I messed around with a "hardened" version of that logic::
```json
"matchJsonata": ["$exists($lookup($$.packageFileContent.pnpm.patchedDependencies, depName & '@' & $replace(currentVersion, '^v', '')))"]
```
**But honestly, I stayed away from it for this PR because:**

It's a bit brittle: JSONata in Renovate can be finicky. If pnpm changes how it strings keys or if the version formatting (like the v prefix) doesn't match perfectly, the rule might just fail silently.

Visibility matters: For a public project, it’s probably better if the gated deps are just right there in the config. It makes it obvious to anyone else why those packages are behaving differently without them needing to decypher JSONata logic.

I went with the manual list for now since it’s battle-tested. I did add matchManagers: ["npm"] to keep it scoped and prBodyNotes to remind whoever’s updating it to check the patch.

If we end up with a ton of patches later, we could always look into a CI script to auto-generate this, but for now, this feels like the safest "set and forget" way.

niStee:
@CommandMC - Good catch on that link, you're right, meant the onboarding one!

**Re: Renovate vs. Dependabot (and the incompatible package issue):**

I noticed those blocks of closed Dependabot PRs for vite recently (like #4294-#4467). That "notification flood" is exactly what this config is meant to fix. Here is how this specific Renovate setup handles those "security noise" and incompatibility issues differently:

**1. Dashboard for Majors (No more surprises)** I’ve set dependencyDashboardApproval: true for majors. Instead of Dependabot just throwing a breaking change PR at you, Renovate puts it on a "to-do list" issue. You only get the PR when you actually click a checkbox to ask for it.

**2. Stability Windows (The "Wait and See" approach)** The config uses stabilityDays: 3 (global) and 7 (majors). Renovate literally waits for a release to be out for a week before bugging you. If a version is buggy or has a "false alarm" security fix, it usually gets yanked or patched before the 7 days are up, so you never even see the bad PR.

**3. Grouping - Less Noise** Instead of 5 different PRs for one ecosystem, this config funnels everything into 5 predictable groups (Prod, Dev, Test, Actions, and Gated Majors). If 5 vite updates come out in a week, you get one PR that just updates itself as new versions release.

**4. Schedule Control** By running only on Sundays at 3am with a prHourlyLimit, it stays out of your way during active dev hours.

**The core difference:** Renovate is "quiet by default" when tuned this way. It’s built to kill the "update fatigue" I see in the closed PR history.

That said, if the team prefers sticking with Dependabot, I totally get it! This was just meant as a lower-noise alternative for the project.

**Re: the patched deps in config**, Can you confirm if @types/node and howlongtobeat-js are currently the only ones maintained with local patches in /patches/**? I gated those from auto-updates based on that assumption.

niStee:
I've refined the configuration after a deeper review of the dependency graph. The initial 'catch-all' groupings were a good start, but I found a few optimizations to reduce friction:

1. Smarter Electron Strategy
I noticed electron-store and electron-updater release independently from electron core. Grouping them all together would block safe helper updates whenever the core framework has a major version bump we aren't ready for.

Solution: I've split them into Core (Isolated), Helpers (Grouped), and Build Tools (Grouped). This keeps critical framework updates isolated for testing without stalling the smaller libraries.

2. UI Framework safety
Libraries like @mui and @emotion-js are now explicitly grouped to ensure their sub-packages stay synchronized, preventing version mismatch errors in the UI.

3. Granular DevDependencies
Instead of one massive 'DevDeps' PR, I've split them into logical buckets (Types, Linting, Testing). This should make the weekly maintenance PRs much faster to skim and merge.

PR #5118, opened 2025-12-25
Add pnpm audit workflow

niStee:
- Add pnpm audit workflow on PRs to main/stable and weekly (Wed 04:15 UTC).
- Use Node 20 and pnpm 10.9.0 (pnpm cache) with pnpm install --frozen-lockfile.
- Run two audits: full pnpm audit --prod (captured to summary) and enforcing pnpm audit --prod --audit-level=high (fails on high/critical). Summary always shows all severities.
- Minimal permissions (contents: read). Actions tag-pinned; Renovate will handle digest pinning #5116.

---

Testing/impact: Workflow-only change; no runtime code touched.

### DIFF
--- a/.github/workflows/pnpm-audit.yml
+++ b/.github/workflows/pnpm-audit.yml
@@ -29,14 +29,17 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
-      - name: pnpm audit (high+ only, report-only)
+      - name: pnpm audit (report all severities)
         run: |
-          pnpm audit --prod --audit-level=high > audit-report.txt
+          pnpm audit --prod > audit-report.txt || true
+
+      - name: pnpm audit (enforce high+)
+        run: pnpm audit --prod --audit-level=high
 
       - name: Publish audit summary
         if: always()
         run: |
-          echo "### pnpm audit (high+ only)" >> "$GITHUB_STEP_SUMMARY"
+          echo "### pnpm audit (full report)" >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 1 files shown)


.github/workflows/pnpm-audit.yml
```diff
@@ -0,0 +1,48 @@
...
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup pnpm 10.9.0
+        uses: pnpm/action-setup@v3
+        with:
+          version: 10.9.0
```
a5ehren, line 21: the pnpm action will pull the version from package.json, you shouldn't have to specify it

.github/workflows/pnpm-audit.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Use Node 20
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
a5ehren, line 26: current version of electron uses node 22, should audit against that

arielj:
I don't think pnpm audit is that useful in our case, at least not as something that should block merging PRs

like... if I'm working on something and some version of a new tool is released fixing some new known vulnerability, I don't want to have to update it in my totally unrelated PR just because pnpm audit says its critical, it's not related to my PR

niStee:
@arielj fair point! 
You shouldn't be blocked on unrelated pull requests. 
I've updated this PR so that it's just for information. Even if a high-priority or critical issue is found, it won't prevent merging (unless we make it a required check in the repository settings, which we shouldn't do). 
It's just a 'heads up' for the team.

PR #5128, opened 2025-12-30
[Feat] Add code to look for vdf files that define compatibility tool

a5ehren:
…locations instead of just assuming that <path>/proton exists. This fixes an issue that cachyos is about to file

---

- [ x ] Tested the feature and it's working on a current and clean install.
- [ x ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)

### DIFF
--- a/src/backend/utils/__tests__/compatibility_layers.test.ts
+++ b/src/backend/utils/__tests__/compatibility_layers.test.ts
@@ -1,14 +1,55 @@
-import {
-  getDefaultWine,
-  getWineExecs,
-  getWineLibs
-} from '../compatibility_layers'
-import { mkdirSync } from 'graceful-fs'
+import { mkdirSync, rmSync, symlinkSync, writeFileSync } from 'graceful-fs'
 import { dirname, join } from 'path'
 import { tmpdir } from 'os'
 import child_process from 'child_process'
 
-jest.mock('../../logger')
+// Mock modules BEFORE importing the module under test
+jest.mock('backend/logger', () => ({
+  logError: jest.fn(),
+  logInfo: jest.fn(),
+  LogPrefix: {
+    GlobalConfig: 'GlobalConfig'
+  }
+}))
+
+// Mock utils with explicit implementation
+jest.mock('backend/utils', () => ({
+  getSteamLibraries: () => Promise.resolve([]),
+  execAsync: () => Promise.resolve()
+}))
+
+// Mock GlobalConfig with the settings needed for tests
+jest.mock('backend/config', () => ({
+  GlobalConfig: {
+    get: () => ({
+      getSettings: () => ({
+        showValveProton: false,
+        customWinePaths: []
+      })
+    })
+  }
+}))
+
+import {
+  getDefaultWine,
+  getWineExecs,
+  getWineLibs,
+  getLinuxWineSet
+} from '../compatibility_layers'
+jest.mock('backend/constants/paths', () => {
+  const pathModule = jest.requireActual<typeof import('path')>('path')
+  const osModule = jest.requireActual<typeof import('os')>('os')
+  const testToolsPath = pathModule.join(osModule.tmpdir(), 'heroic-test-tools')
+  return {
+    ...jest.requireActual('backend/constants/paths'),
+    toolsPath: testToolsPath,
+    configPath: pathModule.join(testToolsPath, 'config.json'),
+    userHome: osModule.tmpdir(),
+    appFolder: testToolsPath,
+    publicDir: pathModule.join(__dirname, '..', '..', '..', 'public'),
+    fixAsarPath: (path: string) => path
+  }
+})
 
 describe('getDefaultWine', () => {
   test('return wine not found', () => {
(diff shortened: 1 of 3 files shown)


src/backend/utils/compatibility_layers.ts
```diff
@@ -169,17 +176,150 @@ export async function getLinuxWineSet(
 
   protonPaths.forEach((path) => {
     if (existsSync(path)) {
-      readdirSync(path).forEach((version) => {
+      readdirSync(path).forEach((item) => {
```
arielj: I wonder if there's a way to simplify this loop using the `glob` method and a pattern like `**/*.vdf` https://nodejs.org/api/fs.html#fspromisesglobpattern-options

then in a single loop we would have a list of all .vdf in the directory and all subdirectories and have a single loop instead of nesting

a5ehren: Good call. I'm not (primarily) a JS/TS dev so I'm not really sure about the utility functions that are available. I'll check it out and see if I can update the PR.

a5ehren:
CI runner is failing because the gh action installs Node 20, but the globSync API does not exist until Node 22.

test:ci task should start passing once #5097 is done

src/backend/utils/compatibility_layers.ts
```diff
@@ -169,17 +176,150 @@ export async function getLinuxWineSet(
...
+                      proton.add({
+                        bin: protonBin,
+                        name: displayName,
+                        type: 'proton'
+                      })
+                    }
+                  }
+                }
+              }
+            }
```
CommandMC: Please use early return / inverted if conditions to avoid excessive indentation like this

For example, instead of
```ts
if (someCondition) {
  // Further logic
}
```
instead use
```ts
if (!someCondition) return // Or "continue", or error handling, depending on context
// Further logic, now with one less indentation level
```

You can further improve the code by using async code (I particularly like [`Promise.try`](<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/try>) to avoid a classic try/catch statement), but I understand this may be difficult if you're not used to it

a5ehren: I just realized that I forgot to push after changing to GlobSync for finding the files. I'll push that and then look at the async.

src/backend/utils/compatibility_layers.ts
```diff
@@ -109,6 +111,133 @@ export function getWineLibs(wineBin: string): {
...
+    const vdfContent = readFileSync(vdfPath, 'utf-8')
+    const parsedVdf = parseVdf(vdfContent) as Record<string, unknown>
+
+    // Extract information from VDF structure
+    const compatTools = parsedVdf?.compatibilitytools as
+      | Record<string, unknown>
+      | undefined
+    const tools = compatTools?.compat_tools as
+      | Record<string, Record<string, unknown>>
+      | undefined
```
CommandMC: I don't like how much `as` is used here (ideally you'd use Zod to actually validate the shape) and how the whole function is essentially a giant try-catch, but that can be resolved later

src/backend/utils/compatibility_layers.ts
```diff
@@ -109,6 +111,133 @@ export function getWineLibs(wineBin: string): {
...
+    }
+
+    // Determine proton binary path
+    // install_path can be absolute (e.g., "/opt/proton-cachyos") or relative
+    const protonBin = isAbsolute(installPath)
+      ? join(installPath, 'proton')
+      : join(itemDir, installPath, 'proton')
+
+    // Only return if the proton binary actually exists
+    if (!existsSync(protonBin)) return null
```
CommandMC: Perhaps add a log message here?

CommandMC (approved):

src/backend/utils/compatibility_layers.ts
a5ehren: sorry, I'm a systems dev dabbling in Typescript. Not used to having libraries for stuff. I'll read about zod and see what it can do.

src/backend/utils/compatibility_layers.ts
a5ehren: Yeah, will do

a5ehren:
The flatpak step needs to re-run, it was a download problem.

src/backend/utils/compatibility_layers.ts
a5ehren: Updated with zod, lmk what you think

src/backend/utils/__tests__/compatibility_layers.test.ts
```diff
@@ -1,14 +1,55 @@
...
-  getWineLibs
-} from '../compatibility_layers'
-import { mkdirSync } from 'graceful-fs'
+import { mkdirSync, writeFileSync } from 'graceful-fs'
 import { dirname, join } from 'path'
 import { tmpdir } from 'os'
 import child_process from 'child_process'
 
-jest.mock('../../logger')
+// Mock modules BEFORE importing the module under test
```
CommandMC, line 6: Jest should automatically move mocks to the start of a file, so the ordering of import & mock should not matter
That being said, I'm fine with keeping this order, tests don't have to be super pretty

CommandMC (approved):

src/backend/utils/compatibility_layers.ts
```diff
@@ -109,6 +112,172 @@ export function getWineLibs(wineBin: string): {
...
+        itemName
+      }))
+      .filter(({ itemName }) => !itemName.startsWith('UMU-Latest'))
+      .filter(({ protonBin }) => existsSync(protonBin))
+      .filter(({ itemDir }) => !processedDirs.has(itemDir))
+      .map(({ protonBin, itemName }) => ({
+        bin: protonBin,
+        name: itemName,
+        type: 'proton' as const
+      }))
```
CommandMC, line 272: I am wondering whether we need this fallback. This only applies if we encountered a Proton without a valid `compatibilitytool.vdf`, right?

src/backend/utils/compatibility_layers.ts
```diff
@@ -109,6 +112,172 @@ export function getWineLibs(wineBin: string): {
...
+ * @returns Set of WineInstallation objects found in the directory
+ */
+function scanProtonPath(path: string): Set<WineInstallation> {
+  if (!existsSync(path)) return new Set()
+
+  try {
+    // Get installations from VDF files
+    const vdfPaths = globSync('{*.vdf,*/*.vdf}', {
+      cwd: path
+    }).filter((vdfPath) => !vdfPath.includes('UMU-Latest'))
```
CommandMC, line 226: Do we really want a `*.vdf` here? Isn't the info we want always in a file called `compatibilitytool.vdf`? As-is, we'd always try to parse `toolmanifest.vdf` as well (which contains irrelevant data, and will fail the schema validation)

PR #5133, opened 2025-12-31
[UI] Remove Extra System Requirements option from Game Sub Menu

redromnon:
There's already a System Requirements section in the Tab view. Thus, it doesn't help to keep the same option in the sub menu.

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/components/DotsMenu.tsx
+++ b/src/frontend/screens/Game/GamePage/components/DotsMenu.tsx
@@ -29,8 +29,6 @@ const DotsMenu = ({ gameInfo, handleUpdate }: Props) => {
 
   const { is_installed, title } = gameInfo
 
-  const hasRequirements = (gameExtraInfo?.reqs || []).length > 0
-
   return (
     <>
       <div className="game-actions">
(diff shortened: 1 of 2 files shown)


flavioislima:

redromnon:

arielj:
I'm not sure if having the duplicated option to see the requirements is a problem, we also have the browse files and browse wine prefix links that are duplicated and I don't think it's an actual issue (users may find it easier to find things inside the 3-dots menu)

like... if a user has a narrow screen, it might be not obvious that there's a System Requirements tab on that right pane

Captura de pantalla 2026-01-03 a la(s) 6 01 16 p  m

redromnon:
Makes sense. Some users may prefer using the three-dots menu. It would prefer to use the `Monitor` icon instead of the `DesktopAccessDisabledIcon` in the menu.

On another note:

> like... if a user has a narrow screen, it might be not obvious that there's a System Requirements tab on that right pane

For this issue, can a scrollbar be added to tell the user that another tab exists? Here's an example of this, where the user wouldn't know the System Requirements tab exists unless there's a scrollbar present.
Screenshot From 2026-01-04 12-15-23

PR #5143, opened 2026-01-03
[UX] Add link to wine version release notes

arielj:
This PR adds a small UX improvement for the Wine Manager by adding a link to open the release notes in github directly so the user doesn't have to go to the actual repo manually.

Captura de pantalla 2026-01-03 a la(s) 5 43 14 p  m

---

### DIFF
--- a/src/frontend/screens/WineManager/index.tsx
+++ b/src/frontend/screens/WineManager/index.tsx
@@ -213,7 +213,7 @@ export default function WineManager(): JSX.Element | null {
           >
             <div className="gameListHeader">
               <span>{t('info.version', 'Wine Version')}</span>
-              <span>{t('wine.link', 'Notes')}</span>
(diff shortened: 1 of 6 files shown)


src/frontend/screens/WineManager/index.tsx
```diff
@@ -213,6 +213,7 @@ export default function WineManager(): JSX.Element | null {
           >
             <div className="gameListHeader">
               <span>{t('info.version', 'Wine Version')}</span>
+              <span>{t('wine.link', 'Notes')}</span>
```
flavioislima: nitty-picky but since its release notes can we do `wine.release_notes` instead?

arielj: updated

flavioislima (approved):
LGTM
We eventually should re-do this wine table, its not sustainable lol

PR #5145, opened 2026-01-04
[UX] Export/Import categories data

### LINKED ISSUE #3416: Exporting the categories to a file for backup purposes
### Problem description

As I understand due to the nature of EGS, the categories are only stored locally on the device (as opposed to Steam where they are linked to the Steam account).

I want to make sure my categories are not gone if I experience a storage failure. Is there any way to export and backup them?

I have found they are listed in `~/.config/heroic/store/config.json` under `customCategories` JSON key. If I add those lines to a newly created `config.json` on a fresh installation, will the launcher automatically pick those up?

(I am on Linux, .deb installation).

### Feature description

Have "export categories" and "import categories" buttons that will respectively create a file with categories that can be then used to import them to a newly created/empty/without custom configurations `config.json` file.

### Alternatives

On Linux, open the  `~/.config/heroic/store/config.json` file, the categories along with games belonging to them will be available under `customCategories` JSON key.

### Additional information

_No response_

arielj:
This PR adds the actions to export the categories to a JSON file and also the option to import a JSON file with categories information.

For now it exports the categories inside the `Games/Heroic` folder, so we can be sure Heroic flatpak will have write access to it.

The buttons are at the bottom of the Manage Categories modal:

Captura de pantalla 2026-01-03 a la(s) 11 40 42 p  m

Closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3416

---


flavioislima (approved):
LGTM

CommandMC (changes requested):
I don't think having the buttons in the modal itself makes sense, since they offer functionality you don't need often. Tucking them somewhere into the settings, or at least making the buttons less prominent, would be ideal

arielj:
> I don't think having the buttons in the modal itself makes sense, since they offer functionality you don't need often. Tucking them somewhere into the settings, or at least making the buttons less prominent, would be ideal

maybe I can hide the 2 buttons inside a `<details>` tag collapsed by default?

what I like about the buttons in the modal is that it makes it really easy to discover them, it's right next to where categories are managed

I think collapsing them and show them only on demand can help clean it up and keep them easily discoverable, I'll update the PR

arielj:
This is how it looks like now, collapsed by default

Captura de pantalla 2026-01-06 a la(s) 10 45 16 a  m
Captura de pantalla 2026-01-06 a la(s) 10 45 22 a  m

CommandMC (changes requested):
Sorry for another nitpick, testing things now:

- One button says "Export to file", the other "Import file". To keep them consistent, I'd either change the one to "Export file", or the other to "Import from file"
- Maybe add icons to the buttons?

src/backend/importers/categories.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    for (const key in jsonContent) {
+      // some checks to ensure we don't break the categories if the
+      // json file is not the correct format
+      // only allow string keys
+      if (typeof key !== 'string') continue
+
+      // only allow an array of strings
+      if (!Array.isArray(jsonContent[key])) continue
+      const gameIds = jsonContent[key].filter((val) => typeof val === 'string')
```
CommandMC, line 29: This feels like an obvious use case for Zod

```ts
// Somewhere at the start of the file
const ImportSchema = z.record(z.string(), z.string().array())

// Inside the function
const jsonContent = JSON.parse(content)
const validatedContent = ImportSchema.parse(jsonContent)
// validatedContent is now of type Record<string, string[]>
```
If we truly want to accept partially-malformed category exports, we could still handle that in Zod with some transforms. IMO we should just not do that though, I don't think that's a common use case

src/preload/api/misc.ts
```diff
@@ -55,6 +55,8 @@ export const pathExists = makeHandlerInvoker('pathExists')
 export const processShortcut = makeListenerCaller('processShortcut')
 export const handleGoToScreen = frontendListenerSlot('openScreen')
 export const handleShowDialog = frontendListenerSlot('showDialog')
+export const exportCategories = makeHandlerInvoker('exportCategories')
+export const importCategories = makeHandlerInvoker('importCategories')
```
CommandMC, line 59: Since you seem to be looking to create more exporters/importers, you may want to create a "category" for those IPC functions. It makes them easier to find & lessens the load on the language server (well, it would if we had more of them)

A "category" is just an object with properties, like this example:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/23de3f919e55a21f189ee7e208e93c3121d5c20f/src/preload/api/settings.ts#L20-L23

(whether you sort them `export.categories` or `categories.export` is up to you)

arielj: yes, the idea is that we could have more importers/exporters for other things in the future, like the app settings for example that users currently have to copy around internal files

PR #5148, opened 2026-01-04
[chore][tech] Update dependencies that do not require significant API changes

a5ehren:
Notable updates that caused a bit of code churn are eslint and prettier.

The new eslint and especially its react rules do a bunch of checks around manual memoization because the React Compiler cannot do anything with those annotations - I've temporarily set a couple of the new options to Warnings until I can actually evaluate the performance impact of React Compiler to see if the code changes are worth it.

---


arielj:
I'm not sure if having all these unrelated changes in the same PR is a good idea

some are comments, others are dependency updates, others move code around, others actually change code

I understand a lot is related to eslint and prettier, but looks off to me to merge the update of a dependency like axios at the same time as adding types for some variables and at the same time of refactoring some methods

a5ehren:
Yeah, totally fair. I can split this into one that is the updates with zero code changes, then do prettier, then do eslint? Basically compartmentalize any update that needs code changes.

PR #5149, opened 2026-01-05
[FIX] Resolve powershell/pwsh reliably and normalize path tests

niStee:
Improve PowerShell spawning on Windows and align test expectations:
- Resolve PowerShell deterministically: prefer known powershell/pwsh paths, cache the resolved binary, and log the spawn command; fall back to PATH only when needed.
- Make PATH lookups more robust: use absolute where.exe when available, cache finds, and handle both powershell/pwsh with the same resolver.
- Normalize test path expectations to handle Windows separators in filesystem and downloader tests.

Notes:
- Hooks are disabled locally on Windows (MSYS bash fork failures); branch was pushed with --no-verify. We ran lint/prettier/codecheck/i18n via aliases before push. Please let CI run the full hook equivalents.
- e2e: 5/6 passing; webview_controls.spec.ts fails due to Google consent redirect URL. Main app flows (login/install/play/etc.) not covered in this run.
- If possible, sanity-check executable resolution in the Steam sandbox, since the finder logic was touched.

---

### DIFF
--- a/src/backend/utils/os/path/index.ts
+++ b/src/backend/utils/os/path/index.ts
@@ -5,17 +5,21 @@ import { access } from 'fs/promises'
 const isWindows = process.platform === 'win32'
 const findCommands = isWindows
   ? [
-      'C:\\Windows\\System32\\where.exe',
+      'C\\Windows\\System32\\where.exe',
       // fall back to PATH lookup; may be missing inside Electron sandboxes
       'where'
     ]
   : ['which']
-const windowsShellPaths = [
-  'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe',
-  'C:\\Windows\\System32\\powershell.exe',
-  'C:\\Program Files\\PowerShell\\7\\pwsh.exe',
-  'C:\\Program Files\\PowerShell\\7\\powershell.exe'
+const commonWindowsShellPaths = [
+  'C\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe',
+  'C\\Windows\\System32\\powershell.exe',
+  'C\\Program Files\\PowerShell\\7\\pwsh.exe',
+  'C\\Program Files\\PowerShell\\7\\powershell.exe'
 ]
+const windowsShellCandidates: Record<string, string[]> = {
+  powershell: commonWindowsShellPaths,
+  pwsh: commonWindowsShellPaths
+}
 const finderCache = new Map<string, string | null>()
 
 async function pathExists(candidate: string) {
@@ -35,7 +39,7 @@ async function runFinder(
     const child = spawn(cmd, [executable], {
       shell: isWindows,
       windowsHide: true,
-      env: { ...process.env } // ADD THIS LINE - inherit parent environment
+      env: { ...process.env }
     })
 
     let settled = false
@@ -64,15 +68,16 @@ async function runFinder(
 async function searchForExecutableOnPath(
   executable: string
 ): Promise<string | null> {
-  if (finderCache.has(executable)) return finderCache.get(executable) ?? null
+  const cacheKey = executable.toLowerCase()
+  if (finderCache.has(cacheKey)) return finderCache.get(cacheKey) ?? null
 
   // Special-case PowerShell so we never depend on PATH for it
-  if (isWindows && executable.toLowerCase() === 'powershell') {
-    for (const candidate of windowsShellPaths) {
+  if (isWindows && windowsShellCandidates[cacheKey]) {
+    for (const candidate of windowsShellCandidates[cacheKey]) {
       // stop at the first existing path
       /* istanbul ignore next -- platform-specific */
       if (await pathExists(candidate)) {
-        finderCache.set(executable, candidate)
+        finderCache.set(cacheKey, candidate)
         return candidate
       }
     }
@@ -81,12 +86,12 @@ async function searchForExecutableOnPath(
   for (const cmd of findCommands) {
     const found = await runFinder(cmd, executable)
     if (found) {
-      finderCache.set(executable, found)
+      finderCache.set(cacheKey, found)
       return found
     }
   }
 
-  finderCache.set(executable, null)
+  finderCache.set(cacheKey, null)
   return null
 }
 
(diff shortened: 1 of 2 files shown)


src/backend/utils/os/path/index.ts
```diff
@@ -1,6 +1,64 @@
...
+import { constants } from 'fs'
+import { access } from 'fs/promises'
 
-const findCommand = process.platform === 'win32' ? 'where' : 'which'
+const isWindows = process.platform === 'win32'
+const findCommands = isWindows
+  ? [
+      'C\\Windows\\System32\\where.exe',
+      // fall back to PATH lookup; may be missing inside Electron sandboxes
+      'where'
```
CommandMC, line 10: `where` also just uses the `%PATH%` environment variable to look for the requested application, so `%PATH%` being unset would break it as well. Also, `C:\Windows\System32` not being on `%PATH%` would cause *many* issues unrelated to Heroic

src/backend/utils/os/path/index.ts
```diff
@@ -1,6 +1,64 @@
...
+    }
+
+    child.stdout.on('data', (output: Buffer | string) => {
+      const first = output.toString().split(/\r?\n/)[0]?.trim()
+      if (first) finish(first)
+    })
+
+    child.on('error', () => finish(null))
+    child.on('close', () => finish(null))
+  })
```
CommandMC, line 60: Please use `genericSpawnWrapper` here

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/7ce2796857233cc2fda50d45f24f0d542b3f1ac5/src/backend/utils/os/processes/index.ts#L16

src/backend/utils/os/path/index.ts
```diff
@@ -10,22 +68,31 @@ const findCommand = process.platform === 'win32' ? 'where' : 'which'
...
-    // this avoids problems inside steam where "which" calls
-    // throws alot of errors and still find the exectuable.
-    // We also prevent endless waiting when stderr, stdout
-    // and error are never called
-    const child = spawn(findCommand, [executable])
-    child.stdout.on('data', (output: Buffer | string) => {
-      resolve(output.toString().trim())
-    })
+  const cacheKey = executable.toLowerCase()
+  if (finderCache.has(cacheKey)) return finderCache.get(cacheKey) ?? null
```
CommandMC, line 72: How long does `where`/`which` take to execute on your machine? Is a cache really necessary?

On my end, `/usr/bin/which` takes around 500µs. I can't imagine Windows takes *that* much longer

src/backend/utils/os/path/index.ts
```diff
@@ -10,22 +68,31 @@ const findCommand = process.platform === 'win32' ? 'where' : 'which'
...
 
-    // if we close and not got any data on stdout, we return null
-    child.on('close', () => resolve(null))
-  })
+  // Special-case PowerShell so we never depend on PATH for it
+  if (isWindows && windowsShellCandidates[cacheKey]) {
```
CommandMC, line 75: This is a confusing way of filtering PowerShell. I don't think we should do this at all (as mentioned in my other comment), but if you *would* do it, making `windowsShellCandidates` an array and calling `windowsShellCandidates.includes(cacheKey)` instead would be clearer

src/backend/utils/os/path/index.ts
```diff
@@ -10,22 +68,31 @@ const findCommand = process.platform === 'win32' ? 'where' : 'which'
...
+    for (const candidate of windowsShellCandidates[cacheKey]) {
+      // stop at the first existing path
+      /* istanbul ignore next -- platform-specific */
```
CommandMC, line 78: We don't use istanbul

src/backend/utils/os/path/index.ts
```diff
@@ -1,6 +1,64 @@
...
+
+async function runFinder(
+  cmd: string,
+  executable: string
+): Promise<string | null> {
+  return new Promise<string | null>((resolve) => {
+    const child = spawn(cmd, [executable], {
+      shell: isWindows,
+      windowsHide: true,
+      env: { ...process.env }
```
CommandMC, line 42: This is the default value

> env: Environment key-value pairs. Default: process.env.

\- https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

src/backend/utils/filesystem/__tests__/unix.test.ts
```diff
@@ -53,7 +54,11 @@ describe('getDiskInfo_unix', () => {
     expect(ret.totalSpace).toBe(100 * 1024)
     expect(ret.freeSpace).toBe(10 * 1024)
     expect(accessSyp).toHaveBeenCalledTimes(2)
-    expect(spawnWrapperSpy).toHaveBeenCalledWith('df', ['-P', '-k', '/foo/bar'])
+    expect(spawnWrapperSpy).toHaveBeenCalledWith('df', [
+      '-P',
+      '-k',
+      path.normalize('/foo/bar')
```
CommandMC, line 60: This path contains neither `.`, `..`, nor multiple `//`, so calling `path.normalize` on it should not be necessary. Why are you making this change?

src/backend/launcher.ts
```diff
@@ -1743,6 +1745,10 @@ async function callRunner(
   const abortController = createAbortController(abortId)
 
   let promise = new Promise<ExecResult>((res, rej) => {
+    logDebug(
+      `About to spawn: bin="${bin}", cwd="${options?.cwd || runner.dir}"`,
+      runner.logPrefix
+    )
```
CommandMC, line 1751: Leftover debugging code?

CommandMC (changes requested):
Summarizing my comments, I do not believe the changes in this PR are helpful

niStee:
> Summarizing my comments, I do not believe the changes in this PR are helpful

After performing a clean setup and running `pnpm install` and `pnpm download-helper-binaries` when switching branches, I could no longer reproduce the original issue. It appears that the issue was caused by my local environment rather than the existing path logic.
To avoid adding unnecessary complexity, I am closing this PR for now, but will reopen it if I identify a clear, reproducible bug in the path handling.

PR #5150, opened 2026-01-05
chore: bump axios/tmp/playwright

niStee:
Bump direct dependencies (axios/tmp/playwright) to address audit warnings. Overrides removed to avoid devtools breakage. No behavioral code changes.

package.json + lockfile updates only. Lint/prettier unchanged.

pnpm test: 2 existing backend test failures on Windows due to path separator expectations; 18/20 suites pass.

pnpm run test:e2e: 5/6 tests pass; webview_controls still fails because the webview URL input never appears (known GitHub/consent/webview issue).

---


arielj:
this is doing kinda the same as https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5083

niStee:
> this is doing kinda the same as https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5083

Indeed

package.json
```diff
@@ -22,6 +22,23 @@
...
+      "js-yaml@<3.14.2": "3.14.2",
+      "js-yaml@>=4.0.0 <4.1.1": "4.1.1",
+      "brace-expansion@1": "1.1.12",
+      "brace-expansion@2": "2.0.2",
+      "@eslint/plugin-kit": "0.3.4",
+      "tmp": "0.2.4",
+      "vite": "6.4.1",
+      "@playwright/test": "1.55.1"
+    }
+  },
```
CommandMC: Pnpm's config no longer lives in this file

> The local configuration file is located in the root of the project and is named pnpm-workspace.yaml.

\- https://pnpm.io/cli/config

You may be using an outdated version of pnpm

CommandMC (changes requested):
As discussed in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5083, using overrides to force dependencies to use newer dependencies themselves is not stable. We should look into updating our own dependencies, and replacing/dropping ones without new versions instead

niStee:
> As discussed in #5083, using overrides to force dependencies to use newer dependencies themselves is not stable. We should look into updating our own dependencies, and replacing/dropping ones without new versions instead

I completely agree that we should update our dependencies. In the meantime, as requested, I have moved the overrides for transitive dependencies, such as `cross-spawn`, to the `pnpm-workspace.yaml` file. However, updates to direct dependencies, such as `axios`, are still appearing in `package.json`.

Please let me know if you think this looks better. :)

arielj:
I'm still not sure why we need the overrides thing here, I already mentioned before that it's doing the same thing as https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5083 and I already left comments there about the overrides maybe not being correct (we don't want to break react devtools for example)

niStee:
> I'm still not sure why we need the overrides thing here, I already mentioned before that it's doing the same thing as #5083 and I already left comments there about the overrides maybe not being correct (we don't want to break react devtools for example)

Thanks, @arielj. I've removed all the overrides and am now only using direct bumps for the axios/tmp/playwright files to avoid risks associated with DevTools. I've updated the title, too.

niStee:
Re‑ran tests after regenerating the lockfile: same 2 Windows path‑separator backend test failures, and the webview_controls e2e still fails due to Google’s consent redirect changing the expected URL; no new regressions from the dependency bumps.

flavioislima:
@niStee can you run `pnpm i` to update the lock-file? It broke after I merged main to fix the conflicts.

PR #5151, opened 2026-01-05
Update types/node to last 22.x version

a5ehren:
Rename the patch, it applies cleanly.

---


a5ehren:
Bad is probably too strong, but I think a better approach (that doesn't require carrying a patch to a package that updates a lot) would be to have the linter block the use of process.platform and direct people to the helpers in `environment.ts` instead?

a5ehren:
hrm that actually ends up being a crapton of code changes. I'll just bump the patch.

PR #5155, opened 2026-01-06
[Tech] Remove unnecessary dependencies

CommandMC:
Cleaning up some dependencies we either don't use or only use in one place (wherever easy substitutions were available)

One package I was unsure about was `sass`. Vite has native support for Sass and I didn't see the package imported anywhere, but I figured it might be required for some other tool (e.g. IDE integration)

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,10 @@
   "devDependencies": {
     "@eslint/js": "^9.29.0",
     "@playwright/test": "^1.52.0",
+    "@testing-library/dom": "^9.3.4",
+    "@testing-library/jest-dom": "^5.17.0",
+    "@testing-library/react": "^14.3.1",
+    "@testing-library/user-event": "^14.6.1",
     "@types/fs-extra": "^11.0.4",
     "@types/graceful-fs": "^4.1.9",
     "@types/ini": "^1.3.34",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -164,6 +164,18 @@ importers:
       '@playwright/test':
         specifier: ^1.52.0
         version: 1.52.0
+      '@testing-library/dom':
+        specifier: ^9.3.4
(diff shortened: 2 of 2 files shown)


flavioislima:
@CommandMC did something break when you removed SASS? I dont think it is being used anywhere, I think it is fine to remove it.

package.json
flavioislima, line 141: This could be removed as well, it won't work for a long time now afaik

CommandMC: The devtools work fine here

image

flavioislima, line 86: This package is not maintained anymore we should replace it in the future. Not in the scope of this PR though.

CommandMC:
Nothing on my end broke with `sass` removed, but I'm using Webstorm (which has support for basically every web technology under the sun built-in). With other editors/tools, this may not be the case

flavioislima:
> Nothing on my end broke with `sass` removed, but I'm using Webstorm (which has support for basically every web technology under the sun built-in). With other editors/tools, this may not be the case

I dont think it is a dependency of any IDE, this is just the SASS compiler, I have no idea when it ended up here. It basically compiles SASS code to CSS and we dont even have a script to do that and vite will compile any SASS code we have. So it is safe to remove.

package.json
flavioislima: true

arielj:
I have a goal to eventually mode all sass files to be css (we kinda only use the nesting and it's now native in modern css)

maybe I should go one day and just rename all the files after the next release :sweat_smile:

arielj:
about the testing library stuff, I know we don't use it currently, but I made this PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4809/files some time ago to actually start adding some frontend tests and it uses some of them

I do think we should have frontend unit testing (we should have way more testing in general :grimacing:)

PR #5159, opened 2026-01-08
[DX] Replace all SCSS files with CSS files

arielj:
This PR replaces all SCSS/SASS files with plain CSS files

We mostly used SASS for the nesting and just a handful of `@mixin` and `@use` that are really easy to replace with CSS `@import`.

In theory, this should also speed up the build time a bit since it won't have to convert the SCSS files into CSS files.

It would be nice to merge this after the next release.

---


a5ehren:
Does this let you remove sass from packages.json?

arielj:
> Does this let you remove sass from packages.json?

that's the idea, yes, I just prefer to remove the dependency in another PR (we have many PRs related to dependencies, it's being removed in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5155/files) so it's easier to put the package back in if it's actually needed for any reason

PR #5160, opened 2026-01-09
[UX] Some QOL improvements for mac settings and support

arielj:
This PR fixes a few this related to settings on Mac, both for the end user and for people in support

I'll add notes in the diff

---

### DIFF
--- a/src/backend/config.ts
+++ b/src/backend/config.ts
@@ -20,13 +20,7 @@ import {
 } from './utils/compatibility_layers'
 import { backendEvents } from './backend_events'
 import { configStore } from './constants/key_value_stores'
-import {
-  isFlatpak,
-  isLinux,
-  isMac,
-  isIntelMac,
-  isWindows
-} from './constants/environment'
+import { isFlatpak, isLinux, isMac, isWindows } from './constants/environment'
 import {
   configPath,
   defaultWinePrefix,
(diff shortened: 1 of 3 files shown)


src/backend/config.ts
```diff
@@ -323,7 +317,7 @@ class GlobalConfigV0 extends GlobalConfig {
       enableUpdates: false,
       addDesktopShortcuts: false,
       addStartMenuShortcuts: false,
-      autoInstallDxvk: isLinux || isIntelMac,
+      autoInstallDxvk: isLinux || isMac,
```
arielj, line 320: we want this setting also for apple silicon macs so wine-crossover and wine-staging is not broken by default

src/backend/launcher.ts
```diff
@@ -401,22 +401,22 @@ function filterGameSettingsForLog(
 
     if (notNative) {
       const wineType = gameSettings.wineVersion
+      delete gameSettings.preferSystemLibs
+      delete gameSettings.autoInstallVkd3d
+      delete gameSettings.autoInstallDxvkNvapi
```
arielj, line 406: this settings are not available on macos

src/frontend/screens/Settings/components/AdvertiseAvxForRosetta.tsx
```diff
@@ -18,8 +18,8 @@ const AdvertiseAvxForRosetta = () => {
     false
   )
 
-  // Only show on macOS when using toolkit wine and not native games
-  if (!isMac || isMacNative || wineVersion.type !== 'toolkit') {
+  // Only show on macOS when using toolkit wine or commercial crossover, and not native games
+  if (!isMac || isMacNative || wineVersion.type === 'wine') {
```
arielj, line 22: this env variable actually works on Crossover too, I couldn't find docs about it but I did try adding the env variable and it seems to work

flavioislima (approved):
LGTM

PR #5161, opened 2026-01-09
[Refactor] Reduce the number of requests that we do at boot time

arielj:
The idea of this PR is to introduce a new way of fetching the latest releases of things so we can avoid doing a lot of unnecessary requests to GitHub when the app starts.

I created this repo https://github.com/Heroic-Games-Launcher/releases-info that runs daily and generates a JSON file with a lot of information about the different tools we use: wine-ge, ge-proton, gptk, dxvk, vkd3d, anticheat files, etc

The main idea is that, at boot, we only fetch that file, and we can get information about the latest releases and versions of the different components and then compare with what we already have locally, and only then fetch the files or version from GitHub if we identify there's something new to fetch.

Updates for DXVK, VKD3D, DXVK mac, and DXMT are not implemented in this PR though. For DXVK/VKD3D I still need to figure out what's the best way. For DXMT we currently don't support it but I'm planning to add support for it soon and I'll be using this new feature to keep that updated.

EDIT: one more thing about this... eventually we could automate the releases-info.json file to be sent to some CDN to avoid GitHub completely on boot unless needed, I think for now it's good enough to reduce the number of requests, but we can investigate that if there's still some issues related to rate limiting

---

### DIFF
--- a/src/backend/anticheat/utils.ts
+++ b/src/backend/anticheat/utils.ts
@@ -1,17 +1,43 @@
 import { writeFile, readFile } from 'fs/promises'
-import { logInfo, LogPrefix, logWarning } from 'backend/logger'
+import { logDebug, logInfo, LogPrefix, logWarning } from 'backend/logger'
 import { AntiCheatInfo } from 'common/types'
 import { runOnceWhenOnline } from '../online_monitor'
 import { axiosClient } from 'backend/utils'
 import { join } from 'node:path'
 import { appFolder } from 'backend/constants/paths'
 import { isMac, isWindows } from 'backend/constants/environment'
+import { createHash } from 'node:crypto'
+import { createReadStream } from 'graceful-fs'
+import { finished } from 'node:stream/promises'
+
+async function createMD5(filePath: string) {
+  const hash = createHash('md5')
+  const rStream = createReadStream(filePath)
+  rStream.pipe(hash)
+  await finished(rStream)
+  return hash.digest('hex')
+}
 
 const anticheatDataPath = join(appFolder, 'areweanticheatyet.json')
 
-function downloadAntiCheatData() {
+async function downloadAntiCheatData(latestFileHash?: string) {
   if (process.env.CI === 'e2e') return
   if (isWindows) return
+
+  const localFileHash = await createMD5(anticheatDataPath)
+
+  if (latestFileHash && localFileHash === latestFileHash) {
+    logDebug(
+      'AreWeAnticheatYet data did not change. Skipping.',
+      LogPrefix.Backend
+    )
+    return
+  }
+
+  logDebug(
+    'AreWeAnticheatYet data changed. Downloading latest file.',
+    LogPrefix.Backend
+  )
 
   runOnceWhenOnline(async () => {
     const url = isMac
@@ -19,8 +45,12 @@ function downloadAntiCheatData() {
       : 'https://raw.githubusercontent.com/Starz0r/AreWeAntiCheatYet/HEAD/games.json'
 
     try {
-      const { data } = await axiosClient.get(url)
-      await writeFile(anticheatDataPath, JSON.stringify(data, null, 2))
+      const { data } = await axiosClient.get<string>(url, {
+        responseType: 'text',
+        transformResponse: [(d) => d] // disable JSON parsing
+      })
+
+      await writeFile(anticheatDataPath, data)
       logInfo(`AreWeAntiCheatYet data downloaded`, LogPrefix.Backend)
     } catch (error) {
       logWarning(
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -1,5 +1,5 @@
 import { initImagesCache } from './images_cache'
-import { downloadAntiCheatData } from './anticheat/utils'
+import { fetchLastestReleases } from './utils/releases'
 import { DiskSpaceData, StatusPromise, WineInstallation } from 'common/types'
 import * as path from 'path'
 import {
@@ -111,7 +111,6 @@ import {
   initStoreManagers,
   libraryManagerMap
 } from './storeManagers'
-import { updateWineVersionInfos } from './wine/manager/utils'
 import { addNewApp } from './storeManagers/sideload/library'
 import {
   getGameOverride,
@@ -204,16 +203,6 @@ async function initializeWindow(): Promise<BrowserWindow> {
       checkRosettaInstall()
     }
   }, 2500)
-
-  if (!isWindows && !isCLINoGui) {
-    setTimeout(async () => {
-      try {
-        await updateWineVersionInfos(true)
-      } catch (error) {
-        logError(error, LogPrefix.Backend)
-      }
-    }, 5000)
-  }
 
   const globalConf = GlobalConfig.get().getSettings()
 
@@ -457,7 +446,7 @@ if (!gotTheLock) {
       backendEvents.emit('languageChanged')
     })
 
-    downloadAntiCheatData()
+    fetchLastestReleases()
 
     initTrayIcon(mainWindow)
 
--- a/src/backend/utils/releases.ts
+++ b/src/backend/utils/releases.ts
@@ -1,0 +1,27 @@
+import { backendEvents } from 'backend/backend_events'
+import { isWindows } from 'backend/constants/environment'
+import { LogPrefix, logWarning } from 'backend/logger'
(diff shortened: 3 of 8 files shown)


src/backend/anticheat/utils.ts
```diff
@@ -1,26 +1,55 @@
...
 
     try {
-      const { data } = await axiosClient.get(url)
-      await writeFile(anticheatDataPath, JSON.stringify(data, null, 2))
+      const { data } = await axiosClient.get<string>(url, {
+        responseType: 'text',
+        transformResponse: [(d) => d] // disable JSON parsing
+      })
+
+      await writeFile(anticheatDataPath, data)
```
arielj, line 53: we have to use a `text` responseType because the original file is not correctly indented and if we get json and save it we get correct indentation, which breaks the md5 comparison since the files would be always different even if the content is the same

src/backend/main.ts
```diff
@@ -205,16 +204,6 @@ async function initializeWindow(): Promise<BrowserWindow> {
...
-  if (!isWindows && !isCLINoGui) {
-    setTimeout(async () => {
-      try {
-        await updateWineVersionInfos(true)
-      } catch (error) {
-        logError(error, LogPrefix.Backend)
-      }
-    }, 5000)
-  }
-
```
arielj, line 217: we don't need this anymore, we only fetch the new wine versions if there's any, and we only run that when the frontend is ready (so it shouldn't run when using heroic in cli mode)

src/backend/utils/releases.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+  'wine-crossover': {
+    tag: string
+    published_at: string
+  }
+  anticheatFiles: {
+    shaMac: string
+    shaLinux: string
+  }
+  dxvk: {
```
arielj: the releases info file in the new repo includes this info, but I'm not handling these tools in this PR cause I'm not sure how to yet, but I'm adding the type anyway so it's there

src/backend/utils/releases.ts
```diff
@@ -0,0 +1,73 @@
...
+  runOnceWhenOnline(async () => {
+    const url =
+      'https://raw.githubusercontent.com/Heroic-Games-Launcher/releases-info/refs/heads/main/release-data.json'
+
+    try {
+      const { data } = await axiosClient.get<ReleasesData>(url)
+      void downloadAntiCheatData(
+        isMac ? data.anticheatFiles.shaMac : data.anticheatFiles.shaLinux
+      )
+      updateWineListsIfOutdated(data)
```
CommandMC: By voiding these promises (once here, once in `updateWineListsIfOutdated`), you make the try-catch a lot less useful. Ideally await the promises instead

src/backend/anticheat/utils.ts
```diff
@@ -1,26 +1,55 @@
...
+
+    const rStream = createReadStream(filePath)
+    rStream.on('data', (data) => {
+      hash.update(data)
+    })
+    rStream.on('end', () => {
+      res(hash.digest('hex'))
+    })
+  })
+}
```
CommandMC: You can simplify this by piping the rStream into the hash and using [`steam/promises`' `finished` function](<https://nodejs.org/api/stream.html#streamfinishedstream-options>):
```suggestion
async function createMD5(filePath) {
    const hash = createHash('md5')
    const rStream = createReadStream(filePath)
    rStream.pipe(hash)
    await finished(rStream)
    return hash.digest('hex')
}
```
(there are probably even leaner ways of writing this, but I'm not that experienced with the Stream API)

src/backend/anticheat/utils.ts
CommandMC, line 53: You could alternatively also use the `stream` responseType and pipe the data into the file (essentially the reverse of the createMD5 function above)
```ts
const wStream = createWriteStream(anticheatDataPath)
const { data } = await axiosClient.get<ReadStream>(url, {
  responseType: 'stream'
})
data.pipe(wStream)
await finished(data)
```
The file is likely not that large though, so keeping it in memory like you're doing it right now is probably fine

src/backend/utils/releases.ts
```diff
@@ -0,0 +1,73 @@
...
+    tag: string
+    published_at: string
+  }
+}
+
+// fetch latest versions of wine/proton/gptk and anticheat data if needed
+export const fetchLastestReleases = () => {
+  if (process.env.CI === 'e2e') return
+  if (isWindows) return
+
```
CommandMC, line 12: Should we maybe also gate this behind some persistent timeout (e.g. "only re-download the file once every x hours")?

src/backend/wine/manager/utils.ts
```diff
@@ -27,8 +28,104 @@ export const wineDownloaderInfoStore = new TypeCheckedStoreBackend(
   }
 )
 
+function getLatestLocalVersions() {
```
CommandMC: The return type of this function is not immediately visible (both to myself and the language server). Ideally add one manually (I assume something like `Record<string, string>` would be fine?)

src/backend/wine/manager/utils.ts
```diff
@@ -27,8 +28,104 @@ export const wineDownloaderInfoStore = new TypeCheckedStoreBackend(
...
+
+  const latestLocalVersions = getLatestLocalVersions()
+  const repositoriesToFetch = []
+
+  // compare dates to know which repositories to fetch
+  if (isLinux) {
+    if (latestLocalVersions?.latestGEProton) {
+      if (
+        Date.parse(latestLocalVersions.latestGEProton) <
+        Date.parse(releasesData['ge-proton'].published_at.replace(/T.*/, ''))
```
CommandMC: Why chop off the time here? There have been cases of multiple releases happening on the same day (say some critical issue is found in a release, a hotfix is created for that right after), so limiting us to a single day may give incorrect results

arielj: we are currently storing just the date, not the whole timestamp

```
{
	"wine-releases": [
		{
			"version": "GE-Proton-latest",
			"type": "GE-Proton",
			"date": "2026-01-01",
			"download": "https://github.com/GloriousEggroll/proton-ge-custom/releases/download/GE-Proton10-28/GE-Proton10-28.tar.gz",
			"downsize": 513714713,
			"disksize": 0,
			"checksum": "https://github.com/GloriousEggroll/proton-ge-custom/releases/download/GE-Proton10-28/GE-Proton10-28.sha512sum"
		},
```

So this ends up comparing `2026-01-01` of the local version with `2026-01-01T22:56:26Z` from the releases-info file and it's always older.

I guess we have to start tracking the releases versions with the whole timestamp instead?

users can still click the refresh button in the Wine Manager to trigger the whole refresh of all wines though if something changes multiple times in a single day, this PR only impacts what happens when heroic boots but shouldn't prevent users from getting the latest at the moment they click

src/backend/utils/releases.ts
arielj: I do want these promises to not make the whole thing fail though

like... if the anticheat download fails I don't want it to prevent the updateWineListsIfOutdated call cause they are unrelated

the try catch is mainly for the axios request that can fail and that should prevent the rest of the promises, but downloadAntiCheatData and updateWineListsIfOutdated should be independent

I also thought about decoupling this even more, just get the data with axios and emit a backend event like `releases-info-ready` and passing the data as event data, and then we can add an event listener in other modules to make it more clear that they are all totally independent as long as the releases info is fetched successfully

src/backend/anticheat/utils.ts
arielj: I replaced this with this version using finished

src/backend/utils/releases.ts
arielj: maybe we can go with this for now? if it creates issues in the future maybe we can add some timeout but this already improves things a lot and it's already more complex than I expected haha

src/backend/wine/manager/utils.ts
arielj: I added this type

src/backend/utils/releases.ts
arielj: I went ahead and replaced this with a backendEvent which feels way better now for me (the previous implementation calling function so coupled didn't feel right)

I think it's now more obvious that these are all independent things and only run when the releases info is downloaded

flavioislima (approved):
Tested here and it works fine. 
I noticed Heroic is a bit more responsive as well which is a plus.

PR #5162, opened 2026-01-10
Update to Electron 39

CommandMC:
Testing the app for a bit on Linux, I can't find any obvious issues. Windows and macOS testing would be appreciated.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
     "@types/tmp": "^0.2.6",
     "@vitejs/plugin-react-swc": "^3.9.0",
     "cross-env": "^7.0.3",
-    "electron": "^36.9.4",
+    "electron": "^39.2.7",
     "electron-builder": "^26.0.12",
     "electron-vite": "^3.1.0",
     "eslint": "^9.29.0",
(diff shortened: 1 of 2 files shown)


arielj (approved):
I tested in macos an it seems to be fine (I clicked around, installed and launched some games, changed settings, etc)

I think we should merge this right after a release, so we then have way more time to keep testing maybe? since we are kinda close to a release

CommandMC:
I would argue we should update *before* the release. Our current release has been out for 6 months, and Electron 36 (our current version) is already in end-of-support mode (has been since October). I don't want to have an EoS version out for potentially another 6 months

arielj:
I'm fine with that, I'm just always scared to update thing close to a release when we had so much time to upgrade not this close (I guess we have to be better at upgrading electron sooner during development)

not sure if we want to wait for someone to test on windows before merging this

BananaWorks07:
Could someone test it on steam deck game mode as well, electron 39 uses Wayland by default now just want to make sure that  electron knows it has to fall back to x11 or if we have to do that manually

CommandMC:
I can confirm Electron 39 works fine in the Game Mode of SteamOS

Issue #4791, opened 2025-07-23
Mangohud and Gamescope flatpaks need to be matched to the runtime.. Heroic needs to indicate this


Sunspark-007:
### Problem description

In the most recent 2.18 update Heroic indicated they lowered the runtime to 23.08 because of issues with the Steam Deck. This is fine.

I feel however, a piece of information needs to be communicated to the user, if not built into Heroic itself.

Namely, that Mangohud and Gamescope runtime versions need to be matched to Heroic. E.g. a 24.x Gamescope won't be recognized or used by Heroic 2.18 on the SD until it is changed.

Discovery store is not the way to do this because it won't show you the versions.

All this has to be done on the command line..

If you uninstall the flatpak of gamescope and mangohud on the command line and then go ahead to install them again on the command line it will give you a choice of 4 versions and in the case of 2.18 it needs to be the 23.x version.

### Feature description

Heroic could offer to the end-user to uninstall the incompatible versions and install the compatible versions that will match Heroic or alternatively tell the user what they need to do.

arielj:
This should already be a thing, is not working for you? https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4588 https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4729

we are showing that those tools need to have the same runtime as heroic

what do you see in the gamescope tab for example if you don't have the correct gamescope installed?

Sunspark-007:
This is what it looks like with version 24 of gamescope installed.

<img width="796" height="212" alt="Image" src="https://github.com/user-attachments/assets/a01de41e-c80a-479c-855c-6c6655c5eb59" />

In my opinion, I feel it could communicate better.

E.g. "We could not find a compatible version of Gamescope. Install Gamescope's flatpak package with runtime version $heroic_runtime_version (23.08 in this case) using the system console and restart Heroic."

As-is, it basically says figure it out.. for people who have only ever used Discover store, etc. they won't understand because from their perspective it IS installed and the store won't let you pick which version you want.

arielj:
looks like a bug, it is supposed to show the runtime version

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4729/files#diff-58f52948912477a1de2daaf437944a10dee55c19049ee9352ba4f790b30c890bR84

Sunspark-007:
I don't have the ability to convert the label from feature-request to a bug. :(

[...]

PR #5174, opened 2026-01-13
[Fix] Get current flatpak runtime so it's not hardcoded

arielj:
So, I ran this:

```bash
flatpak run --command=bash com.heroicgameslauncher.hgl
cat /app/manifest.json
```

and it prints a lot of the flatpak metadata for heroic, starting with:

```json
{
  "id" : "com.heroicgameslauncher.hgl",
  "runtime" : "org.freedesktop.Platform",
  "runtime-version" : "25.08",
  "runtime-commit" : "de826b812cbcd018885f245c76a218d31762a1c03bc60f4af897a2681562c899",
  "sdk" : "org.freedesktop.Sdk",
  "sdk-commit" : "87329bcc381871a8a04dfbba15c8303321ac729b09bd26919d7b2e0d1120d739",
  "base" : "org.electronjs.Electron2.BaseApp",
  "base-version" : "25.08",
  "base-commit" : "3242bb61fa04b9114d23f9a204f0479c2704b6a498e8f92abfa3d9132dbd263b",
  "command" : "heroic-run",
  "modules" : [
    ...
```

We can read that file to get the current runtime version so the constant that we use in other places is not missing.

closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4791

---

### DIFF
--- a/src/backend/constants/environment.ts
+++ b/src/backend/constants/environment.ts
@@ -8,7 +8,7 @@ interface Manifest {
 
 function readFlatpakRuntimeVersion() {
   try {
-    const manifestContent = readFileSync('/app/manifest.json').toString()
+    const manifestContent = readFileSync('/app/manifest.json', 'utf8')
     const manifest = JSON.parse(manifestContent) as Manifest
     return manifest['runtime-version']
   } catch (error) {


src/backend/constants/environment.ts
```diff
@@ -1,5 +1,24 @@
...
+  'runtime-version': string
+}
+
+function readFlatpakRuntimeVersion() {
+  try {
+    const manifestContent = readFileSync('/app/manifest.json').toString()
+    const manifest = JSON.parse(manifestContent) as Manifest
+    return manifest['runtime-version']
+  } catch (error) {
+    console.log(
```
arielj, line 15: I'm using console.log here so we don't need to import the logger just for this constant which can bring many other imports into the mix which can create issues during tests. This line is not that relevant for the users anyway to be in the logs.

a5ehren:
Could this be injected into the translation strings? You did something similar in #5042 but it never got merged.

arielj:
> Could this be injected into the translation strings? You did something similar in #5042 but it never got merged.

#5402 was rebased on top of this branch to include this change

a5ehren:
Whoops, sorry.

src/backend/constants/environment.ts
CommandMC, line 1: I don't think this should be in the `constants` folder. Yes, the runtime version *is* a constant, but since it's more work than just reading an environment variable to find (and you actively have to avoid importing things here to not cause circular imports), it may be better to move it into a `backend/utils/flatpak/index.ts` (or perhaps `.../os/flatpak/...`) file instead

arielj: hmmm not sure what others think, I kinda prefer this to be here since it's only going to be used here and nowhere else ever

a5ehren: Guess it depends on what you want "environment.ts" to mean? If it is "environment variables from the host", then it needs to move. If it is "this is about the environment we run in", than this place makes sense.

flavioislima: I agree with CommandMC on that because then it is easier to use in translations strings as well.
Another improvement could be to have it with the SystemInfo so when we show something like: "Running on Flatpak" or whatever we could include the runtime on it and show it on the settings screen.

PR #5176, opened 2026-01-14
[tech] Update all builder components to their latest versions.

a5ehren:
* Note: You must `pnpm rm electron-builder; pnpm add -D
  electron-builder@latest` to update that package, else the lockfile
  does not properly update all the Peers. It is not clear if this is
  an electron-builder bug or pnpm bug.

* Note 2: You must `pnpm rm electron-vite @vitejs/plugin-react-swc
  vite-plugin-svgr; pnpm add -D electron-vite; pnpm add -D
  @vitejs/plugin-react-swc vite-plugin-svgr` to get an updated version
  of the vite compiler.

Afaict there are no breaking changes or massive updates in any of these packages, but its just another step on the road to React 19.

---


a5ehren:
Fwiw I'm not sure either of the dead-code checkers in this project actually work. Both are marked unsupported/deprecated.

package.json
```diff
@@ -128,11 +128,11 @@
     "@types/semver": "^7.7.0",
     "@types/source-map-support": "^0.5.10",
     "@types/tmp": "^0.2.6",
-    "@vitejs/plugin-react-swc": "^3.9.0",
+    "@vitejs/plugin-react-swc": "^4.2.2",
     "cross-env": "^7.0.3",
     "electron": "^36.9.4",
-    "electron-builder": "^26.0.12",
-    "electron-vite": "^3.1.0",
+    "electron-builder": "26.4.1",
```
CommandMC: You dropped the `^` here, was this intentional?

package.json
```diff
@@ -156,9 +156,7 @@
     "typescript-eslint": "^8.34.1",
     "undici": "^7.8.0",
     "unimported": "^1.31.1",
-    "vite-plugin-svgr": "^4.3.0"
-  },
-  "resolutions": {
-    "ts-morph": "17.0.1"
+    "vite": "^7.3.1",
```
CommandMC: `vite` should not have to be an explicit dependency. It is pulled in by `electron-vite` & never used directly

CommandMC:
> Fwiw I'm not sure either of the dead-code checkers in this project actually work. Both are marked unsupported/deprecated.

`ts-prune` should work, we run that as part of the "Lint" workflow (and it seems to be failing here). Personally I don't use either one apart from that

package.json
a5ehren: nope, I'll fix it

package.json
a5ehren: I would think so, but it keeps pulling 6.x even though electron-vite 5.x sets 7 as allowed. I'll try the rm/add trick that I had to do for electron-builder and see if it does anything in the lockfile.

a5ehren: Yeah, if you don't want vite as an explicit depend but want to update it without running `pnpm up` you have to do the rm/add trick for all its peers.

a5ehren:
> > Fwiw I'm not sure either of the dead-code checkers in this project actually work. Both are marked unsupported/deprecated.
> 
> `ts-prune` should work, we run that as part of the "Lint" workflow (and it seems to be failing here). Personally I don't use either one apart from that

OK, the problem here was me removing the ts-morph override, which put it back to an old version that can't read the code. I've bumped it to match our current TS version and it passes locally. 

It looks like all of these checkers conceded the space to https://knip.dev/ so I'll see about getting a config for that in another PR.

electron.vite.config.ts
```diff
@@ -1,4 +1,4 @@
-import { defineConfig, externalizeDepsPlugin } from 'electron-vite'
```
a5ehren, line 1: These changes are from electron-vite deprecating this plugin and moving it to the main build options

package.json
```diff
@@ -154,10 +154,9 @@
     "typescript": "^5.8.3",
     "typescript-eslint": "^8.34.1",
     "undici": "^7.8.0",
-    "unimported": "^1.31.1",
-    "vite-plugin-svgr": "^4.3.0"
+    "vite-plugin-svgr": "^4.5.0"
   },
   "resolutions": {
-    "ts-morph": "17.0.1"
+    "ts-morph": "26.0.0"
```
a5ehren, line 154: This needs to be bumped to keep up with the TS package - without the override, ts-prune installs v13 of ts-morph, which does not really understand modern TS.

pnpm-workspace.yaml
```diff
@@ -1,15 +1,17 @@
...
 onlyBuiltDependencies:
   - '@fortawesome/fontawesome-common-types'
   - '@fortawesome/fontawesome-svg-core'
   - '@fortawesome/free-brands-svg-icons'
   - '@fortawesome/free-regular-svg-icons'
   - '@fortawesome/free-solid-svg-icons'
+  - '@parcel/watcher'
   - '@swc/core'
   - electron
+  - electron-winstaller
```
a5ehren, line 10: these are new pnpm-install scripts from the bumped dependencies. they don't really matter, but it gets rid of an annoying warning to allow them

a5ehren:
Apologize for the gore on that last force-push, but apparently the JS ecosystem has suddenly decided to favor single-quotes. pnpm and prettier did the same thing when asked? Not sure exactly.

PR #5186, opened 2026-01-17
[UX] Do not color logs to avoid confusing lines

arielj:
Currently, we are coloring the warning and error messages in the logs, but the coloring mostly flags errors and warnings that we know can be completely ignored but it gives users the notion that they need to solve them when an game fails to run.

Like... this is for a game that worked just fine for me:
Captura de pantalla 2026-01-17 a la(s) 8 51 44 p  m

There's a lot more error messages that can be ignored in general and most of us know after all the time providing support but it can be really confusing for others.

But for a user trying to get a game to run, all these red error messages are completely misleading and can make them waste a lot of time.

And I added one more change that should help support: many users copy/paste the content of the log dialog not knowing when it's truncated (we show a not-too-visible message about it being truncated)

I changed it a bit to include the message as part of the logs, so it's way easier to notice, either when they select the content or paste it somewhere, and now it's shown both at the beginning and at the end of the truncated log

---

### DIFF
--- a/src/frontend/screens/Settings/sections/LogSettings/index.tsx
+++ b/src/frontend/screens/Settings/sections/LogSettings/index.tsx
@@ -23,48 +23,22 @@ interface LogBoxProps {
 const LogBox: React.FC<LogBoxProps> = ({ logFileContent }) => {
   const { t } = useTranslation()
   const maxLines = 1000
-  let sliced = false
   let lines = logFileContent.split('\n')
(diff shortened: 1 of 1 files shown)


arielj:
another option could be to keep track of error lines that we know are not real errors to fix and completely exclude them from the logs when displayed, but it would be a lot of work to keep updating that list

PR #5191, opened 2026-01-20
[Feature] Add DXMT to wine staging on M chips

arielj:
This PR adds support in heroic to automatically add a copy of a Wine Staging installation but with [DXMT](https://github.com/3Shain/dxmt) in it.

When we are on an M chip, and a wine staging is installed, we now create a copy of the directory, copy the DXMT dlls inside it, and rename it to `...-DXMT` so it shows up in the wine selector as an option.

Uninstalling a Wine Staging version also deletes the corresponding `-DXMT` copy of it.

Some things to note:
- This does not work the same way as DXVK/VKD3D that copy dlls in the prefix, that doesn't work, the files need to be copied inside the wine bundled directory (that's why I make a copy when installed and it's different than the other tools)
- DXMT has experimental support for intel chips, but it's [experimental](https://github.com/3Shain/dxmt/wiki/Device-System-Runtime-Specifications#device) and I'd rather keep the stable support here automated, maybe in the future we can add an experimental feature flag to enable this on intel macs for users that really want to try this (though DXMT is going to stop supporting intel chips at some point too)
- This currently does NOT support updating DXMT inside the wine copy, but we can eventually implement that (with my other PR that checks the releases of all tools, we can then use that event to update all the `-DXMT` wines)

How to test this:
- Install rocket league
- Install any Wine-Staging version
- If you use the normal wine staging just installed, the game will have the known issue with name plates not displaying
- Now switch to the `-DXMT` version of that same wine staging, delete the prefix (so the DXVK stuff is cleared, they conflict with DXMT), and run the game... now the name plates are visible

---

### DIFF
--- a/src/backend/tools/dxmt.ts
+++ b/src/backend/tools/dxmt.ts
@@ -1,5 +1,5 @@
 import { backendEvents } from 'backend/backend_events'
-import { isIntelMac, isMac, isWindows } from 'backend/constants/environment'
+import { isIntelMac, isMac } from 'backend/constants/environment'
 import { logDebug, logError, LogPrefix, logWarning } from 'backend/logger'
 import { isOnline } from 'backend/online_monitor'
 import { installOrUpdateTool } from '.'
@@ -17,9 +17,6 @@ import { toolsPath } from 'backend/constants/paths'
 
 const DXMT = {
   getLatest: () => {
-    if (isWindows) {
-      return
-    }
     if (!isOnline()) {
       logWarning(
         'App offline, skipping possible DXMT update.',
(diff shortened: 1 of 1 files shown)


src/backend/launcher.ts
```diff
@@ -1174,7 +1174,10 @@ function setupWineEnvVars(gameSettings: GameSettings, gameId = '0') {
   if (!gameSettings.enableEsync && wineVersion.type === 'proton') {
     ret.PROTON_NO_ESYNC = '1'
   }
-  if (gameSettings.enableMsync && isMac) {
+  if (
+    isMac &&
+    (gameSettings.enableMsync || wineVersion.name.endsWith('-DXMT'))
```
arielj, line 1179: we want msync always on for DXMT

src/backend/logger/constants.ts
```diff
@@ -7,7 +7,7 @@ const LogPrefix = {
   Nile: 'Nile',
   Zoom: 'Zoom',
   WineDownloader: 'WineDownloader',
-  DXVKInstaller: 'DXVKInstaller',
+  ToolInstaller: 'ToolInstaller',
```
arielj, line 10: I changed this so it doesn't say `DXVK` for all tools, not really needed for this PR (I could have added `DXMTInstaller`) but it was weird to shows `DXVKInstaller` also for VKD3D and nvapi logs

PR #5195, opened 2026-01-21
[Tech] Remove DXVK FPS limit on linux

arielj:
This PR removes the setting for the DXVK FPS frame rate on linux but not on mac. It was removed for DXVK on linux https://github.com/doitsujin/dxvk/pull/5331 but it's still supported for DXVK on mac https://github.com/Gcenx/DXVK-macOS?tab=readme-ov-file#frame-rate-limit

This replaces https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5055 as that one was removing the feature completely, even for mac. On linux the alternative is to use mangohud for the rate limiting but there's no mangohud for macos.

---


Etaash-mathamsetty (approved):
lgtm

PR #5201, opened 2026-01-22
[DX] Add more logs around getWinePath

arielj:
We sometimes get reports with an error like this:

```
Error: ENOENT: no such file or directory, copyfile '/opt/Heroic/resources/app.asar.unpacked/build/bin/x64/win32/EpicGamesLauncher.exe' -> ''
```

Where the `getWinePath` function failed silently and we get an empty destination, but we have no information at all about why it failed.

This PR fixes that so we can see the stderr output of the wine command when we get an empty wine path to understand what's the actual issue.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -1505,7 +1505,7 @@ async function runWineCommand({
   }
 
   if (!(await validWine(wineVersion))) {
-    return { stdout: '', stderr: '' }
+    return { stdout: '', stderr: 'Invalid wine' }
   }
 
   const env_vars: Record<string, string> = {
@@ -1568,7 +1568,7 @@ async function runWineCommand({
         options.onOutput(data, child)
       }
 
-      stdout.push(data.trim())
+      stdout.push(data)
     })
 
     child.stderr.on('data', (data: string) => {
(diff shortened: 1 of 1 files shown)


src/backend/launcher.ts
```diff
@@ -1568,7 +1568,7 @@ async function runWineCommand({
         options.onOutput(data, child)
       }
 
-      stdout.push(data.trim())
+      stdout.push(data)
```
arielj, line 1571: the `.trim()` here was removing the new lines, so the stdout and stderr were all kinda broken

imLinguin:
I'm wondering whether we should maybe hardcode in-prefix paths instead of relying on Wine. 
It wont be as scalable, but I believe we can assume the directory structure to be rather stable

arielj:
> I'm wondering whether we should maybe hardcode in-prefix paths instead of relying on Wine. It wont be as scalable, but I believe we can assume the directory structure to be rather stable

maybe we could have those paths hardcoded as fallbacks? I do think using wine to get the path can help identify other problems (if getting the path which is like the simplest thing to do fail, other things will most likely fail too)

PR #5208, opened 2026-01-25
[feat] add achievements to game info page

lmReef:
# Overview

Feature suggestion to add achievements to the game info pages.
Currently I've just implemented it for GOG games as a poc but if everyone's keen on it I'll look into supporting Epic/Amazon afterwards.

I tried my best to reuse existing functions in the same way as similar code, but let me know if there's anything that could be done better

I really like achievements and felt like it was the main thing missing from Heroic vs Steam etc for me, so I'm happy to make whatever changes to get this in!

## Key features implemented

- list all achievements in the right side bar on game info pages
  - ~tab selected by default if the game has achievements, otherwise info (same as before)~
- completion % in tab bar
- sorted by date unlocked then by rarity
- epic + legendary rarity have glow on icon
- _edit_: hide name and desc for hidden achievements
- _edit_: caching + updates on game session end

## Additional ideas I could be keen to add

- epic / amazon support
- "sort by" dropdown
- completion % on game thumbnail in library view
  - sort library by achievement %

## Screenshots

unlocked sorted to top:
image

sorted by rarity:
image

no achievements:
image

hidden achievements (last one):
image

---

### DIFF
--- a/src/backend/storeManagers/gog/games.ts
+++ b/src/backend/storeManagers/gog/games.ts
@@ -165,7 +165,7 @@ export async function getAchievements(
       return null
     })
 
-  if (!response) {
+  if (!response?.data?.items) {
     return []
   }
 


flavioislima:

lmReef:

src/backend/storeManagers/gog/library.ts
```diff
@@ -978,6 +979,7 @@ export async function gogToUnifiedInfo(
     art_background: background,
     cloud_save_enabled: false,
     art_icon: icon,
+    achievements: await getGamesAchievements(String(info.external_id)),
```
imLinguin: I dont think this should be part of library refresh. Current refresh loop is too bloated already.

Maybe make it a separate IPC call where you could add a caching layer. Main points to consider with fetching this data:
- Cache responses for some time
- Flush the cache after finishing the game session so it can be reloaded

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -150,13 +157,23 @@ export default React.memo(function GamePage(): JSX.Element | null {
...
+
   const backRoute = location.state?.fromDM ? '/download-manager' : '/library'
 
   const storage: Storage = window.localStorage
 
   const [currentTab, setCurrentTab] = useState<
-    'info' | 'extra' | 'requirements'
-  >('info')
+    'achievements' | 'info' | 'extra' | 'requirements'
+  >(hasAchievements ? 'achievements' : 'info')
```
imLinguin: I think the tab shouldn't be the default - users may wish to not see achievements they can unlock. (very border line case of spoiling the fun that can occur)

src/frontend/screens/Game/GamePage/components/Achievements.tsx
imLinguin, line 1: I think the function should also respect whether the achievements are visible, this is a type of data that can easily spoil the game

lmReef: agreed, I missed this in the api response :+1:

src/frontend/screens/Game/GamePage/index.tsx
lmReef: that makes sense though I do reckon it'd be the minority, what about `info` by default + have the game info pages remember the last open tab or add it to settings?

lmReef:
addressed comments:
- moved api call to new ipc call that runs when you open the game info page + when a game session ends
- cached api response
- reverted default tab to info (I got lost in the sauce trying to find a good place to handle a `lastOpenTab` approach, might do later)
- fixed hidden achievements (see last achievement in screenshot)

image

src/backend/storeManagers/gog/games.ts
```diff
@@ -139,6 +142,36 @@ export async function getExtraInfo(appName: string): Promise<ExtraInfo> {
...
+
+  if (!GOGUser.isLoggedIn()) return []
+  const credentials = await GOGUser.getCredentials()
+  if (!credentials) return []
+  const url = `https://gameplay.gog.com/clients/${appName}/users/${credentials?.user_id}/achievements?locale=${lang}`
+
+  const response: AxiosResponse | null = await axiosClient
+    .get(url, {
+      headers: { Authorization: `Bearer ${credentials.access_token}` }
+    })
```
imLinguin: GOG uses a header for this endpoint instead of search param

```suggestion
  const url = `https://gameplay.gog.com/clients/${appName}/users/${credentials?.user_id}/achievements`

  const response: AxiosResponse | null = await axiosClient
    .get(url, {
      headers: { Authorization: `Bearer ${credentials.access_token}`, 'X-Gog-Lc': lang }
    })
```

src/frontend/screens/Game/GamePage/components/Achievements.tsx
```diff
@@ -0,0 +1,48 @@
...
+              src={x.date_unlocked ? x.image_url_unlocked : x.image_url_locked}
+            />
+            <div
+              className={`achievement-text ${x.date_unlocked ? 'unlocked' : 'locked'}`}
+            >
+              <span className="achievement-title">
+                {x.visible ? x.name : 'Hidden Achievement'}
+              </span>
+              <span className="achievement-desc">
+                {x.visible ? x.description : ''}
```
imLinguin: I think we can show the achievement when it's unlocked. Here it would always show as invisible

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -155,9 +173,19 @@ export default React.memo(function GamePage(): JSX.Element | null {
...
+  useEffect(() => {
+    const updateAchievements = async () => {
+      window.api.clearAchievementCache(appName)
+      const updatedAchievements = await window.api.getAchievements(appName)
+      setAchievements(updatedAchievements)
+      achievementStore.set(appName, updatedAchievements)
+    }
+    if (!isPlaying) updateAchievements()
+  }, [isPlaying])
+
```
imLinguin, line 189: This function will always be called when you first visit the GamePage, making the cache for achievements useless. Why dont we tie the caching logic to be completely in the backend component.
All frontend has to care about is to (re)load the data.

lmReef: Fixed the cache clearing with a `useRef`, my reasoning for the frontend store was that it eliminates the tab "pop-in" caused by the data only becoming available after the initial render (after the `useEffect` hook).

I think it feels a lot nicer playing around with it, I'd be keen to add something similar for the other tabs' pop-in for consistency too but maybe out of scope for this PR.

Also happy to remove for now and look at it another time, let me know which is preferable :+1:

lmReef: thanks heaps for the reviews by the way!

BananaWorks07:
> addressed comments:
> 
> * moved api call to new ipc call that runs when you open the game info page + when a game session ends
> * cached api response
> * reverted default tab to info (I got lost in the sauce trying to find a good place to handle a `lastOpenTab` approach, might do later)
> * fixed hidden achievements (see last achievement in screenshot)
> 
> image

Not sure if you have addressed this or not, but would prefer if the hidden achievements icons were also hidden. Maybe we can have some placeholder image or just blur out the image a ton

lmReef:
> Not sure if you have addressed this or not, but would prefer if the hidden achievements icons were also hidden. Maybe we can have some placeholder image or just blur out the image a ton

Yeah I wondered about this too, it _is_ using the icons returned by the api in `image_url_locked` so I assumed it was on the game developers to use a specific icon for locked + hidden achievements 

Just had a look at the same game on gog and they do use a different icon, should heroic have its own or should I just blur the icon _a lot_ for now?

image

image

src/frontend/screens/Game/GamePage/index.tsx
imLinguin: I think just info tab is okay for now. We can always revisit this in the future

src/frontend/screens/Game/GamePage/index.tsx
lmReef: @imLinguin curious if this approach is acceptable, otherwise I can remove it and move on

imLinguin: I think invalidation logic should be done mostly on the backend.

 I imagine we could add a way to read databases directly from comet, which could let it work while offline as well - and showcase a common state.
But I digress

imLinguin:
Some other UI nit-picks:

## vertical overflow
- when window is smaller the achievements break the layout 
- it seems the max height overwrites some parent sizing


<details>
<summary><b>Images</b></summary>
obraz

obraz

</details>

## tabs
There is no way to scroll through them, when I open system info I may not be able to go back to Install Info

<details>
<summary><b>Images</b></summary>
obraz
obraz
</details>


## Contrast
The subtexts may be hard to read
<details>
<summary><b>Images</b></summary>
obraz
</details>

imLinguin:
Hey @lmReef 
Would you be able to address those few points?
I'm thinking of including this PR in 2.21 milestone

lmReef:
Heya @imLinguin
Yep absolutely! I've been a bit busy recently with a new puppy but will have time to finish it this weekend, does that work for 2.21's timeline?

I'll address the vertical overflow, tabs, and contrast but want clarify about the use of the frontend store - I really like how it fixes the pop-in but also agree there could be more streamlined ways to do this (perhaps a more generic `GameDataStore` for all tabs, or rendering the tabs in a `disabled` state while the data is loaded), should I just remove it for now?

imLinguin:
I think having a store is a good thing. But maybe it should be simply moved into the ipc handler I think as far as frontend is concerned it just needs to reload the data when playing state changes. 

Any data processing like cache should be done in backend module.
We eventually want to completely isolate frontend from node api including electron stores and rely only on ipc.

lmReef:
- [ ] vertical overflow
	- it seems this is the behavior for all tabs (also happens on system requirements tab at smaller breakpoints) controlled by the `max-height` property [here](https://github.com/lmReef/HeroicGamesLauncher/blob/feat/game-info-achievements/src/frontend/screens/Game/GamePage/index.css#L101)
- [x] tabs
	- enabled MUI's tab scroll buttons
- [x] contrast
	- used lighter color vars
- [x] store
	- removed the frontend store

flavioislima:
@lmReef just make sure your PR is updated with  `main` before we can merge it.

lmReef:
@flavioislima Awesome cheers! I'll make sure its in sync with `main` next chance I get

I'm not able to merge/update as it says I'm still blocked by requested changes from @imLinguin's review, and won't let me re-request review for some reason (so will just @ here instead)

PR #5215, opened 2026-01-26
[UX] Add AutoSync Saves global setting

veryprofessionaldodo:
Noticed that it would be useful to have a global app setting to enable autosync saves for all games that don't have any game specific config. 

Even though this doesn't look like it works, it's because the work was almost completely done, it was just not really wired for this part. In the `game_config.ts` file, in the `getSettings` function (line 203) we have two main variables, the default and game settings. The default settings already included the `autoSyncSaves` property (line 213), but the global config setting was never being set, only the game config. 

This MR just adds a new setting that sets that global config, and is only used as a fallback, since `game_config.ts` already considers game-specific configs to override default settings. 

So in the end, here's the behaviour we have:

- game does not have any setting related to autosync saves (`undefined`), autosync is inherited from global config
- game has any setting (`true` or `false`), overrides global config

Important: I did a string to the translation file and I can add to the rest of the files, I just want confirmation that this approach and MR are valid before putting in the time to add that part.

Video of it working below: 

https://github.com/user-attachments/assets/823d3d31-6fbe-43aa-b91e-ebdea6cee198

---


CommandMC:
Save sync requires per-game setup at the moment, so adding a global option for it does not make much sense

veryprofessionaldodo:
@CommandMC Yup, but this would basically be a complimentary fallback. I personally don't want to worry if I have multiple devices and making sure I have "autosync" enabled on all devices, for all games. This is just a soft default that the user might want. 

The video goes through it well, but it's really just complementary. If the user already chose an autosync option, use that, if not we can inherit from that setting instead of it being just false.

So in practice, this is an extension of the game-specific setting, that's why I didn't need to add any code in the `game_config.ts` file

CommandMC:
Enabling autosync right now without going into the cloud saves settings and actually *setting* a save path will not work. This setup is required per-game, no matter whether there is a global autosync option

veryprofessionaldodo:
@CommandMC So let me see if I understood the flow correctly. In my mind there are only two options, either
a) I can install a game, enable autosync for that game before ever launching it, and then launch it as normal, without any issues arising with cloud saves, or
b) I can only enable autosync after the game has already run at least once

If it's a) then this MR will work no problem, if it's b) this MR won't work and I'll have to close it :(

CommandMC:
I don't know how I can explain this differently. Auto-sync will only work once you have a save path configured in the "Cloud Saves Sync" category of the game settings. If a path is not configured, it will not work. The path being configured is not related to running the game, you can set the path before or after. The point is that you *have to* go into that tab manually and set the path (either letting Heroic auto-compute it for you, or setting it yourself)

I plan to change this workflow soon (I have posted some progress of that on Discord), but for now, this is how that feature works

veryprofessionaldodo:
@CommandMC Aaahh, okay, now it makes sense. Yeah, this is not really a great workflow indeed :D In that case you're right, this won't work :( I'll close this merge request in that case, maybe something to re-add in the future if the flow changes and it allows for this sort of feature

PR #5230, opened 2026-01-27
[FIX] Launch options fixes

flavioislima:
This should fix the issue of a infinite loop on the game settings for games that does not have launch options.
Also fix the issue some games, Elite Dangerous only so far, showing only a single option (VR in this case) by default.

---

### DIFF
--- a/src/frontend/helpers/library.ts
+++ b/src/frontend/helpers/library.ts
@@ -318,21 +318,6 @@ async function checkLaunchOptionsAndLaunch({
     runner
   )
 
-  const hasDefaultOption = availableLaunchOptions.some(
-    (option) =>
-      (option.type === undefined || option.type === 'basic') &&
-      'parameters' in option &&
-      option.parameters === ''
-  )
-
-  if (availableLaunchOptions.length === 1 && !hasDefaultOption) {
-    availableLaunchOptions.unshift({
-      name: 'Default',
-      parameters: '',
-      type: 'basic'
-    })
-  }
-
   // If no launch options or only one option, launch directly
   if (!availableLaunchOptions.length || availableLaunchOptions.length === 1) {
     // If there's exactly one option, use it
@@ -470,10 +455,6 @@ async function checkLaunchOptionsAndLaunch({
             key: 'lastUsedLaunchOption',
             value: option
           })
-
-          if (launchCanceled) {
-            return res({ status: 'done' })
-          }
 
           res(
             window.api.launch({
--- a/src/frontend/hooks/useLaunchOptions.ts
+++ b/src/frontend/hooks/useLaunchOptions.ts
@@ -28,19 +28,6 @@ export const useLaunchOptions = ({
           appName,
           runner as Runner
         )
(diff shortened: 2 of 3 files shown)


src/frontend/hooks/useLaunchOptions.ts
```diff
@@ -14,26 +15,27 @@ export const useLaunchOptions = ({
...
         )
-        setLaunchOptions(options)
-
-        // Find and set the previously used option
-        if (lastUsedOption && options.length > 0) {
-          const foundIndex = findLaunchOptionIndex(options, lastUsedOption)
-          if (foundIndex !== -1) {
-            setSelectedIndex(foundIndex)
-          }
+        if (options.length === 1) {
```
CommandMC: Do we really only want to add this if we have one option? I would've expected there to always be a "Default" case (launching the game without any special options)

arielj: I think we might need this for games that have more options and none is the "default" (I saw a report on discord for a game with 2 DLC options and no "no parameters" option, I understand this check wouldn't add the "default" in that case)

maybe what we can check is: if any of the option has no parameters, then that's the default and we can skip this, otherwise we add this

flavioislima: I think it also depends on the runner, this fixes the Epic games, for instance, since the GOG has no issue with the current approach. But at least with the games I have that has options, this works for both.

arielj: my concern is: a game that has more than 1 option, but none of the options it has is the `parameters: ""` option

because this check is only checking if there's a single option, but if there's more we are not adding this "default" one

I think that if we change to check if there's no option with `parameters: ""` then we add this default one it should cover all possibilities I think

flavioislima: @CommandMC @arielj changed the approach now to check for a default option with the "" parameter, I can only test this with elite dangerous on epic right now, so if you guys can test with other games that would be good.

flavioislima: Ignore it, it now introduced a new bug for GOG games, I will take another look tomorrow.

src/frontend/helpers/library.ts
```diff
@@ -283,7 +283,23 @@ async function checkLaunchOptionsAndLaunch({
...
+      option.parameters === ''
+  )
+
+  if (availableLaunchOptions.length === 1 && !hasDefaultOption) {
+    availableLaunchOptions.unshift({
+      name: 'Default',
+      parameters: '',
+      type: 'basic'
+    })
+  }
```
arielj: we are doing this same thing in 2 different places now, we could move it to the `getLaunchOptions` handler

flavioislima: Done!

arielj (approved):
Looks good, I tested only Ghostrunner from epic though, I don't know of any GOG game with this issue or other games from epic

PR #5236, opened 2026-01-27
[tech] Update font and icon packages to latest versions

a5ehren:
Just gets some backend improvements, no visual changes that I can see.
---

- [ x ] Tested the feature and it's working on a current and clean install.
- [ x ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,17 +53,17 @@
     "lint-translations": "esbuild --bundle --platform=node --target=node21 meta/lintTranslations.ts | node"
   },
   "dependencies": {
-    "@fontsource/cabin": "^4.5.10",
-    "@fontsource/rubik": "^4.5.14",
-    "@fortawesome/free-brands-svg-icons": "^6.7.2",
-    "@fortawesome/free-regular-svg-icons": "^6.7.2",
-    "@fortawesome/free-solid-svg-icons": "^6.7.2",
-    "@fortawesome/react-fontawesome": "^0.2.2",
+    "@fontsource/cabin": "^5.2.8",
+    "@fontsource/rubik": "^5.2.8",
+    "@fortawesome/free-brands-svg-icons": "^7.1.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -86,15 +87,16 @@
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-i18next": "^12.3.1",
+    "react-is": "^18.3.1",
```
CommandMC: This addition doesn't seem relevant?

CommandMC (changes requested):

package.json
a5ehren: It was in the instructions for migrating the fortawesome stuff to 7.x, but they also have you do an override to match your react version so yeah, not needed.

package.json
```diff
@@ -53,17 +53,18 @@
...
-    "@fortawesome/free-brands-svg-icons": "^6.7.2",
-    "@fortawesome/free-regular-svg-icons": "^6.7.2",
-    "@fortawesome/free-solid-svg-icons": "^6.7.2",
-    "@fortawesome/react-fontawesome": "^0.2.2",
+    "@fontsource/cabin": "^5.2.8",
+    "@fontsource/rubik": "^5.2.8",
+    "@fortawesome/free-brands-svg-icons": "^7.1.0",
+    "@fortawesome/free-regular-svg-icons": "^7.1.0",
+    "@fortawesome/free-solid-svg-icons": "^7.1.0",
+    "@fortawesome/fontawesome-svg-core": "^7.1.0",
```
CommandMC: I don't think we need this package anywhere, right?

To clarify: We use pnpm, so packages do not have access to other packages we have installed. When you see guides say "Install x, and its dependencies y and z", installing y and z will not do anything here and is thus not required

a5ehren: Ah ok, I just followed the site to make sure it kept working, but they only had npm directions.

PR #5240, opened 2026-01-27
[FIX] Only use fake epic exe when turned on with env variable

arielj:
This reverts a change that is causing issues with Fornite with the fake epic exe wrapper turned on by default.

It changes the logic back to be off by default and only enabled with `USE_FAKE_EPIC_EXE=true` (I'm using `"true"` now while before we were only checking presence, so maybe some user has `USE_FAKE_EPIC_EXE=whatever` and that won't work, but they shouldn't, in the rockstar guide we say to use `"true"` https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Rockstar-Games-from-Epic-Games)

EDIT: changed this to support any value except "0" and "false"

"0", "false", and undefined (not setting the env) won't set the fake epic exe

---

### DIFF
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -890,7 +890,7 @@ export async function launch(
 
   // Use the wrapper EXE to launch games.
   if (
-    commandEnv['USE_FAKE_EPIC_EXE'] === 'true' &&
+    ![undefined, '0', 'false'].includes(commandEnv['USE_FAKE_EPIC_EXE']) &&
     existsSync(fakeEpicExePath)
   ) {
     if (isWindows) {


CommandMC:
I would vote for having it enabled on *any* value again, for consistency with how Proton's variables work

arielj:
> I would vote for having it enabled on _any_ value again, for consistency with how Proton's variables work

My concern is that now there are users setting `USE_FAKE_EPIC_EXE=0` to get fornite to work, if we allow any value then the `0` will then make those user use the epic exe

I guess we could change that it's any value other than `undefined/0/"false"`?

BananaWorks07:
> > I would vote for having it enabled on _any_ value again, for consistency with how Proton's variables work
> 
> My concern is that now there are users setting `USE_FAKE_EPIC_EXE=0` to get fornite to work, if we allow any value then the `0` will then make those user use the epic exe
> 
> I guess we could change that it's any value other than `undefined/0/"false"`?

yeah I would agree with that change

arielj:
Updated to accept anything other than `undefined`, `"0"` and `"false"`

PR #5245, opened 2026-01-28
[UI] Make game sub-menu use Dropdown component 

veryprofessionaldodo:
This is a follow-up to a previous PR to add in dropdown components. Made some more extensions and modifications to make that more usable, and changed the dropdown in the game card to now use the same component. This gives us several advantages, but the biggest one is just better usability with gamepad, as well as more standardisation in UI over several screens. 

I've also modified the default behaviour for the dropdowns with a gamepad. Instead of the dropdown appearing when focusing, it will now only allow opening and closing by clicking it. Mouse behavior is unchanged, just the behavior with a gamepad.

Here's a before and after:

Before:

https://github.com/user-attachments/assets/f930db0c-e05e-40db-8f0e-83ce77c57a74

After:

https://github.com/user-attachments/assets/5cff8a74-a770-479d-9fbc-dabe9e44f0a8

### DIFF
--- a/src/frontend/components/UI/Dropdown/index.scss
+++ b/src/frontend/components/UI/Dropdown/index.scss
@@ -41,7 +41,6 @@
     transition: all 200ms;
     pointer-events: none;
     opacity: 0;
-    max-height: 0;
     overflow: hidden;
     gap: 0.3rem;
     padding: var(--space-sm);
--- a/src/frontend/components/UI/Dropdown/index.scss
+++ b/src/frontend/components/UI/Dropdown/index.scss
@@ -1,5 +1,21 @@
 .dropdownContainer {
   position: relative;
+  border-radius: var(--space-3xs);
+  width: 100%;
+  height: 40px;
+  display: flex;
+  justify-content: space-around;
+  background: var(--input-background);
+  box-shadow: 0px 0px 0px 3px var(--search-bar-border, var(--input-background));
+  font-weight: normal;
+  font-size: var(--text-md);
+  color: var(--text-secondary);
+  border: none;
+  background-repeat: no-repeat;
+  margin: 0;
+  -webkit-box-sizing: border-box;
+  -moz-box-sizing: border-box;
+  box-sizing: border-box;
 
   .button {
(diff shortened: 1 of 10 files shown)


arielj:
Not sure why but when I test this locally, the settings and 3-dots buttons have different backgrounds

image

in the description video they look the same though

src/frontend/components/UI/Dropdown/index.scss
```diff
@@ -1,5 +1,21 @@
 .dropdownContainer {
   position: relative;
+  border-radius: var(--space-3xs);
+  width: 100%;
+  height: 40px;
+  display: flex;
+  justify-content: space-around;
+  background: var(--input-background);
+  box-shadow: 0px 0px 0px 3px var(--search-bar-border, var(--input-background));
```
arielj, line 9: it's a bit weird that this is used to give the element a border, is there a reason to do it like this and not have a padding for example?

src/frontend/screens/Game/GameSubMenu/index.css
```diff
@@ -18,6 +18,47 @@
   }
 }
 
+.game-submenu-container {
+  height: 39px;
+  width: 39px;
+  display: flex;
```
arielj, line 24: this display: flex is already in `.dropdownContainer`

src/frontend/components/UI/Dropdown/index.scss
```diff
@@ -1,5 +1,21 @@
...
+  font-weight: normal;
+  font-size: var(--text-md);
+  color: var(--text-secondary);
+  border: none;
+  background-repeat: no-repeat;
```
arielj, line 14: I don't think we need background-repeat set here, it's a color background anyway

src/frontend/components/UI/SelectField/index.css
```diff
@@ -3,6 +3,32 @@
   grid-template-areas: 'label' 'select';
 }
 
+.selectContainer {
+  grid-area: select;
```
arielj, line 7: this grid-area prop is not being used

PR #5249, opened 2026-01-28
[UX] Allow disabling GOG Presence updates

arielj:
This PR adds an option to disable the GOG presence changes when we open and close heroic.

This is not an essential feature and on mac it makes heroic kinda slow (closing heroic triggers the deletePresence function and running legendary/gogdl binaries takes many seconds, heroic takes like 5 seconds or more just to close when nothing else is happening).

This allows users to disable it if they don't care about showing their presence.

An alternative was to disable the playtime sync, but it was not obvious that the 2 things were related, only clear reading the code (actually, this is a question for @imLinguin, was there a reason to skip the gog presence when playtime sync was disabled?).

---

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -638,6 +638,7 @@
         "default-steam-path": "Default Steam path",
         "defaultWinePrefix": "Set Folder for new Wine Prefixes",
         "disable_controller": "Disable Heroic navigation using controller",
+        "disable_gog_presence": "Disable GOG Presence updates",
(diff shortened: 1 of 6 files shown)


imLinguin:
> An alternative was to disable the playtime sync, but it was not obvious that the 2 things were related, only clear reading the code (actually, this is a question for @imLinguin, was there a reason to skip the gog presence when playtime sync was disabled?).

I didnt want to add new gog specific options and just encapsulated this one.

PR #5250, opened 2026-01-28
Update desktop filename to fix generic icon on AppImage

brenobaptista:
A mismatch between the app_id (`heroic`) and the filename for desktop file (`heroic_games_launcher.desktop`) creates this generic Wayland icon instead of the Heroic icon. First I tried to change the filename to contain the scope to Linux only, but I couldn't figure out where it's set (closest was executableName on electron-builder.yml). I also tried to change the appId only for Linux, but the only field that worked was the root packageName.

Get app id: `WAYLAND_DEBUG=1 ./Heroic-*.AppImage 2>&1 | grep set_app_id`
Extract files from AppImage to investigate: `./Heroic-*.AppImage --appimage-extract`

This problem doesn't happen on the Flatpak version.

Let me know if there's a better solution to this or if it can potentially break other builds.

### Before:
before
before-2

### After:
after
after-2

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -64,7 +64,7 @@ linux:
   icon: build/icon.icns
   desktop:
     entry:
-      Name: Heroic Games Launcher
+      Name: Heroic
       Comment[de]: Ein OSS-Spielelauncher für GOG, Epic Games und Amazon Games
       Comment[pl]: Otwartoźródłowy launcher dla GOG, Epic Games i Amazon Games
       Comment[fr]: Un lanceur de jeux open source pour GOG, Epic Games et Amazon Games
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "heroic_games_launcher",
+  "name": "heroic",
   "version": "2.19.1",
   "versionNames": {
     "stable": "Punk 01 - Shaka",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "heroic",
+  "name": "heroic_games_launcher",
   "version": "2.19.1",
   "versionNames": {
     "stable": "Punk 01 - Shaka",


CommandMC:
I'm a little hesitant on changing the package name, since a lot of other build outputs also depend on it. Are you sure we can't change the desktop file name? I would assume that it comes from electron-builder

brenobaptista:
I tried again today and figured it out. It comes from desktop.entry.Name, which was sanitized from "Heroic Games Launcher" to heroic_games_launcher. So by changing it to Heroic I could revert the package name back.

The problem is that now the app name is consistently Heroic instead of "Heroic Games Launcher". I don't know if that's ok for your branding. Looks fine to me.

An alternative to this is using extraMetadata.name to overwrite the package name during build time, but then we would go back to the previous problem.

But since the desktop filename changed, I'm not sure what will happen for existing users that will auto-update to the new version. I assume it will duplicate.

Do you think it's a good idea to put a comment on desktop.entry.Name explaining it should match the package.name? In case somebody wants to change it in the future and introduce a mismatch again.

### Before:
upstream1
upstream2
image
image

### After:
mine1
mine2
image
image

brenobaptista:
I tested Heroic v2.19.1 in a Debian VM:

- GNOME
  - X11:
    - Window title: heroic
    - Icon: incorrect (shows a generic blue icon, possibly GNOME’s generic Wayland icon)
  - Wayland:
    - Window title: heroic
    - Icon: incorrect (same generic blue icon)

- KDE Plasma
  - X11:
    - Window title: Heroic Games Launcher
    - Icon: correct (nice!)
  - Wayland:
    - Window title: Heroic Games Launcher
    - Icon: incorrect (shows the generic Wayland icon)

Tomorrow I'll test if this PR helps GNOME, I only confirmed the fix on KDE Wayland.

brenobaptista:
Good news! I just tested on the VM and this PR also fixes for GNOME on both X11 and Wayland.

(ignore the weird taskbar on Wayland, I installed GNOME and KDE on same VM to test quickly so this is a side effect of that conflict - I tested yesterday with a GNOME-only VM and the effect was the same)

### Before

X11
gnome-x11-before

Wayland
heroic-wayland-before

### After

X11
gnome-x11-after

Wayland
gnome-wayland-after

brenobaptista:
Closing this PR as bumping Electron from v39.2.7 to at least v39.3.0 fixes the issue without changing anything. This existing PR by dependabot fixes the issue: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5458

Explanation:
- Investigation on this problem from another repository (Joplin): https://github.com/laurent22/joplin/pull/14235
- Backport of https://github.com/electron/electron/pull/49290 is available starting [v39.3.0](https://github.com/electron/electron/releases/tag/v39.3.0)
  - Electron gained support for https://wayland.app/protocols/xdg-toplevel-icon-v1
- I can reproduce the bug on v39.2.7, then v39.3.0 and v39.8.5 work.

Screenshot_20260405_180742
Screenshot_20260405_180847
Screenshot_20260405_180928

PR #5251, opened 2026-01-28
[TECH] Remove Wine-GE from the wine manager options

arielj:
This PR removes the "Wine GE" option from the Wine Manager.

Users can still use Wine-GE if they already have it installed, but in most cases wine-ge is a problem for users since it's so outdated and abandoned. We always recommend ge-proton+umu whenever we see a user using wine-ge in discord.

If someone really needs Wine-GE it can be installed manually or using other tools (I'd rather have proton-sarek or other variants eventually in the Wine Manager instead for example).

We were even wasting resources checking for wine-ge updates in every boot.

I can never remember what to do when removing translations and weblate, so I just removed the translations in this PR.

---


CommandMC (commented):
I wonder whether we can add other Proton versions in its place (e.g. [dwproton](<https://dawn.wine/dawn-winery/dwproton>), [Proton-EM](<https://github.com/Etaash-mathamsetty/Proton/>) or [Proton-CachyOS](<https://github.com/CachyOS/proton-cachyos>). The Wine Manager seems a little empty with just one version

arielj:
> I wonder whether we can add other Proton versions in its place (e.g. [dwproton](https://dawn.wine/dawn-winery/dwproton), [Proton-EM](https://github.com/Etaash-mathamsetty/Proton/) or [Proton-CachyOS](https://github.com/CachyOS/proton-cachyos). The Wine Manager seems a little empty with just one version

I agree, I would like to keep those in different PRs though, we can merge this after the release so it's not too empty and then we add more over time before the next one

flavioislima:
Based on the stress we had before with the removal of the DXVK FPS limit, we should maybe use the settings screen to enable/disable repos from the PR #5306.
We can set Wine-GE to be disabled by default but with the option to enable it.
And then we can have all these other repos as well.

flavioislima:
Something like this:
image

CommandMC:
Wine-GE is now officially broken with DXVK 3.0, so it really does not make sense for it to still be available. If people do want to use it (for OpenGL or DX1-6 games), existing versions will remain on their hard drives & they can still manually download & install it

CommandMC (approved):

imLinguin (approved):

PR #5262, opened 2026-01-30
[UX] Hide main window when quitting

CommandMC:
When quitting, we wait for the GOG presence to be updated before calling `app.exit()`. This introduces a noticeable delay between clicking the close button in Heroic & the main window actually being closed.
To be a little more responsive, we can hide the main window once we know we don't need it anymore. Heroic will of course keep running for a short while after the window is closed, but that should be fine

Somewhat a sister PR to https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5249, although this may be safer to merge

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -237,8 +237,6 @@ async function handleExit() {
   const isLocked = existsSync(join(gamesConfigPath, 'lock'))
   const mainWindow = getMainWindow()
 
-  await gogPresence.deletePresence()
-
   if ((isLocked || isRunning()) && mainWindow) {
     const { response } = await showMessageBox(mainWindow, {
       buttons: [i18next.t('box.no'), i18next.t('box.yes')],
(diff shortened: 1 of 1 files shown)


arielj:
I wonder if we should also remove the tray icon? Not super sure but sharing this as an idea, because now heroic closes but the tray icon stays there for a while

CommandMC:
I think the tray icon is fine to have, maybe even a good indicator that Heroic is still running & "cleaning up" (from a user's perspective)

arielj:
> I think the tray icon is fine to have, maybe even a good indicator that Heroic is still running & "cleaning up" (from a user's perspective)

maybe we can then update the tray icon to remove the options and add a disable entry like "Closing..."?

now you can still click things there while heroic is closing (even try to close it again or try to launch a game)

flavioislima (approved):
LGTM

PR #5263, opened 2026-01-30
Also consider "ThirdPartyManagedProvider" attribute when checking for a third party store

CommandMC:
It seems some newer Ubisoft games set this instead of "ThirdPartyManagedApp". This caused us to attempt to get install info for those games, which will of course never work

Relevant Discord thread: https://discord.com/channels/812703221789097985/1466907169008910480

Can't test this, waiting for OP to confirm fix

---

### DIFF
--- a/src/backend/storeManagers/legendary/library.ts
+++ b/src/backend/storeManagers/legendary/library.ts
@@ -544,7 +544,8 @@ function loadFile(app_name: string): boolean {
   const FolderName = customAttributes?.FolderName
   const canRunOffline = customAttributes?.CanRunOffline?.value === 'true'
   const thirdPartyManagedApp =
-    customAttributes?.ThirdPartyManagedApp?.value || undefined
+    customAttributes?.ThirdPartyManagedApp?.value ||
+    customAttributes?.ThirdPartyManagedProvider?.value
 
   if (dlcItemList) {
     dlcItemList.forEach((v: { releaseInfo: { appId: string }[] }) => {
--- a/src/common/types/legendary.ts
+++ b/src/common/types/legendary.ts
@@ -105,7 +105,15 @@ type CustomAttributeType =
   | 'extraLaunchOption_001_Args'
   | 'extraLaunchOption_001_Name'
   | 'ThirdPartyManagedApp'
+  | 'ThirdPartyManagedProvider'
   | 'AdditionalCommandLine'
+  | 'GameID'
+  | 'MainWindowProcessName'
+  | 'RegistryKey'
+  | 'RegistryPath'
+  | 'neverUpdate'
+  | 'partnerLinkId'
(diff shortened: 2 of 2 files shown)


arielj:
But we can get install info for ubisoft games, it's EA games that we can't. Unless maybe now some games from ubisoft are not installable? because if that's the case I think we'll have to implement something new for them to install ubisoft connect and relaunch like we do for EA

Captura de pantalla 2026-01-30 a la(s) 7 59 55 p  m

arielj:
with this PR that makes this game not installable

Captura de pantalla 2026-01-30 a la(s) 8 09 36 p  m

arielj:
what I notice comparing that metadata with the one that can be installed is that the one from discord doesn't have asset_infos nor base_urls, maybe that's relevant?

Captura de pantalla 2026-01-30 a la(s) 8 14 43 p  m

CommandMC:
Hm, I see, seems this needs further changes then

`legendary info ...` outputs no manifest info at all, so the game is not installable

arielj:
> Hm, I see, seems this needs further changes then
> 
> `legendary info ...` outputs no manifest info at all, so the game is not installable

maybe we have to combine that attribute with the metadata not having `asset_infos`

- Some games are installable with no `asset_infos` like FallGuys (EDIT: I was wrong, FallGuys has `asset_infos`, not sure what I saw before)
- EA Games are all not installable (Dragon Age: Inquisition has `asset_infos` but it reports a 4kb install size)
- Ubisoft games seems to be installable if they have `asset_infos`, and not installable if they don't have any `asset_infos`

CommandMC:
For Ubisoft games, it's usually better to still install them through the Ubisoft Connect app. The game files you get from Epic are often outdated and need to be repaired by UC first

We may be able to use the `partnerLinkType` custom attribute to make this a little easier. For Ubisoft games, this is set to `ubisoft` (and has been for a while, games from 2020 also have it)

arielj:
> For Ubisoft games, it's usually better to still install them through the Ubisoft Connect app. The game files you get from Epic are often outdated and need to be repaired by UC first
> 
> We may be able to use the `partnerLinkType` custom attribute to make this a little easier. For Ubisoft games, this is set to `ubisoft` (and has been for a while, games from 2020 also have it)

I do think the user experience of having to install the other launcher first and then get that launcher install the game is not as good though (the game also gets installed inside the prefix, which is a problem if you ever need to clear the prefix, it's easier to import, install DLCs, etc). I feel it way better to install and update in the first launch when possible at least for me.

Personally I would vote to keep the installation that we have for the ones we can and implement something like the EA integration but for Ubisoft for the ones we can't install.

CommandMC:
Closing as #5510 includes these changes & the TODO bits

PR #5268, opened 2026-01-31
Fix inconsistent naming of proper nouns and Wine prefix

### LINKED ISSUE #4991: Change a few texts
### Problem description

_No response_

### Feature description

"linux" to "Linux" in the "Auto Install/Update DXVK-NVAPI on Prefix" option's description
"flatpak" to "Flatpak" in the notice where Heroic tells you it can't find Gamescope on the Flatpak version of Heroic
"Mangohud" to "MangoHud" in the place where you can enable it
"WinePrefix folder" to "Wine prefix folder" in settings
"SteamDeck Compatibility" to "Steam Deck Compatibility" in the extra info for a library item

### Alternatives

_No response_

### Additional information

_No response_

brenobaptista:
Fixes: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4991

Linux, Flatpak, MangoHud, and Steam Deck are proper nouns, while “Wine prefix” can be translated (for example, as "prefixo").

This user has been controversial on other projects, but I also noticed these issues in Heroic, so I decided to fix them.


src/frontend/screens/Settings/components/MangoHud.tsx
```diff
@@ -5,11 +5,11 @@ import useSetting from 'frontend/hooks/useSetting'
 import ContextProvider from 'frontend/state/ContextProvider'
 import InfoIcon from 'frontend/components/UI/InfoIcon'
 
-const Mangohud = () => {
+const MangoHud = () => {
   const { t } = useTranslation()
   const { platform } = useContext(ContextProvider)
   const isLinux = platform === 'linux'
-  const [showMangohud, setShowMangohud] = useSetting('showMangohud', false)
+  const [showMangoHud, setShowMangoHud] = useSetting('showMangoHud', false)
```
arielj: this should not be changed, it's going to reset the settings stored with the  `showMangohud` hey

arielj (changes requested):
we have to be careful not to change setting keys or it will reset the user's settings

src/frontend/screens/Settings/components/MangoHud.tsx
brenobaptista: Sorry about that, I should be more careful. I'll update it and review again to look for cases like this.

brenobaptista:
@arielj Hi! When you have a chance, could you please take another look? Thanks for your time and feedback!

PR #4608, opened 2025-06-04
[Feature] Allow running games without Wine

arielj:
There are cases where old games won't run through wine on linux/mac but community patches exist to make them work.

In many cases, the patch not only makes the game work but also provide executables to make the game run the windows version natively (without wine).

Some cases are:
- `Star Wars Jedi Knight - Jedi Academy` that doesn't work with Wine but can be executed when using the [OpenJK patches](https://jkhub.org/tutorials/errors-problems/installing-openjk-or-other-clients-on-windows-mac-or-linux-r83), and it provides patches for windows, mac, and linux, so the game can be executed with a linux binary for example to avoid the need of a prefix (saves are stored in the game's install directory so cloud saves should still work).
- Fallout 1 and 2 that don't work on MacOS but the there are the [fallout1-ce](https://github.com/alexbatalov/fallout1-ce) and [fallout2-ce](https://github.com/alexbatalov/fallout2-ce) projects that provide binaries to fix the game.

This PR adds the option to disable wine for these games (with the new `Do not use wine` option in the Wine tab) to be combined with the native executable set as alternative exe in the Advanced tab.

How to test this:
- Download `Star Wars Jedi Knight - Jedi Academy` (I think it was free on GOG at some point?)
- Download the linux x86_64 patch from https://github.com/jacoders/openjk/releases
- Extract the content of the file inside the GameData folder inside the game's install dir
- Set the `openjk` executable that was just copied as alternative exe in the Advanced tab
- Toggle `Do not use wine` in the Wine tab
- Run the game

Check the logs and it should be executed with the `--no-wine` flag and you'll see it's not using wine nor umu nor any of those.

This PR closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/2472

This is what it looks like
image

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #5271, opened 2026-02-01
[Zoom] Fix installation of mac-native games from Zoom Platform

arielj:
This PR fixes installing games from Zoom on a Mac.

Note that games are NOT playable with this PR, there's an issue when we try to launch `.app` files directly that is addressed in this other PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4608/changes#diff-69e9f792ff44c2d37e460ca5076ab07248dd5ac260a056ac8c2cff4870afee58R94

Once that other PR is merged we can fix launching native mac games from zoom. This PR only fixes the installation.

Since this is still all experimental and we clarify only linux I think it's fine to have this with a partial fix to install it correctly and later another fix to actually run the file in another PR.

---

### DIFF
--- a/src/backend/storeManagers/zoom/games.ts
+++ b/src/backend/storeManagers/zoom/games.ts
@@ -23,7 +23,7 @@ import {
   LaunchOption,
   InstallPlatform
 } from 'common/types'
-import { existsSync, rmSync } from 'graceful-fs'
+import { existsSync, readdirSync, rmSync } from 'graceful-fs'
 import { installedGamesStore, libraryStore } from './electronStores'
 import {
   logDebug,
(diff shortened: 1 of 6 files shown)


src/backend/storeManagers/zoom/games.ts
```diff
@@ -345,7 +346,26 @@ export async function install(
...
+        downloadPath,
+        '-d',
+        installPath
+      ])
+      executable =
+        readdirSync(installPath).find((path) => path.endsWith('.app')) ||
+        downloadPath.replace('.zip', '.app')
+      executable = executable.replace(installPath, '')
+    } else {
+      await fs.promises.chmod(executable, '755')
```
arielj, line 366: I'm not sure about this else, I kept the same that's there from linux but I don't have any mac game that is not a .zip, maybe it's not even possible 🤷

src/backend/storeManagers/zoom/games.ts
```diff
@@ -541,7 +558,10 @@ export function isNative(appName: string): boolean {
     return true
   }
 
-  if (isMac && gameInfo.install.platform === 'osx') {
+  if (
+    isMac &&
+    (gameInfo.install.platform === 'osx' || gameInfo.install.platform === 'mac')
```
arielj, line 563: the original code here was using `'osx'` but the few games I got from there that have mac version have a `'mac'` platform name, not 'osx'

I'm leaving 'osx' there just in case, no idea if some games do use that instead

arielj:
I'll update this once https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5392 is merged, since it needs that to run .app games

PR #5282, opened 2026-02-04
[tech] Minify the SVGs with SVGO. Saves ~110kB

a5ehren:
Ran the frontend SVGs though SVGO (npm pkg) to shrink the GE and CW logos. This results in no perceptible difference at full-size, and definitely not at the size actually rendered.

---


arielj:
I'm not sure if this is doing that much, when the app gets compiled all these files are turned into JS code and the SVG is rendered inline:

this is for the down_icon
image

here's the ge-proton logo
image

like... optimizing by removing new lines or removing some unused properties doesn't really add much value since the compiler will change all that anyway (and they will be rendered with newlines too)

I would get this if it was a normal web where the size of assets has some performance impact, but I'm not sure it's doing anything in heroic.

a5ehren:
All I've got otherwise the output from vite. 

before:
`build/assets/App-pZ65zaP0.js                               872.94 kB`

after:
`build/assets/App-D2m94rdT.js                               762.16 kB`

I'm not saying its a big or even noticeable difference in actual use, but it does show up in the files.

PR #5292, opened 2026-02-06
[UI/UX] Move FPS limit option to new "Legacy" Game Settings Tab on Linux

flavioislima:
**Adds a new Tab, Linux only for now, to show deprecated settings that might still be useful in some scenarios:**

image


BananaWorks07:
If we want to put this back should add in a note that it doesn't work on certain versions of dxvk

src/frontend/screens/Settings/sections/SyncSaves/gog.tsx
```diff
@@ -123,7 +123,7 @@ export default function GOGSyncSaves({
           }}
         />
       )}
-      <div className="infoBox saves-warning">
+      <div className="defaults-hint">
```
flavioislima, line 126: Small unrelated change to use same style for all warnings on the settings screen.

src/frontend/screens/Settings/components/SettingsModal/index.tsx
```diff
@@ -10,6 +10,7 @@ import SettingsContext from '../../SettingsContext'
 import useSettingsContext from 'frontend/hooks/useSettingsContext'
 import LogSettings from '../../sections/LogSettings'
 import './index.scss'
+import '../../sections/GamesSettings/index.scss'
```
arielj: is this import here needed? my understanding is that this css file should already be loaded from the other component and it should be global already

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -244,6 +247,19 @@ export default function GamesSettings() {
         <Gamescope />
       </TabPanel>
 
+      {isLinux && (
```
arielj, line 250: I think this tab would be empty for linux native games since the DXVK fps limit check won't be rendered

BananaWorks07: Maybe can move some other settings there too, like prime-run one?

src/frontend/screens/Settings/components/SettingsModal/index.tsx
flavioislima: YEah, me too but the class wasn't working unless imported it directly, weird 🤔

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
flavioislima: @arielj good point, will add a check for linux native games.

@Etaash-mathamsetty sounds good

src/frontend/screens/Settings/components/SettingsModal/index.tsx
flavioislima: Now it worked. Solved.

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
flavioislima: Added the isNative check but I wont add the prime-run for now, lets add it on another PR after discussing if it makes sense or not.

PR #5302, opened 2026-02-10
[FIX] Launch Epic Games while offline for games that supports it

flavioislima:
While trying to play some games offline I noticed that Epic games were always failing to launch and the Play button status won't change. Playing the EXE through the Wine setting made the game work, but not through the play button.

This was because the `isEpicServiceOffline` was always returning an axios error and failing to launch legendary.
Adding the `isOnline` check to the make the games work fine.

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -541,7 +541,8 @@ export default React.memo(function GamePage(): JSX.Element | null {
       launchArguments,
       runner: gameInfo.runner,
       hasUpdate,
-      showDialogModal
+      showDialogModal,
+      notPlayableOffline
     })
   }
 
--- a/public/locales/en/gamepage.json
+++ b/public/locales/en/gamepage.json
@@ -14,6 +14,12 @@
             "title": "Move Game Installation"
         },
         "no": "NO",
+        "offline_warning": {
+            "message": "This game might not work properly offline. Do you want to play anyway?",
+            "playAnyway": "Play Anyway",
+            "title": "Offline Warning"
+        },
+        "ok": "OK",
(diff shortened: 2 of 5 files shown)


src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -149,6 +149,7 @@ export default React.memo(function GamePage(): JSX.Element | null {
     !!gameInfo.thirdPartyManagedApp &&
     !gameInfo.isEAManaged
   const isOffline = connectivity.status !== 'online'
+  const notPlayableOffline = isOffline && !gameInfo.canRunOffline
```
arielj, line 152: we shouldn't disable the button in these cases because there are games that report `canRunOffline = false` but they actually work offline

one example is 20 Minutes Till Dawn, it works fully offline but it reports it doesn't, check https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/3385

arielj: we could show a warning though, but I wouldn't fully disable the button

flavioislima: If we dont disable, when you click it nothing happens.
I will think on a warning dialog with a play anyway option.

flavioislima: image

flavioislima:
> looks good to me
> 
> maybe in the future we can add a `don't ask me again` checkbox per game so if a user knows a game works fine offline (like that case I mentioned with 20 minutes till dawn) they don't get this warning every time
> 
> but not a blocker for this

Makes sense, I will try to do it before merging since we already have this in other places, should be easy to have it.

PR #4865, opened 2025-08-10
[Feat]: Open game config file in folder (#4788)

KaranDuggal:
I have tested this changes on linux(ubuntuOS). it's working fine

Note: i have not tested this changes on windows and Mac

feature request: #4788 

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [x] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

PR #5304, opened 2026-02-11
[UI/UX] Improve Environmental Table settings and a few others

flavioislima:
This improves the UI of the Environmental Table setting on the game settings that is being broken for a while now and also add some new options for Copy, builk edit and clear all.

Another small feat is to open the game settings file when clicking the appname on the game settings.

https://github.com/user-attachments/assets/d533676e-6685-442c-a465-9f2f846a9309

---

### DIFF
--- a/src/frontend/screens/Settings/components/BulkEditModal.tsx
+++ b/src/frontend/screens/Settings/components/BulkEditModal.tsx
@@ -1,0 +1,110 @@
+import { useState } from 'react'
+import { EnviromentVariable } from 'common/types'
+
+interface BulkEditModalProps {
+  initialValue: string
+  onSave: (envs: EnviromentVariable[]) => void
+  onCancel: () => void
+  t: (key: string, fallback?: string) => string
+}
+
+const BulkEditModal = ({
+  initialValue,
+  onSave,
+  onCancel,
+  t
+}: BulkEditModalProps) => {
+  const [currentText, setCurrentText] = useState(initialValue)
+  const [error, setError] = useState('')
+
+  const handleSave = () => {
+    const lines = currentText.split('\n')
+    const envMap = new Map<string, string>()
+    let hasError = false
+
+    lines.forEach((line) => {
+      const trimmedLine = line.trim()
+      if (!trimmedLine) return
+
+      if (trimmedLine.includes('=')) {
+        const [key, ...rest] = trimmedLine.split('=')
+        const trimmedKey = key.trim()
+        const trimmedVal = rest.join('=').trim()
+
+        if (!trimmedKey || !trimmedVal) {
+          hasError = true
+        } else {
+          envMap.set(trimmedKey, trimmedVal)
+        }
+      } else {
+        hasError = true
+      }
+    })
+
+    if (hasError) {
+      setError(
+        t(
+          'options.env_variables.bulk_edit_error',
+          'Some variables are incorrect. Please check that all lines follow the KEY=VALUE format and neither is empty.'
+        )
+      )
+      return
+    }
+
+    const newEnvs: EnviromentVariable[] = Array.from(envMap).map(
+      ([key, value]) => ({ key, value })
+    )
+    onSave(newEnvs)
+  }
+
+  return (
+    <div className="bulk-edit-wrapper">
+      <textarea
+        className="bulk-edit-textarea"
+        defaultValue={initialValue}
+        onChange={(e) => {
+          setCurrentText(e.target.value)
+          setError('')
+        }}
+        placeholder="KEY=VALUE"
+      />
+      {error && (
+        <div className="env-var-error" style={{ marginTop: '10px' }}>
+          {error}
+        </div>
+      )}
+      <div className="bulk-edit-actions">
+        <button
+          className="SvgButton"
+          onClick={onCancel}
+          style={{
+            padding: '8px 16px',
+            borderRadius: '4px',
+            cursor: 'pointer',
+            background: 'var(--background-darker)',
+            color: 'var(--text-color)',
+            border: 'none'
+          }}
+        >
+          {t('common.cancel', 'Cancel')}
+        </button>
+        <button
+          className="SvgButton"
+          onClick={handleSave}
+          style={{
+            padding: '8px 16px',
+            borderRadius: '4px',
+            cursor: 'pointer',
+            background: 'var(--accent)',
+            color: 'white',
+            border: 'none'
+          }}
+        >
+          {t('common.save', 'Save')}
+        </button>
+      </div>
+    </div>
+  )
+}
+
+export default BulkEditModal
--- a/src/frontend/screens/Settings/components/BulkEditModal.tsx
+++ b/src/frontend/screens/Settings/components/BulkEditModal.tsx
@@ -74,32 +74,10 @@ const BulkEditModal = ({
         </div>
       )}
       <div className="bulk-edit-actions">
-        <button
-          className="SvgButton"
-          onClick={onCancel}
-          style={{
-            padding: '8px 16px',
-            borderRadius: '4px',
-            cursor: 'pointer',
-            background: 'var(--background-darker)',
-            color: 'var(--text-color)',
-            border: 'none'
-          }}
-        >
+        <button onClick={onCancel} className="button is-secondary">
           {t('common.cancel', 'Cancel')}
         </button>
-        <button
-          className="SvgButton"
-          onClick={handleSave}
-          style={{
-            padding: '8px 16px',
-            borderRadius: '4px',
-            cursor: 'pointer',
-            background: 'var(--accent)',
-            color: 'white',
-            border: 'none'
-          }}
-        >
+        <button onClick={handleSave} className="button is-primary">
           {t('common.save', 'Save')}
         </button>
       </div>
--- a/src/frontend/screens/Settings/components/BulkEditModal.tsx
+++ b/src/frontend/screens/Settings/components/BulkEditModal.tsx
@@ -31,7 +31,7 @@ const BulkEditModal = ({
         const trimmedKey = key.trim()
         const trimmedVal = rest.join('=').trim()
 
-        if (!trimmedKey || !trimmedVal) {
+        if (!trimmedKey) {
           hasError = true
         } else {
           envMap.set(trimmedKey, trimmedVal)
--- a/src/frontend/screens/Settings/components/BulkEditModal.tsx
+++ b/src/frontend/screens/Settings/components/BulkEditModal.tsx
@@ -1,54 +1,56 @@
-import { useState } from 'react'
+import { useMemo, useState } from 'react'
+import { useTranslation } from 'react-i18next'
 import { EnviromentVariable } from 'common/types'
+import { validateEnvKey } from './EnvVariableRow'
 
 interface BulkEditModalProps {
-  initialValue: string
+  initialEnvs: EnviromentVariable[]
   onSave: (envs: EnviromentVariable[]) => void
   onCancel: () => void
-  t: (key: string, fallback?: string) => string
 }
 
 const BulkEditModal = ({
-  initialValue,
+  initialEnvs,
   onSave,
-  onCancel,
-  t
+  onCancel
 }: BulkEditModalProps) => {
-  const [currentText, setCurrentText] = useState(initialValue)
+  const { t } = useTranslation()
+  const initialText = useMemo(
+    () => initialEnvs.map((env) => `${env.key}=${env.value}`).join('\n'),
+    [initialEnvs]
+  )
+  const [currentText, setCurrentText] = useState(initialText)
   const [error, setError] = useState('')
 
   const handleSave = () => {
     const lines = currentText.split('\n')
     const envMap = new Map<string, string>()
-    let hasError = false
 
-    lines.forEach((line) => {
+    for (const line of lines) {
       const trimmedLine = line.trim()
-      if (!trimmedLine) return
+      if (!trimmedLine) continue
 
-      if (trimmedLine.includes('=')) {
-        const [key, ...rest] = trimmedLine.split('=')
-        const trimmedKey = key.trim()
-        const trimmedVal = rest.join('=').trim()
-
-        if (!trimmedKey) {
-          hasError = true
-        } else {
-          envMap.set(trimmedKey, trimmedVal)
-        }
-      } else {
-        hasError = true
-      }
-    })
-
-    if (hasError) {
-      setError(
-        t(
-          'options.env_variables.bulk_edit_error',
-          'Some variables are incorrect. Please check that all lines follow the KEY=VALUE format and neither is empty.'
+      if (!trimmedLine.includes('=')) {
+        setError(
+          t(
+            'options.env_variables.bulk_edit_error',
+            'Some variables are incorrect. Please check that all lines follow the KEY=VALUE format.'
+          )
         )
-      )
-      return
+        return
+      }
+
+      const [key, ...rest] = trimmedLine.split('=')
+      const trimmedKey = key.trim()
+      const trimmedVal = rest.join('=').trim()
+
+      const keyError = validateEnvKey(trimmedKey, t)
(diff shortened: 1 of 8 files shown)


arielj (dismissed):
In the current version we have a feature that if a user pastes something like `MY_VAR=something` in the variable name field, heroic automatically splits that into name/value and puts the correct thing in each field

that is not working with this change

also in the current version we show the `Variable name cannot be blank` before the user clicks the + button, now they have to click it to see the validation

arielj:
one more thing, before we couldn't add the same env variable more than once (adding it again would replace the previous one), now we can add it multiple times with different value

arielj:
the bulk edit dialog looks too narrow

image

and using it allows creating wrong rows with empty variable name (if you have a line with just `=something` that I can imagine can happen by mistake)

same happens when editing a variable, in the edit you can clear the variable name and the validation does not check that it's empty and saves it anyway

flavioislima:
@arielj I think I addressed all issues:
image

image

image

And the pasting is working back again.

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  gap: var(--space-md-fixed);
+  margin-bottom: var(--space-xs-fixed);
+}
+
+.env-vars-title {
+  font-size: var(--text-lg);
+  color: var(--accent);
```
biliesilva: we should be using accent color here only for links, these are links too?

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+.env-var-key {
+  font-weight: bold;
+  color: var(--accent);
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.env-var-connector {
+  color: var(--text-muted);
```
biliesilva: Do we have this color variable?

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+  grid-template-columns: 1fr auto 1.5fr;
+  flex-grow: 1;
+  gap: var(--space-sm-fixed);
+  align-items: center;
+  overflow: hidden;
+}
+
+.env-var-key {
+  font-weight: bold;
+  color: var(--accent);
```
biliesilva: same here

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+}
+
+.env-var-row:hover .env-var-row-actions {
+  opacity: 1;
+}
+
+.env-var-add-section {
+  margin-top: var(--space-sm-fixed);
+  padding: var(--space-md-fixed);
+  border: 1px dashed var(--accent);
```
biliesilva: the border should be a neutral color

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+  align-items: center;
+  gap: var(--space-xs-fixed);
+}
+
+.bulk-edit-textarea {
+  width: 100%;
+  min-width: 500px;
+  min-height: 300px;
+  background: var(--background);
+  border: 1px solid var(--accent);
```
biliesilva: here

biliesilva (changes requested):

arielj (changes requested):
Testing this again, reporting some issues I see

- The save button in the bulk edit dialog with the default theme is not readable

image

 - there seems to be some infinite loop going on, when I open the settings for a game I get the terminal flooded with using cached install info message

image

- maybe use a "check" icon instead of a floppy disk icon to confirm the edit of an entry?

image

I think it makes more sense to match the X to cancel the edit.

- adding an individual env variable does not allow names with white spaces (which is fine), but the bulk edit option allows spaces

image
image

then if you try to edit that env variable that has spaces, the saving fails silently without any message about it having white spaces

- a ser can copy a variable twice (or more) by mistake and have it many times, it's not clear what's the final result when a variable is there twice (we should probably check if the name is already in use and not allow the copy)

image

this also works differently than adding a variable with a duplicated name, in that case heroic replaces the first one it find instead of adding a duplicate

then if a user bulk edits, it removes the duplicates keeping the last one defined

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
```diff
@@ -1,34 +1,168 @@
...
+    const bulkText = environmentOptions
+      .map((env) => `${env.key}=${env.value}`)
+      .join('\n')
+
+    showDialogModal({
+      title: t('options.env_variables.bulk_edit', 'Bulk Edit'),
+      message: (
+        <BulkEditModal
+          initialValue={bulkText}
+          t={t}
```
arielj: is there a reason to pass the `t` function as a prop here instead of doing `const { t } = useTranslation()` like we do in other components?

src/frontend/screens/Settings/components/BulkEditModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+      {error && (
+        <div className="env-var-error" style={{ marginTop: '10px' }}>
+          {error}
+        </div>
+      )}
+      <div className="bulk-edit-actions">
+        <button
+          className="SvgButton"
+          onClick={onCancel}
+          style={{
```
arielj: maybe we can use the class button is-secondary and button is-primary for consistency with other buttons instead of adding a custom inline style for these (this should fix the issue I reported with the Save not readable in some themes)

there's also no SvgButton class used anywhere else in the app, this is also not rendering a svg, the SvgButton class should be removed from these buttons

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
```diff
@@ -46,50 +180,176 @@ const EnvVariablesTable = () => {
...
-          )
-        } else if (key.match(/ /)) {
-          keyError = t(
-            'options.env_variables.error.space_in_key',
-            `Variable names can't contain spaces`
-          )
-        }
+    <div className="env-vars-container">
+      <div className="env-vars-header">
+        <h3 className="env-vars-title">
```
arielj: the other sections use a `<label>` tag for the section's title, this is the only one using h3 which looks off

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+.env-var-edit-fields {
+  display: flex;
+  flex-grow: 1;
+  gap: var(--space-xs-fixed);
+  align-items: center;
+}
+
+.env-var-edit-fields .Field {
+  margin: 0;
+  flex-grow: 1;
```
arielj: these 2 properties don't seem to be needed, if I turn these styles off with the devtools nothing changes

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+}
+
+.env-var-add-section {
+  margin-top: var(--space-sm-fixed);
+  padding: var(--space-md-fixed);
+  border: 1px dashed var(--accent);
+  border-radius: var(--border-radius-sm);
+  display: flex;
+  flex-direction: column;
+  gap: var(--space-sm-fixed);
```
arielj: these 3 properties doesn't seem to be needed, it looks just fine without flex column and the gap while it adds some spacing, it also doesn't look bad without that gap

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+
+.env-var-add-inputs {
+  display: flex;
+  gap: var(--space-sm-fixed);
+  align-items: center;
+}
+
+.env-var-add-inputs .Field {
+  flex-grow: 1;
+  margin: 0;
```
arielj: same as the edit .Field style, these 2 properties don't seem to do anything for these fields

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+  padding: var(--space-sm-fixed);
+  font-family: monospace;
+  border-radius: var(--border-radius-sm);
+  resize: vertical;
+}
+
+.env-var-error {
+  color: var(--danger);
+  font-size: var(--text-xs);
+  margin-bottom: var(--space-xs-fixed);
```
arielj: I would remove these 2 variables, it looks fine without the margin-bottom (it makes it more consistent with the space around the other edges), and the text is too small with that font-size

src/frontend/screens/Settings/components/EnvVariablesTable.css
```diff
@@ -0,0 +1,176 @@
...
+  gap: var(--space-sm-fixed);
+  margin-top: var(--space-md-fixed);
+  justify-content: flex-end;
+}
+
+.env-vars-container .svg-button:disabled {
+  cursor: not-allowed;
+  opacity: 0.5;
+  filter: grayscale(1);
+}
```
arielj: I'm not opposed to a change like this, but I'd rather make this change somewhere else that is more reusable instead of specific to this single button

we have a similar button right above this for the wrapper that does not change colors when disabled, maybe apply this for the `.svg-button:disabled` selector not only the ones inside `.env-vars-container`?

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
```diff
@@ -1,34 +1,168 @@
...
 
-  const getEnvironmentVariables = () => {
-    const columns: ColumnProps[] = []
-    environmentOptions.forEach((env) =>
-      columns.push({ key: env.key, value: env.value })
-    )
-    return columns
+  const [editingIndex, setEditingIndex] = useState<number | null>(null)
+  const [editKey, setEditKey] = useState('')
+  const [editValue, setEditValue] = useState('')
```
arielj: maybe each row should be its own component?

now there's some weird glitch:
- if you start editing an env
- without saving/canceling you delete all env variables (with the bulk delete)
- then add new variables
- the one matching this index will be added already in edit mode

if each row is its own component then it can have its own "isEditing" state instead of handled by this parent component

I think it will also make this component simpler to maintain/understand, now the return value is doing a lot when I imagine it could be rendering a loop of EnvVariableRow components without having to deal with editing state and all that

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
```diff
@@ -1,34 +1,168 @@
...
+          onClick: () => setEnvironmentOptions([])
+        }
+      ]
+    })
+  }
+
+  const openBulkEdit = () => {
+    const bulkText = environmentOptions
+      .map((env) => `${env.key}=${env.value}`)
+      .join('\n')
```
arielj: maybe pass the env variables to the BulkEditModal as a prop and do this string joining in that component? I think it makes more sense for this to happen there since that component is the one splitting them back to an array of hashes

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
```diff
@@ -46,50 +180,176 @@ const EnvVariablesTable = () => {
...
+                        setEditKey(key)
+                        setEditValue(rest.join('='))
+                      } else {
+                        setEditKey(val)
+                      }
+                      setFormError(validateEnv(val, editValue))
+                    }}
+                    placeholder="NAME"
+                    extraClass={
+                      formError && editingIndex === index ? 'error' : ''
```
arielj: this is adding an `error` class when an input has an error, but it has no real effect, nothing changes visually and it's not used for anything in the javascript

src/frontend/screens/Settings/sections/FooterInfo/index.tsx
```diff
@@ -12,7 +12,16 @@ export default function FooterInfo() {
...
         <span className="appName">
           AppName: &nbsp;
-          <span style={{ userSelect: 'all' }}> {appName}</span>
+          <span
+            style={{ userSelect: 'all' }}
+            title={t(
+              'settings.open-game-settings-file-hint',
+              'Click to open game settings file'
+            )}
+            onClick={() => window.api.showConfigFileInFolder(appName)}
```
arielj: one comment about this, there's another PR that adds a button for this in the 3-dots menu https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4865/changes

arielj (commented):
- when navigating with a controller, if you click the edit of a variable or when you save/cancel the edit, the focus gets lost and the cursor is back to the X that closed the settings dialog.

I'm thinking we should force the focus either on the new input we show when editing or back to focus the "edit" button when saving/canceling the edit.


- when the EDIT of an env variable is visible, the validations of the ADD new variable don't show up

when I mean is:
- add an env variable
- click its edit button
- now type something in the add env variable inputs
- no error about name or value being empty is shown
- save/cancel the edit
- now the error is shows in the add env variable inputs

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
```diff
@@ -46,50 +180,176 @@ const EnvVariablesTable = () => {
...
+          />
+          <span className="env-var-connector">=</span>
+          <TextInputField
+            htmlId="new-env-value"
+            value={newValue}
+            onChange={(val) => {
+              setNewValue(val)
+              setFormError(validateEnv(newKey, val))
+            }}
+            placeholder={t('options.advanced.placeHolderValue', 'VALUE')}
```
arielj: this is changing the value of this translation key here but not in the translation files

I do agree this should be VALUE and not "E.g.: Path/To/ExtraFiles" as it is today, but maybe we can rename the translation key to have a new value for all languages?

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
```diff
@@ -46,50 +180,176 @@ const EnvVariablesTable = () => {
...
+                  />
+                  <span className="env-var-connector">=</span>
+                  <TextInputField
+                    htmlId={`edit-value-${index}`}
+                    value={editValue}
+                    onChange={(val) => {
+                      setEditValue(val)
+                      setFormError(validateEnv(editKey, val))
+                    }}
+                    placeholder="VALUE"
```
arielj: this placeholder is not translated, it should be the same translation used in the add env variable fields

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
flavioislima: fixed

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
flavioislima: done

src/frontend/screens/Settings/sections/FooterInfo/index.tsx
flavioislima: Yeah, I will keep my changes since that PR is stale, but that solution seems better, should not be hard to fix the conflicts there if I merge mine first.

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
flavioislima: fixed

src/frontend/screens/Settings/components/EnvVariablesTable.tsx
flavioislima: Yeah, I needed to change the key name so it will generate a new one on weblate

flavioislima:
Merging this one since all comments were addressed, retested and did some cleanup.

PR #5306, opened 2026-02-11
[UI] Wine manager screen cleanup and add wine settings screen

flavioislima:
The idea would be to also have a setting to add custom Wine/Proton repositories but this would increase the complexity a lot so I might do it in the future.

image

Screenshot_20260211_183949

---


arielj:
hmmm I'm not sure I like the 3 dots thing, in the current version it's really clear when the Latest version has an update available, with this change users have to click the 3-dots to check if there's an update available

I also noticed that if you start shrinking the screen the table breaks
image

this didn't happen before
image

flavioislima:
@arielj did some changes to accommodate and fix the table. One of them is removing the Notes column while keeping the functionality now by clicking on the name version and hiding some of the titles on small width since those are self explanatory:

image

image

arielj:
it looks a bit weird how the release date and size columns are so wide and the version name starts truncating when you shrink the screen, when there's way more room available if the other columns shrink instead

Captura de pantalla 2026-02-22 a la(s) 1 08 24 p  m

it also looks a bit weird how the font size changes when you shrink the screen enough for it to change the layout, the text is more readable and big with the smaller screen

finally, now that there's no clear separation between rows (other than a bit of space), in wide screens it's hard to map with buttons on the right side match which version on the left side of the table:

Captura de pantalla 2026-02-22 a la(s) 1 10 09 p  m

(in the current version we have a line separating versions so it's easier to read what belongs to what)

arielj:
The search input is not working properly for me, check this:
Captura de pantalla 2026-02-22 a la(s) 1 12 49 p  m

it's not filtering GPTK 3.0-2

some combinations of letters and numbers work, but it's not clear what it's matching

arielj:
> The idea would be to also have a setting to add custom Wine/Proton repositories but this would increase the complexity a lot so I might do it in the future.

I don't know if we should do this, we already have issues when we launch heroic too often that it can hit the github rate limiter because we check updates in many repos. If users add more repos here it can create issues and we also have no idea what they will add, maybe adding repos that heroic can't really handle

I'd rather have our own vetted list of repos we add as options in the code.

With this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5161 we could add more repos that we know work fine with heroic and also optimize the checks against github to not hit the rate limiter

flavioislima:
@arielj addressed the rest of the issues now

arielj:
the search function is still not working properly, if I search `32` I would expect `GE-Proton10-32` to be displayed but it shows no results, I have to search `10-32` explicitly, if I search `1` it only shows `GE-Proton10-11`, but if I search `10` it shows all the `10-...` versions

and then if I open the Wine GE tab, none of those combinations work, any number I search return no results, even if I put `8-26` that should match a specific version

arielj:
if I'm in the Wine GE tab and I open the settings and disable it, the tab is removed but I'm left in the list of WineGE versions, I think it should be taken to another tab if I remove the tab I'm on

arielj:
I really like the new download progress, but then when it finished downloading the % resets to 0 while it's unzipping

I think it should stay at 100% until if finished unzipping or we should remove the % while it's unzipping (because now it looks like it's the % of the unzipping step, but it never moves, it's just fixed at 0%)

arielj:
when using a RTL language, the progress should start from the other side, and also the % ends up in the wrong place between the unit and the size, some of the CSS is using left/right alignment instead of logical values like inline start/end to respect the RTL alignments

image

src/frontend/screens/WineManager/components/WineManagerSettingsModal.tsx
```diff
@@ -0,0 +1,98 @@
...
+                  })
+                  .map((repo: WineManagerUISettings) => (
+                    <div key={repo.value} className="repoItem">
+                      <ToggleSwitch
+                        htmlId={`repo-${repo.value}`}
+                        title={repo.type}
+                        value={repo.enabled}
+                        disabled={
+                          (repo.value === 'protonge' && platform === 'linux') ||
+                          (repo.value === 'gpt' && platform !== 'darwin')
```
arielj: I'm not sure if this thing to disable wine repositories is that useful? like... why would someone disable wine-ge? they can just ignore it and don't use it

but also disabling a repository just hides the tab from the wine manager but does not hide the wines of that repository from the wine version selector in the game's settings and it does not skip the checks for new versions when pressing the refresh button (unless I didn't see that)

I think we don't really need this complexity, users can just ignore the tab of a repository if they don't want to use it?

flavioislima: yeah, the idea was to expand the list of wine in the future but I believe this won't really be useful since the defaults already works pretty well.

flavioislima:
@arielj simplified the PR and addresed the issues you pointed out. Let me know if you find any other issues.

arielj:
this issue with the progress between the unit and the size for RTL languages is still present

image

arielj (approved):
I'm approving this though the above issue is present, but I think the rest is all good

EDIT: the linter is complaining

flavioislima:
> this issue with the progress between the unit and the size for RTL languages is still present
> 
> image

Fixed it
image

Issue #4727, opened 2025-07-11
Show if a game supports offline mode or not in the logs


arielj:
### Problem description

This can be useful for support threads.

The information that we have in the metadata is not always correct (some games don't say they work offline but they do), but in many cases that could help to at least identify a possible problem

### Feature description

in the logs, when the `offline mode` is enabled, check if the game reports supporting running offline and show a warning if it does not

PR #5318, opened 2026-02-15
[UX] Log if a game does not support offlineMode but offlineMode is on

arielj:
Closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4727

We see pretty often in support threads that users have offlineMode turned on for online games. We already show a warning in the Settings modal when offlineMode is turned on for a game that reports not supporting it, but users may ignore it and we may not notice it when checking logs.

This PR makes it more explicit: if a game does not  `canRunOffline` and `offlineMode` is turned on, we show a warning in the logs so it's more obvious and clear.

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -790,6 +790,8 @@ async function prepareWineLaunch(
   failureReason?: string
   envVars?: Record<string, string>
 }> {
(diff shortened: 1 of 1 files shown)


Etaash-mathamsetty (approved):
lgtm

Issue #5170, opened 2026-01-12
Instruct the users to share the file from disk if uploading log file fails


arielj:
### Problem description

Currently, when a user wants to share a log, if the request to upload the file to 0x0.st fails, the user gets a message that it failed and sometimes they share that message and we then need to explain how to get the file to upload it manually.

### Feature description

In the dialog that says that the upload failed, also show instructions to upload the log:
- mention that it failed to upload but they can still share it manually
- show a button to open the log in folder
- mention that they can upload the file to discord or to any other sharing service

PR #5319, opened 2026-02-15
[UX] Tell users to manually upload log if upload button failed

arielj:
Closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/5170

During support, it's really common for users to share a screenshot or say the upload of logs didn't work. And it's always the same answer that they have to click SHOW LOG FILE IN FOLDER and upload launch.log manually somewhere.

This adds that comment in that same dialog, saving time for everyone.

image

---

### DIFF
--- a/src/frontend/components/UI/Dialog/index.css
+++ b/src/frontend/components/UI/Dialog/index.css
@@ -108,6 +108,10 @@
   padding: 0 var(--dialog-margin-horizontal) var(--dialog-gap);
 }
 
+.log-upload-result {
+  max-width: 500px !important;
+}
+
 .Dialog__footer {
   display: flex;
   gap: 16px;
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -732,6 +732,7 @@
                 },
                 "error": {
                     "content": "Failed to upload log file. Check Heroic's general log for details",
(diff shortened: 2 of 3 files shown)


src/frontend/components/UI/Dialog/index.css
```diff
@@ -108,6 +108,10 @@
   padding: 0 var(--dialog-margin-horizontal) var(--dialog-gap);
 }
 
+.log-upload-result {
+  max-width: 500px !important;
```
CommandMC, line 112: Correct me if I'm wrong, but a raw px value is not responsive, right? Regardless, I don't think we should add this in the general dialog style file.

We could do this the way MUI intends by passing a `sx` prop from the dialog to the internal paper component. Here's a [patch for that](https://github.com/user-attachments/files/25375881/0001-dialog-maxwidth-mui.patch)

arielj: it's not responsive but I doubt anyone is using a window smaller than 500px for this to be a problem

the issue I wanted to fix is that if it grows you end up with a really wide dialog with a single long like of text that is hard to read, a narrow dialog with just text is easier to the eyes

flavioislima: I think it's fine if its only for this specific dialog.

PR #5321, opened 2026-02-16
Trying to fix disableUMU

Kyuyrii:
I'm trying to fix the use of Proton in Heroic Snap without UMU, since umu-support hasn't been accepted yet.

This modification is to remove some variables that are apparently problematic when using Proton without UMU.

Can you check if the changes were made correctly?

---


Kyuyrii:
This video is a bit messed up, but I think you can understand what these changes are supposed to fix.

[2026-02-16-02-32-19.webm](https://github.com/user-attachments/assets/5144748d-2caa-45d5-9e54-6404b5bdcd9a)

CommandMC:
Running Proton without umu is not a supported use case. If Snap requires umu to be disabled, we're not making a Snap version.

Please hash out the Snap-related issues on their side first. Feel free to re-open if these changes are still relevant then.

PR #5336, opened 2026-02-19
[UX] Fix mix between default prefixes dir and shared prefix dir

### LINKED ISSUE #3274: Default wine prefix settings are confusing

arielj:
This PR makes the final changes so it closes https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3274

This renames the variables and setting keys we use to make it more clear that one is the default prefixes folder (where prefixes are created inside) and the default (now `shared`) wine prefix dir (the one used with the "Use shared Wine prefix").

The new setting value is only applied for new installation or if a user resets Heroic, for existing users this should not change anything (it means that if they already have the base prefixes dir as .../default it will continue like that and prefixes will be created inside it, to not break things for them).

Similar thing happens with the `default` wine prefix config, now it's renamed to `sharedWinePrefix` so it matches better the `Use shared Wine prefix` option. For existing user we still show the previous value they have (they probably have .../default) and for new users we show .../shared instead.

This will solve this issue where users end up with `.../Prefixes/default` as a prefix and also `.../Prefixes/default/some game prefix here` as other prefixes inside that one.

---

### DIFF
--- a/src/backend/config.ts
+++ b/src/backend/config.ts
@@ -23,10 +23,11 @@ import { configStore } from './constants/key_value_stores'
 import { isFlatpak, isLinux, isMac, isWindows } from './constants/environment'
 import {
   configPath,
-  defaultWinePrefix,
+  sharedWinePrefix,
   gamesConfigPath,
   heroicInstallPath,
-  userHome
+  userHome,
+  defaultWinePrefixDir
 } from './constants/paths'
 import { join } from 'path'
 import { spawnSync } from 'child_process'
@@ -293,6 +294,10 @@ class GlobalConfigV0 extends GlobalConfig {
     let settings = JSON.parse(readFileSync(configPath, 'utf-8'))
     const defaultSettings = settings.defaultSettings as AppSettings
 
+    // keep old setting value for backwards compatibility, always use defaultWinePrefixDir in new code
+    if (!defaultSettings.defaultWinePrefixDir)
+      defaultSettings.defaultWinePrefixDir = defaultSettings.defaultWinePrefix
+
     // fix relative paths
     const winePrefix = !isWindows
       ? defaultSettings?.winePrefix?.replace('~', userHome)
@@ -300,7 +305,7 @@ class GlobalConfigV0 extends GlobalConfig {
 
     settings = {
       ...this.getFactoryDefaults(),
-      ...settings.defaultSettings,
+      ...defaultSettings,
       winePrefix
     } as AppSettings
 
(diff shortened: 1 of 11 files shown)


flavioislima (approved):
Tested here and I can confirm that it changes to `shared` only on new installs. On current configs it keeps using the `default` one.

Image

PR #5342, opened 2026-02-20
[FIX] Check alternative folders when installing DXVK

arielj:
Latest changes in the DXVK macos repo made the installation of DXVK fail because we are looking only for the `x32` and `x64` directories and the new asset has the folders `i386-windows` and `x86_64-windows` instead.

This PR fixes that by checking if the folders exists and falling back to the new folder name if the previous does not exist.

---

### DIFF
--- a/src/backend/tools/index.ts
+++ b/src/backend/tools/index.ts
@@ -238,9 +238,17 @@ export const DXVK = {
     const toolPathx32 = `${toolsPath}/${tool}/${globalVersion}/${
       tool === 'vkd3d' ? 'x86' : 'x32'
     }`
-    const dlls32 = readdirSync(toolPathx32)
+    const toolPathx32Alt = `${toolsPath}/${tool}/${globalVersion}/i386-windows`
+    let dlls32: string[] = []
+    if (existsSync(toolPathx32)) dlls32 = readdirSync(toolPathx32)
+    else if (existsSync(toolPathx32Alt)) dlls32 = readdirSync(toolPathx32Alt)
+
     const toolPathx64 = `${toolsPath}/${tool}/${globalVersion}/x64`
-    const dlls64 = readdirSync(toolPathx64)
+    const toolPathx64Alt = `${toolsPath}/${tool}/${globalVersion}/x86_64-windows`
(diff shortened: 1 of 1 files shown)


Gcenx:
The new `-builtin` archive is to be used in another way _not_ to be used like the other archive.

DXMT/DXVK-macOS-builtin can be used via the env `WINEDLLPATH_PREPEND` instead of copying the files over wines dlls.

DXMT layout
Screenshot 2026-02-20 at 1 52 04 PM

DXVK layout
Screenshot 2026-02-20 at 1 52 13 PM

In my case I'd set `export WINEDLLPATH_PREPEND=/opt/dxmt` for DXMT or `export WINEDLLPATH_PREPEND=/opt/dxvk` for DXVK-macOS, this means only needing a single wine package not multiple like nw with the -DXMT & none DXMT variants.

This env works for `wine-crossover` 23.7.1 and the winehq packages making it simpler to swap between different wine packages even game-porting-toolkit without needing to worry about winecfg `Native` dlls needing to be removed and reset for d3d10core/d3d11/dxgi etc.

arielj:
> The new `-builtin` archive is to be used in another way _not_ to be used like the other archive.
> 
> DXMT/DXVK-macOS-builtin can be used via the env `WINEDLLPATH_PREPEND` instead of copying the files over wines dlls.
> 
> DXMT layout Screenshot 2026-02-20 at 1 52 04 PM
> 
> DXVK layout Screenshot 2026-02-20 at 1 52 13 PM
> 
> In my case I'd set `export WINEDLLPATH_PREPEND=/opt/dxmt` for DXMT or `export WINEDLLPATH_PREPEND=/opt/dxvk` for DXVK-macOS, this means only needing a single wine package not multiple like nw with the -DXMT & none DXMT variants.
> 
> This env works for `wine-crossover` 23.7.1 and the winehq packages making it simpler to swap between different wine packages even game-porting-toolkit without needing to worry about winecfg `Native` dlls needing to be removed and reset for d3d10core/d3d11/dxgi etc.

I'm going to merge the workaround from this PR for now and then keep implement this env for a future release, thanks

Issue #5052, opened 2025-11-23
Amazon account cannot switch


MichelN86:
### Describe the bug

I have 2 accounts, 1 for my self and 1 for my wife. She has different games in her account then i have. I wanted to switch account and it wont let me. It keeps logging in in my own account even when i used her password/username. No matter what i do, it seems it's in some kind of loop.

I deleted the folder in the /.config/heroic and that fixed it but i have to set everything up again.

### Add logs

```text
x
```

### Steps to reproduce

Just login to another account from amazon.

### Expected behavior

That it will switch to the account that i entered.

### Heroic Version

Latest Stable

### System Information

- OS [e. g. "Ubuntu"]:

MichelN86:
Also please add vkd3d latest version:
https://github.com/HansKristian-Work/vkd3d-proton/releases/tag/v3.0a

PR #5354, opened 2026-02-22
[FIX] Use correct partition and clear data on epic logout

arielj:
There's a bug that users can't logout from the epic account to log into a new one, it keeps reloging you in with the previous acount.

The issue was that `store` during the login is `undefined` and ends up using a partition called `undefined`, but then the logout of epic clears a partition called `epicstore`.

This PR fixes that by:
- using `epicstore` as the partition name if we are logging in
- also clearing the Data of the partition

I think there might be a similar issue with Amazon from this report https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/5052 but I don't have an Amazon account to check.

---

### DIFF
--- a/src/frontend/screens/WebView/index.tsx
+++ b/src/frontend/screens/WebView/index.tsx
@@ -379,7 +379,7 @@ export default function WebView() {
       <webview
         ref={webviewRef}
         className="WebView__webview"
-        partition={`persist:${store}`}
+        partition={`persist:${startUrl === epicLoginUrl ? 'epicstore' : store}`}
         src={startUrl}
         allowpopups={trueAsStr}
         preload={webviewPreloadPath}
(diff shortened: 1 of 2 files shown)


src/frontend/screens/WebView/index.tsx
```diff
@@ -379,7 +379,7 @@ export default function WebView() {
       <webview
         ref={webviewRef}
         className="WebView__webview"
-        partition={`persist:${store}`}
+        partition={`persist:${startUrl === epicLoginUrl ? 'epicstore' : store}`}
```
flavioislima, line 382: If `store` is undefined on login maybe we should set it somewhere before this? And for the other stores as well?

arielj: hmmm to change that I would have to change the router and I don't understand it that much, I'm worried I'll break something else (the change would affect all stores, not just the epic one), it is intentionally not setting the `store` param for login routes

flavioislima: ok sounds good

PR #5365, opened 2026-02-25
[Fix] False positive "an error has occurred" messages on bleeding-edge Linux distros

CommandMC:
Our generic "error handler" is a little sensitive, which caused it to trigger on this output:
```
legendary version "0.20.41", codename "Freeman Pontifex (Heroic)"
/opt/Heroic/resources/app.asar.unpacked/build/bin/x64/linux/./legendary/requests/__init__.py:113: RequestsDependencyWarning: urllib3 (2.6.3) or chardet (6.0.0.post1)/charset_normalizer (3.4.4) doesn't match a supported version!
```

This warning is thrown since users on Linux can update some dependencies independently of Legendary/Heroic, causing slightly-too-new versions to be used. I don't think this alone is an issue however, so we should just ignore it

85d3fc186b0d7ce45d6e7f59b31fb3ce2e028804 just changes whitespace, 938642ae16b09c7bf539096c050cec4ad9cdff14 is the main change

Can't test this, contacting some affected users to test

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -290,14 +290,18 @@ async function errorHandler({
   const deletedFolderMsg = 'appears to be deleted'
   const expiredCredentials = 'No saved credentials'
   const legendaryRegex = /legendary.*\.py/
-  // this message appears on macOS when no Crossover was found in the system but its a false alarm
-  const ignoreCrossoverMessage = 'IndexError: list index out of range'
+  const ignoreMessages = [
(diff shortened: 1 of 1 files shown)


lathrus:
I downloaded the code with the patch, compiled it myself and can confirm that the patch works correctly.

My OS:
 Kernel: 6.19.3-2-cachyos arch: x86_64 bits: 64
 Desktop: KDE Plasma v: 6.6.1 Distro: CachyOS

PR #5366, opened 2026-02-25
[FIX] Add a launch default option when none is available besides DLCs

flavioislima:
This fixes an edge case where some games has only their DLCs as launch options and doesn't show a default one, making the game fail to launch.

image

---

### DIFF
--- a/src/frontend/components/UI/DialogHandler/components/MessageBoxModal/index.css
+++ b/src/frontend/components/UI/DialogHandler/components/MessageBoxModal/index.css
@@ -27,7 +27,6 @@
 }
 
 .launchOptionsDialog .Dialog__footer {
-  display: flex;
   flex-direction: column;
   gap: var(--space-lg);
 


flavioislima:
**Other games will keep working just fine:**

image

image

src/frontend/hooks/useLaunchOptions.ts
```diff
@@ -49,6 +49,24 @@ export const useLaunchOptions = ({
...
 
+  // add a default option if not present
+  useEffect(() => {
+    if (
+      launchOptions.length > 0 &&
+      !launchOptions.some(
+        (option) =>
+          (option.type === undefined || option.type === 'basic') &&
+          option.name === 'Default' &&
+          option.parameters === ''
```
imLinguin: In case of GOG, games already specify the default that is the game's name. Maybe its worth patching it in `gog/library.ts` - the isPrimary boolean is the indicator of the default.

a patch for `gog/library.ts` could look like
```diff
-        name: task?.name || infoFile.name,
+        name: task.isPrimary ? 'Default' : task?.name || infoFile.name,
```

when isPrimary is true we could also ommit parameters I think

src/frontend/helpers/library.ts
```diff
@@ -318,6 +318,19 @@ async function checkLaunchOptionsAndLaunch({
...
+      (option) => option.type === undefined || option.type === 'basic'
+    )
+  ) {
+    availableLaunchOptions.unshift({
+      type: 'basic',
+      name: 'Default',
+      parameters: ''
+    })
+  }
+
```
arielj: this should be added in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/8bb414b232253fb6f0be54505808437099276a9e/src/backend/main.ts#L1082 for the same reason as the previous PRs, so it's done in a single place

src/frontend/hooks/useLaunchOptions.ts
flavioislima: This made it to duplicate the default options so other changes might be needed. Since GOG games are working fine in the current state we should address this other change later on since this PR is just to fix this small edge case on Epic games.

image

src/frontend/helpers/library.ts
flavioislima: Good catch, fixed.

arielj: one of these was moved to main.ts, but this one is still here in library.ts

flavioislima: Oofff
done

src/backend/main.ts
```diff
@@ -1094,6 +1094,25 @@ addHandler('getLaunchOptions', async (event, appName, runner) => {
...
+    availableLaunchOptions.length > 0 &&
+    !availableLaunchOptions.some(
+      (option) =>
+        (option.type === undefined || option.type === 'basic') &&
+        option.name === 'Default' &&
+        option.parameters === ''
+    )
+  ) {
+    availableLaunchOptions.unshift({
+      name: i18next.t('launch.default', 'Default', {
```
arielj: if looks like we are doing a similar thing twice here?

once if availableLaunchOptions.length === 1 and once more if availableLaunchOptions.length > 0

the first time we are adding Default and the second time we are adding translated Default

I think what you want to do in this PR is replace the `...length === 1` conditional and update the name of the option you add instead

flavioislima: true, the first part became redundant now. removed it

src/frontend/components/UI/DialogHandler/components/MessageBoxModal/index.css
```diff
@@ -25,3 +25,13 @@
   margin-bottom: 10px;
   color: red;
 }
+
+.launchOptionsDialog .Dialog__footer {
+  display: flex;
```
arielj: Dialog__footer is already display: flex

PR #5367, opened 2026-02-26
[SIDELOAD]: add support for proton_shortcuts

imLinguin:
this functionality can be easily extended to other providers, however I think only zoom will benefit most from this. 
In order for shortcuts to appear in Heroic, an app needs to register something in start menu. In EA App desktop and menu shortcuts are one option.

The `proton_shortcuts` feature was originally meant for Steam to support running arbitrary executables like mods in prefixes of games and being able to display launch options for new additions

I have modified executable path constraints to support Windows paths, as well as added a name field for altExe option

### DIFF
--- a/src/backend/storeManagers/sideload/library.ts
+++ b/src/backend/storeManagers/sideload/library.ts
@@ -1,11 +1,16 @@
-import { ExecResult, GameInfo } from 'common/types'
+import { ExecResult, GameInfo, LaunchOption } from 'common/types'
 import { readdirSync } from 'graceful-fs'
 import { dirname, join } from 'path'
 import { libraryStore } from './electronStores'
-import { logWarning } from 'backend/logger'
+import { logError, logWarning } from 'backend/logger'
 import { addShortcuts } from 'backend/shortcuts/shortcuts/shortcuts'
 import { sendFrontendMessage } from 'backend/ipc'
 import { isMac } from 'backend/constants/environment'
+import { getSettings } from './games'
+import { readdir, readFile, stat } from 'node:fs/promises'
+import { parse as iniParse } from 'ini'
+import { DesktopEntry } from 'common/types/shortcuts'
+import { Path } from 'backend/schemas'
 
 export function addNewApp({
   app_name,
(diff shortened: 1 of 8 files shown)


imLinguin:
Sample option needed for EA to register a shortcut (1st checkbox - `Create Desktop shortcut`)

obraz

Options visible in Heroic of EA App. With game installed

obraz

flavioislima:
This should work with Ubisoft connect as well? Tested here and it won't appear in the launch options:
image

src/backend/storeManagers/sideload/library.ts
```diff
@@ -114,7 +119,48 @@ export async function getInstallInfo(): Promise<undefined> {
...
 
-export const getLaunchOptions = () => []
+export async function getLaunchOptions(
+  appName: string
+): Promise<LaunchOption[]> {
+  const gameSettings = await getSettings(appName)
+  if (gameSettings.wineVersion.type !== 'proton') return []
+  if (!gameSettings.winePrefix) return []
+
+  const protonShortcutsPath = join(
```
flavioislima, line 129: it should not check for `pfx` here? 
For me the path is:
`Prefixes/default/Ubisoft Connect/pfx/drive_c/proton_shortcuts/`

flavioislima:
I think the main issue here is that if I add EA APP as sideload, it will show the launch options fine.
But if I then want to add a game that was installed there, as a sideload, since they use the same prefix and proton version, it will show the same shortcuts. 

So If I have a game like Jedi Fallen Order and other 3 or 4 games, and I sideload Jedi FO through Heroic calling the Exe directly, in the launch options it will show the other games as well. Since they are all inside that start menu, am I wrong? But I tested only with Ubisoft for now.

flavioislima (changes requested):

src/backend/storeManagers/sideload/library.ts
imLinguin: Right, my assumption was umu usage, but right, we can add pfx

imLinguin:
@flavioislima If I understood, yeah you are right. If you have multiple Heroic entries pointing to the same prefix it will show those options. I dont think there is a good way to prevent that.

I think this feature could replace such use case of separating entries in library - a good feature for it would be a possibility to make Heroic ask which option to run every time

arielj:
maybe a solution for that is, in the sideload add game dialog, when a prefix is selected, we could list the shortcuts detected inside it

then the user can pick which options they want to keep for that game from those detected and we automatically set some flag that this particular game should not use the proton shortcuts feature for that sideloaded game

arielj:
or maybe we could have an option to convert a Launch Option entry into it's own game entry in the library (and do the same automatic flag to not update the list of shortcuts for that specific game)

I think people have requested a feature like this for games that are actually bundles with multiple games inside that you have to pick which one you run

flavioislima:
I think those are good ideas that will improve the UX in this case @arielj @imLinguin

PR #5372, opened 2026-02-26
[FIX] Clean up release notes by removing URLs and usernames

flavioislima:
The XML is never being updated if on the line with the change there is a URL so now it should clean the URL and remove the username.

---


flathub/update-flathub.ts
flavioislima, line 1: Ignore the lint issues, this file should be ignored since its not part of the main code.

imLinguin (approved):

PR #5392, opened 2026-03-05
[FIX] Pick executable inside .app using metadata

arielj:
Currently, when we sideload a native mac app/game, the code is picking the first file it finds in the `Contents/MacOS` directory but the first file is not necessarily the one that should be executed.

This PR changes that to use the plist file metadata to read the `CFBundleExecutable` metadata to know what file should be used.

This has a few more benefits:
- users pick the .app and we show that in the settings (current logic is to replace the .app config to the detected file which is kinda confusing for the user to see a path inside the .app in the input, now it happens on the fly when launching)
- this will be needed for this PR https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4608/changes#diff-69e9f792ff44c2d37e460ca5076ab07248dd5ac260a056ac8c2cff4870afee58R97 to run games without wine
- this also helps with native mac games from the zoom platform

I tested running a sideloaded native, non-native, and browser game, they all worked fine.

---

### DIFF
--- a/src/backend/storeManagers/index.ts
+++ b/src/backend/storeManagers/index.ts
@@ -106,12 +106,13 @@ export function getTargetExePath(
 
   // we can't execute a `.app` file directly
   // we have to use the path of the script inside the .app directory
-  if (executable && isMac && executable.endsWith('.app')) {
+  if (isMac) {
     try {
       // look for the `CFBundleExecutable` key and the next value
       const plistContent = readFileSync(
-        join(executable, 'Contents', 'Info.plist')
-      ).toString()
+        join(executable, 'Contents', 'Info.plist'),
+        'utf-8'
+      )
       const matchResult = plistContent.match(
         /<key>CFBundleExecutable<\/key>\n\s+<string>(.*)<\/string>/m
       )
--- a/src/backend/storeManagers/index.ts
+++ b/src/backend/storeManagers/index.ts
@@ -14,8 +14,12 @@ import { GameManager, LibraryManager } from 'common/types/game_manager'
 import { logInfo, RunnerToLogPrefixMap } from 'backend/logger'
 
 import { addToQueue } from 'backend/downloadmanager/downloadqueue'
-import { DMQueueElement, GameInfo, Runner } from 'common/types'
+import { DMQueueElement, GameInfo, GameSettings, Runner } from 'common/types'
 import { GlobalConfig } from 'backend/config'
+import LogWriter from 'backend/logger/log_writer'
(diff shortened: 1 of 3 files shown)


src/backend/storeManagers/index.ts
```diff
@@ -89,3 +93,41 @@ export async function initStoreManagers() {
...
+  logWriter: LogWriter
+) {
+  const executable = gameSettings.targetExe || gameInfo.install.executable
+
+  if (!executable) return executable
+  if (!executable.endsWith('.app')) return executable
+
+  // we can't execute a `.app` file directly
+  // we have to use the path of the script inside the .app directory
+  if (executable && isMac && executable.endsWith('.app')) {
```
CommandMC: You're checking for `if (!executable)` above, so `executable` will always be truthy here

```suggestion
  if (isMac && executable.endsWith('.app')) {
```

src/backend/storeManagers/index.ts
```diff
@@ -89,3 +93,41 @@ export async function initStoreManagers() {
...
+    try {
+      // look for the `CFBundleExecutable` key and the next value
+      const plistContent = readFileSync(
+        join(executable, 'Contents', 'Info.plist')
+      ).toString()
```
CommandMC:
```suggestion
      const plistContent = readFileSync(
        join(executable, 'Contents', 'Info.plist'),
        'utf-8'
      )
```

CommandMC (commented):
I don't really like that we have to do all this manual work. A brief search suggests we can use the `open` command to open a .app file directly (leaving the whole executable detection part up to the OS), although that would involve passing arguments differently

arielj:
> I don't really like that we have to do all this manual work. A brief search suggests we can use the `open` command to open a .app file directly (leaving the whole executable detection part up to the OS), although that would involve passing arguments differently

I did try just launching the .app file the same way we launch other executables, but it fails to launch .app files.

I'll try with `open` just to double check, but I did have issues in other PRs related to opening `.app` files (the zoom mac support and the run without wine PRs had this same issue while I was developing them)

I do think it's safer to keep this logic, because we still need this calculation for the no-wine feature and in those cases we tell legendary/gogdl/nile to run a game and those are not handling .app files correctly from my tests, and we would need those tools to be updated first if we want to not depend on this

src/backend/storeManagers/index.ts
arielj: I updated this, I also didn't need the endsWith(.app) check

PR #5394, opened 2026-03-05
Fix: Run EXE installers from their directory in sideload dialog

BrandowLucas:
The "Run Installer First" button in the sideload dialog wasn't setting the working directory properly, causing DLL dependencies to not be found. Wine needs the working dir as a Windows path, not Linux path. Changes:
- Added getDirname IPC helper to get directory from file path
- In runWineCommand, convert Linux cwd to Windows path and use cmd /c cd /d to change directory before running
- SideloadDialog now passes the EXE's directory as startFolder

---


flavioislima:

src/frontend/screens/Library/components/InstallModal/SideloadDialog/index.tsx
```diff
@@ -262,10 +262,12 @@ export default function SideloadDialog({
           appName: app_name,
           config: { ...gameSettings, winePrefix, wineVersion }
         })
+        const workingDir = await window.api.getDirname(exeToRun)
         await window.api.runWineCommand({
           commandParts: [exeToRun],
           wait: true,
           protonVerb: 'runinprefix',
+          startFolder: workingDir,
```
arielj: I wonder if we can avoid doing this rountrip to the backend to get the workingDir to send it back to the backend

maybe here we can send `startFolder: 'workingDir'` (literally that string) and then in the backend we could check `if (startFolder  === 'workingDir') startFolder = path.dirname(commandParts[0])`

BrandowLucas: done.

BrandowLucas:

BrandowLucas:
recheck

src/backend/launcher.ts
```diff
@@ -1528,6 +1528,11 @@ async function runWineCommand({
     PROTON_VERB: protonVerb
   }
 
+  logDebug(
+    `WINEPREFIX env: ${env_vars.WINEPREFIX || 'NOT SET'}`,
+    LogPrefix.Backend
+  )
```
arielj: is this change intentional or you were using it while developing this?

src/backend/main.ts
```diff
@@ -1259,6 +1259,8 @@ addHandler('gamepadAction', async (event, args) => {
 
 addHandler('getShellPath', async (event, path) => getShellPath(path))
 
+addHandler('getDirname', async (event, filePath) => path.dirname(filePath))
```
arielj: you can now remove everything related to this getDirname event/function since it's not used

arielj (changes requested):

src/backend/launcher.ts
BrandowLucas: Yes, I left it there because the WINEPREFIX wasn’t loading correctly for some reason. Keeping it there made it clearer which prefix it was pointing to. I also thought it might be helpful to leave the debug log for reference, but since you mentioned it, I’ll go ahead and remove it.

CommandMC:
Could you clarify what exactly this change achieves? I understand we're not setting the working directory correctly, but DLLs are first searched for in the folder containing the application. The CWD is searched through, but only in a lower priority ([reference](<https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#search-order-for-unpackaged-apps>))

If possible, please provide steps to verify the original issue

BrandowLucas:
> Could you clarify what exactly this change achieves? I understand we're not setting the working directory correctly, but DLLs are first searched for in the folder containing the application. The CWD is searched through, but only in a lower priority ([reference](<https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#search-order-for-unpackaged-apps>))
> 
> If possible, please provide steps to verify the original issue

I’m not at home right now, so I can’t provide screenshots. However, without this 5-line patch, selecting the **run installer first** option causes my mounted DVD game Need for Speed Most Wanted `autorun.exe` to fail with a DLL error, preventing the setup from continuing.

After applying the patch, the `autorun.exe` installer launches correctly and proceeds with the installation without any issues.

PR #5396, opened 2026-03-05
Add explicit gamescope enable toggle

lg28literconvectionmicrowaveoven:
Explicit enabling/disabling of gamescope for when some games misbehave due to compositor-specific behavior. Launches with primary display dimensions if not upscaling.

---

- [ ✔️ ] Tested the feature and it's working on a current and clean install.
- [ ✔️ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ✔️ ] Created / Updated Tests (If necessary)
- [ ✔️ ] Created / Updated documentation (If necessary)

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -359,7 +359,8 @@
             "gameWidth": "The width resolution used by the game. A 16:9 aspect ratio is assumed by gamescope.",
             "upscaleHeight": "The height resolution used by gamescope. A 16:9 aspect ratio is assumed.",
             "upscaleMethod": "The upscaling method gamescope should use.",
-            "upscaleWidth": "The width resolution used by gamescope. A 16:9 aspect ratio is assumed."
+            "upscaleWidth": "The width resolution used by gamescope. A 16:9 aspect ratio is assumed.",
+            "resolution-info": "When enabling gamescope but not enabling upscaling, game resolution is set to the dimensions of the current primary monitor."
         },
         "general": "Sync with EGL if you have a working installation of the Epic Games Launcher elsewhere and want to import your games to avoid downloading them again.",
         "mangohud": "MangoHUD is an overlay that displays and monitors FPS, temperatures, CPU/GPU load and other system resources.",
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -360,7 +360,7 @@
             "upscaleHeight": "The height resolution used by gamescope. A 16:9 aspect ratio is assumed.",
             "upscaleMethod": "The upscaling method gamescope should use.",
             "upscaleWidth": "The width resolution used by gamescope. A 16:9 aspect ratio is assumed.",
-            "resolution-info": "When enabling gamescope but not enabling upscaling, game resolution is set to the dimensions of the current primary monitor."
+            "autores-info": "Game resolution is set to the dimensions of the current primary monitor."
         },
         "general": "Sync with EGL if you have a working installation of the Epic Games Launcher elsewhere and want to import your games to avoid downloading them again.",
         "mangohud": "MangoHUD is an overlay that displays and monitors FPS, temperatures, CPU/GPU load and other system resources.",
@@ -696,6 +696,7 @@
             "enableGamescope": "Enable Gamescope",
             "enableLimiter": "Enable FPS Limiter",
             "enableUpscaling": "Enables Upscaling",
+            "enableAutoRes": "Enable Automatic Resolution",
             "missingMsg": "We could not find gamescope on the PATH. Install it or add it to the PATH.",
             "warningFlatpak": "We could not find a compatible version of Gamescope. Install Gamescope's flatpak package with runtime {{runtimeVersion}} and restart Heroic."
         },
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -692,6 +692,7 @@
         "gamemode": "Use GameMode (Feral Game Mode needs to be installed)",
         "gamescope": {
             "enableForceGrabCursor": "Enable Force Grab Cursor",
+            "enableGamescope": "Enable Gamescope",
             "enableLimiter": "Enable FPS Limiter",
             "enableUpscaling": "Enables Upscaling",
             "missingMsg": "We could not find gamescope on the PATH. Install it or add it to the PATH.",
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -591,10 +591,11 @@ async function prepareLaunch(
   }
 
   if (
-    gameSettings.gamescope?.enable ||
-    ((gameSettings.gamescope?.enableLimiter ||
-      gameSettings.gamescope?.enableUpscaling) &&
-      !isSteamDeckGameMode)
+    (gameSettings.gamescope?.enable !== null
+      ? gameSettings.gamescope?.enable
+      : gameSettings.gamescope?.enableLimiter ||
+        gameSettings.gamescope?.enableUpscaling) &&
+    !isSteamDeckGameMode
   ) {
     const gameScopeBin = await searchForExecutableOnPath('gamescope')
     if (!gameScopeBin) {
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -21,6 +21,7 @@ import {
 import i18next from 'i18next'
 import { existsSync, mkdirSync } from 'graceful-fs'
 import { join, dirname, isAbsolute } from 'path'
+import { screen } from 'electron'
 
 import {
   constructAndUpdateRPC,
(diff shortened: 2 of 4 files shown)


flavioislima:

lg28literconvectionmicrowaveoven:

lg28literconvectionmicrowaveoven:
recheck

public/locales/en/translation.json
```diff
@@ -141,14 +141,16 @@
         "select": {
             "button": "Select",
             "exe": "Select EXE",
+            "exe-placeholder": "Select EXE...",
```
arielj: not sure what's going on with these extra translations, they are not part of heroic's code unless I'm missing something

src/backend/shortcuts/ipc_handler.ts
```diff
@@ -54,7 +54,7 @@ addListener('removeShortcut', async (event, appName, runner) => {
 
   notify({
     body: isMac ? bodyMac : body,
-    title: i18next.t('box.shortcuts.title', 'Shortcuts Removed')
+    title: i18next.t('box.shortcuts.title-removed', 'Shortcuts Removed')
```
arielj: many of these changes are unrelated to the toggle, please cleanup the PR

arielj (changes requested):
tagging @Etaash-mathamsetty for a review here too since there's a similar PR already https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4224

src/frontend/screens/Settings/components/Gamescope.tsx
```diff
@@ -122,276 +123,301 @@ const Gamescope = () => {
 
   return (
```
arielj: maybe, to make this diff way cleaner, the code could be something like...

```
if (!gamescope.enable) {
  return <the code with just the toggle>
}

return {
  the current code with the added button to disable gamescope
```

I think it would make it way more clear what's new and what's just github messing up with the diff

src/backend/launcher.ts
```diff
@@ -658,6 +660,10 @@ async function prepareLaunch(
         if (gameSettings.gamescope.windowType === 'borderless') {
           gameScopeCommand.push('-b')
         }
+      } else {
+        const { width, height } = screen.getPrimaryDisplay().workAreaSize
```
BananaWorks07, line 665: This change is unrelated and shouldn't be part of this MR. Also shouldn't change behavior in a non transparent way like this.

BananaWorks07:
There are just a lot of unrelated changes here that should be removed, and we have the issue my MR had which was handling the case where the setting is not defined in older versions

src/backend/launcher.ts
lg28literconvectionmicrowaveoven: I believe it is related to this MR as gamescope defaults to a 720p game resolution unless explicitly specified with the -w and -h arguments. I'll update the frontend with a message that states this implementation for transparency.

src/backend/shortcuts/ipc_handler.ts
lg28literconvectionmicrowaveoven: the pre-push hooks fail without these changes with translation related complaints. should i just bypass them then?

src/frontend/screens/Settings/components/Gamescope.tsx
lg28literconvectionmicrowaveoven: sure i'll amend that commit

src/backend/shortcuts/ipc_handler.ts
arielj: the problem is that you have a lot of code changing translations or adding translations that are unrelated to gamescope, that's why the pre-commit hook fails, remove that code and then it shouldn't complain

arielj: like... why are you changing the key for the select exe text https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5396/changes#diff-0cdbcc24969ddbf2e8703c58e7c9996d1edfb37e70ca78411986a63bb7c689e4R104 (looks like an AI change to me)

lg28literconvectionmicrowaveoven: yes that last commit was a bandaid fix. if i remove the last commit on this MR that makes all the changes that you're pointing out, it fails to push citing complaints that have nothing to do with the real changes made by the two feature commits right before that.

arielj: the diff in this PR has a lot of changes that are unrelated to the toggle, you have to remove everything that is not related to the gamescope toggle from this PR

don't include changes that have nothing to do with this change

arielj: the way to fix missing translations is to run `pnpm i18n`, and that command does NOT change typescript code, it only update what's needed in the locale files

lg28literconvectionmicrowaveoven: ok done

src/backend/launcher.ts
```diff
@@ -590,9 +591,10 @@ async function prepareLaunch(
   }
 
   if (
-    (gameSettings.gamescope?.enableLimiter ||
+    gameSettings.gamescope?.enable ||
+    ((gameSettings.gamescope?.enableLimiter ||
       gameSettings.gamescope?.enableUpscaling) &&
-    !isSteamDeckGameMode
+      !isSteamDeckGameMode)
```
arielj: there's a problem with this logic, imagine this:

a user enables gamescope, then enables either the limiter or the upscaling

if they then disable gamescope, this will still use gamescope since the `(gameSettings.gamescope?.enableLimiter || gameSettings.gamescope?.enableUpscaling)` condition will be true

lg28literconvectionmicrowaveoven: fixed

src/backend/launcher.ts
```diff
@@ -589,11 +590,7 @@ async function prepareLaunch(
     }
   }
 
-  if (
-    (gameSettings.gamescope?.enableLimiter ||
-      gameSettings.gamescope?.enableUpscaling) &&
-    !isSteamDeckGameMode
-  ) {
+  if (gameSettings.gamescope?.enable && !isSteamDeckGameMode) {
```
arielj: this is now a different problem

if a user already had gamescope running, this will disable it for them with no explanation, they'll have to open the settings and find out they now have to enable gamescope

we should also respect what they had before

I think what you need to do is make the `enable` property be a boolean or null, and default it to null, then if it's null you use the old logic, if it's true/false you use the new logic

arielj: this is the same I reported in the other PR I linked before https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4224#discussion_r1899662104

src/backend/launcher.ts
BananaWorks07: I would disagree, heroic runs on wayland and we have no way of knowing which screen the user wants to run their game on. So better to just let gamescope handle all this instead of us

lg28literconvectionmicrowaveoven: gamescope does not in fact handle this, as i stated, it defaults to a 720p res unless defined. if the user has multi-monitor setups and wishes to specify non-default resolutions, they may do so by enabling upscaling and specifying it there.

src/backend/launcher.ts
lg28literconvectionmicrowaveoven: your logic seems sound, i've implemented it.

lg28literconvectionmicrowaveoven: @arielj should be as you requested now.

src/frontend/screens/Settings/components/Gamescope.tsx
arielj: this is still pending

lg28literconvectionmicrowaveoven: done. the diff should be much cleaner now

src/backend/launcher.ts
arielj: this is still not addressed, don't mark it as resolved, you said you were going to make it clear for the user that this was setting implicit values for width/height

and there's also the question about how this would work with wayland

lg28literconvectionmicrowaveoven: @arielj sorry about that. i've tested this on wayland/hyprland and it's worked perfectly so far. i've tested with multiple monitors too. the latest commit should make it explicit with an info tooltip.

Edu4rdSHL: I agree that this change is unrelated; we should let gamescope use its defaults, and then the user can tweak it as he wants. This, for example, forces a specific resolution by default, but what if the user wants to render at a lower resolution and then upscale it? Please dont touch defaults

Edu4rdSHL (changes requested):

Edu4rdSHL:
Imo, it would be even better if we had simply an "Enable Gamescope" toggle and the "Additional Options" box; that will make things much easier for development and users.

- No need to add every new option to the check
- Users can simply set global or per-game arguments without anything else overwriting things

But we also dont know the number of users with the current options enabled so I think it needs to remain as it's now

PR #5407, opened 2026-03-13
fix(gamepage): prevent achievement list overflow

imLinguin:
A small change that should fix the outstanding issue on large lists of achievements

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.css
+++ b/src/frontend/screens/Game/GamePage/index.css
@@ -6,6 +6,7 @@
   gap: var(--space-md);
   border-radius: 10px;
   height: 100%;
+  max-height: 100vh;
   max-width: 2400px;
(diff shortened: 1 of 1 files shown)


src/frontend/screens/Game/GamePage/index.css
```diff
@@ -6,6 +6,7 @@
   gap: var(--space-md);
   border-radius: 10px;
   height: 100%;
+  max-height: max(100vh, 700px);
```
arielj: I think it's better to do `height: 100vh;`, the 700px looks kinda magic, with 100vh height the responsiveness is also a bit better

imLinguin: I was thinking to make it limited if you really squash the view.
But yeah, we can keep it just 100vh

PR #5409, opened 2026-03-14
[UX] Download known fixes even before downloading a game

arielj:
This PR changes the logic used to fetch game known fixes. Instead of downloading the info per-game during the game's installation, we now download a json file with all the known fixes at boot from https://github.com/Heroic-Games-Launcher/known-fixes/blob/main/known_fixes.json (it's only 16kb and it's only updated eventually only if we add new fixes, so nothing to worry at this point)

This has a lot of advantages both for the end user and for us:
- we can display the wiki links even for games that are not installed, this is really useful so users can find important information before installing a game (like if they want to install a Rockstar game on Mac, they can read they'll need the paid Crossover, so they can avoid downloading the game if they don't have it)
- we don't need to add extra code to the download manager, and we also don't need to remove the file when uninstalling a game
- now with the new method to get releases info from a single repo that reduces the number of requests we do at boot, this file will only be updated if needed based on the md5 of the local file and the md5 of the latest version (before we were trying to download a known fixes file for every game installation and we didn't know if there was a fix or not until we tried)
- with the previous logic, users would only get the known fixes updated for a game by reinstalling the game, now they get the fixes updated when they open heroic if there's anything new
- the pattern is more consistent with how we handle anticheat data (that is kinda similar metadata for games from another source) and the code is more contained in it's own module now
- this also solves a bug where Heroic wouldn't download known fixes for imported games!

---

### DIFF
--- a/src/backend/anticheat/utils.ts
+++ b/src/backend/anticheat/utils.ts
@@ -6,17 +6,7 @@ import { axiosClient } from 'backend/utils'
 import { join } from 'node:path'
 import { appFolder } from 'backend/constants/paths'
 import { isMac, isWindows } from 'backend/constants/environment'
-import { createHash } from 'node:crypto'
-import { createReadStream } from 'graceful-fs'
-import { finished } from 'node:stream/promises'
-
-async function createMD5(filePath: string) {
-  const hash = createHash('md5')
-  const rStream = createReadStream(filePath)
-  rStream.pipe(hash)
-  await finished(rStream)
-  return hash.digest('hex')
-}
+import { createMD5 } from 'backend/utils/releases'
 
 const anticheatDataPath = join(appFolder, 'areweanticheatyet.json')
 
--- a/src/backend/downloadmanager/utils.ts
+++ b/src/backend/downloadmanager/utils.ts
@@ -1,19 +1,13 @@
 import { gameManagerMap } from 'backend/storeManagers'
 import { logError, LogPrefix, logWarning } from 'backend/logger'
-import {
-  downloadFile,
-  isEpicServiceOffline,
-  sendGameStatusUpdate
-} from '../utils'
-import { DMStatus, InstallParams, Runner } from 'common/types'
+import { isEpicServiceOffline, sendGameStatusUpdate } from '../utils'
+import { DMStatus, InstallParams } from 'common/types'
 import i18next from 'i18next'
 import { notify, showDialogBoxModalAuto } from '../dialog/dialog'
 import { isOnline } from '../online_monitor'
 import pathModule from 'path'
-import { existsSync, mkdirSync, rmSync } from 'graceful-fs'
-import { storeMap } from 'common/utils'
+import { existsSync, rmSync } from 'graceful-fs'
 import { gogdlConfigPath } from 'backend/storeManagers/gog/constants'
-import { fixesPath } from 'backend/constants/paths'
 
 async function installQueueElement(params: InstallParams): Promise<{
   status: DMStatus
@@ -82,8 +76,6 @@ async function installQueueElement(params: InstallParams): Promise<{
   }
 
   try {
-    downloadFixesFor(appName, runner)
-
     const { status, error } = await gameManagerMap[runner].install(appName, {
       path: path.replaceAll("'", ''),
       installDlcs,
@@ -183,15 +175,6 @@ async function updateQueueElement(params: InstallParams): Promise<{
       status: 'done'
     })
   }
-}
-
-async function downloadFixesFor(appName: string, runner: Runner) {
-  const url = `https://raw.githubusercontent.com/Heroic-Games-Launcher/known-fixes/main/${storeMap[runner]}/${appName}-${storeMap[runner]}.json`
-  const dest = pathModule.join(fixesPath, `${appName}-${storeMap[runner]}.json`)
-  if (!existsSync(fixesPath)) {
-    mkdirSync(fixesPath, { recursive: true })
-  }
-  downloadFile({ url, dest, ignoreFailure: true })
 }
 
 export { installQueueElement, updateQueueElement }
--- a/src/backend/known_fixes/utils.ts
+++ b/src/backend/known_fixes/utils.ts
@@ -1,13 +1,13 @@
 import { KnownFixesFile, Runner } from 'common/types'
-import { existsSync, readFileSync } from 'graceful-fs'
+import { createWriteStream, existsSync, readFileSync } from 'graceful-fs'
 import { storeMap } from 'common/utils'
 import { appFolder } from 'backend/constants/paths'
 import { logDebug, logInfo, LogPrefix, logWarning } from 'backend/logger'
 import { createMD5 } from 'backend/utils/releases'
 import { runOnceWhenOnline } from 'backend/online_monitor'
 import { axiosClient } from 'backend/utils'
-import { writeFile } from 'fs/promises'
 import { join } from 'path'
+import { pipeline } from 'stream/promises'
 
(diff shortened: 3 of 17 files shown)


src/backend/anticheat/utils.ts
```diff
@@ -6,17 +6,7 @@ import { axiosClient } from 'backend/utils'
...
-import { createReadStream } from 'graceful-fs'
-import { finished } from 'node:stream/promises'
-
-async function createMD5(filePath: string) {
-  const hash = createHash('md5')
-  const rStream = createReadStream(filePath)
-  rStream.pipe(hash)
-  await finished(rStream)
-  return hash.digest('hex')
-}
```
arielj, line 19: I moved this to utils/releases.ts so it can be shared between modules, since it's only used to check md5 for this releases feature only

src/backend/downloadmanager/utils.ts
```diff
@@ -185,13 +177,4 @@ async function updateQueueElement(params: InstallParams): Promise<{
   }
 }
 
-async function downloadFixesFor(appName: string, runner: Runner) {
```
arielj, line 190: we don't need any of that now during the game download

src/backend/utils/uninstaller.ts
```diff
@@ -56,13 +55,6 @@ export const removePrefix = async (appName: string, runner: Runner) => {
   rmSync(winePrefix, { recursive: true })
 }
 
-const removeFixFile = (appName: string, runner: Runner) => {
-  const fixFilePath = join(fixesPath, `${appName}-${storeMap[runner]}.json`)
-  if (existsSync(fixFilePath)) {
-    rmSync(fixFilePath)
-  }
-}
```
arielj, line 78: we don't need to delete any file now during uninstall

src/common/types/ipc.ts
```diff
@@ -22,7 +22,7 @@ import type {
   InstallInfo,
   InstallParams,
   InstallPlatform,
-  KnowFixesInfo,
+  KnownFixesInfo,
```
arielj, line 25: there was a typo before

src/frontend/hooks/hasKnownFixes.ts
```diff
@@ -1,17 +1,16 @@
 import { useEffect, useState } from 'react'
-import { KnowFixesInfo, Runner } from 'common/types'
+import { KnownFixesInfo, Runner } from 'common/types'
 
-export const hasKnownFixes = (appName: string, runner: Runner) => {
-  const [knownFixes, setKnownFixes] = useState<KnowFixesInfo | null>(null)
+export const useKnownFixes = (appName: string, runner: Runner) => {
```
arielj, line 4: react was complaining about this name

src/frontend/hooks/hasKnownFixes.ts
CommandMC, line 1: `useAwaited` can compact this hook down quite a bit:

1. Apply the following patch to allow for parameters to be used with `useAwaited`:
    ```diff
    Index: src/frontend/hooks/useAwaited.ts
    IDEA additional info:
    Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
    <+>UTF-8
    ===================================================================
    diff --git a/src/frontend/hooks/useAwaited.ts b/src/frontend/hooks/useAwaited.ts
    --- a/src/frontend/hooks/useAwaited.ts	(revision Staged)
    +++ b/src/frontend/hooks/useAwaited.ts	(date 1773573707694)
    @@ -1,17 +1,15 @@
     import { useEffect, useState } from 'react'
     
    -export function useAwaited<T>(getter: () => Promise<T>): T | null
    -export function useAwaited<T>(getter: () => Promise<T>, defaultValue: T): T
    -export function useAwaited<T>(
    -  getter: () => Promise<T>,
    -  defaultValue: T | null = null
    +export function useAwaited<T, Args extends unknown[]>(
    +  getter: (...args: Args) => Promise<T>,
    +  ...args: Args
     ): T | null {
    -  const [value, setValue] = useState<T | null>(defaultValue)
    +  const [value, setValue] = useState<T | null>(null)
     
       useEffect(() => {
         // This is `setValue` as long as the component requesting the value is mounted
         let setValueIfMounted = setValue
    -    getter().then(setValueIfMounted)
    +    void getter(...args).then(setValueIfMounted)
         return () => {
           // TODO: Send signal to BE to abort the promise
           setValueIfMounted = () => {
    @@ -19,7 +17,7 @@
             // requesting it no longer exists
           }
         }
    -  }, [])
    +  }, [getter, args])
     
       return value
     }
    ```
2. Change this file to just be
    ```ts
    import { useAwaited } from './useAwaited'
    import { Runner } from 'common/types'
    
    export const useKnownFixes = (appName: string, runner: Runner) => {
      return useAwaited(window.api.getKnownFixes, appName, runner)
    }
    ```

src/backend/known_fixes/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+export function getKnownFixesFor(appName: string, runner: Runner) {
+  if (runner === 'sideload') return null
+  if (!existsSync(fixesPath)) return null
+
+  try {
+    const fixesContent = JSON.parse(
+      readFileSync(fixesPath).toString()
+    ) as KnownFixesFile
+
+    return fixesContent[storeMap[runner] as keyof KnownFixesFile][appName]
```
CommandMC: You can get rid of this type assertion by tweaking the `storeMap` a bit:
```diff
Index: src/common/utils.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/common/utils.ts b/src/common/utils.ts
--- a/src/common/utils.ts	(revision Staged)
+++ b/src/common/utils.ts	(date 1773574285279)
@@ -1,9 +1,9 @@
 import { Runner } from './types'
 
-export const storeMap: { [key in Runner]: string | undefined } = {
+export const storeMap = {
   legendary: 'epic',
   gog: 'gog',
   nile: 'amazon',
   sideload: undefined,
   zoom: 'zoom'
-}
+} as const satisfies Record<Runner, string | undefined>
```

After that, this should apply:

```suggestion
    return fixesContent[storeMap[runner]][appName]
```

src/backend/known_fixes/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+    const url =
+      'https://raw.githubusercontent.com/Heroic-Games-Launcher/known-fixes/main/known_fixes.json'
+
+    try {
+      const { data } = await axiosClient.get<string>(url, {
+        responseType: 'text',
+        transformResponse: [(d) => d] // disable JSON parsing
+      })
+
+      await writeFile(fixesPath, data)
```
CommandMC: This would be an obvious candidate to use `responseType: 'stream'` and stream the data in as it comes in

I believe I left a comment about doing this on another PR of yours, but just in case:

```suggestion
      const response = await axiosClient.get(url, { responseType: 'stream' })
      const fileStream = createWriteStream(fixesPath)
      await pipeline(response.data, fileStream)
```

src/backend/utils/releases.ts
```diff
@@ -25,3 +28,13 @@ export const fetchLastestReleases = () => {
...
   })
 }
+
+export async function createMD5(filePath: string) {
+  if (!existsSync(filePath)) return ''
+
+  const hash = createHash('md5')
+  const rStream = createReadStream(filePath)
+  rStream.pipe(hash)
+  await finished(rStream)
```
CommandMC: Now that I'm looking at this code again & know a bit more about the Node Stream API, a small suggestion

```suggestion
  await pipeline(rStream, hash)
```

CommandMC (approved):

PR #5411, opened 2026-03-15
[UX] Add back prime check with more information

### LINKED ISSUE #5469: Re-add "Use Dedicated Graphics Card" option
### Problem description

There was an option under the "Other" tab, for a particular game, called "Use Dedicated Graphics Card".
I've updated to version 2.20.1 from 2.18.x and it's now gone.
I'm running an old game (The Nations) and it only seems to work with that option enabled.

Hope there's a way to bring this back.

### Feature description

Re-add the option, perhaps with a disclaimer or under a hidden menu?

### Alternatives

_No response_

### Additional information

_No response_

arielj:
This PR adds the option to configure Nvidia Optimus and AMD hybrid GPUs directly from the UI instead of the user having to set the PRIME env variables.

This was removed because users would use it incorrectly, now this PR adds a lot more information for the user to understand when they should is it and when not, with a link to [this new Wiki page](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/wiki/Hybrid-GPUs:-NVIDIA-Optimus-and-AMD-Dynamic-Switchable-Graphics) with even more details and the lists of GPUs that might need this, and explaining this should NOT be used for DirectX games.

With the new name of the option and all the extra context, this might fix the issue of users enabling this incorrectly and at the same time be clear enough so users that need it can use it.

I did not ad this in the Legacy tab because it's technically not Legacy, it's not that OpenGL is planned to automatically detect the correct GPU, so I put it back in Other.

image

---


src/frontend/screens/Settings/components/NvidiaPrime.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <div className="toggleWithInfo">
+      <ToggleSwitch
+        htmlId="nvidiaPrime"
+        value={nvidiaPrime}
+        handleChange={toggleNvidiaPrime}
+        title={t(
+          'setting.nvidiaprime.description',
+          'Force use of NVIDIA Optimus or AMD CrossFire dGPU. ONLY use this for OpenGL games.'
```
CommandMC: Vulkan is also affected by this, so this would have to say "... Vulkan/OpenGL games."

src/frontend/screens/Settings/components/NvidiaPrime.tsx
```diff
@@ -0,0 +1,84 @@
...
+            'setting.nvidiaprime.confirmation.message',
+            'Only one graphics card was detected in this system. Please note that this option is intended for multi-GPU systems with headless GPUs (like laptops). On single-GPU systems, the GPU is automatically used & enabling this option can cause issues. Do you really want to enable this option?'
+          ),
+          buttons: [
+            { text: t('box.yes'), onClick: () => setNvidiaPrime(true) },
+            { text: t('box.no') }
+          ],
+          type: 'MESSAGE'
+        })
+        return
```
CommandMC: I would suggest changing this to just hide the checkbox completely if only one GPU is detected

CommandMC (changes requested):
On the Wiki: I don't think having a list of models makes sense. Most of these graphics cards have desktop & laptop variants, so just the model number alone is not enough to verify Optimus being in use

src/frontend/screens/Settings/components/NvidiaPrime.tsx
arielj: done

src/frontend/screens/Settings/components/NvidiaPrime.tsx
arielj: done

src/frontend/screens/Settings/components/NvidiaPrime.tsx
```diff
@@ -0,0 +1,66 @@
...
+          className="underlined"
+          href=""
+          onClick={(event) => {
+            event.preventDefault()
+            window.api.openExternalUrl(WIKI_URL)
+          }}
+        >
+          {t(
+            'help.nvidiaprime.link',
+            'Check the list of GPUs and more information in our wiki'
```
CommandMC: This still refers to there being a list of GPUs, which is no longer the case

src/frontend/screens/Settings/components/NvidiaPrime.tsx
```diff
@@ -0,0 +1,66 @@
...
+  useEffect(() => {
+    void window.api.systemInfo
+      .get()
+      .then((result) => setNumberOfGPUs(result.GPUs.length))
+  }, [])
+
+  // if not linux, hide this setting
+  // if only 1 GPU is detected, hide it, but only if this was unset
+  // we want to still show it if less than 1 GPU and is set so users can disable it
+  if (!isLinux || (numberOfGPUs < 2 && !nvidiaPrime)) {
```
arielj, line 27: I added one more change here, if a user has the setting enabled and for some reason their setup changes and they don't have 2 GPUs, we have to still show the option so users can disable it

I'm not sure exactly how this works but we had a report of a user changing from nvidia to amd with this setup already turned on and maybe this is needed just in case for people to turn off the setting (I think users will still have 2 GPU if replaced, but if the dGPU is removed and we still have the env variables set I think it can create issues)

src/frontend/screens/Settings/components/NvidiaPrime.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <div className="toggleWithInfo">
+      <ToggleSwitch
+        htmlId="nvidiaPrime"
+        value={nvidiaPrime}
+        handleChange={() => setNvidiaPrime(!nvidiaPrime)}
+        title={t(
+          'setting.nvidiaprime.description',
+          'Force use of NVIDIA Optimus or AMD Dynamic Switchable Graphics dGPU. ONLY use this for OpenGL and Vulkan games.'
```
imLinguin, line 39: OpenGL and Vulkan are the only APIs on Linux. I think its redundant to mention them

arielj: what I meant here is that it's not needed when using DXVK, I can remove that or replace it with something different

what I want to avoid is users enabling it for DXVK, what I meant there is that when the game itself uses opengl or vulkan instead of vulkan through dxvk, does that make more sense? not sure how to explain it better for the user

flavioislima: MAybe it could be "Native OpenGL and Vulkan..."

flavioislima (approved):
lgtm

CommandMC (approved):

PR #5427, opened 2026-03-23
[Mac] Update DXMT in installed Wines if new version detected

arielj:
When I added the support for DXMT in one of the latest versions, since DXMT files were copied into copies of the wine file it was not being updated when new DXMT versions were released (only way to update DXMT was to reinstall the wine staging version).

This PR fixes that by using the new releasesInfo event and if the latest DXMT version is newer than the latest one installed locally it updates all DXMT wines with the new files.

I know we should actually use the `WINEDLLPATH_PREPEND` env variable for both DXVK-MacOS and DXMT to not have to copy files inside the wine file but until we implement that we can do this.

---

### DIFF
--- a/src/backend/tools/dxmt.ts
+++ b/src/backend/tools/dxmt.ts
@@ -14,6 +14,7 @@ import {
   writeFileSync
 } from 'graceful-fs'
 import { toolsPath } from 'backend/constants/paths'
+import { wineDownloaderInfoStore } from 'backend/wine/manager/utils'
 
 const DXMT = {
   getLatest: () => {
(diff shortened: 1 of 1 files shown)


src/backend/tools/dxmt.ts
```diff
@@ -116,6 +74,57 @@ const DXMT = {
         )
       }
     }
+  },
+
+  copyLatestDXMTFiles: (pathToWine: string) => {
+    const dxmtVersion = DXMT.getCurrentDXMTVersion()
```
arielj, line 80: this function is the same code as before, but extractes to reuse it, nothing changed in the files copy logic

src/backend/tools/dxmt.ts
```diff
@@ -116,6 +74,57 @@ const DXMT = {
...
+    filesToCopy.forEach((file) => {
+      logDebug(
+        `Copying ${join(pathToDXMT, file)} in ${join(wineInternalPath, file)}`,
+        LogPrefix.ToolInstaller
+      )
+      copyFileSync(join(pathToDXMT, file), join(wineInternalPath, file))
+    })
+  },
+  addDXMTSuffixInPlist: (pathToWine: string) => {
+    const infoFilePath = join(pathToWine, 'Contents', 'Info.plist')
```
arielj, line 116: same here, nothing changed in the plist edit code, just extracted it for readability

PR #5435, opened 2026-03-28
[UX] Support multiple profiles

arielj:
This PR adds a new feature to be able to open heroic with different profiles. This is NOT for opening multiple accounts simultaneously, it's also not for opening 2 instances of Heroic at the same time with different profiles.

## How to use it

When running heroic, we can define the `HEROIC_PROFILE` env variable to define a profile name to use internally by heroic and isolate configuration files and other files like cache stores, etc.

Passing no env variable is the "default" profile (files are stored at the root of the ~/.config/heroic folder, like it is now without this feature).

Passing the environment variable adds new folder under ~/.config/heroic with the profile name defined to create all files for that profile inside that folder.

image

## Main changes

In the backend, the stores and other configuration files are configured to use the subfolder both in the frontend and the backend.

In the frontend:

Since we use the window's localStorage, I added a wrapper class that prefixes the profile name for all keys

image

Also, the webview partitions are prefixed for each profile to avoid sharing the login when opening the stores.

Finally, the configuration for the stores also uses the subfolders.

## some comments

I'm opening this PR mainly to get feedback of the approach and help with some tests. Currently the profile is only configured through the command line with the environment variable. I was thinking about keeping it like this for now and eventually we can implement an option to switch profiles in the frontend maybe? but I don't know how that would work yet.

The builds are failing (not sure why it can't import env but it can in other files), I didn't want to spend too much time checking that if this approach is not good.

I'll add more notes in the diff. The changes are not that many different changes but a few un many places

---

### DIFF
--- a/src/backend/anticheat/utils.ts
+++ b/src/backend/anticheat/utils.ts
@@ -7,7 +7,7 @@ import { join } from 'node:path'
 import { appFolder } from 'backend/constants/paths'
 import { isMac, isWindows } from 'backend/constants/environment'
 import { createHash } from 'node:crypto'
-import { createReadStream } from 'graceful-fs'
+import { createReadStream, existsSync } from 'graceful-fs'
 import { finished } from 'node:stream/promises'
 
 async function createMD5(filePath: string) {
@@ -24,20 +24,22 @@ async function downloadAntiCheatData(latestFileHash?: string) {
   if (process.env.CI === 'e2e') return
   if (isWindows) return
 
-  const localFileHash = await createMD5(anticheatDataPath)
+  if (existsSync(anticheatDataPath)) {
+    const localFileHash = await createMD5(anticheatDataPath)
 
-  if (latestFileHash && localFileHash === latestFileHash) {
+    if (latestFileHash && localFileHash === latestFileHash) {
+      logDebug(
+        'AreWeAnticheatYet data did not change. Skipping.',
+        LogPrefix.Backend
+      )
+      return
+    }
+
     logDebug(
-      'AreWeAnticheatYet data did not change. Skipping.',
+      'AreWeAnticheatYet data changed. Downloading latest file.',
       LogPrefix.Backend
     )
-    return
   }
-
-  logDebug(
-    'AreWeAnticheatYet data changed. Downloading latest file.',
-    LogPrefix.Backend
-  )
 
   runOnceWhenOnline(async () => {
     const url = isMac
--- a/src/backend/constants/key_value_stores.ts
+++ b/src/backend/constants/key_value_stores.ts
@@ -1,10 +1,36 @@
+import { env } from 'process'
 import { TypeCheckedStoreBackend } from '../electron_store'
+import { join } from 'path'
+
+export const storesPath = env.HEROIC_PROFILE
+  ? join(env.HEROIC_PROFILE, 'stores')
+  : 'stores'
+
+export const cacheStoresPath = env.HEROIC_PROFILE
+  ? join(env.HEROIC_PROFILE, 'store_cache')
+  : 'store_cache'
+
+export const gogStorePath = env.HEROIC_PROFILE
+  ? join(env.HEROIC_PROFILE, 'gog_store')
+  : 'gog_store'
+
+export const nileStorePath = env.HEROIC_PROFILE
+  ? join(env.HEROIC_PROFILE, 'nile_store')
+  : 'nile_store'
+
+export const sideloadAppsStorePath = env.HEROIC_PROFILE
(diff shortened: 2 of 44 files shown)


src/backend/anticheat/utils.ts
```diff
@@ -7,7 +7,7 @@ import { join } from 'node:path'
 import { appFolder } from 'backend/constants/paths'
 import { isMac, isWindows } from 'backend/constants/environment'
 import { createHash } from 'node:crypto'
-import { createReadStream } from 'graceful-fs'
+import { createReadStream, existsSync } from 'graceful-fs'
```
arielj, line 10: This is unrelated to this PR, it's a bug from my previous PRs when heroic opens for the first time. I will make another PR fixing this but I'm including the change here so it's easier to test this PR.

src/backend/constants/key_value_stores.ts
```diff
@@ -1,10 +1,36 @@
+import { env } from 'process'
 import { TypeCheckedStoreBackend } from '../electron_store'
+import { join } from 'path'
+
+export const storesPath = env.HEROIC_PROFILE
```
arielj, line 5: this is the first change, all stores in the backend are configured here to set the profile subfolder if defined

src/backend/constants/paths.ts
```diff
@@ -21,13 +21,22 @@ if (process.env.CI === 'e2e') {
 export const flatpakHome = env.XDG_DATA_HOME?.replace('/data', '') || homedir()
 export const userHome = isSnap ? env.SNAP_REAL_HOME! : homedir()
 
+// general `heroic` folder for shared elements like anticheat data or downloaded tools
 export const appFolder = join(configFolder, 'heroic')
-export const userDataPath = app.getPath('userData')
+// profile-specific subfolder for isolated data like configurations, installed games, etc
+export const profileFolder = env.HEROIC_PROFILE
+  ? join(configFolder, 'heroic', env.HEROIC_PROFILE)
+  : appFolder
```
arielj, line 29: here we add different folder for the shared files and for the profile files

src/backend/storeManagers/legendary/user.ts
```diff
@@ -65,7 +66,10 @@ export class LegendaryUser {
       return
     }
 
-    const ses = session.fromPartition('persist:epicstore')
+    let partition = 'persist:epicstore'
+    if (env.HEROIC_PROFILE) partition = `persist:${env.HEROIC_PROFILE}-epictore`
```
arielj, line 70: we also need to prefix the partition name for the different webviews to avoid login from other profiles

src/backend/tools/dxmt.ts
```diff
@@ -57,9 +57,12 @@ const DXMT = {
     }
   },
   getCurrentDXMTVersion: () => {
-    return readFileSync(join(toolsPath, 'dxmt', 'latest_dxmt'))
-      .toString()
-      .split('\n')[0]
```
arielj: this is another change unrelated to this PR, adding it here to avoid errors but I'll make another PR to fix this properly before the next release

src/frontend/components/UI/ChangelogModal/index.tsx
```diff
@@ -10,8 +10,9 @@ type Props = {
   dimissVersionCheck?: boolean
 }
 
-const storage = window.localStorage
-const lastChangelog = storage.getItem('last_changelog')?.replaceAll('"', '')
+const lastChangelog = window.storage
+  .getItem('last_changelog')
+  ?.replaceAll('"', '')
```
arielj, line 15: window.storage is now defined in the root's index.tsx as a wrapper around the real localStorage

src/frontend/helpers/electronStores.ts
```diff
@@ -9,6 +9,31 @@ import {
 } from 'common/types/electron_store'
 import { GameInfo } from 'common/types'
 
+const profilePrefix =
+  window.heroicProfile === 'default' ? '' : `${window.heroicProfile}/`
+
+const storesPath = window.heroicProfile ? `${profilePrefix}stores` : 'stores'
```
arielj, line 15: similar change for the store paths made in the backend

I'm not sure if this works on windows or if we have to use the separator in the other direction (the browser doesn't have anything like path.join to join crossbrowser)

it would be great if someone on windows could test this

src/frontend/helpers/profile_local_storage.ts
```diff
@@ -0,0 +1,20 @@
+// Wrapper around the window's localStorage to prefix the profile name to all keys
+export class ProfileLocalStorage {
```
arielj, line 2: an alternative for this would be to move all the things we store in the localStorage of the browser into a new electron store or use the configStore

Issue #4982, opened 2025-10-12
Feature Request/Community Tool: Highly Accurate Python Script for Auto-Categorization (Epic, GOG, Amazon)


kiodo1981:
### Problem description

The current process for assigning categories (genres) to games in Heroic Launcher is **manual and time-consuming**, especially for users with large, multi-store libraries (Epic, GOG, Amazon).

Heroic **lacks a reliable, automated feature** to fetch genres from external databases (like Steam, RAWG, etc.) and apply them automatically to the `config.json` file, forcing users to categorize hundreds of titles individually.

### Feature description

**Solution: Python Script for Automated Genre Tagging**

This script resolves the manual categorization issue by:

1.  **High-Accuracy Lookup:** Fetches genres from Steam, RAWG, Wikipedia, and uses aggressive DuckDuckGo (DDG) logic as a robust fallback.
2.  **Full Library Support:** Automatically applies categories to Epic, GOG, and Amazon titles in `config.json`.
3.  **Result:** Demonstrated a 100% success rate on a library of 1000+ titles, including localized games.
4.  **Proposal:** Integrate this logic directly into Heroic Launcher to offer a native "Auto-Categorize" function.

[Tool Genre Heroic.py](https://github.com/user-attachments/files/22870400/Tool.Genre.Heroic.py)

[Tool Genre Heroic.py](https://github.com/user-attachments/files/22870403/Tool.Genre.Heroic.py)

### Additional information

**Solution: Python Script for Automated Genre Tagging**

This script resolves the manual categorization issue by:

1.  **High-Accuracy Lookup:** Fetches genres from Steam, RAWG, Wikipedia, and uses aggressive DuckDuckGo (DDG) logic as a robust fallback.
2.  **Full Library Support:** Automatically applies categories to Epic, GOG, and Amazon titles in `config.json`.
3.  **Result:** Demonstrated a 100% success rate on a library of 1000+ titles, including localized games.
4.  **Proposal:** Integrate this logic directly into Heroic Launcher to offer a native "Auto-Categorize" function.

T0nd0Tara:
Maybe a better option would be not to run a script "manually", but rather let heroic to automatically fetch that data via the ways you listed

That way it can be totally automatic, and will add new games to the categories straight away

The reasons I think you should not automatically let Heroic run the script comes down to 2 things:
[...]

PR #5438, opened 2026-03-28
[UX] add automatic genre filters

T0nd0Tara:
Adds the genres of each game as filters.
As epic lets you get 2 games every week, it's hard to follow which games you have exactly.
This feature fetches the genres from GoG and Amazon/Nile, and falls back to the batch api of PCGamingWiki.

image

CLOSES:
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4982


flavioislima:

T0nd0Tara:

arielj (commented):
I tried this locally and it says that it's going to fetch data for 667 games from PCGamingWiki (in the background without the user knowing)

I'm not sure what I think about this feature (pcgamingwiki matching by name is prone to errors, and also making 600 requests against their API is probably not great for us resources wise and also for them for use to abuse that api), but this definitely should not be on by default if we add it and we should be transparent to the user about what's happening.

src/backend/genres/cache.ts
```diff
@@ -0,0 +1,289 @@
...
+const GENRES_PATH = join(appFolder, 'genres.json')
+const PCGW_API = 'https://www.pcgamingwiki.com/w/api.php'
+const BATCH_SIZE = 50
+
+// gameId (appName_runner) -> genres[]
+type GenresCache = Record<string, string[]>
+
+let cache: GenresCache = {}
+
+export function loadCache(): GenresCache {
```
arielj: we already have cache management implemented using electron-store, look at src/backend/cache.ts and how it's used in other places

T0nd0Tara:
> pcgamingwiki matching by name is prone to errors

It already happens for non GoG games `getInfoFromPCGamingWiki` seem to not get a `gogID` for non `gog` games.
I can split the data I believe such that if I couldn't fetch the genres from GoG, it will locate it by it's id instead of it's name. But nothing we can do about the other runners really.

> making 600 requests against their API is probably not great for us resources wise

I'm using their batch api, and fetching only what is needed. But as it does seem a bit of a worry, I am doing those requests concurrently, so to reduce traffic to their api I can make the calls synchronous.

That means for 600 games, there are actually only 12 requests, if that still seem like a lot I can just not fetch them on login.

And it's only for the first login, the data is cached so it won't need to fetch the genres for the same games.


That being said, for GoG and Amazon I try to fetch from their api first, meaning if all of you games are from them, there are going to be a lot of requests to their api, should I just use pcgamingwiki for all of the games?
My worry is that it wont match the genres inside the game's page. (as there we first fetch from the runner's api)

src/backend/genres/cache.ts
T0nd0Tara: fixed in [550a691](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5438/commits/550a6915a3c32d1fd1638290155ebefa8d34ebc3)

flavioislima:
I have a PR that does something similar but the data is downloaded and cache on user interaction only.
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5333

I believe its fine because this is something that several other apps does. 
As long as we cache the info, I believe its fine since it wont run again except for new games.

PR #5445, opened 2026-03-31
feat: add store page link to library context menu and game details page

i3130002:
## Summary
- Adds a "Store Page" option to the right-click context menu on game cards in the library, navigating to the game's store page in the built-in WebView
- Adds a shopping cart icon button to the game details page top bar (left of the "..." menu), also linking to the store page
- Both only appear for non-sideloaded games that have a `store_url` (Epic, GOG, etc.)

## Files changed
- `src/frontend/screens/Library/components/GameCard/index.tsx` — new context menu item
- `src/frontend/screens/Game/GamePage/index.tsx` — new top bar button
- `src/frontend/screens/Game/GamePage/index.css` — button styling

## Test plan
- [ ] Right-click a game with a store URL (Epic/GOG) → "Store Page" option visible and opens store in WebView
- [ ] Right-click a sideloaded game → option not shown
- [ ] Open game details page → shopping cart icon appears left of "..." for games with a store URL
- [ ] Sideloaded games / games without `store_url` → no store page button shown

_Due to epic game login issues I could not verify the changes._

### DIFF
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -7,7 +7,8 @@ import {
   Info,
   Star,
   Monitor,
-  EmojiEvents
+  EmojiEvents,
+  ShoppingCart
 } from '@mui/icons-material'
 
 import { Tab, Tabs } from '@mui/material'
@@ -425,6 +426,15 @@ export default React.memo(function GamePage(): JSX.Element | null {
                     </NavLink>
                     <div className="topRowWapperInner">
                       {!isBrowserGame && <SettingsButton gameInfo={gameInfo} />}
+                      {!isSideloaded && gameInfo.store_url && (
+                        <NavLink
+                          className="store-page-icon"
+                          to={`/store-page?store-url=${gameInfo.store_url}`}
+                          title={t('submenu.store')}
+                        >
+                          <ShoppingCart />
+                        </NavLink>
+                      )}
                       <DotsMenu
                         gameInfo={gameInfo}
                         handleUpdate={handleUpdate}
--- a/src/frontend/screens/Game/GamePage/index.tsx
+++ b/src/frontend/screens/Game/GamePage/index.tsx
@@ -7,8 +7,7 @@ import {
   Info,
   Star,
   Monitor,
-  EmojiEvents,
-  ShoppingCart
+  EmojiEvents
 } from '@mui/icons-material'
 
 import { Tab, Tabs } from '@mui/material'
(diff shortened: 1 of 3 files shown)


flavioislima:

i3130002:

i3130002:
# Screenshots:
## Store in games library
image

## Store in game details
image





They both work and I can see myself reaching them frequently.

src/frontend/screens/Game/GamePage/index.tsx
```diff
@@ -425,6 +426,15 @@ export default React.memo(function GamePage(): JSX.Element | null {
                     </NavLink>
                     <div className="topRowWapperInner">
                       {!isBrowserGame && <SettingsButton gameInfo={gameInfo} />}
+                      {!isSideloaded && gameInfo.store_url && (
```
arielj: I don't think this should be here, there's already the option to open the store page in the 3-dots menu, I don't think this is needed that often to justify this specific button (I have literally haven't used that option in years for example)

arielj (changes requested):

src/frontend/screens/Game/GamePage/index.tsx
i3130002: Hello @arielj , thabks for your comment. I can see that we have different needs when it comes to reviewing games. 

My personal scenario:
I have a ton of games from Epic for free and I usually wonder which one is good, has the controller support, game size, system requirements, what the game even is and etc. These usually force me to check the store page. It happens a lot for me to reach for it and when you check 20 games, it's way nicer to do it from the right click. But I can also see that there is no need to do so if we have all the information in the game details page (including videos and images, corrrct game size, controller, game type, etc)

arielj: I'm not opposed to having the button in the right click menu (no idea what others think about it, we shouldn't make that menu too long either to not be annoying to use), I'm talking about this button next to the 3-dots menu in the game page that I don't think it's justified to be there when it's already in the 3-dots menu right next to it

i3130002: Great point @arielj, 

The reason I added that one too is that most of the time because of my muscle memory I just click on the Game and expect to see the store page (Just like when you browse the games in stores) But the details page we have is limited and I need to go to the actual store page to even see the game images/videos. 

I added that button to make it easier to use. 


image

System requirements has a dedicated section which is reasonable and I think having one for "Add to Steam" and "Store Page" can be useful too. (We have a lot of space up there and some comfort buttons doesn't hurt.) 

Although I'm happy with even the right click button. 

Appreciate it

arielj: but then... why not add a button for each of the options in the 3-dots menu? some users may use them often too, it's too easy to overdo that and end up with noise for most users for a preference of some users, what's why I would keep this just in the 3-dots menu, if the reason is "I expect to see it there" if feels like a preference and we would have to accept that for any other request

if we had more requests by many users for something like that then it would make more sense to be more accessible as a general usability problem, but I haven't seen that requested

i3130002: Great. I will remove it and hopefully we can merge the right click menu. 

:pray:

i3130002: Please merge

Issue #5134, opened 2026-01-01
"Verbose Logging" is harmful for game FPS. Should it be off by default?


Arcitec:
### Describe the bug

For a while now, the verbose logging flag has been enabled here:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/a5347f20a231e3b95aee1c10b389cf085af47e3e/src/frontend/screens/Settings/components/VerboseLogs.tsx#L8

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/a5347f20a231e3b95aee1c10b389cf085af47e3e/src/backend/config.ts#L362

I have personally seen a game slow down to unplayable framerates with Verbose Logging, and come back up to working framerates after disabling the logging. I think the game was Nightingale.

Anyway, logging leads to extra traces being written all the time which has an impact on framerate. Since this is a gaming launcher, I think it should be off by default. Heavy logging makes sense as an opt-in feature after the user has an issue.

### Steps to reproduce

1. Disable verbose logging.

### Expected behavior

Verbose game log off by default to focus on game performance.

### Heroic Version

Latest Stable (Flatpak)

### System Information

- OS [e. g. "Ubuntu"]: Fedora Workstaton 43

Arcitec:
The default was changed as part of a refactor here, which was meant to get more logs from new users:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4220

Arcitec:
It's possible that this logging is only for Heroic and doesn't affect Proton/Wine, VKD3D, DXVK, etc. In that case, verbose logs should be almost zero-cost and you can quickly close this issue. But I bring it up because I personally saw a game (again, I think it was Nightingale) go from 10-ish FPS to a smooth 60+ by turning off verbose logging. But perhaps it was the mere action of restarting the game that fixed it instead of the logging. 🤷‍♂️ 

In short: If it only affects "Settings: Log" then it's safe to use by default. But if it also makes game logs verbose (Proton/Wine, VKD3D, DXVK), then it's awful and should be disabled. 😉

arielj:
Verbose logs do have a performance impact only for games that have a lot of output.

I first implemented the option to disable logs to improve performance, then at some point I change the default to be disabled by default but it made support way harder because most users would send empty logs not understanding the logs were disabled

[...]

PR #5452, opened 2026-04-02
[UX] Show a message to the user suggesting disabling verbose logs

arielj:
This might be a solution for https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/5134

Currently, we have logs enabled by default to make support and troubleshooting easier, but when a game is already working correctly there's no real need to keep the verbose logs enabled and for some games with too much output this can really hurt performance.

This PR adds a message next to the play button suggesting the user to disable logs if they are enabled and the total played time is more than 5 minutes (I imagine that if there's a 5 minutes play time it means the game runs, which I think it's good enough).

This way, users can see the message and turn off the verbose logs and also learn about this and the possible impact.

Captura de pantalla 2026-04-02 a la(s) 12 55 56 a  m
(the screenshot shows the message for a game with more than 1 minute, but I changed to 5 after taking the screenshot)

---

### DIFF
--- a/src/frontend/screens/Game/GamePage/components/MainButton.tsx
+++ b/src/frontend/screens/Game/GamePage/components/MainButton.tsx
@@ -7,6 +7,7 @@ import {
   CloudQueue,
   Download,
   Error,
+  Info,
   Pause,
   PlayArrow,
   Stop,
@@ -15,6 +16,7 @@ import {
 import classNames from 'classnames'
 import { GameInfo } from 'common/types'
 import useSetting from 'frontend/hooks/useSetting'
+import { timestampStore } from 'frontend/helpers/electronStores'
 
 interface Props {
   gameInfo: GameInfo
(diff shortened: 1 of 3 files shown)


src/frontend/screens/Game/GamePage/components/MainButton.tsx
```diff
@@ -183,81 +186,96 @@ const MainButton = ({
...
+    <>
+      {showDisableLogsHint && (
+        <span className="disableLogsHint">
+          <Info />
+          {t(
+            'disableLogsHint',
+            'Looks like the game is working. You might want to disable verbose logs for better performance. Click the ^ icon and use the "Play Now" button without logs'
+          )}
+        </span>
       )}
```
arielj, line 190: the only change in this return value is this added element above `.playButtons`, the rest did't change, it's just the diff messed up because of the updated indentation adding a `<></>` wrapper

PR #5424, opened 2026-03-21
[UX] Add Proton-Sarek to wine manager

arielj:
This PR add Proton-Sarek to the Wine Manager

I also updated the releases-info repo to include this https://github.com/Heroic-Games-Launcher/releases-info/blob/main/release-data.json#L18

I tested:
- installing different versions of proton-sarek
- running a game with different versions of proton-sarek

It seems to be fine running through umu but let me know if it should be disabled for this

---

Use the following Checklist if you have changed something on the Backend or Frontend:

- [ ] Tested the feature and it's working on a current and clean install.
- [ ] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ ] Created / Updated Tests (If necessary)
- [ ] Created / Updated documentation (If necessary)

Issue #5425, opened 2026-03-21
Be more explicit to the user if their GPU does not support Vulkan 1.3+


arielj:
### Problem description

Currently, it's not clear for users when they cannot use modern DXVK when their GPU does not support Vulkan 1.3+.

Most of the times they end up asking for support and someone has to know their GPU is old or maybe something shows up in the logs.

### Feature description

Since we [already check the vulkan support](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/backend/tools/index.ts#L810) we could use that information and display it in better places than the general logs.

I imagine:
- [this PR](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5424) adds Proton-Sarek, we would pick this as the default proton on new installs
- we can display a warning in the GE-Proton tab in the wine manager
- we can display a warning if GE-Proton is selected in the Wine Version selector for games and games defaults
- we can store this check result somewhere global and print that information also in the game logs if GE-Proton is used to run a game
- maybe there are more places where this can be useful

This should make it way more user-friendly.

### Alternatives

Another option could be to even hide GE-Proton, but I don't like hiding things from users because they think something is not working, a warning explaining they should us it is probably better.

bojogon7:
### Similar issue

Automatically choose correct translation layer depending on GPU

### Problem description

Depending on GPU you can have various performance impact by using incorrect translation layer. 

### Feature description

Based on GPU choose recommended translation layer.

When GPU supports: 
Vulkan 1.4 and up = DXVK 3.0+
Vulkan 1.3 = DXVK 2.7.1
Vulkan 1.1 to 1.2 = [DXVK-Sarek](https://github.com/pythonlover02/dxvk-Sarek) 
OpenGL and Vulkan 1.0 = WineD3D

### Alternatives

Give some recommendation in proton settings like: highlight recommended settings. 

### Additional information

For old GPUs default proton could be proton-cachyos as it have DXVK-sarek. 

### Maybe can also be considered automatic DXVK for GPU brands, like: 
- NVIDIA [dxvk-nvapi](https://github.com/jp7677/dxvk-nvapi)
- Intel [dxvk-igdext](https://github.com/GameTechDev/dxvk-igdext)

Etaash-mathamsetty:
[...]

PR #5463, opened 2026-04-05
[UX] Add proton-cachyos to the wine manager

arielj:
This PR adds proton-cachyos as an option to download in the Wine Manager, this replaces https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5424

The benefits of proton-cachyos are:
- it updates more often
- it includes DXVK alternatives [through env variables](https://github.com/CachyOS/proton-cachyos?tab=readme-ov-file#proton-cachyos-config-options) (dxvk-sarek, d7vk, dxvk-low-latency)
- it has more users in general since it's the default in cachyos, so issues are identified faster

I updated the releases-info repo to also check cachyos https://github.com/Heroic-Games-Launcher/releases-info/blob/main/release-data.json#L34

I'll add some extra notes in the diff

I think we can later add some feature for users to enable the DXVK alternatives from the UI without the need of env variables, but I think that would be out of the scope of this PR, we can instruct users to use the env variable.

---

### DIFF
--- a/src/backend/tools/index.ts
+++ b/src/backend/tools/index.ts
@@ -815,7 +815,7 @@ function getDxvkUrl(): string {
   }
   if (any_gpu_supports_version([1, 1, 0])) {
     logInfo(
-      'The GPU(s) in this system only support Vulkan 1.1/1.2, falling back to DXVK 1.10.3',
+      'The GPU(s) in this system only support Vulkan 1.1/1.2, falling back to DXVK 1.10.3 for Wine, or use DXVK-Sarek with proton-cachyos',
       LogPrefix.ToolInstaller
     )
     return 'https://api.github.com/repos/doitsujin/dxvk/releases/tags/v1.10.3'
@@ -851,7 +851,7 @@ function getVkd3dUrl(): string {
   }
   if (any_gpu_supports_version([1, 1, 0])) {
     logInfo(
-      'The GPU(s) in this system only support Vulkan 1.1/1.2, falling back to VKD3D 2.6',
+      'The GPU(s) in this system only support Vulkan 1.1/1.2, falling back to VKD3D 2.6 for Wine, or use DXVK-Sarek with proton-cachyos',
       LogPrefix.ToolInstaller
     )
     return 'https://api.github.com/repos/Heroic-Games-Launcher/vkd3d-proton/releases/tags/v2.6'
--- a/src/backend/tools/index.ts
+++ b/src/backend/tools/index.ts
@@ -818,7 +818,7 @@ function getDxvkUrl(): string {
   }
   if (any_gpu_supports_version([1, 1, 0])) {
     logInfo(
-      'The GPU(s) in this system only support Vulkan 1.1/1.2, falling back to DXVK 1.10.3',
+      'The GPU(s) in this system only support Vulkan 1.1/1.2, falling back to DXVK 1.10.3 for Wine, or use DXVK-Sarek with proton-cachyos',
       LogPrefix.ToolInstaller
     )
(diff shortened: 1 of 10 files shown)


src/backend/wine/manager/downloader/utilities.ts
```diff
@@ -46,15 +59,31 @@ async function fetchReleases({
...
-            if (asset.name.endsWith('sha512sum')) {
-              release_data.checksum = asset.browser_download_url
-            } else if (
-              asset.name.endsWith('tar.gz') ||
-              asset.name.endsWith('tar.xz')
-            ) {
-              release_data.download = asset.browser_download_url
-              release_data.downsize = asset.size
+          // for proton-cachyos we want to always get the main x86 asset,
+          // the other logic sometimes gets the V3, V4 or arm asset incorrectly
```
arielj: proton-cachyos releases has many assets (arm and different x86 versions). The recommendation is to not use the `V3/V4` assets since they may be less stable, we have to ensure we install the stable asset for x86

src/frontend/screens/WineManager/index.tsx
```diff
@@ -25,10 +22,6 @@ const WineItem = lazy(
   async () => import('frontend/screens/WineManager/components/WineItem')
 )
 
-const configStore = new TypeCheckedStoreFrontend('wineManagerConfigStore', {
-  cwd: 'store'
-})
```
arielj: I had to remove the use of this electron store because there's no way to update its stored value and it ignores new repositories added.

I understand this was saved in a store to not have to loop through the other OS repositories when rendering the tabs? but it's not really helping

tuananh:
@arielj from user's perspective, everything seems to work fine when testing.

- on first load, when select proton-cachyos, the list doesn't seeem to automatically populated. i have to click refresh button.

src/backend/tools/index.ts
```diff
@@ -815,7 +815,7 @@ function getDxvkUrl(): string {
   }
   if (any_gpu_supports_version([1, 1, 0])) {
     logInfo(
-      'The GPU(s) in this system only support Vulkan 1.1/1.2, falling back to DXVK 1.10.3',
```
imLinguin, line 821: This route shouldnt be called when using Proton. What's the intention here?
Do we want a separate feature for enabling this automatically for Proton CachyOS?

arielj: We are calling `DXVK.getLatest()` when heroic opens (and it calls this internally), independent of what we have configured for a game, this prints in the general logs information for the user to know they have a GPU that does not support Vulkan 1.3+.

Then if they see that in the logs they know they have to use proton-cachyos instead of ge-proton.

arielj: this is also called when running a game with wine I think when it installs DXVK, but I'm adding this information for that use case of the DXVK update when heroic is started

arielj: I do think we should have something automated in another PR, yes, and make this more visible for the user, I opened an issue here https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/5425

We could even use this information to default to proton-cachyos with the dxvk-sarek env variable instead of ge-proton if the gpu does not support vulkan 1.3+, but I think all that would be a different feature (would mean moving this check somewhere more global to check once and reuse in other places and also the frontend)

tuananh:
Just a friendly ping. This would be a very cool feature to add.

CommandMC (dismissed):
Nit: Please use `Proton-CachyOS` (with that capitalization) to refer to this version everywhere

Otherwise, good to go as soon as the conflicts are fixed

CommandMC (approved):

PR #5466, opened 2026-04-05
[FIX] How Long To Beat 🎶One More Time...🎶

flavioislima:
It seems now HLTB changed how they load their website at first (NEXTJS SSR maybe) and then the old way of getting the game info is broken. So now we use the HLTB ID from PCGamingWiki and parse the HTML result of it. For me it seems faster than before and it is working fine.

image

image

---


arielj:
I tested this and noticed that many games don't show the HLTB data because the pcgamingwiki for those games doesn't have the HLTB info

some I noticed (I tested around 10 games):
- 20 minutes till dawn ([hltb](https://howlongtobeat.com/game/109172), [wiki](https://www.pcgamingwiki.com/wiki/20_Minutes_till_Dawn))
- Zoeti ([hltb](https://howlongtobeat.com/game/127217), [wiki](https://www.pcgamingwiki.com/wiki/Zoeti))
- [REDACTED] ([hltb](https://www.pcgamingwiki.com/wiki/Redacted), [wiki](https://howlongtobeat.com/game/155773))
- `>Observer_` (this one is weird because it does have the HLTB link in the wiki, but I guess it fails to find that page in pcgamingwiki)

Not sure what to do about that, but I can imagine people complaining that they can't see the info for some of their games.

I wonder if we should have some mapping somewhere else of games and their hltb id to not depend on pcgamingwiki? but it sounds like too much work

flavioislima:
> I tested this and noticed that many games don't show the HLTB data because the pcgamingwiki for those games doesn't have the HLTB info
> 
> some I noticed (I tested around 10 games):
> 
> * 20 minutes till dawn ([hltb](https://howlongtobeat.com/game/109172), [wiki](https://www.pcgamingwiki.com/wiki/20_Minutes_till_Dawn))
> * Zoeti ([hltb](https://howlongtobeat.com/game/127217), [wiki](https://www.pcgamingwiki.com/wiki/Zoeti))
> * [REDACTED] ([hltb](https://www.pcgamingwiki.com/wiki/Redacted), [wiki](https://howlongtobeat.com/game/155773))
> * `>Observer_` (this one is weird because it does have the HLTB link in the wiki, but I guess it fails to find that page in pcgamingwiki)
> 
> Not sure what to do about that, but I can imagine people complaining that they can't see the info for some of their games.
> 
> I wonder if we should have some mapping somewhere else of games and their hltb id to not depend on pcgamingwiki? but it sounds like too much work

yes, unfortunately we wont have a perfect solution that will work for every game, at least not for now, this should fix for the majority of them and then we can investigate it more in the future or someone else can try a different approach.

arielj (approved):

arielj:
I was thinking that maybe we can use gog.com website to get the hltb time for gog games? that should make it way more reliable at least for gog

for example, I checked SteamWorld Dig 2 and it does not show the HLTB info in heroic, but they have that in gog.com and we are linking to the correct page in the Store Page option

I think it will have some benefits for gog games:
- we'll have information for more games
- the match will be correct instead of using a search, since the gog store link is reliable
- we need only one request to get the times instead of one to pcgamingwiki to find the id and then another to hltb to get the times

flavioislima:
> I was thinking that maybe we can use gog.com website to get the hltb time for gog games? that should make it way more reliable at least for gog
> 
> for example, I checked SteamWorld Dig 2 and it does not show the HLTB info in heroic, but they have that in gog.com and we are linking to the correct page in the Store Page option
> 
> I think it will have some benefits for gog games:
> 
> * we'll have information for more games
> * the match will be correct instead of using a search, since the gog store link is reliable
> * we need only one request to get the times instead of one to pcgamingwiki to find the id and then another to hltb to get the times

But do we already have this info on the gog gameinfo? 
We could do that in another PR unless it is pretty simple to add to this one.

arielj:
> But do we already have this info on the gog gameinfo? We could do that in another PR unless it is pretty simple to add to this one.

we have the gog store url already

the change would be that, for gog games, instead of fetching PCgamingwiki and parse it to get the HLTB id, we can fetch the store page and parse it to get the time values already

I can try that later to add to this PR if you are ok with that

flavioislima:
> > But do we already have this info on the gog gameinfo? We could do that in another PR unless it is pretty simple to add to this one.
> 
> we have the gog store url already
> 
> the change would be that, for gog games, instead of fetching PCgamingwiki and parse it to get the HLTB id, we can fetch the store page and parse it to get the time values already
> 
> I can try that later to add to this PR if you are ok with that

yeah, I am fine with it. np.

PR #5467, opened 2026-04-06
[Tech]  Update electron to v41.1.1

flavioislima:
Updated Electron to latest version:
https://releases.electronjs.org/release/v41.1.1

---

### DIFF
--- a/flatpak/com.heroicgameslauncher.hgl.desktop
+++ b/flatpak/com.heroicgameslauncher.hgl.desktop
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Name=Heroic Games Launcher
-Exec=heroic-run --ozone-platform-hint=auto %u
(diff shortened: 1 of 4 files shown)


imLinguin:
Have you checked if there aren't any [breaking changes](https://www.electronjs.org/docs/latest/breaking-changes) that affect us?

flavioislima:
> Have you checked if there aren't any [breaking changes](https://www.electronjs.org/docs/latest/breaking-changes) that affect us?

yeah, I also tested and will test a bit more.
So far the things that could affect us are the `clipboard` change but we are already using the contextBridge from backend. 
And the `ozone` flag that now can be removed.

Other than that I think its fine. But we should all test to see if something broke before next release.

imLinguin (approved):

PR #5470, opened 2026-04-07
[FIX] Do not watch legendary file to refresh the lilbrary

arielj:
We are always starting a watcher for changes in the legendary installed json file.

This has a few issues:
- it triggers a refresh of the installed games in a different way than the rest of the stores (it's done with an async watcher instead of sync right after the installation ends)
- we are always trying to add the watcher even if the file does not exist, showing an error in the logs (this added confusion to users that do not use Epic games)
- the code had a workaround for a known issue with the  `watch` function that it could be triggered twice while the file was being written, it was extra complexity
- the watcher was always added, even when not doing anything that would change this file like when running a game with no GUI, using resources for no reason

This PR removes that and uses a similar approach as the other stores: once the game is installed, we refresh the installed games in the backend and the frontend was already triggering a refresh.

I tried installing and uninstalling a game and it updates the frontend correctly.

---

### DIFF
--- a/src/backend/storeManagers/legendary/games.ts
+++ b/src/backend/storeManagers/legendary/games.ts
@@ -20,7 +20,7 @@ import {
   getGameInfo as getLegLibraryGameInfo,
   changeGameInstallPath,
   installState,
-  refresh
+  refreshInstalled
 } from './library'
 import { LegendaryUser } from './user'
 import {
@@ -663,7 +663,7 @@ export async function install(
   }
   addShortcuts(appName)
 
-  refresh()
+  refreshInstalled()
 
   return { status: 'done' }
 }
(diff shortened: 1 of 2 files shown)


src/backend/storeManagers/legendary/games.ts
```diff
@@ -654,6 +654,8 @@ export default class LegendaryGame implements Game {
     }
     this.addShortcuts()
 
+    libraryManagerMap['legendary'].refreshInstalled()
```
arielj, line 657: calling this instead of `refresh` since we don't need all the other actions `refresh` and `refreshLegendary` do, we just installed the game so we only need to refresh based on legendary's installed.json

CommandMC (approved):

PR #5472, opened 2026-04-10
fix: clear GOG presence when disabling the setting

### LINKED ISSUE #5443: Disabling GOG presence updates does not unset the current logged in state
### Describe the bug

As the title says, disabling the GOG presence does not update the GOG presence state.

### Add logs

```text
Nothing relevant, just that there's a:

(14:50:02) [INFO]:    [Gog]:             GOG presence set

but no

GOG presence deleted
```

### Steps to reproduce

1. Open Heroic with GOG presence updates enabled (in advanced settings).
2. Disable the GOG presence update.
3. Close Heroic
4. The user still shows up as Online on gog's website

### Expected behavior

When the user disables the GOG presence, it should be deleted immediately, no you have to way until the server decides to change your status, because closing heroic won't delete the presence since the presence updates are disabled.

Also when the GOG presence is disabled, enabling it should set the presence immediately without having to way for a restart.

### Screenshots

_No response_

### Heroic Version

Latest Stable

### System Information

not relevant

### Additional information

A similar thing happens when enabling the GOG presence updates, the change has no effect until the next time the user opens heroic, but you still see a `GOG presence deleted` message on exit anyway.

One more kinda annoying thing related to this, the GOG presence update messages in the logs should probably be written to the GOG logs? it's kinda annoying that after some time the general logs gets flooded with the presence set messages

reverb256:
## Problem
When `Disable GOG Presence updates` is enabled and the app is closed, the user still shows as **Online** on gog.com. The presence is never cleared.

## Root Cause
`deletePresence()` in `src/backend/storeManagers/gog/presence.ts` has a `disableGOGPresence` guard that causes it to return early when the setting is enabled — exactly when presence needs to be deleted.

## Fix
Removed the `disableGOGPresence` check from `deletePresence()`. The `disablePlaytimeSync`, `isLoggedIn()`, and `isOnline()` guards remain as valid reasons to skip the API call.

Closes #5443

### DIFF
--- a/src/backend/storeManagers/gog/presence.ts
+++ b/src/backend/storeManagers/gog/presence.ts
@@ -4,6 +4,7 @@ import { axiosClient } from 'backend/utils'
 import { GOGUser } from './user'
 import { isOnline } from 'backend/online_monitor'
 import { GlobalConfig } from 'backend/config'
+import { backendEvents } from 'backend/backend_events'
 
 interface PresencePayload {
   application_type: string
@@ -63,13 +64,13 @@ async function setPresence() {
   }
 }
 
-async function deletePresence() {
+async function deletePresence(force = false) {
   try {
     const { disablePlaytimeSync, disableGOGPresence } =
       GlobalConfig.get().getSettings()
     if (
       disablePlaytimeSync ||
-      disableGOGPresence ||
+      (!force && disableGOGPresence) ||
       !GOGUser.isLoggedIn() ||
       !isOnline()
     )
@@ -90,6 +91,18 @@ async function deletePresence() {
     logError(['Failed to delete gog presence', e], LogPrefix.Gog)
   }
 }
+
+
+// React immediately when the user toggles GOG presence on/off in settings
+backendEvents.on('settingChanged', ({ key, newValue }) => {
+  if (key === 'disableGOGPresence') {
+    if (newValue) {
+      deletePresence(true)
+    } else {
+      setPresence()
+    }
+  }
+})
 
 export default {
   setCurrentGame,
--- a/src/backend/storeManagers/gog/presence.ts
+++ b/src/backend/storeManagers/gog/presence.ts
@@ -92,7 +92,6 @@ async function deletePresence(force = false) {
   }
 }
 
-
 // React immediately when the user toggles GOG presence on/off in settings
 backendEvents.on('settingChanged', ({ key, newValue }) => {
   if (key === 'disableGOGPresence') {
--- a/src/backend/storeManagers/gog/presence.ts
+++ b/src/backend/storeManagers/gog/presence.ts
@@ -97,9 +97,9 @@ async function deletePresence(force = false) {
 backendEvents.on('settingChanged', ({ key, newValue }) => {
   if (key === 'disableGOGPresence') {
     if (newValue) {
-      deletePresence(true)
+      void deletePresence(true)
     } else {
-      setPresence()
+      void setPresence()
     }
   }
 })


flavioislima:

reverb256:

arielj:
While this fixes the basic problem, I think we should have a better solution that calls deletePresence as soon as we turn that off (now we still have the issue that when the presence is set, it's going to still show the presence even after disabling it until you close heroic).

A better solution would be to:
- when heroic starts, if presence is turned on, call the setPresence
- when a user turns off the presence, call deletePresence
- when a user turns on the presence, call setPresence
- when heroic closes, if presence is turned on, call deletePresence

it would require listening to the config changed events and a bit of logic changes, but I think it will be a better fix, since the change in this PR solves only one of the different things listed in that issue

reverb256:
@arielj Thanks for the thorough review! You're right — my initial fix only addressed the symptom.

I've updated the PR to handle all 4 scenarios:

1. **Startup** → `setPresence` already called in `main.ts` on boot ✅
2. **Disable** → Now calls `deletePresence` immediately via `settingChanged` listener ✅
3. **Enable** → Now calls `setPresence` immediately via `settingChanged` listener ✅  
4. **Shutdown** → `deletePresence` already called in `utils.ts` on exit ✅

The approach follows the existing pattern used by `plausible.ts` and `logger/index.ts` — listening to `backendEvents.settingChanged`. I also fixed a latent bug where `clearInterval(interval)` didn't reset the reference, so re-enabling wouldn't recreate the refresh timer.

Please take another look when you get a chance.

src/backend/storeManagers/gog/presence.ts
```diff
@@ -1,9 +1,10 @@
...
-import { logError, logInfo, LogPrefix } from 'backend/logger'
+import { backendEvents } from 'backend/backend_events'
+import { GlobalConfig } from 'backend/config'
+import { LogPrefix, logError, logInfo } from 'backend/logger'
+import { isOnline } from 'backend/online_monitor'
 import { axiosClient } from 'backend/utils'
+import { app } from 'electron'
 import { GOGUser } from './user'
-import { isOnline } from 'backend/online_monitor'
-import { GlobalConfig } from 'backend/config'
```
arielj, line 6: Don't rearrange these things, it adds noise to the PR to not know at first glance what's an actual change and what's just reordering. Reordering is not needed.

I imagine this is made by AI, please don't just push any change made by AI, cleanup the changes and only send the minimal changes needed. Don't throw the burden of cleaning the AI generated code to the reviewers.

src/backend/storeManagers/gog/presence.ts
```diff
@@ -65,20 +66,14 @@ async function setPresence() {
 
 async function deletePresence() {
   try {
-    const { disablePlaytimeSync, disableGOGPresence } =
-      GlobalConfig.get().getSettings()
-    if (
-      disablePlaytimeSync ||
-      disableGOGPresence ||
```
arielj, line 72: removing this check would result in heroic always trying to delete the gog presence, even when not needed at all every time heroic is closed, we shouldn't do that (it executes the gogdl binary to check for auth information and makes a request to gog even when not needed)

instead, maybe add a new optional parameter to the `deletePresence` function to be able force the call even if the presence is turned off and pass `true` when the deletePresence function is called when the setting changes instead of removing this check

reverb256:
@arielj Good points, both addressed in the latest push:

1. **Import reordering noise** — removed, imports kept in original order with only the new `backendEvents` import added in place.

2. **Guard removal in deletePresence** — reverted. Added a `force` parameter instead, so the normal code path (shutdown, etc.) still checks `disableGOGPresence` and avoids unnecessary GOG API calls. Only the `settingChanged` handler passes `force=true` to bypass the guard when the user explicitly toggles the setting off.

Also fixed a latent bug where `clearInterval(interval)` didn't reset the reference, so re-enabling presence wouldn't recreate the refresh timer.

reverb256:
Thanks for the feedback @arielj — good catches, both.

**Condition reordering:** Fixed. The guard now stays at its original position, just wrapped inline with `(!force && disableGOGPresence)` so the diff only shows the actual logic change.

**Unrelated `interval = undefined` fix:** Removed from this PR. I'll submit that separately if needed.

The force parameter approach you suggested is exactly right — existing callers keep their guard, only the `settingChanged` listener passes `force=true` to bypass it.

src/backend/storeManagers/gog/presence.ts
```diff
@@ -91,6 +92,18 @@ async function deletePresence() {
   }
 }
 
+
```
arielj: there's an extra new line here, remove that please so it's not changed for other people in the future if they touch this file

arielj:
@reverb256, the lint job keeps complaining about style issues https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/actions/runs/24394668774/job/71249342911

do you have the pre-commit hooks configured to run the prettier checks before pushing code to github? do not skip that

reverb256:
Thanks for the review @arielj! I've cleaned up the PR:

1. **Import reordering** — Reverted, imports are back in original order. The earlier commits had formatter noise which I've removed.
2. **Line 72 check** — The `disableGOGPresence` check is still there on line 80: `(!force && disableGOGPresence)`. The only change in `deletePresence` is that it now also gets called with `force=true` from the settings toggle listener (line 100), so the existing check correctly skips deletion when the user has presence disabled AND it's not a forced toggle.
3. **Extra blank line** — Removed.

The full diff should now be minimal — just the `void` additions on the floating promises and the settings toggle listener.

arielj:
thanks, it works fine

next time PLEASE answer like a person and not just with AI generated output, I don't like talking to an AI

arielj (approved):

PR #5483, opened 2026-04-13
[FIX] Always pick staging asset instead of devel for Wine Staging on MacOS

arielj:
When we fetch the information of Wine releases, we always pick the first wine asset. In the Wine Staging repo, sometimes the `wine-devel` package is listed first instead of the `wine-staging` package, so we end up downloading the wrong package (like in version 11.6).

We want to always download Wine-Staging for consistency, as the Staging package has an extra battle.net patch that is not included in Wine-Devel (check the release notes https://github.com/Gcenx/macOS_Wine_builds/releases/tag/11.6), so if a user is already using Battle.net we don't break that if they change to a newer Wine version.

For users with this information already cached this won't change though, only new elements added to that versions cache will have this fix, but I think it's good enough, it's only one patch different for one specific app, anyone using Battle.net can be instructed to clear that file and get the new version if needed.

---

### DIFF
--- a/src/backend/wine/manager/downloader/utilities.ts
+++ b/src/backend/wine/manager/downloader/utilities.ts
@@ -19,6 +19,19 @@ function getVersionName(type: string, tag_name: string): string {
   }
 }
 
+interface ReleasesData {
+  data: {
+    tag_name: string
+    published_at: string
+    html_url: string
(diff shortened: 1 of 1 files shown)


flavioislima (approved):
lgtm

PR #5491, opened 2026-04-18
[FIX] Use dpaste.com instead of 0x0.st to upload logs

arielj:
0x0.st doesn't seem to work anymore

This PR replaces that service with dpaste.com, at least temporarily we can use 0x0.st again or we find an alternative that has the same features (dpaste.com does not have the option to delete files unless we create an account and have an API key, which makes no sense since the API key will be shared with the world)

I left the code for the 0x0.st integration and deleting logs commented out, removing it means a lot of changes and I rather keep this as simple as possible before a release and also maybe 0x0.st starts working again.

---

### DIFF
--- a/src/backend/logger/uploader.ts
+++ b/src/backend/logger/uploader.ts
@@ -33,6 +33,7 @@ async function sendRequestToApi(
 
 const KiB = 1024
 const MiB = KiB * 1024
+const EXPIRY_DAYS = 2
 
 /**
  * Reads the first `size` bytes of a file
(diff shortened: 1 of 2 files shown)


src/backend/logger/uploader.ts
```diff
@@ -59,19 +62,24 @@ async function uploadLogFile(
...
-  formData.set('file', fileBlob, filename)
-  formData.set('expires', '24')
+  // const filename = path.basename(fileLocation)
+  // const fileBlob = new Blob([fileContents])
+
+  // const formData = new FormData()
+  // formData.set('file', fileBlob, filename)
+  // formData.set('expires', '24')
+
+  const formData = `content=${fileContents.toString()}&expiry_days=2`
```
CommandMC: This changes the validity timeframe to 2 days (instead of one). I guess that's fine to do in general, but if we want to make that a general change, we'd also have to adapt `getUploadedLogFiles` below

arielj: right, I forgot I put 2 days there instead of 1 haha, but it was intentional (1 day seems to short for support threads in my experience)

I updated the other part now

CommandMC (approved):

PR #5508, opened 2026-04-23
feat: add a nice background on console mode

imLinguin:
This adds a background for console mode. See attached video

[Nagranie ekranu_20260422_184611.webm](https://github.com/user-attachments/assets/3963002d-a24d-451f-9e43-38efdcffd5e2)


src/frontend/screens/ConsoleMode/index.tsx
```diff
@@ -375,6 +376,18 @@ export default function ConsoleMode() {
         </div>
       </div>
 
+      <BackgroundImage
+        url={
```
flavioislima, line 380: This is wild, can you make it a variable outside of the component?

Issue #4703, opened 2025-06-30
Duplicating games


Twig6943:
### Problem description

Same feature as lutris

![Image](https://github.com/user-attachments/assets/ab6e8d84-af38-4cf7-9d0c-b3b7b2ea6804)

### Feature description

.

### Alternatives

.

kmturley:
Relates to #3010

Issue #3010, opened 2023-08-26
Don't show duplicate of games if have same in different stores


The-Ducktor:
### Problem description

having to so that when you have a game on say fallout on gog and also having it on epic it would be nice it it was only one icon in the combined view

### Feature description

combine together when you have the same game from deferent stores

Lariaa:
I think this is  a bit tricky to do. Spelling is not always the same, there are games, demos, dlcs, artwork and other things that have to match.
I also think we need to consider priorities, what if a game is not installed, where should we install from?
If it is installed from more than one store, where should we run it from?
Should you get access to gog artwork if it is not available from epic? Not sure what kind of collisions and edgecases we get here.
The Galaxy Client is doing this already so if we have a gog account in the mix we can tap into its data.
In the future if more platforms are involved we might have to fallback to partial name matching though.

The-Ducktor:
fair points i was thinking it could be a hierarchy of sorts. but i can see it would be hard to implement in terms of the different cover art and such.

arielj:
I know there was some ideas and work for this. The first issue is matching the same game in different stores, the next issue is how to handle that in a non-confusing way for the end user.

For now, as a workaround, you can always hide the ones you don't want to show: right click > hide. And you can filter hidden games using the `eye` icon at the top right of the library.

imLinguin:
Game title matching isn't that reliable, sure we can calculate some sort of similarity score and match it on some high threshold. 

Here are the options I suggest for this feature:
- build custom server able to join game IDs from multiple stores (based on igdb and games metadata from APIs both public and those that don't need authorization), data could be automatically updated thanks to igdb's websockets

- adapt GOG Galaxy account linking and implement library synchronization based completely on Galaxy's gamesdb

I'm personally more in favor of the first one, we could quickly solve bugs in data, add missing releases etc.. Also we want to avoid any privacy issues the second one may have (uploading owned games list to external service).

UI and UX of this is an after thought for now in my opinion

bgiesing:
Ultimately, at bare minimum even if auto matching is too hard, it should be possible to simply select two or more library entries and click a button to combine them manually. That would work for a lot of people even if not automated.

Of course first, that would require implementing a multi-select feature in the first place (which could also be used to bulk install, move, categorize, hide, etc.)

harinath:
[...]

PR #5519, opened 2026-04-24
[DRAFT] Deduplicate games

kmturley:
Relates to: #3010 #4703

This feature cleans/normalizes titles to ignore inconsistent casing, special characters, store-specific tags, and maintain differences between base games, sequels, and DLCs.

Dup[licate games are grouped together in the GameList, with multiple store logos. On the GamePage it allows users to toggle between different store versions of the game.

https://github.com/user-attachments/assets/eb4f3647-94a7-4159-9349-11ee4855e3e5

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "dist:win": "electron-vite build && electron-builder --win",
     "dist:flatpak": "pnpm dist:linux appimage && pnpm flatpak:prepare && pnpm flatpak:build",
     "lint": "eslint --cache .",
-    "lint-fix": "eslint --fix --ext .tsx,ts ./src",
+    "lint-fix": "eslint --fix --quiet --ext .tsx,ts ./src",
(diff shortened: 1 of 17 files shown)


CommandMC:
Could you describe how much of this is still a draft? What parts should we review already, can we help you in any way?

kmturley:
Things left to do:
- Validate the user experience is not confusing for users
- Decide whether to also combine games in search results or not
- Test against existing features filter/sort etc to ensure it is compatible.

Drackrath:
> Decide whether to also combine games in search results or not

I would do so! 

Stumbled upon this PR due to my interest in including Steam Games to this. Please ping me once this is merged.

Can you verify wheter its possible to also merge Imported games as duplicates?
I for example have the edge case that I have 2 installs of the witcher 3, one with and one without mods.

NMinker:
Maybe add something to dim a store if its been previously hidden. Have priority be the one that isn't hidden.

PR #5522, opened 2026-04-24
[Feat] Heroic Backup System and Wizard

### LINKED ISSUE #1896: Backup/Restore settings
### Problem description

Currently, if you install heroic in 2 different system (like linux and window, or mac, or a steam deck), you need to setup up everything again (hidden/favorite games, defaults, etc). Another option for more tech-savy users is to copy some specific files from `.config` and put them in the right place.

### Feature description

It should make migrations/reinstalls easier if we support an `export/backup` and `import/restore` functionality to export all the relevant settings to a file that we can then import somewhere else.

The export feature would only export relevant settings that makes sense to share and not things like credentials/auth code for example, for security reasons.

The import feature would take care of importing things into the right folder, since the folder is not the same for all systems. It could also detect if a game is configured to use a specific Wine version and suggest the user to install that.

This should help people keep heroic installations in sync more easily.

### Alternatives

We can start with simpler export/import featurs, like... maybe just a way to export hidden/favorites, then another way to export general settings, then another to export specific game settings.

### Additional information

_No response_

flavioislima:
A pretty complete backup system for Heroic that backups Logins, Settings, Game Settings, checks for missing paths for Games and Wineprefixes, check for missing wine versions, and more. With options to override and fix missing locations.
Should also convert from Flatpak to other linux installs and vice-versa.

image

image

image
image

image

image

image

### Not planned for this MVP: Proper Game files backup.

---


mglourido:
Tested on Linux (CachyOS) with Epic Games. No errors or crashes. Export, import and rollback all work as expected.Everything seems to be working correctly.

Backup structure (full backup):
```
credentials/
  legendary/
globalSettings/
  config.json
  fixes/
    Sugar-epic.json
libraryCache/
  gog_library.json
  legendary_library.json
  nile_library.json
  legendary/
    installed.json
    metadata/
perGameSettings/
  default.json
  Sugar.json
  3e02273b543f4ff0a1c24d3b534a9ac3.json
wineMetadata/
  wine-downloader-info.json
manifest.json
```

No categories folder present, as you mentioned.

One suggestion: it would be nice if the import also accepted an uncompressed folder. If someone extracts the zip to inspect or modify it, or wants to import just a single file, they have to zip it back up before importing.

Overall looks good to me!

flavioislima:
@MateoGonzalezLourido nice, thanks a lot for that.
I tested myself between flatpaks and non-flatpak builds and they work fine as well.
Not tested between platforms, for instance: Windows x Linux or Mac x Linux, for instance. I am not even sure if we should cover that, maybe we could but in a limited scope.
I will figure it out your suggestions as well, if its not too much I will do it on this PR.

mglourido:
I’m not sure if this is already possible, but is there currently a way to trigger this backup functionality outside the UI, for example through a CLI command?

If not, exposing it through the CLI could be useful for external automation scripts, such as periodically syncing Heroic backups to cloud storage without needing to interact with the launcher UI.

flavioislima:
> I’m not sure if this is already possible, but is there currently a way to trigger this backup functionality outside the UI, for example through a CLI command?
> 
> If not, exposing it through the CLI could be useful for external automation scripts, such as periodically syncing Heroic backups to cloud storage without needing to interact with the launcher UI.

This is a good idea, it doable, but I will do it after this is merged since this PR is already big.

prettyv:
If I'm not mistaken this doesn't currently include backing up of playtimes, right? The (optional) playtime synchronization feature currently only supports GOG and couldn't work for sideloaded games anyway, so I feel this would be good to have included in a backup as well. Especially after I just recently painstakingly manually consolidated playtimes into Heroic's `timestamp.json` from the various other launchers I had used in the past (a playtime import feature from other launchers would have been great, but that'd probably be quite a lot of work and a different topic). 😅

PR #5536, opened 2026-04-28
Feat: tray icon acts as toggle

### LINKED ISSUE #5533: Close back to tray when clicking the tray icon again
### Problem description

**Current behavior**
Clicking tray icon with primary mouse button brings the application from tray into foreground when close to tray is enabled. Clicking the tray icon again does nothing.

When close to tray is not enabled, primary mouse button does nothing.

### Feature description

Making it close back to tray would be handy in case of accidentally clicking the icon. It would also be consistent action mirroring which improves usability.

### Alternatives

_No response_

### Additional information

_No response_

Ishidawg:
Closes _(if it's done correctly)_ https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/5533

Click on tray icon now acts like a toggle.
If the `Exit to System Tray` is enabled, when clicking on tray icon, it will show the window if the windows is hidden, otherwise it will show. Also, if `Exit to System Tray` is not enabled, then it does nothing  when clicking.

I've run:
- `pnpm prettier`
- `pnpm codecheck`
- `pnpm test`

I tested on Linux using the `.appImage` generated by `pnpm dist:linux`.

### DIFF
--- a/src/backend/tray_icon/tray_icon.ts
+++ b/src/backend/tray_icon/tray_icon.ts
@@ -37,14 +37,11 @@ export const initTrayIcon = async (mainWindow: BrowserWindow) => {
   appIcon.on('click', () => {
     const { exitToTray } = GlobalConfig.get().getSettings()
 
-    if (!exitToTray) return
-
-    if (mainWindow.isVisible()) {
+    if (exitToTray && mainWindow.isVisible()) {
       mainWindow.hide()
-      return
+    } else {
+      mainWindow.show()
     }
-
-    mainWindow.show()
   })
 
   backendEvents.on('languageChanged', async () => {
--- a/src/backend/tray_icon/tray_icon.ts
+++ b/src/backend/tray_icon/tray_icon.ts
@@ -35,6 +35,15 @@ export const initTrayIcon = async (mainWindow: BrowserWindow) => {
 
   // event listeners
   appIcon.on('click', () => {
+    const { exitToTray } = GlobalConfig.get().getSettings()
+
+    if (!exitToTray) return
+
+    if (mainWindow.isVisible()) {
+      mainWindow.hide()
+      return
+    }
+
     mainWindow.show()
   })
 


src/backend/tray_icon/tray_icon.ts
```diff
@@ -35,6 +35,15 @@ export const initTrayIcon = async (mainWindow: BrowserWindow) => {
 
   // event listeners
   appIcon.on('click', () => {
+    const { exitToTray } = GlobalConfig.get().getSettings()
+
+    if (!exitToTray) return
```
CommandMC: Returning early here removes some current functionality. If you click the tray icon without the exit to tray option being enabled, Heroic will focus itself (on platforms where that's possible), or at least flash in the taskbar. This is in line with how other applications handle this case (observed in Discord, Ubisoft Connect), so it's probably what the user expects

src/backend/tray_icon/tray_icon.ts
```diff
@@ -35,6 +35,15 @@ export const initTrayIcon = async (mainWindow: BrowserWindow) => {
...
+
+    if (mainWindow.isVisible()) {
+      mainWindow.hide()
+      return
+    }
+
     mainWindow.show()
```
CommandMC:
```suggestion
    if (exitToTray && mainWindow.isVisible()) {
      mainWindow.hide()
    } else {
      mainWindow.show()
    }
```
Untested, but this should re-add the functionality I've mentioned in my other comment

CommandMC (changes requested):

CommandMC:
Just a note here, you can test your changes easily by running `pnpm start`. You do not need to build a whole AppImage (unless you want to send that AppImage somewhere else, of course)

src/backend/tray_icon/tray_icon.ts
Ishidawg: Ok, I didn't noted that before, thanks for explaining <3

src/backend/tray_icon/tray_icon.ts
Ishidawg: I tested here, works great <3

Ishidawg:
> Just a note here, you can test your changes easily by running `pnpm start`. You do not need to build a whole AppImage (unless you want to send that AppImage somewhere else, of course)

Oh, cool. I just do the `pnpm dist:linux` cuz I like to test on both, compiled and not. Thanks for the advise.

CommandMC (approved):

imLinguin (approved):

PR #5546, opened 2026-05-01
[Console] Add ability to install games from console mode

veryprofessionaldodo:
This PR adds some new things: 
- added ability to install games from within console mode, and sort by installed;
- shows small badge for installing games to have some indication of progress;

Also refactored some things. Extracted launching and installing functions to their own components.

Here's a video of it in action: 

https://github.com/user-attachments/assets/ef251026-d5c2-4ccf-8f57-d2f65ae96168

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -233,7 +233,7 @@
             "all": "All",
             "sideload": "Other"
         },
-        "games": "Installed Games",
+        "games": "Installed games",
         "hints": {
             "change_store": "Change store",
             "launch": "Launch",
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -233,7 +233,7 @@
             "all": "All",
             "sideload": "Other"
         },
-        "games": "Installed games",
+        "games": "Installed Games",
         "hints": {
             "change_store": "Change store",
             "launch": "Launch",
--- a/src/frontend/screens/ConsoleMode/InstallOverlay/index.scss
+++ b/src/frontend/screens/ConsoleMode/InstallOverlay/index.scss
@@ -2,28 +2,24 @@
 .consoleInstallOverlay {
   position: fixed;
   inset: 0;
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  justify-content: center;
   gap: 16px;
-  background: color-mix(
-    in srgb,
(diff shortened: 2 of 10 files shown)


flavioislima:
Thanks for doing that.
Some feedback so far:
Would be good to have a progress, maybe only the percent, with the installing text.
And also, right now there is no way to know which games are installed or not, so would be nice to differentiate them maybe by using a different opacity level, a uninstalled label or using the same style as on the normal library.

veryprofessionaldodo:
@flavioislima Good feedback! However, I personally am not really a big fan of making uninstalled games "less appealing", so I checked how Steam does it in Big Picture Mode, and it really is just like we do it here. Every click on a game leads to the Game Page, where then a user can determine whether to launch or install the game, depending on the state. I think that's an elegant solution, if we provide an easy way for users to filter by installed games. 

In my opinion, since there's not a lot of textual clutter outside of the games themselves, the small text boxes up top are visible enough to give the user a clear solution if they want to see what games are installed.

Do agree with the percentages though, will see if I can add in that information in the badge. Ideally, next up we would have a standalone page for viewing download progress.

flavioislima:
> @flavioislima Good feedback! However, I personally am not really a big fan of making uninstalled games "less appealing", so I checked how Steam does it in Big Picture Mode, and it really is just like we do it here. Every click on a game leads to the Game Page, where then a user can determine whether to launch or install the game, depending on the state. I think that's an elegant solution, if we provide an easy way for users to filter by installed games.
> 
> In my opinion, since there's not a lot of textual clutter outside of the games themselves, the small text boxes up top are visible enough to give the user a clear solution if they want to see what games are installed.
> 
> Do agree with the percentages though, will see if I can add in that information in the badge. Ideally, next up we would have a standalone page for viewing download progress.

Yeah, I know on Steam they looks the same but it is not because they do it that it is the ideal UI/UX. Steam has terrible UI/UX for several things, we just got used to it.

For me its is a mess, I have 700+ games in my library and then when navigating without knowing which ones are installed is a mess, even though we have the filter, when enabling it and disabling it, without indicator, it looks too crowded. At least an icon would be good to have.

flavioislima (approved):
Ok, lets merge it like that and we can improve on another iteration.

Thanks

PR #5595, opened 2026-05-19
[Console] Several QoL fixes for Console Mode

veryprofessionaldodo:
This MR improves some console mode things, but it's mostly minor stuff:
- made ConsoleCard selectively listen to `onMouseEnter`. Previously, if the mouse was somewhere in the active zone, but we were using a gamepad to scroll through the list, whenever the card would be behind the cursor it would automatically focus itself, making for a weird UX. Now we only trigger the mouse focus event when if the mouse is above a card, and starts moving;
- allow `focusedIndex` to be less than 0. When this happens, we assume we're in the top bar, and give no focus to any card. When we press down to return back to the game list, it returns to the card we lost focus in. This way we can actually have no card be focused, which we previously didn't allow;
- return focus to card when dismissing any install or update prompts. Previously the focus would be lost in these situations, and we would have to scroll back to return to where we previously were. Now it automatically returns to where we were;
- simplified logic to prevent "back" action button from being triggered to prevent future clause chains. Now we can just add `console-ignore-back` to the `classList` to prevent that behavior from triggering;

Here's how it's working at the moment: 

https://github.com/user-attachments/assets/76184aaf-b6cc-4811-8deb-4831605a8d10


eemmmmg:
Do you think you could implement this?
https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/5596
The lack of this option is a dealbreaker for Console Mode for me

PR #5640, opened 2026-06-05
[Perf] Lazy enrichment and race condition fixes for the download queue

### LINKED ISSUE #5634: [Performance] Lazy enrichment and race condition fixes for the download queue
# Lazy enrichment and race condition fixes for download queue

## Problem description

The download manager has several related issues stemming from how `addToQueue` and `initQueue` interact:

### 1. Blocking enrichment on add (original issue)

When an auto-update is detected on startup (or any element is added to the queue), there is a noticeable delay before the game appears in the download queue. The UI gives no feedback during this time.

The cause is `addToQueue` blocking on `getInstallInfo` before inserting the entry. This call launches an external CLI process (e.g. `legendary info`, `gogdl info`) to fetch the exact download size, which can take several seconds. Nothing is shown to the user until it resolves.

This also means every element added to the queue pays the cost of manifest resolution upfront — even if the element won't be processed for a long time, or gets cancelled before it ever runs.

### 2. Race conditions between `addToQueue` and `initQueue`

Both functions follow a read → modify → write pattern against `downloadManager` (electron-store on disk). Because `addToQueue` contains an `await` (`getInstallInfo`) between its read and write, any other operation that touches the queue during that window (e.g. `initQueue` finishing an element and calling `removeFromQueue`) can have its write silently overwritten.

In practice this means:
- An element removed by `initQueue` can reappear after `addToQueue` flushes its stale snapshot.
- An element updated by `initQueue` (e.g. `startTime`) can be reverted.

### 3. No guard against concurrent `initQueue` executions

```ts
if (isIdle()) {
  void initQueue()
}
```

If two `addToQueue` calls arrive before the first `initQueue` sets `queueState = 'running'`, two loops start processing the same queue in parallel.

### 4. `resumeCurrentDownload` doesn't reset state

```ts
function resumeCurrentDownload() {
  autoPaused = false
  void initQueue()
}
```

`queueState` is still `'paused'` when `initQueue` starts. The loop processes one element, hits `isPaused()`, and exits — the rest of the queue is never processed.

### 5. Missing error handling in the processing loop

If `installQueueElement` or `updateQueueElement` throws, the exception propagates out of `initQueue` unhandled. `queueState` stays stuck on `'running'` and `currentElement` points to a zombie element. The queue is effectively dead until app restart.

### 6. Minor issues

- `stopCurrentDownload` uses `currentElement!` (non-null assertion) without a guard — crashes if called when nothing is downloading.
- `removeFromQueue` calls `downloadManager.delete('queue')` followed immediately by `downloadManager.set('queue', elements)` — the delete is redundant.

### Core change: lazy enrichment

Move the `getInstallInfo` call (and GOG redist detection) out of `addToQueue` and into `initQueue`, executed just before each element is processed.

**`addToQueue` becomes effectively synchronous** — it reads the queue, appends the element with a provisional size (`?? MB`), writes, and notifies the frontend. No `await` between read and write eliminates the race condition window.

**`initQueue` enriches each element just-in-time** — right before starting the install/update, it calls `getInstallInfo` to resolve the real size and detect GOG redists. This means:
- Only the element about to be processed pays the manifest cost.
- Cancelled elements never fetch their manifest at all.
- The data is as fresh as possible (relevant if the user changed install options while queued).

The queue entry progresses through visible states:
1. Update detected → game added to queue instantly, showing a placeholder size.
2. Queue reaches this element → enrichment runs, size updates to the real value.
3. Download starts → existing flow takes over.

### Concurrency and robustness fixes

- **`processingLock` flag** on `initQueue` to prevent concurrent executions.
- **`queueState = 'idle'`** set in `resumeCurrentDownload` before calling `initQueue`.
- **`try/catch` around each element's processing** inside the loop, catching errors and setting `status = 'error'` instead of letting the exception kill the queue.
- **`finally` block** in `initQueue` that resets `currentElement`, `processingLock`, and `queueState` regardless of how the loop exits.
- **Null guard** in `stopCurrentDownload`.
- **Remove redundant** `downloadManager.delete('queue')` in `removeFromQueue`.

## Alternatives

An event-driven rewrite (class-based, in-memory state, `EventEmitter` for signaling) was considered. It would solve all of the above more cleanly, but the scope is significantly larger. The lazy enrichment approach achieves the main goals (responsiveness, reduced race conditions, no wasted work) with minimal structural changes to the existing code.

## Additional information

- The download itself still starts only after enrichment resolves, so there is no risk of starting an incomplete or incorrect update.
- The total wall-clock time may be similar or even slightly better (no wasted manifest fetches), but the perceived responsiveness improves significantly.
- All proposed changes are backward-compatible — no type changes or API surface changes required.

mglourido:
Fixes #5634

Fixes a blocking delay when adding elements to the download queue and addresses several race conditions. See the linked issue for full details.

---
How to test
Setup: pnpm install && pnpm start

1. Basic install — size visible immediately (no ?? MB)

Open the install dialog for any game → note the size shown
Close without installing
Queue the same game for install
Expected: size appears immediately in the queue (cache hit), no ?? MB

2. "Update all" — no UI freeze

Have 3+ games with pending updates
Click "Update all"
Expected: all entries appear instantly, UI stays responsive
Before this PR: app froze ~1-2s per game sequentially

3. First-time install — background analysis fills ?? MB

Queue a game without opening the install dialog first
Expected: briefly shows ?? MB, updates to correct size within a few seconds

4. Pause / Resume

Start a download → Pause → Resume
Expected: download continues correctly

5. Cancel current download

Start a download → Cancel
Expected: queue advances to next item normally

6. Cancel queued item (not active)

Queue 2+ games → cancel one that isn't downloading
Expected: item disappears, active download unaffected

7. Full queue completion

Queue 2-3 small items and let them finish
Expected: automatic progression, queue empties correctly

---


flavioislima:

mglourido:

mglourido:
Note: the lint errors shown in CI (`await-thenable` in `ipc_handler.ts`) are pre-existing in the `main` branch and unrelated to this PR. I verified locally that my changes introduce 0 new lint errors.

imLinguin:
These seem to come from the definition change to `addToQueue` function.

mglourido:
The fix was correct locally but wasn't properly committed before pushing, so the PR was missing the `async` signature on `addToQueue`. Resolved by explicitly committing the change. Prettier formatting has also been fixed.

mglourido:
Note for future improvement: `cancelCurrentDownload` calls `removeFromQueue` once per DLC in a loop — if `installDlcs` has N entries, that's N disk writes and N frontend notifications. I plan to address this in a separate PR by filtering the queue once and notifying the frontend once.

mglourido:
# Download Manager: lazy enrichment + size cache
 
## Problem with the original code
 
`addToQueue` called `getInstallInfo` with `await` inside its own body — blocking the entire process while the Legendary or gogdl CLI responded (1-2 seconds per game). When using "update all" with multiple games, those calls stacked sequentially: the UI froze until all of them finished.
 
Additionally, `initQueue` had no mutual exclusion mechanism (`processingLock`) and used the `element` variable directly inside the while loop, which could mutate during iteration (race condition).
 
## Refactoring: lazy enrichment
 
Size calculation is extracted from `addToQueue` into `enrichElement`, which runs just before each element starts processing. `addToQueue` becomes instantaneous.
 
Side effects of this change:
- Queued elements show `?? MB` while waiting for their turn (in the original, the size was visible from the moment they were added)
- `processingLock` is added to prevent concurrent re-entries into `initQueue`
- `currentProcessing` is introduced as an immutable snapshot inside the loop
## Solution to `?? MB`: `sizeCache.ts`
 
Persistent cache (electron-store, `size-cache.json` file) of up to 10 entries with a 12-hour TTL. Stores the formatted size (e.g. `"2.5 GB"`) associated with a key that uniquely identifies a download configuration.
 
The result is the same visual behavior as the original — correct size visible from the moment it's added to the queue — but without blocking `addToQueue` again.
 
### Cache key
 
```
type :: appName :: runner :: platform :: installLanguage :: sdlList :: branch :: build
```
 
- **`type`** (`install` / `update`): an update is a delta patch, different size from a full install
- **`platform`**: normalized internally (`Mac` → `osx`, `Windows` → `windows`) so that the store manager value and the queue element value are consistent
- **`installLanguage`**: GOG has different sizes per language (`perLangSize`)
- **`sdlList`**: Legendary allows selecting optional packages, each combination has a different size. Sorted before inclusion so that `[A,B]` and `[B,A]` are equivalent
### Where the cache is populated
 
| Point | When | Result |
|---|---|---|
| IPC handler `getInstallInfo` (`main.ts`) | When the install dialog is opened | Size available before pressing install |
| `enrichElement` (`downloadqueue.ts`) | When processing each element | Size saved for next time |
| `analyzeElementSize` (`downloadqueue.ts`) | Serialized background task, for waiting elements | Fills in `?? MB` without blocking the active download |
 
### Where it's queried
 
- **`addToQueue`**: before assigning `?? MB`, checks for a cache hit → the element enters the queue already with the correct size
- **`enrichElement`**: if there's a hit and the game is not GOG Windows (which needs the full response to handle redist dependencies), skips the `getInstallInfo` call
## Background analysis (`analyzeElementSize`)
 
For elements waiting in the queue (updates that don't go through the install dialog), a **serialized** background analysis is launched — one at a time — to avoid saturating CLI calls. When it finishes, it updates the element in the queue and notifies the frontend via `changedDMQueueInformation`.
 
The Legendary and GOG store managers already cache the result of `getInstallInfo` internally, so if the dialog was opened before, the background analysis call is a file read, not a CLI call.
 
## Summary of improvements vs original code
 
| Scenario | Original | This branch |
|---|---|---|
| Adding 1 game to queue | Blocks ~1-2s (CLI call) | Instantaneous |
| "Update all" (5 games) | Blocks ~5-10s sequentially | Instantaneous, background analysis |
| Second install of the same game | Always blocks | Instantaneous (cache hit) |
| Size visible in queue | Always correct (but blocking) | Correct from cache, `?? MB` only on first time |
| Race condition in `initQueue` | Present | Fixed with `processingLock` and `currentProcessing` |

mglourido:
The Flatpak build timed out — Is this a real network failure or a problem with my code? The code looks fine to me.

mglourido:
Testing report — New queue vs original Heroic (ARK: Survival Evolved + 4 DLCs)
Startup behavior

New: all DLC entries appear in the queue simultaneously and instantly with correct sizes, without delaying the main installation.
Original: DLC entries appeared progressively as the main game was installing, consuming resources that could otherwise go to the active download.

Transition between downloads
No noticeable difference — transitions from one DLC to the next are instantaneous in both versions.
Duration

New: 14m 40s
Original: 15m 50s

Note: the new version was run first, which may introduce slight variability in the comparison due to network conditions. Both runs were done without prior caches.
Conclusion
The most significant improvement is UX — the new queue feels considerably more responsive and fluid. The time difference is not drastic but consistent. The new implementation is also more robust against race conditions and edge case bugs.
Note: several fixes in this PR (race condition prevention, resume behavior, error handling) could not be reflected in this test as they require specific conditions such as auto-updates on startup, network interruptions, or concurrent queue operations. The test covers the core install flow only.

mglourido:
Additional manual testing
Beyond the installation test above, the following queue functionalities were manually verified to work correctly with the new implementation:

Pause and resume downloads
Cancel current download
Cancel queued items
Queue progression after completion of each item
App behavior during and after queue completion

During this testing, some bugs in the new implementation were identified and fixed — these are already reflected in the PR commit history. Additionally, some existing bugs and potential future improvements unrelated to this PR were identified and will be reported and addressed separately

MateoGonzalezLourido:
Ready for review. Apologies for the scope creep — the PR ended up larger than initially planned as I found related issues worth addressing together.Happy to split it if preferred.

mglourido:
Ready for review. Apologies for the scope creep — this ended up touching more files than originally planned, but the changes are concentrated in a few areas:
Core change: downloadqueue.ts — lazy enrichment, race condition fixes, and cancellation/pause logic. This is where the bulk of the diff lives.
New files: sizeCache.ts (install size cache) and downloadManager.test.ts (707 lines of tests covering the new logic).
Cleanup: resolved pre-existing lint/async warnings in the original download queue code.
Supporting changes: type additions in common/types.ts, electron_store.ts, and ipc.ts to support the queue and cache. Everything else (main.ts, frontend components, redist.ts) are minimal wiring changes.
Happy to walk through any part of it or split if preferred.

flavioislima:
Can you please write a "HOW TO TEST" step by step on this and other PRs you opened? this will improve the review time. Thanks

mglourido:
Added in this PR, in the description.

In the code, I added a cache to avoid recalculating the size twice in the same installation flow and to fetch data already used in the library menu. This doesn't mean it fails if the cache is missing (it recalculates on install), but I implemented this cache persisting it to disk with a TTL. I'm not sure if I should store it in memory with a Map() instead — it might be overcomplicating something that doesn't need it.

if this gets merged, I plan to open separate PRs for some additional improvements and features I kept out of this one to avoid scope creep.

mglourido:
These are merge correction commits.

mglourido:
This is a guard I was missing that would continue the current download if it was paused and add things to the queue.

PR #5656, opened 2026-06-10
[Feat] Detect and clean up incomplete game downloads from the library #5655

### LINKED ISSUE #5655: [Feat] Detect and clean up incomplete game downloads from the library
### Problem description

When a game installation is stopped and the user chooses to keep the downloaded files, Heroic removes the entry from the download queue. The game no longer appears as installed, the Download Manager no longer tracks it, and the partial files remain on disk with no way to find or remove them from the UI other than deleting them manually. The same happens when legendary exits with an error mid-download.

**Current behavior:**
No cleanup option is shown for games with incomplete installations. The Download Manager only tracks active, paused, or queued downloads — once a download is cancelled with files kept, or fails with an error, the entry moves to the "Completed" section as "(Canceled)" with no action to remove the leftover files. The game card in the library shows no indication that partial files exist on disk.

### Feature description

Games with incomplete or interrupted installations should be visually identifiable in the library and have a cleanup option available, the same way fully installed games have an uninstall option.

### Alternatives

_No response_

### Additional information

_No response_

mglourido:
Fixes #5655 

When a user stops an installation and chooses to keep the downloaded files, Heroic calls cancelCurrentDownload({ removeDownloaded: false }), which removes the entry from the persistent download queue. On the next launch, initQueue finds nothing — the game never re-appears in the Download Manager, is not marked as installed, and the partial files on disk are invisible to the user with no cleanup path in the UI.

The same applies when legendary exits with an error mid-download: installQueueElement resolves with status: 'error', the entry moves to the DM "Completed" section as (Canceled), and the files are left on disk with no action to remove them.

This PR adds: a visual badge on game cards that have partial files on disk, a library filter to audit them, and a cleanup action in the uninstall modal that cancels any active/queued download for that game before deleting the folder.

### `downloadqueue.ts`
- `pauseCurrentDownload`, `cancelCurrentDownload` and `stopCurrentDownload` properly converted to `async/await` — they were previously void and could leave processes running at the time files were deleted.
- New function `cancelDownloadForApp(appName)`: cancels if it's the active download, or removes it from the queue if it was pending.
- New function `cleanUpPartialInstall(appName, runner, folder, folderName)`: generic, runner-agnostic cleanup — cancels the active/queued download and deletes the folder from disk. **Any current or future runner benefits without having to implement anything extra.**
### `uninstaller.ts`
- `uninstallGameCallback` now accepts an optional `partialInstallFolder` parameter.
- If the game is not installed but a partial folder exists, routes to the specialized cleanup flow instead of the normal `uninstall` flow (which assumes a full installation).
- Respects an optional `game.cleanUpPartialInstall()` method if the runner implements it; otherwise falls back to the generic one.
- Now returns `Promise<boolean>` — the frontend knows whether cleanup succeeded before clearing localStorage.
### `game_manager.ts`
- `Game` interface extended with optional method `cleanUpPartialInstall?(partialInstallFolder: string): Promise<void>` — a clear contract for runners that need custom teardown.
### `ipc.ts`
- `uninstall` in `AsyncIPCFunctions` now accepts `partialInstallFolder?: string` and returns `Promise<boolean>` instead of `Promise<void>`.
### `utils.ts` — `removeFolder`
- Refactored by extracting an internal `scheduleDelete` helper, removing duplicated logic.
- Adds `force: true` to `rmSync` — prevents errors on Windows when the process that was writing the file hasn't released the handle yet.
- Try/catch in the timer: a permissions error no longer propagates as an unhandled async exception.
### `legendary/library.ts`
- Guard against empty `res.stdout` in `getInstallInfo` — throws a clear error instead of calling `JSON.parse("")` and crashing.
---
 
### New hook `useHasPartialInstall` (`src/frontend/hooks/useHasPartialInstall.ts`)
- Reactive hook that detects whether a game has a partial install tracked in localStorage.
- Listens to the `storage` event to update badges and filters **in real time** without requiring a page reload.
- When it detects the game has become fully installed (`isInstalled = true`), automatically clears the stale localStorage marker.
- Returns `{ hasPartialInstall, partialInstallFolder }`.
### `hasProgress.ts` — Refactor
- Removed the complex percentage recalculation logic (`calculatePercent` / `previousProgress`), which was a leftover from an earlier era and produced incorrect values.
- The hook now returns only `[currentProgress]` — simpler and correct.
### `InstallProgress.ts`
- New function `clearInstallProgress(appName, runner)`: resets the store state to 0% instead of deleting the key — prevents the progress bar from getting stuck at the last value when the user reinstalls.
### `helpers/library.ts`
- `getPartialInstallFolder(appName)`: reads the partial install marker from localStorage with invalid JSON handling.
- `clearPartialInstall(appName)`: removes the marker and dispatches a synthetic `StorageEvent` so `useHasPartialInstall` reacts immediately (even within the same tab).
- `install()`: now writes the folder marker to localStorage **before** calling the backend (not only when the user manually stops). If the install fails, the marker is automatically cleared via `.catch(clearPartialInstall)`.
- `handleStopInstallation`: dispatches a `StorageEvent` on pause or cancel to keep the reactive hook in sync.
---
 
### `GameCard`
- Imports `useHasPartialInstall` to track partial install state.
- New visual badge with a `Storage` icon (MUI) overlaid on the game cover, with tooltip `"Partial install on disk — right-click to clean up"`.
- Badge shifts down if the card already shows a status badge (CSS `:has(.gameCardStatus)`), avoiding overlap.
- The "Uninstall" context menu item now appears for games with a partial install too (`isInstalled || hasPartialInstall`).
- Passes `partialInstallFolder` to `UninstallModal`.
### `GameCard/index.css`
- Styles for `.partialInstallBadge`: absolute positioning, semi-transparent background, `--status-warning` color, correct z-index, hidden in list view.
### `UninstallModal`
- Accepts `partialInstallFolder?: string` prop.
- For partial installs: shows a differentiated message (`"Do you want to clean up the incomplete download of..."`), hides the Wine prefix and settings checkboxes (irrelevant for an incomplete download), and bypasses the "another Epic game is installing" lock (the backend cancels the download before deleting).
- Only clears localStorage if the backend confirmed success (`uninstalled === true`).
### `GameSubMenu`
- Uninstall button now visible when a partial install exists, even if the game is not installed.
- Passes `partialInstallFolder` to `UninstallModal`.
### `GamePage`
- `previousProgress` removed from `hasProgress` (hook refactor), passes `null` explicitly where it was passed before.
### `LibraryFilters`
- New toggle: **"Show partially installed only"** — filters the library to show only games with an incomplete download.
- Resets together with the rest of the filters when clicking "Reset Filters".
### `Library/index.tsx`
- `showPartiallyInstalledOnly` state with localStorage persistence (`show_partially_installed_only`).
- Reactive filtering: excludes installed games and shows only those with a partial folder in localStorage.
- Exposes the state and setter via `LibraryContext`.
### `LibraryContext.tsx`
- Initial values for `showPartiallyInstalledOnly` and `setShowPartiallyInstalledOnly`.
### `frontend/types.ts`
- `LibraryContextType` extended with `showPartiallyInstalledOnly` and `setShowPartiallyInstalledOnly`.
---
 
## i18n
 
- New key `gamepage:box.uninstall.partial-message` — confirmation message for partial install cleanup.
- New key `label.game.partial-install` — badge tooltip.
- New key `header.show_partially_installed_only` — library filter label.
- Translation updates across 80+ locales (via merge with `main`).
---
 
## What this PR adds over the existing solution
 
| | Previous solution | This PR |
|---|---|---|
| Partial file cleanup | Only if the user stops from the UI | Also when the process crashes or Heroic closes |
| Visual badge | No | Yes, reactive in real time |
| Library filter | No | Yes, with persistence |
| Differentiated modal | No | Yes, without irrelevant checkboxes |
| Runner compatibility | Legendary only | All runners (generic fallback by default) |
| Progress on reinstall | Could stay stuck at previous % | Resets to 0% correctly |
| Async correctness | Several unawaited functions | Fixed throughout the call chain |

---
How to test

  This branch detects, flags, and cleans up partial installs (downloads that were started but never finished). The implementation is runner-agnostic: it works for any store that uses Heroic's download queue (Epic, GOG, Amazon) without the runner having to implement anything.

  Prerequisites

- A logged-in account from any store (Epic, GOG, or Amazon) with at least one medium/large game, so you have time to cancel the download partway through.
- Ideally, test with at least two different stores to validate the generic nature of the refactor.
- Local build of the branch:
pnpm install
pnpm start

Scenario 1 — Create a partial install by cancelling the download

1. Start installing a game (any store).
2. At around 20–40%, cancel the download and choose to keep the downloaded files.
3. Expected: the game stays not installed but a partial-install badge (disk icon) appears in the top-left of the GameCard.

Scenario 2 — Badge and availability of the cleanup

1. Right-click the partial install's card → Uninstall should appear even though it isn't "installed".
2. Same behavior from the GamePage submenu.
3. The badge must not appear if the game is installing, queued, or fully installed.

Scenario 3 — Correct cleanup modal

1. Open the Uninstall modal from a partial install.
2. Expected:
  - Message like "Do you want to clean up the incomplete download of '...'?" (not the regular uninstall text).
  - The Wine prefix and settings checkboxes are not shown, nor is the default-prefix warning.
3. Confirm → the active/queued download is cancelled, the leftover folder is deleted (after a short delay), the badge disappears, and progress resets to 0%.

Scenario 4 — Clean up a partial install while another Epic download is running

1. Have an Epic partial install and, in parallel, install another Epic game.
2. Clean up the partial install.
3. Expected: it's allowed (the backend cancels the partial install's download before deleting); the "can't uninstall while an Epic game is installing" block does not apply to the partial-install case.

Scenario 5 — Library filter

1. Filters → enable "Show partially installed only".
2. Expected: only not-installed games with a partial folder are listed, regardless of store. The state persists across reloads and "Reset filters" clears it.

Scenario 6 — Persistence across a crash / abrupt close

1. Start a download and close Heroic without cancelling.
2. Reopen Heroic.
3. Expected: the game is still flagged as a partial install (the marker is written when the download starts, not only when it's cancelled manually).

Scenario 7 — Reinstall after cleanup

1. After cleaning up a partial install, reinstall the same game.
2. Expected: the bar starts from 0% (it doesn't stay "stuck" at the previous percentage) and any leftover marker is removed once it completes.

Regressions to check

- Multi-store: repeat scenarios 1–3 on a different store (e.g. if you tested Epic, also validate GOG or Amazon) to confirm the generic cleanup works across all runners.
- Normal uninstall of a fully installed game: the modal still shows the Wine/settings checkboxes and the usual text.
- DLC uninstall: the DLC message is intact.
- Normal download progress: advances correctly from 0→100 (the previousProgress calculation was removed).

---


mglourido:
If anyone can help me test this feature, I'd appreciate it. It works for me, but I have all my files in the usual Heroic path. If it works for someone using different drives, etc., then great.

Even so, if it works for me, it should work for everyone; I'm not creating anything new.

mglourido:
I love this feature; it adds so much convenience.

mglourido:
Post-added changes:

Added an icon on the game card to indicate that local game files are present on disk

Uninstalling is now allowed while a game is queued, downloading or updating (The download is cancelled and then uninstalled, thus bypassing the restriction that prevents this.)

Download cache is now reset on uninstall, matching Legendary's behavior

Removed install status icon when a game is queued or actively downloading

CommandMC:
When pausing/cancelling a game installation, Heroic asks the user whether they want to keep the downloaded files. Your situation seems to fix the case where they choose "Yes", but didn't actually want "Yes". I don't think this situation needs fixing.

mglourido:
This covers more than just that case. The cancel dialog only appears for intentional cancellations initiated through Heroic. It doesn't handle:

Network failures or Legendary crashes mid-download
System restarts while downloading
Cases where the user cancels from outside Heroic (e.g. killing the process)

In all of these, partial files are left on disk with no cleanup option in the UI.

Also add a way for the user to see that they have partial installations in the library without having to enter the game's installation menu.

CommandMC:
> In all of these, partial files are left on disk with no cleanup option in the UI.

That's not the case though. These downloads are still visible in the Download Manager, and will automatically be re-attempted when launching Heroic again.

> Also add a way for the user to see that they have partial installations in the library without having to enter the game's installation menu.

This is also possible in the Download Manager.

mglourido:
Thanks for the clarification! That's partially correct — paused and crashed downloads do stay in the DM queue and get re-attempted on restart. However, for the Stop → "Yes, keep files" case, cancelCurrentDownload calls removeFromQueue, which removes the element from the persistent store. On the next launch, initQueue finds nothing, so the game doesn't appear in the DM queue at all — only in the Completed section as "(Canceled)", with no cleanup action available.
Same gap applies when Legendary errors out mid-download: removeFromQueue is called, files stay on disk, no cleanup path in the DM.
That's the specific case this PR addresses. If I'm wrong about this, I genuinely couldn't find a way to achieve what this PR does through the existing DM.

arielj:
To be honest, I think all the code needed for this feature is overkill when a user could just manually delete the folder and only affects some edge cases.

Like... having a specific library filter JUST for this for example seems really weird.

Maybe it makes more sense to have some kind of section in the "Advanced" settings (next to the reset heroic and reset cache maybe?) to detect if there are leftover files from partial installation and instruct the user that they can delete them. Then it would be a way simpler feature with less code and it's closer to "clear cache" where it makes more sense than in the library.

mglourido:
I just added the filter because it wasn't difficult.And that way they don't search for downloads through the library and they all appear together.

The download queue could be improved to save partial downloads there (instead of in the history), similar to how Steam handles it. Although Steam does allow uninstalling partial downloads, similar to how it's implemented here. This is just a suggestion, and I'm removing everything from the library.

PR #5659, opened 2026-06-11
[Fix] Remove redundant legendary list call on startup  #5641

### LINKED ISSUE #5641: perf: avoid redundant legendary list --third-party call on startup
### Problem description

During startup, legendary list --third-party runs twice in quick succession:

refreshLibrary() → calls legendary list --third-party and writes assets.json
listUpdateableGames() → calls legendary list --third-party again, reads the same assets.json

The second call is redundant in this flow since the data was just fetched seconds earlier. On my system it adds ~3 seconds to startup.
I understand listUpdateableGames() is designed to be self-contained — it can't assume a prior refresh happened, which makes sense for manual refresh calls and other entry points. This is about optimizing the startup path specifically, not changing that contract.

### Feature description

Cache-aware approach: have refreshLibrary() record a timestamp after running the CLI. listUpdateableGames() checks the timestamp and skips the CLI call if assets.json was written within the last N seconds, falling back to the full call otherwise.
This keeps listUpdateableGames() self-contained (it still works standalone), just avoids the duplicate work when called right after a refresh.

### Alternatives

_No response_

### Additional information

Happy to submit a PR if this approach sounds reasonable, or adapt to a different solution if maintainers prefer.

mglourido:
Fixes #5641
LegendaryLibraryManager.refresh() already calls refreshLegendary(), which runs legendary list --third-party and writes the up-to-date assets.json. However, listUpdateableGames() was running that exact same legendary list --third-party command again (plus its own login/online/Epic-offline guards) right before comparing versions.

Since listUpdateableGames() runs as part of the startup library refresh, this meant legendary list was being executed twice on startup — a redundant network round-trip that slowed things down for no benefit.

This PR removes the redundant command and its guards from listUpdateableGames(). The method now just reads the local installed.json and assets.json (already refreshed by refresh()) and compares versions to find updateable games.

- Removed: the second legendary list --third-party call, plus the duplicate isLoggedIn / isOnline / isEpicServiceOffline / abort / error handling.
- Kept: a lightweight isLoggedIn() && isOnline() early-return guard, and the local-file version comparison.

Net result: one fewer legendary list invocation per startup, faster library refresh, identical update-detection behavior.

---
How to test

1. Have at least one installed Epic/Legendary game that currently has an update available (or just one installed game to confirm no false positives).
2. Clear the cache so the refresh path runs fully from scratch:
  - Close Heroic.
  - Remove the legendary cache files so they get regenerated: delete (or back up) assets.json and installed.json's sibling cache under the legendary config dir, e.g. ~/.config/legendary/ (or use Heroic's Settings → Advanced → Clear Cache).
3. Start Heroic with logs enabled and watch the legendary log during startup.
4. In the logs, confirm:
  - Refreshing library... / Refreshing Epic Games... appears and legendary list runs once (no longer a second "Checking for game updates" list call).
  - Game list updated, got N games & DLCs shows the correct game count.
  - Updateable games are still detected — look for Update is available for <appName>: <current> != <latest> and the final updateable-games log line.
5. Confirm the library loads correctly and games show their update badge where applicable — i.e. the startup state updates the same as before, just without the duplicate network call.
6. 
---

### DIFF
--- a/src/backend/storeManagers/legendary/library.ts
+++ b/src/backend/storeManagers/legendary/library.ts
@@ -56,10 +56,6 @@ const fallBackImage = 'fallback'
 const allGames: Set<string> = new Set()
 let installedGames: Map<string, InstalledJsonMetadata> = new Map()
 const library: Map<string, GameInfo> = new Map()
-// Timestamp (ms) of the last successful `legendary list` run; used to skip
-// redundant calls in listUpdateableGames when refresh just ran.
-const LIST_TTL_MS = 30 * 1000
-let lastLegendaryListTimestamp = 0
 
 export default class LegendaryLibraryManager implements LibraryManager {
   async init() {
@@ -110,8 +106,6 @@ export default class LegendaryLibraryManager implements LibraryManager {
 
     if (res.error) {
       logError(['Failed to refresh library:', res.error], LogPrefix.Legendary)
-    } else {
-      lastLegendaryListTimestamp = Date.now()
     }
     this.refreshInstalled()
     return res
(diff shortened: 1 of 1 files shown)


CommandMC:
Debouncing the `list` call is not how you resolve this problem. Instead, look into why the debounce is necessary in the first place (why are we trying to refresh multiple times on startup?)

mglourido:
If I'm not mistaken , refresh() and listUpdateableGames() are not designed to share work with each other. I applied the TTL to avoid having to change how they work internally. So can I restructure how they work so that one detects the data is already up to date without needing the TTL?

mglourido:
There is no code path that calls listUpdateableGames() without a prior refresh() on the same runner.

mglourido:
Startup seems to work (both hot and after deleting assets.json — I think that should be enough coverage, since going offline shouldn't change the behavior). There are no side effects because refresh() is always called first, so the relevant code in listUpdateableGames() can be removed directly. Where I had doubts was with checkGameUpdates — whether there could be issues when multiple stores update at the same time, but I don't think there should be any problems as long as refresh() is always called first (though checkGameUpdates could potentially be refactored — that would require deciding what to do with autoUpdate for example).

PR #5660, opened 2026-06-11
[Perf] parallel load library games

mglourido:
Game library data is now loaded asynchronously, preventing the UI from blocking during startup.

## Performance impact

- **Cold cache**: operations that could take several seconds now complete in milliseconds.
- **Warm cache**: load time is roughly the same, but the UI no longer freezes while data is being fetched.

---


CommandMC:
This is largely a repeat of the second part of my comment here: https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5661#issuecomment-4709756968

1. The UI is not driven by the Backend process, so it using synchronous code does not cause it to freeze
2. If you're gonna rewrite things to be async, at least use actual async APIs ([`fs/promises`](<https://nodejs.org/api/fs.html#promises-api>) in this case)

PR #5661, opened 2026-06-11
[Perf] Wine: parallel download and async checksum verification

mglourido:
Two improvements to the Wine/runtime installation flow:

**Replace duplicated `downloadFile` implementation**

Wine packages were being downloaded entirely into memory before being written to disk. `runtimes.ts` had its own `downloadFile` that handled this, with no retry or abort support. It now uses the existing `downloadFile` from `backend/utils.ts`, which streams directly to disk, supports abort signals, and handles retries. This also results in a significant reduction in code.

**Async checksum verification**

Checksum verification after installation is now performed asynchronously, preventing the UI from freezing. Additionally, `hashFile` is now only called when there is an actual checksum to verify (GE-Proton, Wine-GE). For versions without a checksum (Wine-Staging-macOS and others), the file is skipped entirely.

---

### DIFF
--- a/src/backend/wine/manager/downloader/main.ts
+++ b/src/backend/wine/manager/downloader/main.ts
@@ -302,8 +302,8 @@ async function installVersion({
   })
 
   // Check if download checksum is correct
-  const downloadChecksum = await hashFile(tarFile)
   if (sourceChecksum) {
+    const downloadChecksum = await hashFile(tarFile)
     if (!sourceChecksum.includes(downloadChecksum)) {
       unlinkFile(tarFile)
       throw new Error('Checksum verification failed')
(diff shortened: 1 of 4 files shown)


CommandMC:
> Wine packages were being downloaded entirely into memory before being written to disk. runtimes.ts had its own downloadFile that handled this, with no retry or abort support

First off, this function does not download Wine, it handles extra components (like umu, dxvk, vkd3d, and dxvk-nvapi). Secondly, these runtimes do not need the extra functionality provided by the general function, they are small (<10MB) and installation is not supposed to be cancel-able. I am also not sure whether the Lutris API we pull from even supports parallel downloads (or whether we should use them, considering they place significantly more load on the server)

> Checksum verification after installation is now performed asynchronously, preventing the UI from freezing

1. Electron runs code in multiple processes. The backend not using asynchronous functions does not mean the UI freezes.
2. Checksum verification takes a minuscule amount of time; even if the UI would freeze, it would not be noticeable
3. If you're gonna refactor this into async code, at least use actual [Promise-based `stream` API](<https://nodejs.org/api/stream.html#streams-promises-api>)

PR #5662, opened 2026-06-11
[Fix] several memory leaks corrected

mglourido:
## Summary
Fixes several memory leaks identified in the backend:

aborthandler.ts — for...in over MapIterator (silent bug)

callAllAbortControllers() used for...in over abortControllers.keys(), which iterates the enumerable properties of a MapIterator — none. The loop body never executed, meaning active downloads and background operations were never aborted on exit. Fixed by using for...of.

shortcuts/utils.ts and tools/index.ts — AbortControllers never cleaned up

Several callers created AbortController entries via createAbortController but never called deleteAbortController after the operation completed. These entries accumulated indefinitely in the global abortControllers Map. Fixed by wrapping the operations in a try/finally block that calls deleteAbortController.

progress_bar.ts — duplicate event listeners on pause/resume cycles

Each non-done/non-queued gameStatusUpdate event added a new progressUpdate-{appName} listener without removing the previous one. On cleanup, only one instance was removed. Fixed by calling removeAllListeners before re-registering and on cleanup.

game_config.ts — instances Map never cleared

GameConfig.instances grew indefinitely as games were accessed, with no eviction path. Added evict(appName), called on uninstall to drop a single entry, and evictAll(), called on logout to clear all cached instances across platforms. Evicted instances are reloaded lazily from disk on the next GameConfig.get() call — each config file is at most ~1 KB, so the cost is negligible.

---

### DIFF
--- a/.idea/icon.svg
+++ b/.idea/icon.svg
@@ -1,0 +1,1 @@
+../src/frontend/assets/heroic-icon.svg
--- a/src/backend/__tests__/progress_bar.test.ts
+++ b/src/backend/__tests__/progress_bar.test.ts
@@ -87,16 +87,15 @@ describe('progress_bar', () => {
     })
 
     it('stops listening for progress updates', () => {
-      jest.spyOn(backendEvents, 'off')
+      jest.spyOn(backendEvents, 'removeAllListeners')
 
       sendGameStatusUpdate({
         appName: 'Test',
         status: 'done'
       })
 
-      expect(backendEvents.off).toBeCalledWith(
-        'progressUpdate-Test',
-        expect.any(Function)
+      expect(backendEvents.removeAllListeners).toBeCalledWith(
+        'progressUpdate-Test'
       )
     })
   })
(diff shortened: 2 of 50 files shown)


mglourido:
Fix async error handling in downloadImage

downloadImage was not async and did not await downloadFile, so the try/catch only caught synchronous throws — promise rejections were silently lost.

In steamhelper.ts the return value was checked immediately after the call, meaning error was always undefined even when the download failed. In getIcon, the download was fire-and-forget, returning the icon path before the file existed on disk — a silent race condition for any caller that used the path immediately.

Changes:
- Made downloadImage async and awaited downloadFile
- Replaced the .finally() chain with a try/catch/finally block — deleteAbortController is now guaranteed to run in all cases
- Added await in steamhelper.ts so download errors are correctly pushed to the errors array
- Added await in getIcon so the path is only returned once the file is actually on disk

mglourido:
Ready for review

PR #5666, opened 2026-06-12
[codex] Add Epic account switching

### LINKED ISSUE #5456: A multi-account support
### Problem description

Currently, I own two Epic Games accounts. To play a game hosted on my secondary account, I have to manually log out of the first and re-authenticate with the second, which is a tedious process.

### Feature description

Support for multiple accounts within a single Heroic Games Launcher instance.

### Alternatives

_No response_

### Additional information

_No response_

colonelpanic8:
Fixes #5456.

## Summary

This adds multi-account support for Epic/Legendary without replacing the currently saved account while logging in to another one.

- store Epic accounts separately from Legendary's active profile and track the active account in Heroic config
- add Login-page UI to view saved Epic accounts, switch the active account, add another account, and remove inactive accounts
- run add-account login in an isolated temporary Legendary config, then promote it only after auth succeeds
- preserve and self-heal Legendary install metadata when switching accounts so installed games stay recognized
- add a Nix flake dev environment for setup, start, lint, tests, and the combined check workflow

## Validation

- `corepack pnpm lint` passed with warnings only; no lint errors
- `corepack pnpm exec eslint --cache . --quiet` passed
- `corepack pnpm prettier` passed
- `corepack pnpm find-deadcode` passed
- `corepack pnpm codecheck` passed
- `corepack pnpm test:ci` passed: 22 suites, 142 tests
- `corepack pnpm exec electron-vite build` passed, with existing Rollup/chunk-size warnings
- `nix run path:/home/imalison/Projects/HeroicGamesLauncher#check --extra-experimental-features 'nix-command flakes' --no-update-lock-file` passed
- `git diff --check` and cached diff whitespace checks passed
- Manually added a second Epic account, switched to it, refreshed the library, preserved the previous account, and launched Rocket League successfully under the new active account

## Checklist

Notes on unchecked items: manual validation was done against an existing developer install so we could verify install-metadata preservation and launching an already-installed Rocket League copy. Login, account switching, library refresh, and play were manually tested; install, uninstall, and move were not manually exercised.


flavioislima:

colonelpanic8:

flavioislima:
@colonelpanic8 can you add more details on how to test and also write what could break and what could go wrong? This will speed up our reviews.

colonelpanic8:
> @colonelpanic8 can you add more details on how to test and also write what could break and what could go wrong? This will speed up our reviews.

Sure. I'll try to give this some more thought, but basically, you need to go to the manage accounts option and add a second account and then toggle between things and click around and try to see if anything does not reflect the new account state? afaict things work pretty well.

colonelpanic8:
Thanks — I did a deeper code-review pass, rebased this onto the latest `main`, and pushed an additional hardening commit (`f79bd356`).

### Suggested test plan

It is most useful to test with two Epic accounts that have visibly different libraries/ownership, plus at least one already-installed game and, if possible, one third-party imported install.

1. Start from an existing single-account Heroic profile. Confirm its Epic library and installed games are correct, then restart Heroic once.
2. Open **Manage Accounts**, add account B, and confirm:
   - B becomes active;
   - account A remains listed;
   - the displayed username and library change to B;
   - existing installed games remain detected.
3. Switch A → B → A several times. After each switch:
   - refresh the Epic library;
   - verify the username, owned games, and account-specific library match the selected account;
   - launch an already-installed game;
   - confirm installed and third-party-installed entries remain present.
4. Restart Heroic with each account active and verify that the same account and library come back.
5. Exercise failure paths:
   - try to add an account that is already saved;
   - cancel/interrupt login;
   - try login while offline;
   - if practical, make the config directory temporarily unwritable and verify the current account remains usable.
6. Remove an inactive account and confirm the active account is unaffected. Then remove the active account and confirm it logs out cleanly; if Legendary logout fails, the account should remain active rather than disappearing from the UI.
7. Start a library refresh and then switch accounts. The switch should wait for the existing refresh and then refresh the newly selected account.
8. Verify unrelated state survives account changes: GOG/Amazon login sessions and Heroic UI preferences should not be cleared.

### What could break / highest-risk areas

- The implementation swaps Legendary profile directories, so the biggest risk is losing or partially replacing `user.json`/auth state on filesystem errors.
- `installed.json` and `third-party-installed.json` must remain shared/preserved across switches; losing either can make installed games appear missing.
- A stale library refresh could otherwise repopulate the previous account's games after a switch.
- Epic web cookies need to be cleared without clearing Heroic's default Electron session or other stores' state.
- Duplicate, corrupt, or incomplete saved profiles must never be activated.
- Logout failure must not be reported as successful removal.
- A remaining area worth stress-testing is switching while another long-running Legendary operation (install/update/launch) is already in flight; normal frontend library refreshes are now coordinated, but an unrelated runner process crossing the account boundary is the hardest concurrency case.

### Hardening added during review

The latest commit now:

- stages and validates profile copies before swapping them, with backup/rollback and crash recovery;
- writes install metadata into the staged copy, so metadata-write failure does not leave the active profile half-switched;
- serializes add/switch/remove/logout account mutations and globally disables account controls while one is running;
- waits for an active frontend library refresh before switching/removing;
- validates the saved account ID before activation;
- clears only Epic's persistent web partitions, not Electron's default session;
- propagates failed logout/removal correctly to the renderer;
- recovers each install-metadata file independently from the newest valid JSON copy.

Validation after the rebase/fixes:

- 22 test suites, 143 tests passed;
- focused Legendary account suite: 9 tests passed;
- TypeScript/codecheck passed;
- Prettier and dead-code checks passed;
- ESLint passed with 0 errors (repository warnings only);
- production Electron/Vite build passed with the existing Rollup/chunk warnings.

arielj:
just a comment, I made a draft PR some time ago to support "profiles" instead of support swithing accounts

then one can define the profile by an env variable passed to the heroic command which I personally think is safer than swithing accounts because each profile is isolated, config files are not swapped

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5435

colonelpanic8:
> just a comment, I made a draft PR some time ago to support "profiles" instead of support swithing accounts
> 
> then one can define the profile by an env variable passed to the heroic command which I personally think is safer than swithing accounts because each profile is isolated, config files are not swapped
> 
> #5435

That seems cool and useful, but personally, I'd rather avoid having to restart the app to do my switches. No reason we can't have both features right?

arielj:
> > just a comment, I made a draft PR some time ago to support "profiles" instead of support swithing accounts
> > then one can define the profile by an env variable passed to the heroic command which I personally think is safer than swithing accounts because each profile is isolated, config files are not swapped
> > #5435
> 
> That seems cool and useful, but personally, I'd rather avoid having to restart the app to do my switches. No reason we can't have both features right?

technically, any number of features could be part of heroic, but personally I would not want to maintain 2 features for the same end goal (switch accounts more easiliy)

one thing I'm not sure about this PR is that it's specific to Epic/Legendary, while the profiles feature can be used to login to acounts for any supported store (current and future, since there's no store-specific code)

also, the profiles feature could have a profile selector in the UI and automatically restart heroic if someone doesn't want to restart the app manually (I do think restarting the app is safer to make sure state is properly cleared). I intentionally didn't add that in my draft PR to have more feedback before spending more time, but it can be made more user-friendly for sure

J0shua-code:
updates on this?

colonelpanic8:
Rebased onto current `main` (`9b89d3fc`) and resolved the conflicts. Both commits are preserved; head is now `ce5fa8a4`.

### Conflict resolutions

- **`legendary/constants.ts`**: took the post-Snap-removal `legendaryConfigPath` from `main` and kept the new `legendaryAccountsPath` / `legendaryTempLoginPath` next to it.
- **`legendary/library.ts`**: #5814 removed the block that sets `LEGENDARY_CONFIG_PATH` for every Legendary call (the comment above it still says it should be set, so this looks unintentional on `main`). This PR relies on Legendary using Heroic's `legendaryConfigPath`, and on being able to override it for the isolated add-account login, so I restored it without the old `SNAP` check: the env var is set to `legendaryConfigPath` unless the caller already supplied one. Happy to split that into its own fix if you'd prefer it land separately.
- **`legendary/user.ts`**: logout now goes through `clearEpicWebSession()`, which clears the unified `persist:epic` partition from #5752 (and the legacy `persist:epicstore` one).
- **`WebView/index.tsx`**: kept the unified partition scheme from #5752 as the default, and only use an isolated non-persistent `epic-login-<key>` partition for the add-account flow so the new login doesn't reuse the active account's cookies. Also re-added the `useRef` import that #5707 dropped.

### Other fixes found while rebasing

- The translation keys for the accounts manager were nested under `login.epic`, which is already a plain string in the catalog, so `i18next --fail-on-update` (the pre-push hook) rejected them. Renamed them to `login.epic_accounts.*` and added the new keys (plus `login.error` / `login.error_title`) to `public/locales/en/translation.json`.

### Validation

- `pnpm codecheck`, `pnpm lint` (warnings only), `pnpm prettier`, `pnpm i18n --fail-on-update` all pass
- `jest src/backend/storeManagers/legendary`: 9 tests pass

PR #5671, opened 2026-06-13
[Perf] GOG. Use native du/PowerShell for install size calculation

mglourido:
Replaces the pure-JS recursive file walk used to calculate installed game size with native OS tools where available. Only GOG games, installed.

Why

The previous implementation walked every file in the game directory using Node's fs APIs. For large games with thousands of files this is noticeably slow, blocking the library refresh.

How

    Linux: runs du -sb — reports bytes directly and is orders of magnitude faster than a JS walk.
    macOS: runs du -sk and converts 1K-blocks to bytes.
    Windows (large games only, ≥ 5 GB compressed download size): runs PowerShell Get-ChildItem | Measure-Object -Sum — only triggered for large games where the startup cost of PowerShell is worth it.
    Fallback: if the native command fails or the conditions above aren't met, falls back to the original JS recursive walk (now getPathDiskSizeJS).

The download_size from the GOG manifest is passed as a hint to decide whether to invoke PowerShell on Windows.

---

### DIFF
--- a/src/backend/storeManagers/gog/games.ts
+++ b/src/backend/storeManagers/gog/games.ts
@@ -439,7 +439,10 @@ export default class GOGGameManager implements GameManager {
       return { status: 'error' }
     }
 
-    const sizeOnDisk = await getPathDiskSize(join(path, gameInfo.folder_name))
+    const sizeOnDisk = await getPathDiskSize(
+      join(path, gameInfo.folder_name),
+      installInfo.manifest.download_size
+    )
     const install_path = join(path, gameInfo.folder_name)
 
     const installedData: InstalledInfo = {
@@ -1200,6 +1203,7 @@ export default class GOGGameManager implements GameManager {
     )
     const gameObject = installedArray[gameIndex]
 
+    let downloadSizeHint: number | undefined
     if (gameData.install.platform !== 'linux') {
       const installInfo = await libraryManagerMap['gog'].getInstallInfo(
         appName,
@@ -1221,6 +1225,7 @@ export default class GOGGameManager implements GameManager {
       gameObject.branch = updateOverwrites?.branch
       gameObject.language = overwrittenLanguage
       gameObject.versionEtag = etag
+      downloadSizeHint = installInfo.manifest.download_size
     } else {
       const installerInfo =
         await libraryManagerMap['gog'].getLinuxInstallerInfo(appName)
@@ -1232,7 +1237,10 @@ export default class GOGGameManager implements GameManager {
     if (updateOverwrites?.dlcs) {
       gameObject.installedDLCs = updateOverwrites?.dlcs
     }
-    const sizeOnDisk = await getPathDiskSize(join(gameObject.install_path))
+    const sizeOnDisk = await getPathDiskSize(
+      join(gameObject.install_path),
+      downloadSizeHint
+    )
     gameObject.install_size = getFileSize(sizeOnDisk)
     installedGamesStore.set('installed', installedArray)
     libraryManagerMap['gog'].refreshInstalled()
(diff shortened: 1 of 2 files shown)


CommandMC:
> this is noticeably slow

How have you measured this slowness? Do you have some before/after benchmarks?

mglourido:
# Benchmark: `du -sb` vs JS recursive walk

**Branch:** `perf/native-du-disk-size`  
**Platform:** Linux  
**Rounds per scenario:** 5  
**Measured with:** `performance.now()` inside `getPathDiskSize` (`src/backend/utils.ts`)

## Results

| Scenario                      | Files  | `du -sb` avg | JS walk avg | Speedup |
|-------------------------------|--------|--------------|-------------|---------|
| Indie  (~100 MB,   200 files) |    200 |       1.6 ms |      2.0 ms |    1.3x |
| Medium (~500 MB, 1,000 files) |  1,000 |       1.8 ms |      7.4 ms |    4.1x |
| AAA    (~2 GB,   5,000 files) |  5,000 |       3.9 ms |     34.9 ms |    8.9x |
| AAA+   (~4 GB,  10,000 files) | 10,000 |       7.5 ms |     75.2 ms |   10.0x |

## Why the gap grows with file count

`du -sb` reads filesystem metadata already maintained by the kernel — the cost is roughly constant per directory, not per file.

The JS walk (`getPathDiskSizeJS`) issues one `lstat()` syscall per file, so its cost scales linearly with file count. A typical AAA title ships 5,000–15,000 files, putting it firmly in the **8–10x** range.

## Scope

This optimization applies to **GOG games only**. Legendary (Epic) and Nile (Amazon) store the install size in their own JSON state files written at install time, so they never call `getPathDiskSize` at runtime.

## Disclaimer

I don't have GOG games installed, so I could not run this benchmark against a real game directory. These numbers were produced by a synthetic test (`scripts/benchmark-disk-size.mjs`) that creates directories with the specified file counts and sizes on a tmpfs/ext4 filesystem. The script was written with AI assistance (Claude).

The relative trend — `du` cost staying nearly flat while JS walk scales linearly with file count — reflects real kernel behaviour and is consistent with published literature on `lstat()` syscall overhead. That said, actual numbers on a real game directory (especially on spinning disk or a network mount) will differ. If a maintainer with GOG games installed can run `node scripts/benchmark-disk-size.mjs /path/to/game`, the timing logs added to `getPathDiskSize` will also appear in `~/.local/state/Heroic/logs/heroic.log` after any install or update.

CommandMC:
- `du` caches results, have you accounted for that in your test?
- I do not believe this complexity is required for an absolute time difference of <100ms

mglourido:
> * `du` caches results, have you accounted for that in your test?

It only ran once, I don't think there were any caches involved, but it's true that I didn't take that into account.Although if ```du``` use caches then that's also a point in its favor, right?

Based on the test data, it's not a drastic change, but if someone installs 125GB games, it would be noticeable, and this is per game installed, not all of them in total.

mglourido:
Would you accept a reduced version that only uses `du -sb` on Linux, without the Windows threshold or the `hintBytes` parameter? The implementation would be minimal, and the fallback to `walk JS` would be maintained if `du` fails.

CommandMC:
> Based on the test data, it's not a drastic change, but if someone installs 125GB games, it would be noticeable, and this is per game installed, not all of them in total.

The function runs on install/update. If the user is installing/updating a game this large, an extra second to compute the install size would not be noticeable

Two can play the extrapolation game. `du` execution time seems to grow slightly exponentially. Using a 2nd order polynomial fitted to your provided data, extrapolating for 125GB (250000 files), we can expect a runtime of ~1.6s. The JS walk on the other hand seems to grow linearly. Plugging 250000 into a linear regression yields ~1.8s.

mglourido:
Makes sense — large games tend to scale in file size rather than file count, so the JS walk doesn't become a real bottleneck there. Closing this; the implementation is here if it ever becomes relevant.

PR #5672, opened 2026-06-13
[Perf] Avoid unnecessary network refresh after Legendary operations

mglourido:
- Adds a `refreshLocal()` method to `LegendaryLibraryManager` that refreshes the library reading only local files, skipping the `legendary list` network call.
- Adds a `localOnly` option to `RefreshOptions`, `LibraryManager`, IPC types, and `GlobalState.refreshLibrary()`.
- After any Legendary operation completes or errors (install, uninstall, repair), the library refresh now uses `localOnly: true`. Game install state is always written to local files by Legendary before signaling completion, so the network call was redundant. Updates are excluded from this — they still trigger a full network refresh.

## Problem

After any Legendary operation finished, `refreshLibrary` triggered a full refresh which called `legendary list` against the Epic Games API. This was unnecessary: install state lives in `installed.json` and game metadata in `legendaryMetadata/*.json`, both written locally by Legendary before it signals completion. The network call only serves to sync the account's game catalogue, which doesn't change as a result of local operations.

## Solution

`refreshLocal()` calls `loadGamesInAccount()`, `refreshInstalled()`, and `loadAll()` — all purely local file reads — then saves the result to the library store. The `localOnly` flag is optional on `LibraryManager` so other store managers (GOG, Amazon) are unaffected. Game updates still go through the full network refresh since `checkForUpdates` requires up-to-date remote data.

---

## How to test
1. Build and run the app from this branch (yarn start / pnpm start).
2. Have an Epic (Legendary) account logged in with at least one installed game available.

Main scenario — Legendary uses local-only refresh
1. Install (or update) an Epic game.
2. When it finishes, watch the logs. You should see:
  - Refreshing library locally...
  - Game list updated locally, got X games & DLCs
3. Confirm there is no full legendary list/network recall after the install completes (no Game list updated, got ... line from the full refresh() path, and no extra network round-trip to Epic).
4. The newly installed game should still appear correctly in the library (installed state, size, etc.), since refreshInstalled + loadAll still run from local cache.

Regression checks
1. Other stores still do a full refresh: install/update a GOG, Amazon, or Zoom game and confirm the post-install refresh still goes through the normal refresh() path (localOnly is false for non-Legendary runners).
2. Manual full refresh still works: trigger a manual library refresh (refresh button) and confirm Legendary still performs the full network refresh and logs Game list updated, got X games & DLCs.
3. First-load cache population unaffected: with empty caches, GOG/Zoom/Nile still call refreshLibrary({ library }) and populate correctly (no localOnly).
4. Updating-game path: finish an update (not just install) of a Legendary game and confirm the game is removed from the updates list and the library reflects the new state.

What to look for
- Noticeably faster return-to-idle after installing a Legendary game (no redundant full library recall).
- No regressions in installed-state detection or DLC listing.

---

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -907,21 +907,18 @@ if (existsSync(legendaryInstalled)) {
   })
 }
 
-addHandler('refreshLibrary', async (e, library?, localOnly?) => {
-  if (library !== undefined && library !== 'all') {
-    const manager = libraryManagerMap[library] as LibraryManager
-    if (localOnly && manager.refreshLocal) {
-      await manager.refreshLocal()
-    } else {
-      await manager.refresh()
-    }
-  } else {
-    const allRefreshPromises = []
-    for (const manager of Object.values(libraryManagerMap)) {
-      allRefreshPromises.push(manager.refresh())
-    }
-    await Promise.allSettled(allRefreshPromises)
+addHandler('refreshLibrary', async (e, library, localOnly) => {
+  async function doRefresh(manager: LibraryManager) {
+    if (localOnly && manager.refreshLocal) return manager.refreshLocal()
+    return manager.refresh()
   }
+
+  const managers =
+    library !== undefined && library !== 'all'
+      ? [libraryManagerMap[library]]
+      : Object.values(libraryManagerMap)
+
+  await Promise.allSettled(managers.map(doRefresh))
 })
 
 // get pid/tid on launch and inject
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -104,6 +104,7 @@ import {
   initStoreManagers,
   libraryManagerMap
 } from './storeManagers'
+import type { LibraryManager } from 'common/types/game_manager'
 import {
   setGameOverrides,
   getGameOverrides,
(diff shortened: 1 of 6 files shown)


CommandMC:
`localOnly` does not work when a `library` is `undefined`/`'all'`. It isn't a problem for now, but may cause issues in the future. This patch fixes that:
[0001-Fix-Respect-localOnly-if-library-is-undefined-all.patch](https://github.com/user-attachments/files/28961950/0001-Fix-Respect-localOnly-if-library-is-undefined-all.patch)

mglourido:
Applied your patch with a couple of changes:

- Switched `Promise.all()` to `Promise.allSettled()` — with `all`, if any library refresh fails, the rest keep running untracked in the background with no way to handle their completion.
- Also took the opportunity to slim down the main function a bit while I was at it.

---

One thing I noticed in the original fix: `refreshLocal()` is essentially a copy of `refresh()`. Rather than having `refresh()` call `refreshLocal()` (which felt semantically odd), I extracted the shared logic into `applyLocalData()`. Now `refreshLocal()` only calls that, and `refresh()` calls it too after updating the data — keeps the flow clear without the awkward dependency direction.

---

Since `refreshLibrary` was picking up two new optional parameters, I also took the chance to refactor its signature to use an options object.

mglourido:
Ran e2e tests locally and all passed. Ready for review

mglourido:
`loadAll` is too generic — a new reader has to look at the body (or the JSDoc) to figure out that it iterates over all games and loads their metadata from disk into memory.

Suggestion: something like `loadAllMetadataFiles` or `loadLibraryFromDisk` makes the domain and the data source obvious without needing extra documentation.

PR #5676, opened 2026-06-14
Use `Game`s instead of manually passing around `appName` & `runner`

CommandMC:
This builds upon the work in 806091888af779152877cd380052b2dabfe22def and avoids passing around `appName` & `runner`, instead using `Game` directly .

For the Frontend, as outlined in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/4894#discussion_r2292203213, the IPC system will automatically convert a `Game` sent by the Backend to a `GameHandle` (a thin container object, wrapping an `id` and a `runner`). On the way back (Frontend -> Backend), the handle is converted back to its game.

No actual functionality is changed, although there is at least one bug I've noticed and had to fix: The `Runner` type is now built from the schema of the same name, and that schema did not include `zoom` yet. In turn, Zoom games could not be launched via the protocol.

This shouldn't really be as big of a change as it looks (the argument types are typechecked, and for the IPC system changes, it'd be pretty obvious if something wasn't working right), so testing shouldn't be that bad.

---


Drackrath:
I like this PR, needs to be merged before any other new game Launcher Library is being added.

arielj:
I want to review this, but it's 129 files 😭 

I get that they are all related changes, but I think it would make it way easier to review/test if different actions are refactored in different PRs?

like... one refactoring the actions related to gog, one for download related actions, etc?

maybe it's too much work, I just think it would make it easier to get all these changes merged, because I understand there's no behavior changes but it's kinda impossible to test all those changes, and I also understand most changes are similar but they are a lot of different changes (even though repeated)

Drackrath:
Yea I'd also like to have this PR merged. Though I am familiar with the changes done, I don't think that splitting this does much to lower the general workload, it'd instead increase it.

Maybe pinpointing the identical changes in a graphical view would help in a review?

I tested it through my PR's related to this, so I can somewhat confirm that bugs aren't really introduced here. It's just a major refactor.

PR #5682, opened 2026-06-16
Make windows related files openable within just a double click

Twig6943:
~~This is work in progress the diff might be a bit rough.~~ Everything should be good but if something isn't let me know.

Showcase:
Screenshot_2026-06-16_at_4 19 41_PM

Screenshot_2026-06-16_at_4 20 02_PM_1

image

image

This pr adds the ability to open .exe/.msi/.bat files within just a double click. Tested on linux & mac and works fine. ~~I don't know how to do the registiration of file extensions on macos so I'll have to see about that.~~ Turns out electron builder seems to handle that.

~~Disclaimer: AI was used a bit in `exe_handler.ts` ~~ basically got rewritten
- [x] Created / Updated documentation (If necessary) (Don't think anything needs to be added as it's very straightforward)

For those on linux that'd like to test this just use am --launcher to add in the appimage thats in `dist/` after compiling


src/backend/exe_handler.ts
```diff
@@ -0,0 +1,107 @@
...
+  pendingPickerData = null
+  return data
+}
+
+function getSideloadGames(): GameInfo[] {
+  const games: GameInfo[] = libraryStore.get('games', [])
+  return games.filter(
+    (g) => g.runner === 'sideload' && g.title && !g.browserUrl
+  )
+}
```
CommandMC: - Why limit this to sideloaded games?
- This currently does not check the platform of the game. I assume you'd also want to exclude native games (since they don't have a Wineprefix)

src/backend/main.ts
```diff
@@ -615,6 +634,17 @@ app.on('open-url', (event, url) => {
   }
 })
 
+app.on('open-file', (event, filePath) => {
+  event.preventDefault()
```
CommandMC: You probably want to move this into the `if` block to not block Electron's default handler for this event (not sure if there *is* a default handler in this case, but it's always a good idea to only "consume" the event if you actually do something with it)

CommandMC (commented):
- I would prefer if the state handling would be done with a Zustand store (so the Backend can just push the info as soon as it has it, instead of having to wait for the Frontend). Happy to help out there if you're unfamiliar with how to do that.
- Checking if the path starts with `/run/user` may be a good idea. This will happen when the path is inaccessible in the Flatpak sandbox, and will cause *only* the selected file to be visible to Heroic. If the file relies on other resources alongside it, this will cause problems.

Twig6943:
> * I would prefer if the state handling would be done with a Zustand store (so the Backend can just push the info as soon as it has it, instead of having to wait for the Frontend). Happy to help out there if you're unfamiliar with how to do that.

Yea help on that would be appreciated


> * Checking if the path starts with `/run/user` may be a good idea. This will happen when the path is inaccessible in the Flatpak sandbox, and will cause _only_ the selected file to be visible to Heroic. If the file relies on other resources alongside it, this will cause problems.

Added

CommandMC:
I've refactored this to fix a few things. I originally only wanted to move to GlobalStateV2, but kept noticing small things, and now we're here.

1. Used `GlobalStateV2` for state handling
2. Used MUI for everything (that's just my preference though)
3. Changed the dialog title from "Select Wine Prefix" to "Open EXE"
4. Added a description
5. Non-native and non-installed games are not shown anymore
6. Due to (5), the component needs some time to know which games are installed & non-native. While filtering is happening, the dialog will load:
    grafik
7. If the EXE path is not accessible in Flatpak, the dialog will show a warning:
    grafik
8. While the EXE is running, the dialog will show it. It will automatically close once it is done:
    grafik
9. If a relative path is provided, it will now be resolved properly

Issue #4168, opened 2024-12-11
Disable `hidraw` (make Dualshock/DualSense act like Xbox 360 controllers/XInput)


Anuskuss:
### Problem description

There are a few games that only support Xbox 360 controllers (e.g. Rise of the Tomb Raider) and in the past I've used the wrapper functionality to start [ds360](https://github.com/yoyossef/ds360) (which in turn runs [xboxdrv](https://github.com/xboxdrv/xboxdrv) `--mimic-xpad`) which has always been rather ugly and something that I would've expected from my frontend to be able to do.
Turns out that Wine/Proton is already capable of emulating `XInput` but only if you first [disable `hidraw`](https://wiki.archlinux.org/title/Gamepad#Using_with_Wine). I wouldn't expect the average user to know this (I've being using Wine/Proton + Heroic/Lutris for years and didn't know this) and it's usefulness to be great enough that Heroic should handle this for you.

### Feature description

Add a checkbox in the game config which allows you to **Disable hidraw** / **Force XInput**. Doing it manually via `regedit` is prone to errors (e.g. you might forget that you even set it) and not user friendly (if you even know that it exists to begin with).

### Additional information

This is also needed for the Silent Hill 2 remake where you're unable to open your inventory while using a DualSense (touchpad is acting like left mouse click, SHARE and PSBUTTON are unmappable).

Anuskuss:
I'd also love to see more Registry tweaks being exposed in Heroic like e.g. [Wine Wayland](https://wiki.archlinux.org/title/Wine#Wayland) instead of the user having to do that manually.

mbtns:
I support his request, would also reduce the need for the Steam dependence just for controller support cf.  ["Add to Steam" issue on non-Steam Deck Linux systems such as mine.](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/3653).

Anuskuss:
Btw both of my suggestions can now be handled by [proton-ge](https://github.com/GloriousEggroll/proton-ge-custom) so I have reverted my registry edits and have since been using `PROTON_ENABLE_WAYLAND=1` and `PROTON_DISABLE_HIDRAW=1`.
Still I think it makes sense for Heroic to offer those (and many more generally useful tweaks).

ItsLaguna:
[...]

PR #5687, opened 2026-06-19
[Feat] add disableHidraw settings for PlayStation controller support

fschaupp:
Implements https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4168

Adds a new setting to disable hidraw and enable SDL for PlayStation 4/5 controllers, making them work as XInput devices. This improves controller detection in games that expect Xbox 360 controllers.

The setting is available on Linux and macOS (non-native games) and configures Wine registry keys accordingly during game launch.

---


flavioislima:

fschaupp:

CommandMC:
If you are on Linux, modern Proton versions have env vars for this purpose (`PROTON_{ENABLE,DISABLE}_HIDRAW` and `PROTON_USE_SDL`/`PROTON_PREFER_SDL`). If you are on macOS, controller support is a whole other can of worms (and a simple registry entry like this likely does not help)

I would prefer not adding another checkbox to the settings for an option that only helps a small amount of people in an unlikely situation. If you really want to have this option, you could always set it globally (either using Heroic's global env vars, or some other mechanism)

fschaupp:
It's just a bit of a double edged sword. I experienced HGL as a tool for the average gamers. If I can make that by telling my friends, go there and tick in the advanced section the box to make it work with your controller, I think thats reasonable easy. If I need to tell them to put some environment variables somewhere, I'm not so shure anymore - especially when they come from windows, where that is very not common. 

To move on with that, I could recreate the feature using the PROTON_ environment variables, as the WINE route of playing games will get more rare I guess.

CommandMC:
My problem here is that this is very much a "Wine domain"-specific fix. Heroic should not know this much about Proton/Wine IMO (it already knows *way too much*), its concern should be with implementing storefronts

To play on your argument a bit, if a novice user opens the settings and sees a million checkboxes they don't understand, would that not put them off Linux as well?

I'll talk to some of the Proton folks & check whether they have suggestions on how we can make this a smoother experience

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
```diff
@@ -228,6 +229,7 @@ export default function GamesSettings() {
         )}
         <VerboseLogs />
         <DisableUMU />
+        {isLinux && <DisableHidraw />}
```
BananaWorks07: this is not linux specific (also will work on mac). However, on mac the toggle will just set reg keys to apply something that are already set. The same is true for upstream wine, so you are better off using PROTON_DISABLE_HIDRAW=1 (for upstream proton) and PROTON_USE_SDL=1 (custom proton). They can be used together without issues.
edit: there are a handful of controllers on the prefer hidraw list so you need the registry path for mac and linux on wine

src/frontend/screens/Settings/components/DisableHidraw.tsx
```diff
@@ -0,0 +1,45 @@
...
+  if ((isLinux && isLinuxNative) || (isMac && isMacNative)) {
+    return <></>
+  }
+
+  const [disableHidraw, setDisableHidraw] = useSetting('disableHidraw', false)
+
+  return (
+    <div className="toggleRow">
+      <ToggleSwitch
+        htmlId="disableHidrawToggle"
```
BananaWorks07, line 26: I would prefer if the setting was called use SDL gamepad input or something like that

fschaupp:
> My problem here is that this is very much a "Wine domain"-specific fix. Heroic should not know this much about Proton/Wine IMO (it already knows _way too much_), its concern should be with implementing storefronts
> 
> To play on your argument a bit, if a novice user opens the settings and sees a million checkboxes they don't understand, would that not put them off Linux as well?

I see where you come from. I used Lutris quite a lot too where basically every unholy available option has it's checkbox - at least in the advaced view. That's very overwhelmig and a good reason to not do that in HGL.

Adding the option in the menu wouldn't change the fact, that someone needed to start the game (successfully - maybe tinkering to get to that point), only to notice, the beloved gamecontroller does not work.
So...
> 
> I'll talk to some of the Proton folks & check whether they have suggestions on how we can make this a smoother experience

...would be the absolute best option! 🥳 Maybe detecting the attached controllers/last used ones in UMU or Proton or so..

Despite all of this: there are now, today, people attempting it and failing (see the ones commented on the issue, or at least being subscribed to it, + the ones just finding it). This is an attempt to at least make it easier resolvable until then.

---
Nonetheless, it was a first attempt at fixing it and contributing to HGL. I really love where it is now and the direction it's going - independent of whether the feature lands or not, so declining it is fine for me too - no hard feelings 😉

Edit: my comment was put into the quote markdown

src/frontend/screens/Settings/components/DisableHidraw.tsx
fschaupp: oh, good idea ^^. so basically rewording that and the description to gamepad instead of controller?
hm.. that would maybe make users looking for a possible fix by just trying out options discover it more likely.

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
fschaupp: > edit: there are a handful of controllers on the prefer hidraw list so you need the registry path for mac and linux on wine

So if I understood that correctly, on MAC its always enabled using the registry keys, right?
Meaning, if I keep the isLinux gate for the checkbox and migrate the functionality to the Proton-flags instead, it should work out for both sides of the Force?
May the Force be with you :3

flavioislima (approved):
I think its fine to have this under the Advanced Settings, its like the name says "Advanced", so it is for small edge cases in general.

Thanks for the contribution

src/frontend/screens/Settings/sections/GamesSettings/index.tsx
BananaWorks07: No, you need to set the reg keys on mac as well

fschaupp: alright, i removed the rendering barrier then

imLinguin (changes requested):
Please move this to Advanced Settings

fschaupp:
Hi @imLinguin Pawel!
Isn't it already in the Advanced section?
grafik

src/backend/launcher.ts
```diff
@@ -999,6 +999,86 @@ async function prepareWineLaunch(
     await download('battleye_runtime')
   }
 
+  // Set or remove DisableHidraw and Enable SDL registry keys based on setting
+  if (gameSettings.disableHidraw) {
```
BananaWorks07, line 1003: As I said before we should not use registry keys on Proton. If you set the registry key, the user cannot override it using the environment variable. Therefore, the environment variable should be used with Proton.

BananaWorks07 (changes requested):

src/backend/launcher.ts
fschaupp: but you said, 
> edit: there are a handful of controllers on the prefer hidraw list so you need the registry path for mac and linux on wine

and 
> No, you need to set the reg keys on mac as well

So, do you want to have the settings to apply always the PROTON_ environment settings instead, but meant, I should keep the checkbox enable for macs too?

imLinguin:
@fschaupp you are right, it is already there. huh... not sure how I drew that conclusion

src/backend/launcher.ts
BananaWorks07: no, proton is not the same as wine. if the wine version type is a proton based runner use PROTON_DISABLE_HIDRAW (and remove the registry keys if present). If the runner is wine then use the registry keys. I hope that makes sense

fschaupp: alright, now i get finally what you actually want the endresult to be. i'm very aware of the difference between proton and wine. it just did not yet come through what you want me to change.
now that being sorted out, i'll have a look into that maybe tomorrow.

Bebop182:
It is not present on my end on heroic 2.22.1 and latest proton GE runner

CommandMC:
This PR is not merged yet, so its changes will not appear in a standard installation of Heroic

Arcitec:
Editing the registry key is the universal way of solving this for all versions of Wine/Proton.

I am not a fan of the PROTON-specific env variables. They won't work in all cases.

PR #5705, opened 2026-06-23
Make Proton-CachyOS the default Proton version

CommandMC:
This PR makes Proton-CachyOS the default version shown when opening the Wine Manager and the default when Heroic auto-downloads a Wine version if none are found.  It also adds a logo for it (to be shown in the Wine Version selector).

Latest GE-Proton has issues with launching the EA app (and I believe Ubisoft Connect too?). I understand GE is still working on getting video playback just right, so a new release may take some more time (and there's no guarantee that those issues are fixed then). Defaulting to a version with shortcomings like that seems irresponsible.

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -949,7 +949,7 @@ export async function downloadDefaultWine() {
   const isMacOSUpToDate = await isMacSonomaOrHigher()
   const release = availableWine.find((version) => {
     if (isLinux) {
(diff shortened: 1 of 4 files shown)


flavioislima (approved):

CommandMC:
> so a new release may take some more time

Welp, this aged well
https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/GE-Proton11-1

PR #5712, opened 2026-06-24
Move `sanitize-html` import into the backend

CommandMC:
To quote [their README](https://github.com/apostrophecms/apostrophe/tree/main/packages/sanitize-html#requirements)

> sanitize-html is intended for use with Node.js

Importing it in the Frontend lead to a lot of warnings about unavailable modules. It still *worked*, but it might not continue to work, and those warnings were just rather annoying

The Frontend part of this is a little messy. I would've loved to use `useAwaited` here, but that'd need the same changes I made in 0bd1edd9362f666cdcf03723acc34aed16ec281c, and I didn't want to unleash that merge conflict onto future me

---

To test:

Open Heroic with the dev console enabled. On `main`: Observe a lot of warnings about modules being externalized. On this branch: Observe the warnings no longer appearing.

Open the changelog for any GOG game. Observe the changelog still appearing like before (that was the one use we had of this dependency).

---

### DIFF
--- a/src/backend/utils/ipc_handler.ts
+++ b/src/backend/utils/ipc_handler.ts
@@ -11,6 +11,7 @@ import {
 import { hasExecutable } from './os/path'
 import { formatSystemInfo, getSystemInfo } from './systeminfo'
 import { isIntelMac } from 'backend/constants/environment'
+import sanitizeHtml from 'sanitize-html'
 
 addListener('abort', (event, id) => {
   callAbortController(id)
@@ -32,3 +33,12 @@ addHandler('hasExecutable', async (event, executable) => {
 addHandler('isIntelMac', () => {
   return isIntelMac
 })
+
+addHandler('sanitizeHtml', (_e, input, options) =>
+  sanitizeHtml(
+    input,
+    options ?? {
+      disallowedTagsMode: 'discard'
+    }
+  )
+)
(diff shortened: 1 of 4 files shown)


arielj (commented):
I wonder if it would be better to add a "santiziedChangelog" property in the ExtraInfo directly from the backend, so then we don't have to call the backend again to sanitize it? (then we wouldn't need the handler, awaiting, etc, it would already be a sanitized string prop)

CommandMC:
I've made that change in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5692/changes/3f560c17618cf4e2fbfdc0ae6bf7e010e2e1bf3f now. I could also make it here, but since that PR is removing ExtraInfo, that'd just cause an unnecessary merge conflict for future me

arielj:
> I've made that change in [3f560c1](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/commit/3f560c17618cf4e2fbfdc0ae6bf7e010e2e1bf3f) now. I could also make it here, but since that PR is removing ExtraInfo, that'd just cause an unnecessary merge conflict for future me

should this PR be closed? they are both going to conflict anyway (I guess this one could be merged first, but it's not that it's something critical and really needs to be merged sooner)

CommandMC:
Well, the conflict that would happen now would be easy to resolve (it just affects one file). I'm fine with this PR getting closed, or with it getting merged now & me fixing the conflict

PR #5725, opened 2026-06-30
[UX] Library games :  New Animation hover gamecard + accessibility settings

mglourido:
## Overview

This PR refreshes the Library game card hover interaction to make it feel more polished while addressing the temporary blur caused by scaling the entire card too aggressively.
It also improves the resolution and pixelation of the components and images.

It's not a polish of the current one, but a completely new animation, which involves changes beyond just CSS.

https://github.com/user-attachments/assets/b252da15-b26a-4832-a1e2-364cad620df3

## Main Changes

- Reworked the game card hover animation to combine subtle card elevation, controlled scaling, improved shadow depth, and internal cover zoom.
- Increased Legendary/Epic cover image request size from `300x400` to `600x800` so cover art has enough resolution for hover zoom without looking blurry.
- Added internal cover zoom while keeping the image clipped inside the card frame.
- Improved card layout so the cover area and action footer stay visually connected without unwanted gaps.
- Refined title rendering so text remains sharp while the card scales, using a larger rendered font size with transform compensation.
- Limited title wrapping to three lines to avoid unstable word jumps during hover.
- Removed duplicated game names from action button tooltips because the title is already visible on hover.
- Standardized action button and SVG sizes with integer pixel values to reduce subpixel blur.
-A resin/varnish effect was added, which visually enhances the image and adds a visually pleasing effect.

## UX Impact

- Hover feedback feels more premium and responsive.
- Cover art remains sharper during animation.
- Text readability is improved without reducing configured font size.
- Action buttons feel more compact and aligned with the card.
- The blur caused by image scaling is completely eliminated

## Accessibility Notes

- The title font size is not reduced, preserving user font-size preferences.
- Button tooltips still describe the action, but no longer repeat the game title.
- Keyboard/gamepad focus continues to share the same visual treatment as hover through `:focus-within`. Compatibility is not affected.

## Accessibility Settings

- (legendary) An option has been added that allows users to choose the resolution for the game covers (low (a new resolution to reduce resource consumption), medium (current resolution), high (new resolution)). This option is necessary?  Doing this does not implement complex logic and has no repercussions. By default, the original resolution will be retained, giving the option to change it.The current resolution is the fastest for the servers because it's the most used. The higher resolution isn't much slower, but it consumes more resources. That's why I'm taking the opportunity to allow the option to adjust the resolution. This allows the user to use the fastest resolution, the one that consumes the fewest resources or looks the best.

- An option has been added to remove the game store logo from the game cover.

- An option has been added to reduce the animations in this new animation.

- An option has been added to remove the animation while maintaining essential elements (displaying the game title and providing color for the game).

---


mglourido:
I'd like feedback on the animation; I'm not sure if I should give more feedback on making it bigger or if keeping it subtle is fine.

mglourido:
## Things that are not currently included:

-Accessibility settings

-Improved accuracy regarding the required resolution for homepages

-Technical cleanup of the CSS code and its classes

mglourido:
One idea is to use the colors from the game covers instead of dark shadows. It creates a more vibrant effect and doesn't disrupt the new animation.

I didn't implement it, but I did a quick AI-generated test to see how it would look.

https://github.com/user-attachments/assets/5353f008-a8be-4d3b-8954-0cb6b483bd02

PR #5731, opened 2026-07-04
 [Fix] patch extract-zip dependency to mitigate CVE-2026-56876 (Symlink Path Traversal)

Jonatas-Goncalves:
This PR introduces a patch to the `extract-zip` dependency to address **CVE-2026-56876**, an arbitrary file write/read vulnerability caused by unvalidated symlink targets during zip extraction.

Upon investigating the upstream extract-zip repository, I noticed it hasn't received any updates in about 6 years and appears to be abandoned. Since an official upstream release fixing this CVE is highly unlikely, applying a local pnpm patch is the safest and most immediate solution to protect Heroic's users.

### The Issue
Currently, `extract-zip` does not validate the target path of symlinks inside zip archives. If a malicious zip file contains a symlink with a relative path (e.g., `../../../../etc/passwd`), `extract-zip` will extract it without validation, allowing the symlink to point outside the intended extraction directory. 

### The Solution
This patch intercepts the symlink creation process and resolves the absolute path of the symlink target. It then verifies if the resolved target stays within the boundaries of the intended extraction directory (`this.opts.dir`). If it attempts to escape the directory, the extraction throws an `Out of bound symlink target` error, successfully blocking the path traversal exploit.

### Security Impact
Since Heroic Games Launcher handles zip extractions (for tools, updates, or game components), mitigating this vulnerability at the dependency level ensures that malicious archives cannot compromise the user's system via directory traversal.

---

### DIFF
--- a/patches/extract-zip@2.0.1.patch
+++ b/patches/extract-zip@2.0.1.patch
@@ -1,0 +1,31 @@
+diff --git a/index.js b/index.js
+index 23384ea16bb848616886f356d8d69f94325535bf..91c4a60e341b5ffb9793f941f9f86c5d303d76d8 100644
+--- a/index.js
++++ b/index.js
+@@ -125,10 +125,22 @@ class Extractor {
+     const readStream = await promisify(this.zipfile.openReadStream.bind(this.zipfile))(entry)
+ 
+     if (symlink) {
+-      const link = await getStream(readStream)
+-      debug('creating symlink', link, dest)
+-      await fs.symlink(link, dest)
+-    } else {
++      let link = await getStream(readStream)
++      link = link.toString().trim()
++
++      // --- CVE-2026-56876 Fix ---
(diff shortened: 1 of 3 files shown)


flavioislima:

Jonatas-Goncalves:

CommandMC (changes requested):
Heroic does not depend on `extract-zip` directly, so fixing this here does not make sense. We only have it because Electron needs it. Electron only uses it at build time, and does not pass untrusted paths to the function, so the CVE does not apply here.

Jonatas-Goncalves:
Thanks for checking this and for the clear explanation. You're absolutely right — since extract-zip is only a build-time dependency for Electron and doesn't impact the runtime environment or process untrusted user input here, this CVE is a false positive for Heroic.

Closing this PR now. Appreciate your time!

PR #5736, opened 2026-07-06
[Tech] Point Plausible analytics to the self-hosted instance

flavioislima:
## What

Moves Plausible Analytics from the public `plausible.io` cloud to the Heroic team's **self-hosted** instance at `analytics.heroicgameslauncher.com`.

## Changes

- **Self-hosted endpoint** — events now POST to `https://analytics.heroicgameslauncher.com/api/event`.
- **Native OS detection** — instead of sending the operating system as a custom property, we now send a browser-like `User-Agent` so Plausible parses the OS natively into its built-in "Operating System" report. Country is likewise derived natively from the request IP (and discarded/anonymized by Plausible).
- **Custom props kept only where necessary** — `arch`, `distro` and `distroVersion` remain custom properties because Plausible has no native dimension for them. All other system info (stores connected, packaging format, Steam Deck, app version) is unchanged.
- **Consent dialog** — updated to state that Plausible is self-hosted by the Heroic team.

## Notes

The integration remains 100% backend-side and opt-in; no browser script tag is used. The site domain stays `heroic-games-client.com` for data continuity.

### DIFF
--- a/public/locales/en/translation.json
+++ b/public/locales/en/translation.json
@@ -24,7 +24,7 @@
         "enable": "Enable",
         "info": {
             "pt1": "In order to improve the app, Heroic collects 100% anonymous data.",
-            "pt2": "Heroic uses the open-source Plausible Analytics platform to gather basic data: App Version, OS, Stores Connected and Country.",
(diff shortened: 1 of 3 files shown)


flavioislima:
Works!
image

PR #5743, opened 2026-07-09
Hide `ProgressDialog`'s indeterminate progress bar if it does not make sense

CommandMC:
ProgressDialog used to have a fixed `LinearProgress` at its end. This led to a few confusing situations, since there was not necessarily something running in the background if it was shown (making the user think something is happening even though it is not).

This component was used in 3 places:
- the Epic & GOG save sync screen, to show the output of Legendary/GOGDL
- the Winetricks dialog

For the save sync screens, the dialog was only shown after the tool already quit, so showing a progress bar did not make sense at all there. For Winetricks, the progress is now only shown while we're loading available verbs, installing one, or the GUI is open

---

### DIFF
--- a/src/frontend/components/UI/Winetricks/index.tsx
+++ b/src/frontend/components/UI/Winetricks/index.tsx
@@ -111,7 +111,7 @@ export default function Winetricks({ onClose, runner }: Props) {
         appName,
         runner
       })
-      .then(() => setGuiOpen(false))
+      .finally(() => setGuiOpen(false))
   }
 
   const dialogContent = (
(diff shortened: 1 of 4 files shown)


src/frontend/components/UI/Winetricks/index.tsx
```diff
@@ -102,12 +102,16 @@ export default function Winetricks({ onClose, runner }: Props) {
...
-      runner
-    })
+    setGuiOpen(true)
+    window.api
+      .callTool({
+        tool: 'winetricks',
+        appName,
+        runner
+      })
+      .then(() => setGuiOpen(false))
```
arielj:
```suggestion
      .finally(() => setGuiOpen(false))
```

in case something fails when calling the tool and `then` does not trigger

arielj (approved):

PR #5747, opened 2026-07-14
[Fix] Correct reversed A/B and X/Y buttons on Nintendo controllers

flavioislima:
### Problem
On Nintendo Switch Pro Controllers and Joy-Cons, the A/B and X/Y button actions were reversed from what the UI shows: the glyphs indicate **A = select, B = back**, but pressing **A went back** and **B selected**.

### Cause
Nintendo controllers place their face buttons in swapped physical positions compared to Xbox, and Chromium reports them with the "standard" mapping by physical position:

- `buttons[0]` = bottom button → labeled **B** on a Switch
- `buttons[1]` = right button → labeled **A**

Two input paths assumed the Xbox convention (`buttons[0]` = main action, `buttons[1]` = back), so physical **A** fired *back* and **B** fired *select*.

### Fix
- Add a dedicated `checkSwitchPro` layout that swaps the face buttons (and X/Y), routed for Nintendo's vendor id `057e`. GameCube adapters (`057e:0337`) are still handled by the earlier GameCube branch, and Steam-Input-masked controllers (`28de`) are unaffected.
- Make Console Mode's confirm/back button indices layout-aware via `getActionButtonIndex` / `getBackButtonIndex`, so the overlays (ConfirmDialog, InstallOverlay, LaunchOverlay) match the on-screen glyphs.

Pressing the physical **A** button now selects and **B** goes back, matching the labels shown in the UI.

### DIFF
--- a/src/frontend/helpers/gamepad.ts
+++ b/src/frontend/helpers/gamepad.ts
@@ -8,7 +8,8 @@ import {
   checkPS3Clone1,
   checkStandard,
   checkN64Clone1,
-  checkGenius1
+  checkGenius1,
+  checkSwitchPro
 } from './gamepad_layouts'
 import { VirtualKeyboardController } from './virtualKeyboard'
 
@@ -493,6 +494,8 @@ export const initGamepad = () => {
           checkN64Clone1(buttons, axes, index, checkAction)
         } else if (controller.id.match(/0583.*a009/i)) {
           checkGenius1(buttons, axes, index, checkAction)
+        } else if (controller.id.match(/057e/i)) {
+          checkSwitchPro(buttons, axes, index, checkAction)
         } else {
           // if not specific, fallback to the standard layout, seems
           // to be the most common for now and if not exact it seems
--- a/src/frontend/helpers/gamepad.ts
+++ b/src/frontend/helpers/gamepad.ts
@@ -9,7 +9,7 @@ import {
   checkStandard,
   checkN64Clone1,
   checkGenius1,
-  checkSwitchPro
+  checkNintendo
 } from './gamepad_layouts'
 import { VirtualKeyboardController } from './virtualKeyboard'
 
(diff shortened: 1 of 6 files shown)


src/frontend/helpers/gamepad_layouts/nintendo.ts
```diff
@@ -47,6 +47,54 @@ export function checkGameCube(
   checkAction('rightClick', X.pressed, controllerIndex)
 }
 
+// Nintendo Switch Pro Controller / Joy-Cons (Vendor: 057e)
+// Chromium reports these with the "standard" mapping by physical position,
+// so buttons[0] is the bottom button (labeled B on Switch) and buttons[1] is
+// the right button (labeled A). We swap the face buttons so the A/B and X/Y
+// actions match the Nintendo labels shown in the UI.
+export function checkSwitchPro(
```
CommandMC: This function runs for all Nintendo controllers (`057e` is the USB vendor ID for Nintendo, which is the only thing checked for in `gamepad.ts`). So this function should probably be named `checkNintendo` (or at least `checkNintendoSwitch`, since Joy-Cons are also explicitly supported)

src/frontend/helpers/gamepad.ts
```diff
@@ -493,6 +494,8 @@ export const initGamepad = () => {
           checkN64Clone1(buttons, axes, index, checkAction)
         } else if (controller.id.match(/0583.*a009/i)) {
           checkGenius1(buttons, axes, index, checkAction)
+        } else if (controller.id.match(/057e/i)) {
```
CommandMC: Somewhat tying into my other comment: This check lets every controller manufactured by Nintendo through. This may include [other devices](<https://usb-ids.gowdy.us/read/UD/057e>). Hardening the check to only let explicitly supported controllers through (I assume in this case it would be device IDs `2006`, `2007` and `2009`) may be wise

CommandMC (changes requested):

CommandMC (approved):

PR #5748, opened 2026-07-14
[Feat] Add Humble Bundle deals to the Discounts screen

flavioislima:
- Added Humble Bundle deals using our affiliate link.
- Moved the Workflows for GMG and now HB to a new repo: https://github.com/Heroic-Games-Launcher/deals-listing
- Added a new List View at the discount screen.

https://github.com/user-attachments/assets/b36ec3a8-7657-411d-b186-a4846dc35d57


src/backend/discounts/humble.ts
```diff
@@ -0,0 +1,171 @@
...
+
+interface ImpactCatalogItemsResponse {
+  '@page'?: string
+  '@numpages'?: string
+  Items?: ImpactCatalogItem[]
+}
+
+const formatPrice = (amount: number, currency: string): string => {
+  try {
+    return new Intl.NumberFormat('en-US', {
```
CommandMC: Shouldn't this respect the user's current language (`i18next.language`)?

src/backend/discounts/humble.ts
```diff
@@ -0,0 +1,171 @@
...
+      style: 'currency',
+      currency
+    }).format(amount)
+  } catch {
```
CommandMC: In which cases can this throw?

src/backend/discounts/humble.ts
```diff
@@ -0,0 +1,171 @@
...
+  const first = await fetchPage(url, 1)
+  const numPages = Math.min(
+    parseInt(first['@numpages'] ?? '1', 10) || 1,
+    MAX_PAGES
+  )
+
+  const items = [...(first.Items ?? [])]
+  for (let page = 2; page <= numPages; page++) {
+    items.push(...((await fetchPage(url, page)).Items ?? []))
+  }
```
CommandMC: Deals are hosted in a single .json file in our repo, so I fail to see how there could be multiple pages here

src/backend/discounts/humble.ts
```diff
@@ -0,0 +1,171 @@
...
+  return HUMBLE_CURRENCIES.includes(requested)
+    ? requested
+    : HUMBLE_FALLBACK_CURRENCY
+}
+
+interface HumbleCache {
+  products: CatalogProduct[]
+  fetchedAt: number
+}
+const cache = new Map<string, HumbleCache>()
```
CommandMC: This should be a `CacheStore`

src/frontend/screens/Discounts/helpers.ts
```diff
@@ -228,19 +228,37 @@ const withAffiliate = (storeLink: string): string => {
...
+// Our Humble affiliate deeplink base. Humble's impact catalog links embed the
+// real store destination in their `u` param; we re-wrap that destination with
+// our own tracking link so clicks are attributed to Heroic.
+const HUMBLE_AFFILIATE_LINK = 'https://humblebundleinc.sjv.io/heroic'
+
+const withHumbleAffiliate = (storeLink: string): string => {
+  try {
+    const destination = new URL(storeLink).searchParams.get('u') ?? storeLink
+    return `${HUMBLE_AFFILIATE_LINK}?u=${encodeURIComponent(destination)}`
+  } catch {
```
CommandMC: When can this throw?

src/frontend/screens/Discounts/helpers.ts
CommandMC: Renaming this to `withGogAffiliate` seems sensible now

src/frontend/screens/Discounts/index.tsx
```diff
@@ -183,16 +190,22 @@ export default function Discounts() {
...
           window.api.getGogDiscounts(localeSettings, hideOwned, wishlistOnly),
           window.api.getGmgDiscounts(gmgCurrency).catch((err: unknown) => {
             window.api.logError(`Failed to fetch GMG discounts: ${String(err)}`)
             return [] as CatalogProduct[]
+          }),
+          window.api.getHumbleDiscounts().catch((err: unknown) => {
+            window.api.logError(
+              `Failed to fetch Humble discounts: ${String(err)}`
+            )
+            return [] as CatalogProduct[]
```
CommandMC: This type assertion is unnecessary

```suggestion
            return []
```

CommandMC (dismissed):

src/backend/discounts/humble.ts
flavioislima: All prices on this screen are en-US formatted on purpose — the GOG catalog is also queried with `locale: 'en-US'` (see `getLocaleSettings`), so its price strings come back en-US formatted too. If I localize only Humble/GMG we'd end up with mixed formats on the same screen (`$5.99` next to `59,99 €`). Doing it properly means localizing the GOG query plus both feeds together, so I'd rather do that in a follow-up for the whole screen.

src/backend/discounts/humble.ts
flavioislima: `Intl.NumberFormat` throws a `RangeError` when the currency code isn't a well-formed ISO 4217 code. `Currency` comes straight from the feed, so one malformed item would otherwise break the whole normalize pass — the fallback returns a plain `USD 5.99`-style string instead.

src/frontend/screens/Discounts/helpers.ts
flavioislima: `new URL()` throws a `TypeError` on malformed URLs, and `storeLink` comes from the feed. Same pattern as `withGogAffiliate` right above — a bad link falls back to the untagged URL instead of breaking the render.

src/backend/discounts/humble.ts
flavioislima: Good catch, fixed.

src/backend/discounts/humble.ts
flavioislima: fixed

src/frontend/screens/Discounts/helpers.ts
flavioislima: done

flavioislima:
@CommandMC fixed your comments

src/backend/discounts/humble.ts
flavioislima: Although, i am double checking in the API, also for GOG.

flavioislima: Yeah, its actually weird. GOG uses the format for the currency, but since it does not support all languages we support, we are using US as fallback instead of trying to hardcode every single one. Let me know if you hve a better idea, but maybe using the user locale won't fix it since humble, gmg and gog won't support all of them and will always break for someone as I see it.

src/backend/discounts/humble.ts
CommandMC: Hm, alright. Personally I would've used [`Intl.supportedValuesOf('currency')`](<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf#supported_currency_identifiers>) to filter out bad values, but this is fine too

src/backend/discounts/humble.ts
CommandMC: This is just about formatting the currency though. You'd still get USD if USD is provided, it'll just be formatted correctly for the user's locale

src/frontend/screens/Discounts/helpers.ts
CommandMC: Alright, but realistically, would you ever get a bad `storeLink`? I would assume that there would be some checks on Humble's/Impact's side to prevent this

flavioislima: I have no idea, but a try catch also does not hurt to have, right? 😅 
But I can remove, no problem.

src/backend/discounts/humble.ts
flavioislima: Hmm... but then it will be USD formated as EUR... its weird and wrong, no? I am not sure, I can change it if you feel it makes sense.

src/frontend/screens/Discounts/helpers.ts
CommandMC: Well, in this case, it doesn't really make sense (since you just return `storeLink` un-altered if it's an invalid URL, so consumers of this function will get an invalid URL in that case). Between showing an invalid URL and throwing an error, I'd prefer the error

src/backend/discounts/humble.ts
CommandMC: `currency` (returned from the deals feed) is the actual currency code the function will use to format. The language only dictates how the number and currency are formatted (whether it's spelled out "USD" vs "$", for example)

flavioislima: Did the change

src/frontend/screens/Discounts/helpers.ts
flavioislima: Removed

CommandMC (approved):

PR #5752, opened 2026-07-16
fix: unify webview session partitions so login credentials carry over to stores

kairosci:
Store webviews (Epic, GOG, Amazon, Zoom) do not share the authentication session with the login flow. A user logging in through "Manage Accounts" is prompted to log in again when opening any store page.

The root cause was a partition mismatch in Electron:
- Epic login used `persist:epicstore` while the store webview used `persist:epic`, two completely isolated sessions.
- For GOG, Amazon and Zoom the login page used `persist:undefined` while the store used their own partition, again isolated.

This fix unifies the login and store partitions per runner so the cookies set during the OAuth flow are available when the store webview loads, and makes session cleanup on logout consistent across all runners.

- [x] Tested the feature and it is working on a current and clean install.

### DIFF
--- a/src/backend/storeManagers/nile/user.ts
+++ b/src/backend/storeManagers/nile/user.ts
@@ -9,6 +9,7 @@ import { existsSync, readFileSync } from 'graceful-fs'
 import { configStore } from './electronStores'
 import { clearCache } from 'backend/utils'
 import { nileUserData } from './constants'
+import { session } from 'electron'
 
 function authLogSanitizer(line: string) {
   try {
@@ -99,6 +100,10 @@ export class NileUser {
 
     configStore.delete('userData')
     clearCache('nile')
+    const ses = session.fromPartition('persist:amazon')
+    ses.clearStorageData().catch(() => {})
+    ses.clearCache().catch(() => {})
+    ses.clearAuthCache().catch(() => {})
   }
 
   static async getUserData(): Promise<NileUserData | undefined> {
--- a/src/backend/storeManagers/gog/user.ts
+++ b/src/backend/storeManagers/gog/user.ts
@@ -7,7 +7,7 @@ import { isOnline } from '../../online_monitor'
 import { GOGCredentials, UserData } from 'common/types/gog'
 import { libraryManagerMap } from '../index'
 import { clearCache } from 'backend/utils'
-import { app } from 'electron'
+import { app, session } from 'electron'
 import { gogdlAuthConfig } from './constants'
 
 function authLogSanitizer(line: string) {
@@ -130,6 +130,10 @@ export class GOGUser {
     if (existsSync(gogdlAuthConfig)) {
       unlinkSync(gogdlAuthConfig)
     }
+    const ses = session.fromPartition('persist:gog')
+    ses.clearStorageData().catch(() => {})
+    ses.clearCache().catch(() => {})
+    ses.clearAuthCache().catch(() => {})
     logInfo('Logging user out', LogPrefix.Gog)
   }
 
--- a/src/backend/storeManagers/legendary/user.ts
+++ b/src/backend/storeManagers/legendary/user.ts
@@ -68,7 +68,7 @@ export class LegendaryUser {
       return
     }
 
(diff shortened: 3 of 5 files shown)


flavioislima:

kairosci:

Drackrath:
I agree with this

CommandMC (changes requested):
Most of this PR covers a situation that does not happen in practice

1. I do not remember why we separated the two webview partitions, but I recall there _being_ a reason. I wouldn't have a problem with undoing that though (if there are truly still issues with unifying them, those should be documented)
2. The other, larger part of this PR covers a situation where you're logged in on Legendary/GOGDL/Nile, but not in the stores
    1. This cannot happen during regular Heroic usage (you log in through "Manage Accounts", which logs you into both the stores & CLI)
    2. In the off-chance that this does happen (if the user does not open the store page for a while, for example), the user simply has to log in on the store page again. I don't think we need such a complicated synching mechanism to avoid this non-issue.

kairosci:
1. ok
2. The problem occurs in both Legendary and Managed accounts. In neither case does the shop have access to the account; I have to log in manually.
Also, every time I open the app, it takes me to the "No Logged-in" page in the shops.

CommandMC:
> The problem occurs in both Legendary and Managed accounts. In neither case does the shop have access to the account; I have to log in manually.

Well, that's what unifying the two partitions would resolve, no?

>  Also, every time I open the app, it takes me to the "No Logged-in" page in the shops.

I can't reproduce this. Please open an issue about this.

Drackrath:
>>Also, every time I open the app, it takes me to the "No Logged-in" page in the shops.

>I can't reproduce this. Please open an issue about this.

@CommandMC I can indeed reproduce this, but the Login expires after only few hours. Not directly, and mainly on Windows.

kairosci:
On Flatpak (Fedora, but also Ubuntu), every time I kill the process, the store is no longer connected.

src/backend/storeManagers/nile/user.ts
```diff
@@ -99,6 +100,10 @@ export class NileUser {
 
     configStore.delete('userData')
     clearCache('nile')
+    const ses = session.fromPartition('persist:amazon')
```
CommandMC, line 103: Should we maybe store these partition names in constants somewhere, so the same type of `epic`/`epicstore` mixup can't happen again?

(not necessary for this commit, just thinking out loud here)

CommandMC (approved):

src/backend/storeManagers/nile/user.ts
kairosci: That might be a good idea. The question is, how would you prefer to save them? In localStorage or in an embedded database?

In my honest opinion, a local database containing the configurations, names, and key metadata would be an interesting option to propose.

CommandMC: Oh, I was thinking far simpler than that. We could just store them as constants somewhere in `src/common/`

kairosci: Yeah, that wouldn't be bad either, even though I liked the storage idea better; it seems a little cleaner to me; but as a starting point, it's still better than nothing. Should I propose it in a new PR?

kairosci:
Anyway, after testing it, it seems to work; I don't have to log in twice; it seems to log me in automatically to the various stores.

Do you think there's anything that needs to be corrected or improved here, or is it okay with you?

kairosci:
@Drackrath Could you test it yourself too and let me know if issue is resolved? Thanks!

src/backend/storeManagers/nile/user.ts
CommandMC: Well, localStorage isn't shared between the two processes (Node does not enable it by default at all), and any way of dynamically exchanging info between them at runtime requires async operations, which are annoying in the Frontend

> Should I propose it in a new PR?

Sure!

CommandMC:
I've already approved this PR. I was just waiting in case @Drackrath wanted to report whether their issue got resolved

Drackrath:
All good, tested and resolved, can be merged.

PR #5754, opened 2026-07-16
fix(backend): flush session cookies to prevent store disconnects on kill

### LINKED ISSUE #5753: Store disconnected after killing process on Flatpak
On Flatpak (Fedora, but also Ubuntu), every time I kill the process, the store is no longer connected (logged-in).

kairosci:
This pull request ensures that Electron session cookies and DOM storage data are immediately written to disk upon any change. This prevents store credentials from being lost when the process is forcefully killed on Flatpak or during scheduled token refreshes on Windows. Closes #5753.

### DIFF
--- a/src/backend/main.ts
+++ b/src/backend/main.ts
@@ -321,6 +321,23 @@ if (!gotTheLock) {
 
     handleProtocol(argv)
   })
+
+  // Proactively save session cookies to prevent store disconnects on hard crash or ungraceful shutdown
+  app.on('session-created', (ses) => {
+    let timeout: NodeJS.Timeout | null = null
+    ses.cookies.on('changed', () => {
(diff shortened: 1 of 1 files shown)


CommandMC:
> when the process is forcefully killed

I don't consider this a supported use case

> during scheduled token refreshes on Windows

Are you sure your proposed change resolves this issue? I don't see why it would, but I assume you tested this?

kairosci:
I have tried to kill app several times, and I am logged in, it seems to works for me.

CommandMC:
Alright, but I was asking whether this also fixes the Windows issues.

I don't consider killing the Heroic process to be a supported use case. You will cause plenty of other issues doing that as well.

@Drackrath you were able to reproduce the issue on Windows, correct? Could you test this PR to see if it helps?

PR #5755, opened 2026-07-16
Update to Electron 43 and re-add React Dev Tools extension

CommandMC:
Electron 41 (our current version) lacks some functionality required by the Dev Tools extension, so we have to update to 43 for it to work. This also seems like a good idea to do before the release, regardless of the dev tools.

This allows us to drop `react-devtools` (the standalone version), which cleans up our Vite config & should close a lot of our security alerts (since `react-devtools` depends on an ancient version of Electron)

To test:
Run `pnpm start` and open the Chromium dev tools. You should see the React dev tools ("Components" & "Profiler" tabs)

Functionality outside of a dev environment is not impacted, so this should be very safe to merge

---


CommandMC:
E2E tests seem to fail due to the electron install script not running properly. I can somewhat reproduce this locally; manually running `pnpm exec install-electron` makes it install. Not quite sure why this is happening, can anyone else repro?

arielj:
> E2E tests seem to fail due to the electron install script not running properly. I can somewhat reproduce this locally; manually running `pnpm exec install-electron` makes it install. Not quite sure why this is happening, can anyone else repro?

I wonder if that's related to the issue with electron failing to install in newer node versions, maybe you can try this? (it worked for me locally to install electron with node 26) https://github.com/electron/electron/issues/51619#issuecomment-4521813516

this in package.json:
```json
  "overrides": {
    "yauzl": "^3.3.1"
  }
```

CommandMC:
Well, the script runs when ran manually, it just doesn't get ran as part of `pnpm install`

imLinguin (approved):

PR #5772, opened 2026-07-23
Fix: show Rosetta warning when Rosetta is not installed

### LINKED ISSUE #5774: Info Dialog when Rosetta 2 is missing
### Problem description

I've got my Mac updated to Golden Gate (macOS 27) and then I got the following error when opening a game: "Error: spawn Unknown system error -86"
The internet said that this is caused by Rosetta 2 not installed, so I installed it. And I think that maybe a little Info Dialog could show up when this error occurs and tells the user that Rosetta 2 has to be installed.
It is really not that big of an issue, just a little improvement that maybe can reduce some support tickets.

### Feature description

An Dialog that tells the user that he has to install Rosetta 2 to play this game, or something like that.

### Alternatives

_No response_

### Additional information

_No response_

flavioislima:
We had the code to check for if rosetta was installed or not on macOS but its was not working anymore after some regression, I am not sure, maybe never worked. I tested this one now and it works fine. I got the issue when updated to macOS 27 beta.

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -1325,11 +1325,15 @@ export async function checkRosettaInstall() {
     return
   }
 
-  const { stdout: rosettaCheck } = await execAsync(
-    'arch -x86_64 /usr/sbin/sysctl sysctl.proc_translated'
-  )
-
-  const result = rosettaCheck.split(':')[1].trim() === '1'
+  let result = false
+  try {
+    const { stdout: rosettaCheck } = await execAsync(
+      'arch -x86_64 /usr/sbin/sysctl sysctl.proc_translated'
+    )
+    result = rosettaCheck.split(':')[1].trim() === '1'
+  } catch {
+    // the spawn itself fails when Rosetta is not installed
+  }
 
   logInfo(
     `Rosetta is ${result ? 'available' : 'not available'} on this system.`,
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -1325,15 +1325,12 @@ export async function checkRosettaInstall() {
     return
   }
 
-  let result = false
-  try {
-    const { stdout: rosettaCheck } = await execAsync(
-      'arch -x86_64 /usr/sbin/sysctl sysctl.proc_translated'
-    )
-    result = rosettaCheck.split(':')[1].trim() === '1'
-  } catch {
-    // the spawn itself fails when Rosetta is not installed
-  }
+  // the spawn itself fails when Rosetta is not installed
+  const result = await execAsync(
+    'arch -x86_64 /usr/sbin/sysctl sysctl.proc_translated'
+  )
+    .then(() => true)
+    .catch(() => false)
 
   logInfo(
     `Rosetta is ${result ? 'available' : 'not available'} on this system.`,


src/backend/utils.ts
```diff
@@ -1325,11 +1325,15 @@ export async function checkRosettaInstall() {
     return
   }
 
-  const { stdout: rosettaCheck } = await execAsync(
-    'arch -x86_64 /usr/sbin/sysctl sysctl.proc_translated'
-  )
-
-  const result = rosettaCheck.split(':')[1].trim() === '1'
+  let result = false
+  try {
```
flavioislima: On a Mac without Rosetta, the check command doesn't return "0" — it fails to run at all.

So we need the Try/Catch here to make the check work, if we remove it it fails since the command fails.

src/backend/utils.ts
```diff
@@ -1325,11 +1325,15 @@ export async function checkRosettaInstall() {
...
+    const { stdout: rosettaCheck } = await execAsync(
+      'arch -x86_64 /usr/sbin/sysctl sysctl.proc_translated'
+    )
+    result = rosettaCheck.split(':')[1].trim() === '1'
```
CommandMC: While you're at it, could you add some example output to a comment so this string manipulation is easier to understand for us non-macOS developers?

I believe it would be

```suggestion
    // Example output: `sysctl.proc_translated: 1`
    result = rosettaCheck.split(':')[1].trim() === '1'
```

CommandMC (commented):
One thing to note about this: macOS 27 apparently has on-the-fly translation of x86_64 without Rosetta (see the "Important" note [here](<https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment#overview>)). I don't know if this also applies for our use case though

src/backend/utils.ts
arielj: do we even need this comparison?

I mean: if rossetta is not installed the spawn fails, so if it doesn't fail it means rosetta is installed?

it could be:

```ts
let result = false;
try {
  await execAsync(
    'arch -x86_64 /usr/sbin/sysctl sysctl.proc_translated'
  )
  result = true
} catch {
}
```

CommandMC: Well, if you're gonna golf, golf
```ts
const result = await execAsync('arch -x86_64 /usr/sbin/sysctl sysctl.proc_translated')
    .then(() => true)
    .catch(() => false)
```
I am also not sure whether we need the `sysctl` part of the command at all, I assume the `arch -x86_64` wrapper just fails with any command provided if Rosetta isn't installed. So perhaps something simple like `arch -x86_64 echo 'Success'` works too?

flavioislima:
> One thing to note about this: macOS 27 apparently has on-the-fly translation of x86_64 without Rosetta (see the "Important" note [here](<https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment#overview>)). I don't know if this also applies for our use case though

I'm on 27 beta and I needed to install it. That is why I found the issue in heroic since all my games stopped working after the update.

src/backend/utils.ts
flavioislima: @CommandMC I am not expert in macOS as well, but the sysctl is the most reliable way of checking this so its better to keep it.

but I simplified the check like you and @arielj said.

CommandMC: Sure, but if you're not looking at stdout at all, I would assume that it does not matter which command you run on x86. The important part would just be the fact that *any* command *can* run on x86 (which implies Rosetta being installed). But as mentioned, I cannot verify this.

CommandMC (approved):

Issue #5704, opened 2026-06-23
Enable Heroic to grab NSFW/Adult content trough SteamGridDB


thisaray:
### Describe the bug

When searching for game artwork in Heroic using a SteamGridDB API key, NSFW/adult content is not displayed, even though these images are available when signed in

### Add logs

```text
/
```

### Steps to reproduce

1. Log into SteamGridDB directly and verify adult/NSFW content is visible
2. Set a valid SteamGridDB API key in Heroic settings
3. Search for the same game artwork in Heroic
4. Observe that the NSFW content is not displayed in search results

### Expected behavior

Heroic should display NSFW/adult artwork from SteamGridDB

### Heroic Version

Latest Stable

### System Information

CPU: 32x AMD Ryzen 9 5950X 16-Core Processor
Memory: 33.55 GB (used: 14.21 GB)
GPUs:
  GPU 0:
    Name: NVIDIA Corporation AD102 [GeForce RTX 4090]
    IDs: D=2684 V=10de SD=40bf SV=1458
    Driver: nvidia
OS: Arch Linux 7.0.12-arch1-1 (linux)

The current system is not a Steam Deck
We are not running inside a Flatpak container
We are not running from an AppImage

Software Versions:
  Heroic: 2.22.0 Hajrudin
  Legendary: 0.20.43 Riding Shotgun (Heroic)
  gogdl: 1.2.1
  comet: comet 0.2.0
  Nile: 1.1.2 Will A. Zeppeli

Moon-yungg:
Fixed it with PR: #5778 

Enabled by a Checkbox in advanced settings:

<img width="1115" height="437" alt="Image" src="https://github.com/user-attachments/assets/eec09957-d23a-4040-bf18-9efe5c0ae085" />

PR #5778, opened 2026-07-25
Add toggle to enable NSFW content in SteamGridDB search results

Moon-yungg:
Add an option to send the adults flag to steamgriddb so you can choose nsfw images

Refering to issue #5704 

---

image


imLinguin (approved):

thisaray:
This is not 100% the scope of my question, but seeing as you did the gruntwork already: 
could you maybe expand the scope to also include the "epilepsy" and "humor" tags?
https://www.steamgriddb.com/api/v2#tag/GRIDS/operation/getGridsByGameId

Moon-yungg:
@thisaray Sure. I think we should use one toggle for everything so you don't have 3 toggles. Or maybe a multi select box of what kind you want to see. Any idea on how to implement it visually?

Moon-yungg:
@thisaray @imLinguin Added the other categories to the options too.

PR #5791, opened 2026-08-01
Auto Sync Cloud Saves when launched from steam

mhunsber:
Adds `--download --upload` to the steam launch arguments if the user has the game set to auto sync saves.
This ensures that if the user launches the game from steam it actually performs the cloud sync.

Note that this only applies **when** the user hits the add to steam button. This could lead to the following confusions:

- If the user doesn't have autosync enabled when they add to steam, it won't have these arguments.
- If the user later unchecks autosync after they've added to steam, it will still have these arguments.

For that reason I can see why this might not be desired. Maybe this should just be a docs update?

---

- [ x] Tested the feature and it's working on a current and clean install.
- [ x] Tested the main App features and they are still working on a current and clean install. (Login, Install, Play, Uninstall, Move games, etc.)
- [ x] Created / Updated Tests (If necessary)
- [ x] Created / Updated documentation (If necessary)


flavioislima:

mhunsber:

PR #5799, opened 2026-08-03
[Tech] Add Zoom runner to Plausible analytics

flavioislima:


src/backend/utils/plausible.ts
```diff
@@ -102,6 +105,7 @@ export async function startPlausible() {
     gog: !!GOGUser.isLoggedIn(),
     epic: !!LegendaryUser.isLoggedIn(),
     amazon: !!NileUser.isLoggedIn(),
+    zoom: existsSync(zoomTokenPath) && zoomConfigStore.get('isLoggedIn', false),
```
arielj: for consistency with the other stores, I'd suggest creating an "isLoggedIn" function in the storeManager/zoom module with this content

arielj (approved):
Approving this, but just left a suggestion in case you want to include it

src/backend/utils/plausible.ts
flavioislima: done

PR #5800, opened 2026-08-03
Harden GitHub Actions and add a continuous check

Totara-thib:
Hi, drive-by CI hardening in three commits, one logical change each.

**Commit 1 pins every action to its commit sha**, versions kept as comments. A tag like `@v8` is a movable pointer: whoever controls the action, or anyone who compromises it, can re-point it and your next run executes their code with the job's token. The exposed spots here are the release paths: the AUR deploy, the flathub release (which used `ad-m/github-push-action@master`, a moving branch), and the CLA workflow. Same pattern as the tj-actions/changed-files incident (CVE-2025-30066). All shas were resolved from the upstream repos and cross checked against their release tags. Two things worth knowing: `ad-m`'s master is currently exactly its v1.3.0 release commit, so behavior is unchanged, and `contributor-assistant/github-action` is archived upstream, pinning contains the risk but replacing it at some point would be better still, that call is yours. The pins stay maintainable: a dependabot config with the `github-actions` ecosystem bumps them automatically, one small block.

**Commit 2 adds `permissions:` blocks to the five workflows that had none.** The scopes are derived from what each workflow does with the token, not guessed: the draft-release and flathub workflows do all their writing through your dedicated WORKFLOW_TOKEN secret, so their default token drops to `contents: read`; create-pr-from-stable and update-bug-template push branches and open PRs with the default token, so they get `contents: write` and `pull-requests: write`. The rest already declare scoped permissions.

**Commit 3 adds [Plumber](https://github.com/getplumber/plumber) to CI**, the tool I used to find the issues in the first place, so none of this quietly drifts back:

- Scans the workflows on each push to main and on each PR, fails when something regresses: an unpinned action, a job without permissions, a known CVE. The gate passes at 85 points, so one small finding does not block your PRs.
- The config is a small overlay inheriting the CLI's built-in baseline: it trusts the actions you already use and turns off two controls with a comment saying why (the archived CLA action until you replace it, and the flatpak and archlinux build containers that track distro tags by design). Findings land in the security tab as SARIF.
- Adds a score badge to the README. It works like OpenSSF Scorecard's published results: runs publish the score to score.getplumber.io, the badge shows the state of main, a failed publish never fails your CI, and it reads UNKNOWN in gray until the first run.

One heads up: if the org uses an Actions allowlist in settings, patterns written against tags (like `owner/action@v6`) stop matching once refs are shas and workflows refuse to start. Entries need to be `owner/action@*` in that case.

To be fully transparent: I work on Plumber. If you do not want the tool in your CI, no hard feelings, say so and I remove that commit from the PR, the two hardening commits are the part that matters and they stand entirely on their own.

---

### DIFF
--- a/.github/workflows/build-base.yml
+++ b/.github/workflows/build-base.yml
@@ -30,7 +30,7 @@ jobs:
             arch: x64
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
       - uses: ./.github/actions/install-deps
       - name: Build Windows x64 Portable version
         run: pnpm dist:win portable --x64
@@ -56,7 +56,7 @@ jobs:
         run: pnpm dist:linux AppImage --x64 --arm64 --publish=never
         if: runner.os == 'Linux'
       - name: Upload built version
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: ${{ (matrix.os == 'windows-2025'  && 'win-x64-portable'  ) ||
                     (matrix.os == 'windows-11-arm' && 'win-arm64-portable') ||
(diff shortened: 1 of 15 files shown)


flavioislima:

Totara-thib:

Totara-thib:
recheck

PR #5811, opened 2026-08-08
fix(macOS): quote generated shortcut launch URL

marcodallagatta:
## Problem
macOS shortcuts are generated as small `.app` bundles that execute a `run.sh` script. The script passed its `heroic://launch?...&runner=...` URL to Bash without quotes.

In Bash, `&` is a control operator, so `&runner=...` was treated as a separate background command instead of part of the URL. Heroic therefore received an incomplete launch URL and the generated shortcut could not launch the game correctly.

## Fix
- quote the generated `heroic://` launch URL so it is passed as one argument
- add a regression test that exercises the macOS shortcut generator and checks the resulting `run.sh`

## Testing
- `pnpm test:ci`
- `pnpm codecheck`
- `pnpm exec prettier --check src/backend/shortcuts/shortcuts/shortcuts.ts src/backend/shortcuts/shortcuts/__tests__/shortcuts.test.ts`


flavioislima:

arielj:
I think the change looks right, but the test is kinda overkill for what it's testing (it's mocking so many things, creating a file, etc, and then just checking that the url has double quotes)

So maybe, instead, some part of this https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5811/changes#diff-b579c04c5d1726b12575f15d9012c905141b35d12f20fc67c017b01c59036499R227-R229 can be extracted as a simple function that returns the string and then test that method?

I would be fine with a test that complicated if we were testing more things about the shortcuts, but at this point it feels too much.

marcodallagatta:
Hi @arielj, you're right. What about extracting the mac launch command construction into a helper and replace the mocked test with a direct unit test for the generated command?

arielj:
> Hi @arielj, you're right. What about extracting the mac launch command construction into a helper and replace the mocked test with a direct unit test for the generated command?

yes, that's what I'm suggesting, extracting some part of that code without all those sideeffects into its own function and test that function

arielj (approved):
thanks!

PR #5814, opened 2026-08-09
chore: Remove remnants of our Snap support

CommandMC:
It seems that the Snap maintainers do not want to grant umu the same permissions as Steam (which it needs for the same reasons as Steam). Without umu, it does not make sense to port Heroic to Snap (as Proton essentially requires it and falling back to Wine is not an option).

---


flavioislima (approved):
LGTM

PR #5818, opened 2026-08-09
fix: Search for a `heroic` executable on PATH if our executable is not called "heroic"

CommandMC:
Details are in the comment at

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/7da85bc52f70f59671913b1b48a7984e0a12b1c3/src/backend/shortcuts/nonesteamgame/nonesteamgame.ts#L268-L276

This is preparation for a potential official `heroic-games-launcher` package for Arch Linux

Testing this isn't really practical. The added test verifies that (1) the old behaviour is still there and (2) the new behaviour is working though.

---

### DIFF
--- a/src/backend/__mocks__/electron.ts
+++ b/src/backend/__mocks__/electron.ts
@@ -15,9 +15,10 @@ const dialog = {
 
 const app = {
   // app override
-  getPath: jest.fn().mockImplementation((path: string) => {
+  getPath: (path: string) => {
+    if (path === 'exe') return join(appBasePath, 'heroic')
     return join(appBasePath, path)
-  }),
+  },
   getVersion(): string {
     // TODO: What should we return here?
     return '1.0.0'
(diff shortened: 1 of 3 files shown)


src/backend/__mocks__/electron.ts
```diff
@@ -15,9 +15,10 @@ const dialog = {
 
 const app = {
   // app override
-  getPath: jest.fn().mockImplementation((path: string) => {
+  getPath: (path: string) => {
+    if (path === 'exe') return join(appBasePath, 'heroic')
     return join(appBasePath, path)
-  }),
+  },
```
CommandMC, line 21: I am not quite sure why I had to remove the `jest.fn()` wrapper around this function, but with it there, the function was only called for the `userData` arg (somewhere else in the codebase), not `exe`

keenanweaver:
We have to patch Heroic in NixOS to get it working there as well (see here: https://github.com/NixOS/nixpkgs/pull/550824)

Does this address the same issue or is it separate?

CommandMC:
I assume it would? I am not sure how you package Heroic on NixOS. The problem on Arch is that system Electron should be used when building a package for one of its main repositories, and `app.getPath('exe')` only returns `/usr/bin/electronXX` in that case. If you have that same problem on NixOS, this change will resolve it, yes.

flavioislima (approved):
LGTM

PR #5827, opened 2026-08-15
fix: Make sure to set Wine-related options when adding a sideloaded game

### LINKED ISSUE #5820: Heroic using shared prefix no matter what i do
### Describe the bug

After the latest update heroic will use shared prefix folder for all my new games
And it wont let me make separate prefix for each game
And the game launch but without logs and it opens but in heroic it still says (launching)

### Add logs

```text
No logs
```

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### Expected behavior

It used to make a separate prefix for every game by default 

### Screenshots

_No response_

### Heroic Version

Latest Stable

### System Information

- OS {cachyos}

### Additional information

_No response_

CommandMC:
In https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5599, I had assumed that `setWinePrefix`, `setWineVersion` and `setCrossoverBottle` are `useSetting` return values, and thus, setting them will automatically save the set option for the game. However...

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/34924446b69f14e9ec9cdfcce3ff8c9157e19ed4/src/frontend/screens/Library/components/InstallModal/index.tsx#L49-L52

...they are not. Thus, we have to make sure to save them when the user clicks "Finish" in the dialog.

Closes #5820

---

### DIFF
--- a/src/backend/utils.ts
+++ b/src/backend/utils.ts
@@ -1630,7 +1630,10 @@ const axiosClient = axios.create({
   httpsAgent: new https.Agent({ keepAlive: true })
 })
 
-export const writeConfig = (appName: string, config: Partial<AppSettings>) => {
+export const writeConfig = async (
+  appName: string,
+  config: Partial<AppSettings>
+) => {
   logInfo(
     `Writing config for ${appName === 'default' ? 'Heroic' : appName}`,
     LogPrefix.Backend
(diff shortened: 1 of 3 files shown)


arielj (approved):
Seems to be working fine :+1:

I think it would be nice, at some point in the future, to be able to send the settings in the handleInstall call, but that's out of the scope of this

PR #5832, opened 2026-08-17
feat: Use umu to provide the Steam Runtime, and automatically enable it for native GOG games

CommandMC:
The Steam Runtime is useful in two cases:
1. Proton is used to run a Windows game
2. A Linux-native game (potentially built for older distros) is ran on a modern distro

We already use umu to cover the first case. Let's also use it for the 2nd.

The existing "Use Steam Runtime" checkbox is replaced by a dropdown to select the Steam Runtime to use for the game:

grafik

> [!NOTE]
> This also removes the "Use Steam Runtime" option completely for non-native games. Using umu is strictly more compatible. We were running Proton in SteamRT3, not 4 (like modern versions expect)

When the game is then launched, umu is used to provide the runtime:

grafik

---

GOG games will also automatically enable `umu-scout` when they are installed/imported. Scout is the default runtime for Linux-native Steam games, and it seems to provide the most period-accurate runtime for most games.  
We may want to expand this to automatically check some heuristics and decide the runtime based on that, but this is fine as a first implementation.

---

### DIFF
--- a/src/backend/migration/migrations/config.ts
+++ b/src/backend/migration/migrations/config.ts
@@ -1,0 +1,35 @@
+import { readdir } from 'fs/promises'
+import { parse } from 'path'
+
+import { isLinux } from 'backend/constants/environment'
+import { gamesConfigPath } from 'backend/constants/paths'
+import { GameConfig } from 'backend/game_config'
+import { logDebug } from 'backend/logger'
+
+import type { Migration } from '..'
+
+export class UmuSteamRuntimeMigration implements Migration {
+  identifier = 'umu-steam-runtime'
+  async run() {
+    if (!isLinux) return true
+
+    for (const entry of await readdir(gamesConfigPath, {
+      withFileTypes: true
+    })) {
+      const filename = entry.name
+      if (!entry.isFile() || !filename.endsWith('.json')) continue
+      const appName = parse(filename).name
+      const config = GameConfig.get(appName)
+      const settings = (await config.getSettings()) as {
+        useSteamRuntime?: boolean
+      }
+      // NOTE: This will also set the option for non-native games, but the
+      //       option is not actually used for them, so it should be fine
+      if (settings.useSteamRuntime) {
+        config.setSetting('steamRuntime', 'umu-scout')
+        logDebug(['Enabled new Steam Runtime option for', appName])
+      }
+    }
+    return true
+  }
+}
--- a/src/frontend/screens/Settings/components/SteamRuntime.tsx
+++ b/src/frontend/screens/Settings/components/SteamRuntime.tsx
@@ -1,7 +1,6 @@
 import { useCallback, useContext } from 'react'
 import { useTranslation } from 'react-i18next'
 import { SelectField } from 'frontend/components/UI'
-import ContextProvider from 'frontend/state/ContextProvider'
 import useSetting from 'frontend/hooks/useSetting'
 import SettingsContext from '../SettingsContext'
 import InfoIcon from 'frontend/components/UI/InfoIcon'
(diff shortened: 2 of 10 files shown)


imLinguin (approved):
I like what I see. I'll try few native games myself that I know had many issues

CommandMC:
(forgot to make the "None" option translatable)

arielj:
I wonder if we should have some logic to migrate the `useSteamRuntime` setting before removing it.

If someone had `useSteamRuntime` enabled for a native game to make it work, the game will stop working until they manually go and set the new `steamRuntime` config, right?

I think we could keep the `useSteamRuntime` option in the settings json (not the GUI, but in the type), and use it as a fallback to set the scout runtime if `steamRuntime` is unset and `useSteamRuntime` was true. We would have to differentiate `false` (set to None) from `undefined` (never set) though.

Then in some future version of heroic we can remove it.

CommandMC:
I have now added a migration to set `steamRuntime` to "umu-scout" if `useSteamRuntime` is enabled

src/frontend/screens/Settings/components/SteamRuntime.tsx
```diff
@@ -1,48 +1,93 @@
...
+          setSteamRuntime(false)
+      }
+    },
+    [setSteamRuntime]
+  )
 
-  if (!isLinux || !showSteamRuntime) {
-    return <></>
-  }
+  if (!(isLinux && isLinuxNative)) return <></>
```
arielj: I think this only needs to check `if (!isLinuxNative)` here? we are already checking `isLinux` to set `isLinuxNative` in the useSettingsContext hook

arielj (approved):
Looks good.

I tested migrating an older config and a few native games. I can see the runtime and umu used in the logs when activated and nothing used when set to None.

I jsut I left one comments with a tiny thing.

src/backend/migration/migrations/config.ts
```diff
@@ -0,0 +1,35 @@
...
+  async run() {
+    if (!isLinux) return true
+
+    for (const entry of await readdir(gamesConfigPath, {
+      withFileTypes: true
+    })) {
+      const filename = entry.name
+      if (!entry.isFile() || !filename.endsWith('.json')) continue
+      const appName = parse(filename).name
+      const config = GameConfig.get(appName)
```
arielj, line 22: One thing I noticed with this is that if there's any invalid config file in the config directory, the migration fails and it tries on every boot until the broken files are removed/fixed and it can finish.

I don't know how I got invalid config files, I had some files with just `{}` inside it from 2024, maybe it's just some issue that only happens with weird cases because of development.

Anyway, maybe it's worth adding a try/catch here for each file that gets processed just in case anyone has some corrupted file for some random reason (I really don't know how I got those there 2 years ago).

CommandMC: I do also have files like that, but they did not seem to cause issues in my case. Could you get a backtrace on the error?

If there are config files that are otherwise corrupted, I don't think we should just ignore them.

arielj: Sorry, I deleted the weird files and can't get this to happen again.

Now I'm thinking that maybe the ones with `{}` were not the problem, I had one file that had a different format in the name `<appname>_legendary.json` which probably came from some old test that had some content and I deleted it too, maybe the issue was no the content itself but the name not having the expected format (so appName in that code would end up with an invalid appName).

src/frontend/screens/Settings/components/SteamRuntime.tsx
CommandMC: Yes, this is correct.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/51a36d70428884a640183f4b3737ae7ee4befdd3/src/frontend/hooks/useSettingsContext.ts#L27-L28

Done in https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5832/changes/9bf8bb3352dc1e6ac2fdfa04ec14dc41d3a8093a

Issue #5154, opened 2026-01-06
Option to run without GUI when opening heroic:/ intents via xdg


kuba-orlik:
### Problem description

I'm running games from Heroic via Steam ("add to Steam" option within Heroic itself). I'm using Steam's Big Picture mode. What bugs me is that when I launch the game, Heroic Launcher's UI starts first, and it makes it more fiddly to then exit the game once I'm done playing.

### Feature description

Add a toggle in the settings to make launching games in Heroic via `heroic:/` urls skip displaying the Heroic window and jump straight into the game

Alternatively, add support for a `&gui=false` parameter in the `heroic:/` url that will keep that configurable per-url.

### Alternatives

I've changed the .desktop file that XDG uses to figure out what command to run when opening a `heroic:/` url to alway add the `--no-gui` flag. But now when launching Heroic from the apps list... well the GUI doesn't show up.

PR #5833, opened 2026-08-18
fix: Don't change main window visibility state when launching a game

julien-f:
## Summary

Launching a game via a `heroic://launch/...` request while Heroic is already running (e.g. a desktop shortcut, or Steam's "Add to Steam" integration) goes through the `second-instance` handler, which called `mainWindow.show()` whenever `hideWindowOnProtocolLaunch` was off — which is the default (#5154 / #5501). This forced the main window open, and un-minimized it, even if the user had left it minimized or hidden by some other means. The only way to avoid that was enabling `hideWindowOnProtocolLaunch`, but that setting does the opposite extreme: it always *hides* the window on launch, even if it was open — there was no way to just leave the window's current state alone.

Concretely, with `minimizeOnLaunch` enabled but `hideWindowOnProtocolLaunch` left off (a common combination, since these are two independent settings), launching a game from a shortcut caused the window to visibly flash open and then immediately hide again:
1. `second-instance` forces `mainWindow.show()`
2. moments later, `launcher.ts`'s `minimizeOnLaunch` handling calls `mainWindow.hide()`

## What changed

Added `isGameLaunchArgs()` to detect that a `second-instance` event is specifically a game-launch request (`heroic://launch/...`), and skip the forced `show()` in that case instead of gating it on `hideWindowOnProtocolLaunch`. Starting a game via a protocol/shortcut launch now never changes the main window's current visibility state — open stays open, minimized stays minimized, hidden stays hidden.

The existing opt-in "force hide" behavior (`hideWindowOnProtocolLaunch` setting / `gui=false` URL param) is untouched and still works as before, for anyone who wants the window to always hide on a protocol launch even if it happened to be open.

### Scope

This only touches the `second-instance`/protocol-launch path. Launching a game from inside the already-open Heroic UI (clicking Play) goes through a separate IPC path (`window.api.launch` → `launcher.ts`) that never calls `mainWindow.show()` in the first place — it already left the window's state alone by default, and is unaffected by this change.

## Test plan

- [x] Added unit tests for `isGameLaunchArgs` in `src/backend/__tests__/protocol.test.ts`
- [x] `pnpm codecheck` passes
- [x] `pnpm test:ci` passes (139 tests)
- [ ] Manually verified: with Heroic running and minimized, launching a game via a `heroic://launch` shortcut keeps the window minimized instead of flashing open
- [ ] Manually verified: with Heroic running and visible, launching a game via a shortcut leaves the window open (no unexpected hide)
- [ ] Manually verified: default second-instance behavior (reopening the app, no game-launch URL) still shows/focuses the window as before

### DIFF
--- a/src/backend/__tests__/protocol.test.ts
+++ b/src/backend/__tests__/protocol.test.ts
@@ -25,7 +25,11 @@ jest.mock('../config', () => ({
   }
 }))
 
-import { handleProtocol, shouldHideWindowForProtocolArgs } from '../protocol'
+import {
+  handleProtocol,
+  isGameLaunchArgs,
+  shouldHideWindowForProtocolArgs
+} from '../protocol'
 import { app, dialog } from 'electron'
 import { libraryManagerMap } from '../storeManagers'
 import { getMainWindow } from '../main_window'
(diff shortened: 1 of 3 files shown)


flavioislima:

julien-f:

PR #5843, opened 2026-08-25
fix: remove unnecessary proton prefix creation

liusen373:
When Heroic launches a game through Proton, it first calls prepareWineLaunch to do some preparation work, and then calls runWineCommand to actually start the game.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/37a9bef678a837240477e97b708069dab4517666/src/backend/storeManagers/storeManagerCommon/games.ts#L165-L167

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/37a9bef678a837240477e97b708069dab4517666/src/backend/storeManagers/storeManagerCommon/games.ts#L254-L265

Looking at the code, prepareWineLaunch unconditionally calls verifyWinePrefix. The literal purpose of this function is to ensure that the Wine prefix exists, but what it actually does is forcibly recreate the Wine prefix, regardless of whether the prefix already exists.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/37a9bef678a837240477e97b708069dab4517666/src/backend/launcher.ts#L884

In fact, runWineCommand already checks whether the prefix exists when it runs, and creates it if necessary. Therefore, the verifyWinePrefix call in prepareWineLaunch is completely redundant and should be removed. This can reduce game launch delay by about 1/3 to 1/2.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/37a9bef678a837240477e97b708069dab4517666/src/backend/launcher.ts#L1533-L1540

---

### DIFF
--- a/src/backend/launcher.ts
+++ b/src/backend/launcher.ts
@@ -881,7 +881,6 @@ async function prepareWineLaunch(
     )
   }
 
-  await verifyWinePrefix(gameSettings)
   const experimentalFeatures =
     GlobalConfig.get().getSettings().experimentalFeatures
 


flavioislima:

liusen373:

CommandMC:
> and then calls runWineCommand to actually start the game

This is only the case for sideloaded games. The storemanagers for Epic, GOG, and Amazon may assume that a valid prefix already exists (to perform per-storefront preparation), so removing it is not safe.

> In fact, runWineCommand already checks whether the prefix exists when it runs, and creates it if necessary

Correct, and it *only* runs `verifyWinePrefix` if prefix files are missing. So if `verifyWinePrefix` was already ran before, `runWineCommand` will not call it again. So, there is no overhead.

liusen373:
> > In fact, runWineCommand already checks whether the prefix exists when it runs, and creates it if necessary
> 
> Correct, and it _only_ runs `verifyWinePrefix` if prefix files are missing. So if `verifyWinePrefix` was already ran before, `runWineCommand` will not call it again. So, there is no overhead.

The verifyWinePrefix in runWineCommand is not the issue; the problem is the verifyWinePrefix call in prepareWineLaunch.

liusen373:
@CommandMC 
>> and then calls runWineCommand to actually start the game
>
>This is only the case for sideloaded games. The storemanagers for Epic, GOG, and Amazon may assume that a valid prefix already exists (to perform per-storefront preparation), so removing it is not safe.

Indeed, only sideloaded games are launched via runWineCommand.

I checked the code for the other store managers. For games from GOG, Nile, and Legendary, there is an additional setup step after installation, and this step ensures the prefix is initialized. So overall it should be safe.

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/37a9bef678a837240477e97b708069dab4517666/src/backend/storeManagers/legendary/setup.ts#L13-L39

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/37a9bef678a837240477e97b708069dab4517666/src/backend/storeManagers/nile/setup.ts#L83-L86

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/37a9bef678a837240477e97b708069dab4517666/src/backend/storeManagers/gog/setup.ts#L76-L106

However, I'm not very familiar with the store managers other than sideload, so further confirmation is still needed.

CommandMC:
> The verifyWinePrefix in runWineCommand is not the issue; the problem is the verifyWinePrefix call in prepareWineLaunch.

Alright, but `prepareWineLaunch` verifying that the Wine prefix is there is intended functionality. The point is that storefront implementations can call that function & then be sure that a Wineprefix exists at its intended path. A storefront may (and I say may here, since we will soon have plugins, so we cannot audit all storefront code) directly operate on the Wineprefix (for example, editing registry values by opening `user.reg`). It being called is not an implementation detail, it is part of how the function is supposed to work.

liusen373:
> Alright, but `prepareWineLaunch` verifying that the Wine prefix is there is intended functionality. The point is that storefront implementations can call that function & then be sure that a Wineprefix exists at its intended path. A storefront may (and I say may here, since we will soon have plugins, so we cannot audit all storefront code) directly operate on the Wineprefix (for example, editing registry values by opening `user.reg`). It being called is not an implementation detail, it is part of how the function is supposed to work.

So, the intended solution would be to add check-and-initialize logic to prepareWineLaunch similar to what runWineCommand does?

In any case, unconditionally reinitializing the prefix is too much.

liusen373:
@CommandMC 

Additional info: Based on the source code of umu-launcher and Proton, if you want to run a program with Proton, any manual prefix creation is unnecessary — Proton will automatically create the prefix at runtime.

https://github.com/Open-Wine-Components/umu-launcher/blob/e2b203a1fdd2af9f35166f5713cb3f85d72587e2/tools/umu-run-posix#L98-L108

https://github.com/GloriousEggroll/proton-ge-custom/blob/125149cb5168097625030d9c49de2d44acde354a/proton#L1117-L1141

https://github.com/GloriousEggroll/proton-ge-custom/blob/125149cb5168097625030d9c49de2d44acde354a/proton#L2638-L2639

https://github.com/GloriousEggroll/proton-ge-custom/blob/125149cb5168097625030d9c49de2d44acde354a/proton#L2793

I removed verifyWinePrefix from both prepareWineLaunch and runWineCommand, compiled and packaged Heroic, and it can still launch new games normally.

CommandMC:
The point is not that the prefix is created once we launch the actual game. The point is that we need to do setup *before* we run the game. For example, as part of the Rockstar Launcher "EpicGamesLauncher.exe" wrapper, we copy the executable to `<prefix>/drive_c/windows/command/`. We have to do that after the prefix is there, but before we launch the game.

CommandMC:
> So, the intended solution would be to add check-and-initialize logic to prepareWineLaunch similar to what runWineCommand does?
> In any case, unconditionally reinitializing the prefix is too much.

Alright, I can get behind that. I think the best course of action would be calling `verifyWinePrefix` in `prepareWineLaunch` unconditionally like before, and then doing the "Do we actually have to do anything?" check in `verifyWinePrefix`

Edit: Sorry, got the function names wrong at first here. Been a long day.

liusen373:
> Alright, I can get behind that. I think the best course of action would be calling `verifyWinePrefix` in `prepareWineLaunch` unconditionally like before, and then doing the "Do we actually have to do anything?" check in `verifyWinePrefix`

To preserve this discussion, I'll keep this PR open for now. Once you've finished that work, please go ahead and close it.

arielj:
> > Alright, I can get behind that. I think the best course of action would be calling `verifyWinePrefix` in `prepareWineLaunch` unconditionally like before, and then doing the "Do we actually have to do anything?" check in `verifyWinePrefix`
> 
> To preserve this discussion, I'll keep this PR open for now. Once you've finished that work, please go ahead and close it.

I understand that idea command mentions would replace this PR, right? the conversation is preserved even if the PR is closed, so if you are not planning to do that change I'd suggest closing this PR instead of leaving it open

PR #5863, opened 2026-09-01
[Fix] Console mode control QoL Fixes like Gate launch cancel hold, add close combo

### LINKED ISSUE #5520: Steam controls clash with Heroic Console mode
### Describe the bug

If you launch a game through the new awesome console mode on the steam deck after launching Heroic through Steam, it gives you the option to cancel the launch by holding down the B button for 3 seconds. The issue is this prompt to cancel the launch never goes away even when the game has already launched and heroic still registers button presses while the game is focused and running. This means that if something in the game requires you to hold down the B button for more than 3 seconds (in my case when playing Animal Crossing New Horizons it is the sprint button) it causes issues as it tries to exit the game and it switches to the heroic window.

### Add logs

```text
N/A
```

### Steps to reproduce

1. Tested on Steam Deck, add your games to Heroic and launch Heroic through Steam
2. Enable Console mode and launch a game
3. Hold down the B button for 3 seconds or more whilst in the game

### Expected behavior

Once the game has launched through Console mode, the game should no longer be registered as "launching" or "playing" and controller input should not function within the launcher at all while the game is in focus. No problems when not using Console mode. When Steam is closed completely and I tested in Desktop mode it all worked fine so it looks like Steam controller input software clashes with heroic.

### Screenshots

_No response_

### Heroic Version

Latest Stable

### System Information

Steam OS
Steam Deck OLED

### Additional information

_No response_

### LINKED ISSUE #5541: The setting "disable controller" shouldn't apply in Console Mode
### Problem description

I'm running Heroic on Steam Deck, on Desktop Mode displayed on my TV (Deck is docked) and controlling it with my Xbox 360 controller conveniently set up with Steam Input for it to send mouse and keyboard inputs. 

I always keep the Heroic setting "disable controller" enabled because if not, Heroic receives at once both mouse/keyboard inputs AND controller inputs.

I _could_ use a pure controller setup, but so far I have not been happy with the way Heroic controls with a controller...

### Feature description

...but then Console Mode was released and it does sound like a great way to navigate Heroic using controller. Hence my request for Heroic NOT to read the "disable controller" setting when in Console Mode, since this mode is made specifically for controller and that setting should therefore not apply in this.

### Alternatives

_No response_

### Additional information

_No response_

### LINKED ISSUE #5577: Add a way to close the currently running game with game controller button combo
### Problem description

I like the newly added Console Mode of Heroic but I need a feature that I already use with Steam Big Picture (also Steam Deck gaming mode) or with Steam Remote Play.

Sometimes a game has no quit button at all (sideloaded web browser apps), sometimes I want to quit the game immediately without navigating to the title menu/main menu.

On Steam I can just display Steam Overlay at any moment and select "Exit Game" from there.

With Steam Remote Play there is another type of overlay only available for the remote connection. I can close the game (or terminate the stream) from there even more quickly.

**To put it shortly: I want to be able to close the game by pressing a game controller button combo. Basically, send an ALT+F4 by using the controller.**

I am using Linux and I was able to recreate this feature by using Input Remapper. But something like this would improve the Heroic Console Mode for me.

### Feature description

Make it possible to close the currently running game using a game controller button combo.

For example, [on an Xbox 360 controller](https://rbwhitaker.wdfiles.com/local--files/monogame-xbox-controller-input/ControllerAnatomy.png): BACK+RT. Or BACK+START.

### Alternatives

Use Input Remapper and map ALT and F4 to specific buttons or button combo.

### Additional information

Steam Big Picture or Steam Deck use Steam Overlay to offer this type of functionality. Steam Remote Play offers similar one with its own stream overlay.

Drackrath:
Fixes: #5520 
Fixes: #5577
Fixes: #5541

Note:

Changed Close-Combo to View + R2
This should be possible to edit for the user in the future. Seperate PR 

Starting Console mode now is also: View + R2 for 3s

Focuses Heroic with: View + R2 on Wayland & KDE Desktops

Hardened Disabled control options in console mode. Now only View + R2 works (close)

Added override config option in Settings.

---


src/frontend/screens/ConsoleMode/components/ConsoleComboListener/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const navigate = useNavigate()
+  const { startHold, stopHold } = useCancelOnHold({
+    active: true,
+    holdMs: 3000,
+    onCancel: () => navigate('/console')
+  })
+  useGamepadComboHold([BTN_SELECT, BTN_R2], (held) => {
+    if (held) {
+      // Surface Heroic during the countdown
+      window.api.focusMainWindow()
```
flavioislima, line 15: This listener is mounted at the root for the whole non-console app and fires `focusMainWindow` on the very first frame View+R2 is seen, not after the 3s hold. When a game was launched from the normal Library and Heroic is in the background, a momentary View tap while holding RT (common for map/aim in many games) will restore/show/focus Heroic over the fullscreen game, and on Plasma Wayland force-activate it via KWin.

`gamepad.ts` already gates its own input handling on `isFocused` for exactly this reason (Chromium keeps delivering gamepad state to unfocused windows). `useGamepadComboHold` has no such gate, so please either require `document.hasFocus()` / the existing focus flag before starting the hold, or only call `focusMainWindow` once the hold actually completes.

src/frontend/screens/ConsoleMode/hooks.ts
```diff
@@ -78,6 +84,46 @@ export function useGamepadButtonHold(
...
+  enabled = true
+) {
+  const handlerRef = useRef(onChange)
+  handlerRef.current = onChange
+
+  // Stable key avoids re-running on rerenders
+  const comboKey = buttonIndices.join(',')
+
+  useEffect(() => {
+    if (!enabled) return
```
flavioislima, line 101: `useGamepadButtonPress` and `useGamepadButtonHold` were updated to respect `isControllerNavDisabled()`, but `useGamepadComboHold` and `isGamepadButtonHeld` don't check it. With `disableController=true` and the new setting off, A/B/L1/R1/R2 are correctly dead but View+RT still enters console mode from anywhere, quits console mode, and kills a playing game via the LaunchOverlay combo. Same guard as the sibling hooks should be enough.

src/backend/utils/kwin_activate.ts
```diff
@@ -0,0 +1,74 @@
...
+      return stdout.trim()
+    } catch {
+      continue
+    }
+  }
+  throw new Error('qdbus not available')
+}
+
+// KWin denies focus stealing on Wayland
+export async function kwinActivateWindow(caption: string): Promise<void> {
```
flavioislima, line 27: This fallback can't work in the Flatpak build, which is the main Plasma Wayland / Steam Deck desktop target it's aimed at: `qdbus`/`qdbus6` are exec'd inside the sandbox (the runtime is `org.freedesktop.Platform`, no qdbus, so ENOENT), the manifest has no `--talk-name=org.kde.KWin`, and the script is written to sandbox-private temp that host KWin can't read. Both spawns fail, `main.ts` swallows the error, and nothing is logged.

It would need `flatpak-spawn --host` plus the D-Bus permission and a host-visible script path, or alternatively just skip the fallback when running in Flatpak and log that.

flavioislima (commented):
Thanks for this, the hold-to-quit and launch cancel gating are a clear improvement. Three things I'd like addressed before merging, see inline.

flavioislima (changes requested):

src/backend/utils/kwin_activate.ts
Drackrath: I'd rather just log it, as I do not use flatpak at all, and I have little incentive for the use of flatpak builds...

src/frontend/screens/ConsoleMode/hooks.ts
Drackrath: The Idea behind this, was to still allow the console-mode even if controller *Navigation* is disabled. Argueably this does not navigate here, and still enables Console-Mode, which also has Controller support enabled by default. This would drop my intended behaviour.... Do we still apply it your way?

Drackrath: I could also put that in a condition that if the controller is enabled by default in console-mode, this would also enable the start of console-mode with the View+RT, and if it's off, it wouldn't

src/backend/utils/kwin_activate.ts
flavioislima: ok add a TODO and check if flatpak you log it.

src/frontend/screens/ConsoleMode/hooks.ts
flavioislima: OK, makes sense then lets keep as it is.

PR #5871, opened 2026-09-06
[UX] Do not allow moving a game inside itself

arielj:
This PR adds one safety check before moving a game: it does not allow moving a game inside a subfolder of the current folder.

This is important because moving a game moves all the files from the current install dir into the new install dir and then it deletes the original dir. So since files are move into a subdirectory that last deletion removes all game files after they are moved.

---


src/backend/main.ts
```diff
@@ -958,35 +958,72 @@ addHandler(
...
+    if (path.startsWith(install.install_path!)) {
+      // we don't want to move a game in a subfolder of the current install directory
+      // it will cause all game files to be deleted since moving moves all files to the
+      // new directory and deletes the old one
+      validNewPath = false
+
       notify({
         title,
-        body: i18next.t('notify.error.move', 'Error Moving Game')
+        body: i18next.t('notify.error.error', 'Error Moving Game')
```
CommandMC: The key change here seems like a downgrade. `notify.error.move` is not *great*, but `notify.error.error` has no info about it being move-related

CommandMC (commented):
The duplication of logging functions is a little unfortunate. Maybe you can define a local function and then call it with either error? So:

```ts
function onMoveError(error: unknown) {
  notify(...)
  logError(...)
}

if (path.startsWith(...)) {
  onMoveError(`New install path...`)
} else {
  const moveRes = ...
  if (moveRes.status === 'error') {
    onMoveError(moveRes.error)
  }
)
```

src/backend/main.ts
arielj: haha, I have no idea how this happened, some kind of typo/drag-and-drop by mistake

CommandMC (approved):

PR #5881, opened 2026-09-11
fix: Set launch env vars for sideloaded Windows games

shawly:
Sideloaded Windows games launch without `HEROIC_APP_NAME`, `HEROIC_APP_RUNNER`,
`HEROIC_APP_SOURCE`, `GAMEID` and the known-fixes variables. Native sideloaded
games get all of them.

`launchGame()` builds the env inside the `if (isNative)` branch and hands it to
`callRunner()`. Windows games fall through to `runWineCommand()`, which does
accept an `env` option, but the call site only passes wrappers and log writers.
So the env is built and then dropped.

This moves the env up next to `wrappers` and passes it to both branches, which
is what the gog, legendary, nile and zoom managers already do.

### How I ran into it

I use a wrapper command that calls `ludusavi wrap --infer heroic`. It reads
`HEROIC_APP_NAME` and `HEROIC_APP_RUNNER` to work out which game is starting.
It works for native games and reports every sideloaded Windows game as
unrecognized. Setting both variables by hand makes it work, so the lookup is
fine and the variables simply never arrive.

### One behaviour change worth flagging

Sideloaded Windows games now get `GAMEID=umu-0` instead of nothing. That is the
fallback the other runners start from before they swap in a real umu id, so it
looks like the right default here, but shout if you would rather I left it out.

---


flavioislima:

shawly:

PR #5891, opened 2026-09-15
Add support for building ARM64 flatpak

matte-schwartz:
Bit of a personal side project over the past few days: add support for building ARM64 flatpak versions of Heroic. Will need an accompanying change to the flatpak packaging repo. Not my typical language so review would be appreciated. Tested with a locally-built version as a clean install.

---


matte-schwartz:
Also, just to clarify, this has all been tested on Steam Frame running SteamOS. I don't have any ARM handhelds to test this on unfortunately.

PR #5898, opened 2026-09-15
[chore] use scalable icon

Drsheppard01:
OS: CachyOS

I use fuzzel as launcher and i get this message

```shell
warn: render.c:1528: /home/drsheppard01/.local/share/icons/heroic.png: PNG is too large (1024x1024); downscaling using a less precise filter (fast)
```
This is, of course, a Fuzzel issue, but since the repository contains an [excellent vector icon](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/main/src/frontend/assets/heroic-icon.svg) for Heroic, I decided to make this change.

This change will also slightly reduce the size of the installation file


CommandMC:
Are you sure Electron-Builder supports this? [Their documentation](<https://www.electron.build/docs/linux#icon-configuration>) does not mention being able to use a .svg

Drsheppard01:
Thx for quickly reply. Shortly: yes, I tested it. 

electron-builder follows the freedesktop Icon Theme specification: https://specifications.freedesktop.org/icon-theme/latest/#directory_layout and .svg is listed as a valid option

The [example](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/templates/icons/proton-native/proton-native.svg) also includes an icon in SVG format.

CommandMC:
Hm, I see.

I am wondering though, how did you install Heroic? Our AUR package, the CachyOS package, and the Flatpak all have proper icons (in both svg and png formats) set.

Drsheppard01:
From AppImage

arielj:
I wonder if we should remove the other icons

From https://www.electron.build/docs/features/icons-and-images/#tldr--recommended-build-layout it says:

```
SVG is best
SVG icons are rasterized at 1024px before conversion, giving every generated size a clean starting point. Use SVG when your icon is available in vector form.
```

and mentions the PNG as the alternative to SVG and all the other icns, ico and different sizes are all optional, but looks like with SVG we can just use the SVG and remove unused files

Drsheppard01:
For Linux need 3 icons in next file structure:

/usr/share/hicolor/scalable/{APP ID}.svg or {APPNAME}.svg
/usr/share/hicolor/scalable/symbolic/light.svg 
/usr/share/hicolor/scalable/symbolic/dark.svg

There are no standardized naming conventions for tray icons, but it would be best if users had two icons—one for the light theme and one for the dark theme (similar to the current setup, but in SVG).

PR #5915, opened 2026-09-19
fix: Don't trust `isUmuSupported`'s return value if we're running a native game

CommandMC:
A native game's Wine Version might still be set (due to... fun issues in the config system), which will make `isUmuSupported` return true. In turn, this causes us to use umu for those games. In principle this even *works*, but it should certainly not happen without the user's involvement.

To test this: Add a native Linux game to Heroic. On main, note that umu is used by default. On this PR, note that umu is no longer used by default.

---


BananaWorks07 (commented):
This looks fine and would fix any issue, but we really should fix whatever issue config system has that is causing this

BananaWorks07 (approved):

PR #5917, opened 2026-09-19
feat: clear browser cache as part of Clear Heroic Cache

### LINKED ISSUE #5911: Allow an option to clear the browser cache
### Problem description

When I logged in to GOG on Heroic while my VPN is enabled, it rightfully did not show my local currency. This is the default GOG behavior, and just turning off VPN and signing in again is not enough for this: cookies need to be cleared. After clearing cookies and logging in without VPN, web site works fine. However, I can't do the same on the Heroic Launcher; therefore I get stuck with wrong currency. 

I believe this is a rare case, as I just opened a GOG account, and logged in to Heroic just afterwards and cleared GOG cookies to set my local currency later. 

### Feature description

`Clear Heroic Cache` can additionally clear browser cache as well, or `delete browser cookies` may be added as an additional option. I understand that it might be intended that does exist to persist logins. But it could likely solve some other potential issues like this. 

### Alternatives

The current workaround is manually deleting files under /config, which I did to get this to work, without reinstalling. 

### Additional information

_No response_

Abhiroop3917:
Closes #5911

Browser cache (cookies, localStorage, IndexedDB) is now cleared automatically when the user clicks "Clear Heroic Cache" in Settings > Advanced, using session.defaultSession.clearStorageData().

The info box has been updated to list browser cache as one of the items that gets cleared.

---


Abhiroop3917:
@CommandMC could you please review this PR when you get a chance? This addresses issue #5911 by clearing the browser cache automatically as part of the existing "Clear Heroic Cache" action, as you suggested.

src/backend/main.ts
```diff
@@ -685,6 +685,10 @@ addListener('clearCache', (event, showDialog, fromVersionChange = false) => {
   }
 })
 
+addListener('clearBrowserCache', async () => {
+  await session.defaultSession.clearStorageData()
```
CommandMC: - I'm not sure if all session data qualifies as "cache"
- Are you sure doing this for the default session works? Webviews use separate storage partitions
- I would just do this in the existing `clearCache` handler

Abhiroop3917:
Thank you for the review @CommandMC! You're right on all three points:

1. **Session data vs cache** — `clearStorageData()` without options clears everything (cookies, localStorage, IndexedDB etc.), which is too broad. We should pass `{ storages: ['cachestorage', 'shadercache'] }` to limit it to actual cache only.

2. **Wrong session** — After checking the codebase, the webviews use named partitions (`persist:epic`, `persist:gog`, `persist:amazon`, `persist:zoom`) via `session.fromPartition()`, not `defaultSession`. So our current implementation clears the wrong session entirely. We need to clear each of these partitions individually.

3. **clearCache handler** — Agreed, I'll move this logic into the existing `clearCache` function in `utils.ts` and remove the separate IPC listener.

I'll push a fix addressing all three points.

PR #5918, opened 2026-09-19
fix: Don't try to `chmod` native Linux/macOS binaries on exit

CommandMC:
When running a sideloaded native game, Heroic will ensure it's executable:

https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/861874cc4e180a5a1a2c16df24bb32092b89f93a/src/backend/storeManagers/storeManagerCommon/games.ts#L207-L218

To restore the previous state, we then tried to `chmod` the binary again once it is no longer running. This has two issues:
1. If the binary was already executable before, we end up making it *not* executable afterwards
2. If the binary is read-only, this will cause an exception (this is especially problematic since wrappers often are read-only)

Two options to fix this, either add a "Did we run chmod before" variable, or just don't attempt to restore at all. I don't think it's a problem if Heroic automatically makes binaries executable, so I chose to just remove the restore step entirely.

To test: Add a native game to Heroic. Configure it to use a wrapper at a read-only location (e.g. GameMode, Mangohud, system umu). Launch the game, then quit it. On main: In the log, see "An exception occurred when launching the game" error mentioning "EPERM". On this PR: See no error

---


arielj (approved):
I wonder if we should actually have some popup and let the user know when they pick something not executable and ask them if they want Heroic to make it executable.

Sounds more secure to let the user know we are changing that when they approve it (and users that don't read will just click yes yes anyway the first time).

Anyway, that's not in the scope of this PR, just though it could be a good idea in the future.

